PR #2, opened 2019-01-11
Get revision and service listing skeletons in.

sixolet:
This adds the kn program, with basic commands for service and revision listing
in.

kn service list
kn revision list

They use the genericclioptions library from kubernetes to support jsonpath,
yaml, and json output. The default is to just list the names of the objects, for
now, until we add in some of the kubectl-based libraries for creating tables in
another commit.

This is deliberately bare-bones; I am looking to get the basic skeletons of how
this repository is laid out down.

No tests yet. Soon.


cppforlife:
@sixolet what do you think about pattern a lot of go CLIs go with: have a cmd/kn/ directory with main.go and pkg/kn/commands/ that contains implementation of commands (similar to https://github.com/kubernetes/kubernetes/blob/master/cmd/kubectl/kubectl.go)

cppforlife:
i think we should also figure out how to move some of this stuff into objects instead of global variables. i've heard for example from riff folks that they want to use CLI commands as a library as well to add more of their own commands. might be a bit cleaner.

we should also figure out how to only have one exit point (one `os.Exit(1)`) and propagate errors instead of just existing.

sixolet:
That structure of files sounds fine and I'll make this PR so; this is just the structure generated for me by the cobra commandline, rather than something I decided was best.

I'll look at the exiting and non-global suggestions and see if they fit in this PR or a separate follow-on PR or two.

sixolet:
K, I think making stuff less global is going to wait; the examples I can find of Cobra best practices all have their commands global and things initialized using module init functions. Let's have conversation about how to unglobalize stuff after this is in?

pkg/kn/commands/revision_list.go
```diff
@@ -0,0 +1,73 @@
...
+
+// listCmd represents the list command
+var revisionListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List available revisions.",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		// use the current context in kubeconfig
+		config, err := clientcmd.BuildConfigFromFlags("", kubeCfgFile)
+		if err != nil {
+			panic(err.Error())
```
cppforlife: return err here

cppforlife:
> I think making stuff less global is going to wait

ok. will wait for that.

> the examples I can find of Cobra best practices all have their commands global and things initialized using module init functions

knctl does it the same way as kubectl 

- https://github.com/cppforlife/knctl/blob/master/pkg/knctl/cmd/service/list.go
- https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/scale/scale.go

ultimately all commands are combined into root cmd: 

https://github.com/cppforlife/knctl/blob/master/pkg/knctl/cmd/knctl.go

vdemeester:
Agreeing with @cppforlife on the non global variable thing. In `docker/cli` we're doing something close to what `kubectl` does : [pull](https://github.com/docker/cli/blob/master/cli/command/image/pull.go#L26), combined in [commands.go](https://github.com/docker/cli/blob/master/cli/command/commands/commands.go).

That said, we can iterate in follow-ups if needed :angel:

cppforlife:
/lgtm as a starting point. 

> No tests yet. Soon.

i think this should be next. writing some unit tests may actually change some of the structure.

PR #10, opened 2019-01-29
Refactor to allow testing; write first tests.

sixolet:
* Make commands take a params object
* Provide params for capturing output
* Provide params for mocking serving api
* Write tests with captured output and mocked serving api

### DIFF
--- a/pkg/kn/commands/root.go
+++ b/pkg/kn/commands/root.go
@@ -32,7 +32,7 @@ var kubeCfgFile string
 
 // Parameters for creating commands. Useful for inserting mocks for testing.
 type KnParams struct {
-	output         io.Writer
+	Output         io.Writer
 	ServingFactory func() (serving.ServingV1alpha1Interface, error)
 }
 
@@ -63,8 +63,8 @@ Serving: Manage your services and release new software to them.
 Build: Create builds and keep track of their results.
 Eventing: Manage event subscriptions and channels. Connect up event sources.`,
 	}
-	if p.output != nil {
-		rootCmd.SetOutput(p.output)
+	if p.Output != nil {
+		rootCmd.SetOutput(p.Output)
 	}
 	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.kn.yaml)")
 	rootCmd.PersistentFlags().StringVar(&kubeCfgFile, "kubeconfig", "", "kubectl config file (default is $HOME/.kube/config)")
@@ -83,7 +83,7 @@ func initKubeConfig() {
 	if kubeCfgFile == "" {
 		home, err := homedir.Dir()
 		if err != nil {
-			fmt.Println(os.Stderr, err)
+			fmt.Fprintln(os.Stderr, err)
 			os.Exit(1)
 		}
 		kubeCfgFile = filepath.Join(home, ".kube", "config")
@@ -99,7 +99,7 @@ func initConfig() {
 		// Find home directory.
 		home, err := homedir.Dir()
 		if err != nil {
-			fmt.Println(os.Stderr, err)
+			fmt.Fprintln(os.Stderr, err)
 			os.Exit(1)
 		}
 
@@ -112,7 +112,7 @@ func initConfig() {
 
 	// If a config file is found, read it in.
 	if err := viper.ReadInConfig(); err == nil {
-		fmt.Println("Using config file:", viper.ConfigFileUsed())
+		fmt.Fprintln(os.Stderr, "Using config file:", viper.ConfigFileUsed())
 	}
 }
 
--- a/pkg/kn/commands/root.go
+++ b/pkg/kn/commands/root.go
@@ -16,19 +16,44 @@ package commands
 
 import (
 	"fmt"
+	"io"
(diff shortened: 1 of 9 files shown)


pkg/kn/commands/root.go
```diff
@@ -16,19 +16,44 @@ package commands
...
+func NewKnCommand(params ...KnParams) *cobra.Command {
+	var p *KnParams
+	if len(params) == 0 {
+		p = &KnParams{}
+	} else if len(params) == 1 {
+		p = &params[0]
+	} else {
+		panic("Too many params objects to NewKnCommand")
+	}
+	p.Initialize()
```
cppforlife: callee of NewKnCommand should probably call Initialize() instead of this constructor knowing about it. as a nice side effect tests wont keep on calling it every time as well.

pkg/kn/commands/service_list_test.go
```diff
@@ -0,0 +1,111 @@
...
+		func(a client_testing.Action) (bool, runtime.Object, error) {
+			action = a
+			return true, response, nil
+		})
+	cmd.SetArgs(args)
+	err = cmd.Execute()
+	if err != nil {
+		return
+	}
+	for {
```
cppforlife: `output = strings.Split(buf.String(), "\n")`

pkg/kn/commands/service_list_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/knative/serving/pkg/apis/serving/v1alpha1"
+	serving "github.com/knative/serving/pkg/client/clientset/versioned/typed/serving/v1alpha1"
+	"github.com/knative/serving/pkg/client/clientset/versioned/typed/serving/v1alpha1/fake"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	client_testing "k8s.io/client-go/testing"
+)
+
+func fakeList(args []string, response *v1alpha1.ServiceList) (action client_testing.Action, output []string, err error) {
```
cppforlife: as we add more unit tests like this, i think we will end up splitting setting up fakes (specific to command) from calling the command (generic) and getting its output.

pkg/kn/commands/service_list_test.go
```diff
@@ -0,0 +1,111 @@
...
+				TypeMeta: serviceType,
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "bar",
+				},
+			},
+		},
+	})
+	if err != nil {
+		t.Error(err)
+		return
```
cppforlife: t.Fatal instead of t.Error+return

pkg/kn/commands/root.go
```diff
@@ -84,6 +112,18 @@ func initConfig() {
 
 	// If a config file is found, read it in.
 	if err := viper.ReadInConfig(); err == nil {
```
cppforlife: unrelated to the PR, but we should exit if ReadInConfig fails. ideally though all the exits are consolidated via return err.

pkg/kn/commands/root.go
sixolet: Disagree. The goal is to minimize the interface, so when you call NewKnCommand you're ready to go.

pkg/kn/commands/root.go
```diff
@@ -16,19 +16,44 @@ package commands
...
+
+func (c *KnParams) Initialize() {
+	if c.ServingFactory == nil {
+		c.ServingFactory = GetConfig
+	}
+}
+
 // rootCmd represents the base command when called without any subcommands
-func NewKnCommand() *cobra.Command {
+func NewKnCommand(params ...KnParams) *cobra.Command {
```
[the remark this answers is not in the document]
sixolet: So that NewKnCommand can have no parameters. Following the practice detailed [here](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis), which I liked for the entry we expect people to use to a lib.

pkg/kn/commands/service_list_test.go
sixolet: Possibly. I suspect the exact format of private methods in the tests will vary and change; I just wanted to make sure we parameterized the commands.

pkg/kn/commands/root.go
sixolet: ok

pkg/kn/commands/root.go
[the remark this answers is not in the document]
sixolet: NB: Open to going to the full "functional options" route too, but this seemed a reasonable compromise between easy default API and simple configuration API.

pkg/kn/commands/service_list_test.go
sixolet: Done. Thanks

pkg/kn/commands/service_list_test.go
sixolet: Done. Thanks.

pkg/kn/commands/root.go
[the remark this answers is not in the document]
cppforlife: k, lets keep it this way for now.

PR #22, opened 2019-02-05
Add a Makefile

cdrage:
Adds a makefile for standard procedures such as creating the binary +
running tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,11 @@
+default: bin
+
+# Build the project
+.PHONY: bin
+bin:
+	go build ./cmd/...
+
+# Run all available tests
+.PHONY: test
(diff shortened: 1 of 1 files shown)


sixolet:
I'd prefer not to include a makefile unless we need something more complicated than a `go build` to build it, or a `go test` to test it.

Other opinions?

sixolet:
Hum. Tests are passing for me. Anyway, we should get CI testing up in this repo.

cdrage:
@sixolet Just opened up issue #23 with regards to tests..

cppforlife:
@cdrage how about just adding bash script such as `hack/build.sh` and `hack/test.sh`

Issue #17, opened 2019-01-31
pickup config from KUBECONFIG


cdrage:
At the moment the client doen't pick anything up from the KUBECONFIG env.

```sh
dev/k8s/knative-client  master ✗                                                                                                                                                                                                                                          1d ⚑  ⍉
▶ export KUBECONFIG=~/admin.conf

dev/k8s/knative-client  master ✗                                                                                                                                                                                                                                           1d ⚑  
▶ ./kn service list             
Error: Get https://192.168.42.79:8443/apis/serving.knative.dev/v1alpha1/namespaces/default/services: dial tcp 192.168.42.79:8443: connect: no route to host
Usage:
  kn service list [flags]

Flags:
      --allow-missing-template-keys   If true, ignore any errors in templates when a field or map key is missing in the template. Only applies to golang and jsonpath output formats. (default true)
  -h, --help                          help for list
  -o, --output string                 Output format. One of: json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-file. (default "jsonpath={range .items[*]}{.metadata.name}{\"\\n\"}{end}")
      --template string               Template string or path to template file to use when -o=go-template, -o=go-template-file. The template format is golang templates [http://golang.org/pkg/text/template/#pkg-overview].

Global Flags:
      --config string       config file (default is $HOME/.kn.yaml)
      --kubeconfig string   kubectl config file (default is $HOME/.kube/config)
  -n, --namespace string    Namespace to use. (default "default")

Get https://192.168.42.79:8443/apis/serving.knative.dev/v1alpha1/namespaces/default/services: dial tcp 192.168.42.79:8443: connect: no route to host
```

nimakaviani:
+1 on this.

navidshaikh:
This can be closed now.

PR #27, opened 2019-02-28
Pickup config path from KUBECONFIG

tzununbekov:
Use `$KUBECONFIG` env variable as config path if set,  #17

### DIFF
--- a/pkg/kn/commands/root.go
+++ b/pkg/kn/commands/root.go
@@ -82,7 +82,12 @@ func InitializeConfig() {
 }
 
 func initKubeConfig() {
-	if kubeCfgFile == "" {
+	if kubeCfgFile != "" {
+		return
+	}
+	if kubeEnvConf, ok := os.LookupEnv("KUBECONFIG"); ok {
+		kubeCfgFile = kubeEnvConf
+	} else {
 		home, err := homedir.Dir()
 		if err != nil {
 			fmt.Fprintln(os.Stderr, err)


cppforlife:
dont think we have integration tests yet, so it might be harder to test /cc @sixolet

tzununbekov:
> Looks okay to me. @tzununbekov would you like to add some test cases for this ?

To test this case I should probably add tests for all init functions in `root.go`. I can do that, but it may take some time.

navidshaikh:
I have tested the PR manually, works fine for me.
I think the error message upon not finding correct KUBECONFIG can be improved.
For eg take a look at
```
KUBECONFIG='' ./kn service list
ERROR: logging before flag.Parse: W0302 11:05:09.100519   14619 client_config.go:548] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
ERROR: logging before flag.Parse: W0302 11:05:09.100769   14619 client_config.go:553] error creating inClusterConfig, falling back to default config: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
Error: invalid configuration: no configuration has been provided
Usage:
  kn service list [flags]

Flags:
  -h, --help                          help for list
  -o, --output string                 Output format. One of: json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-file. (default "jsonpath={range .items[*]}{.metadata.name}{\"\\n\"}{end}")

Global Flags:
      --config string       config file (default is $HOME/.kn.yaml)
  -n, --namespace string    Namespace to use. (default "default")

invalid configuration: no configuration has been provided
```

Do we need to print the help usage after error message ? Dev has to locate what went wrong by parsing the stderr/stdout printed. I think, upon error, only error message needs to be printed, while user always has choice to do `kn --help`

cppforlife:
agreed with @navidshaikh. would be nice to remove help usage. also lines starting with 'ERROR: logging before' seem to be some kind of internal error messaging that's popping up. can they be eliminated as well?

tzununbekov:
Looks like printing help on error is default cobra framework behavior. I can take a look on how we can change it, but I don't think that it's related to this PR

navidshaikh:
Filed https://github.com/knative/client/issues/38 to control the default behavior of cobra to print usage and https://github.com/knative/client/issues/37 to print the error messages properly.

PR #34, opened 2019-03-05
Incorporate reference to knative/test-infra for setting up automated tests.

sixolet:
It doesn't quite all work with go 1.11/1.12 modules yet, but it's getting there.

This should set up the requisite test script and also the requisite files in the vendor directory. With go modules, the vendor directory isn't otherwise necessary, so I've limited the files in it.

I think I still have a few go.mod / go.sum issues to work out, but I wanted to create the pull request now and see where we get.


.gitignore
```diff
@@ -37,3 +37,12 @@ tags
 
 # VSCode specific
 .vscode
+# Ignore everything in vendor except github.com/knative/test-infra
```
cppforlife: can we just move github.com/knative/test-infra into a separate directory (not vendor). .gitignore looks pretty weird.

sixolet: This not only includes the shell script, it includes a little bit of go code, which has to be on an appropriate path. I'll file a bug on the test-infra repo to support modules better, but this was the best compromise I could think of between the module way and the vendor way. This way, `git mod vendor` does the right thing to the repo, but we don't check in all the vendored deps.

adrcunha: I already filed https://github.com/knative/test-infra/issues/546, feel free to add more details since you know firsthand the limitations.

PR #36, opened 2019-03-06
Pick up default namespace from config

### LINKED ISSUE #7: pick up default namespace from .kube/config instead of hardcoding it in kn

tzununbekov:
Namespace parameter moved to root persistent flags. Function `setNamespace()` sets namespace from config context on CLI initialization. Resolves #7

### DIFF
--- a/pkg/kn/commands/root.go
+++ b/pkg/kn/commands/root.go
@@ -17,6 +17,7 @@ package commands
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 
@@ -27,10 +28,24 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
 	"k8s.io/client-go/tools/clientcmd"
+	"sigs.k8s.io/yaml"
 )
 
-var cfgFile string
-var kubeCfgFile string
+var (
+	cfgFile     string
+	kubeCfgFile string
+	namespace   string
+)
+
+type config struct {
+	Contexts []struct {
+		Name    string
+		Context struct {
+			Namespace string
+		}
+	}
+	CurrentContext string `json:"current-context"`
+}
 
 // Parameters for creating commands. Useful for inserting mocks for testing.
 type KnParams struct {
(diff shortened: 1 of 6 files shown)


tzununbekov:
this PR includes changes from #27, so #27 should be merged first I think

cppforlife:
@tzununbekov we should use client-go functionality for this. see knctl for example: https://github.com/cppforlife/knctl/blob/master/pkg/knctl/cmd/core/config_factory.go#L77

tzununbekov:
Looks like I missed this client-go function, thanks for pointing out. I'll update PR

sebgoa:
@tzununbekov can you update this PR with the client-go function

cppforlife:
@tzununbekov still up for making a change to use client-go?

tzununbekov:
nice, client now has tests!
yes, I'm still up to make these changes though it's turned out to be a bit more code than I expected :slightly_smiling_face: 
I think I'll be able to push it tomorrow (eastern hemisphere) and I also see several more issues which I can help with

bbrowning:
Looks like the feedback to use client-go was addressed. @tzununbekov is this ready for review, or were those other issues you mentioned something that would be part of this PR?

tzununbekov:
> Looks like the feedback to use client-go was addressed. @tzununbekov is this ready for review, or were those other issues you mentioned something that would be part of this PR?

@bbrowning this one is ready - I will submit new PRs for other issues

bbrowning:
I pinged Navid on Slack about how to use Prow commands instead of the GitHub UI to indicate "looks good to me". But, since it's late his local time, I'll nudge this one along. I pulled the changed locally, built the updated `kn` CLI, and verified that the default namespace in my kubeconfig gets used as well as that I can manually override that with the global namespace flag.

/lgtm

pkg/kn/commands/root.go
```diff
@@ -78,11 +82,16 @@ Eventing: Manage event subscriptions and channels. Connect up event sources.`,
 func InitializeConfig() {
 	cobra.OnInitialize(initConfig)
 	cobra.OnInitialize(initKubeConfig)
-
+	cobra.OnInitialize(setNamespace)
 }
 
 func initKubeConfig() {
```
evankanderson: It would be nice if there were a generic "initKubeConfig" method from the Kubernetes/kubectl libraries.

Add a TODO?

evankanderson: I think `clientcmd` will do a lot of this resolution if you don't feed it a kubeCfg file. Not sure if empty string argument will trigger its default resolution, or if you need to use a different method.

pkg/kn/commands/root.go
```diff
@@ -129,3 +138,21 @@ func GetConfig() (serving.ServingV1alpha1Interface, error) {
...
+		return
+	}
+	namespace = "default"
+	config := clientConfig(kubeCfgFile)
+	if ns, _, err := config.Namespace(); err == nil {
+		namespace = ns
+	}
+}
+
+func clientConfig(path string) clientcmd.ClientConfig {
```
cppforlife: here is a better version i cooked up for [kapp](https://github.com/k14s/kapp): https://github.com/k14s/kapp/blob/master/pkg/kapp/cmd/core/config_factory.go#L71-L82. it properly deals with multi-value KUBECONFIG values and picks up default env vars based on cmdclient config. (need to update knctl to do so as well...)

tzununbekov:
@evankanderson @sixolet If I understand your comments correctly, we should try to use kubectl methods where it's possible, so I pushed update which makes kn use kubectl config flags. If this makes sense, we should probably move namespace parameter from persistent flags to `KnParams` or somewhere else. Otherwise I'd bring everything back and borrow config handling from https://github.com/k14s/kapp

Issue #37, opened 2019-03-06
Extra error messages showing up before actual error message


navidshaikh:
```
KUBECONFIG='' ./kn service list
ERROR: logging before flag.Parse: W0302 11:05:09.100519   14619 client_config.go:548] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
ERROR: logging before flag.Parse: W0302 11:05:09.100769   14619 client_config.go:553] error creating inClusterConfig, falling back to default config: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
Error: invalid configuration: no configuration has been provided
[..]
```

Lines starting with 'ERROR: logging before' seem to be some kind of internal error messaging that's popping up. It should be fixed.

PR #42, opened 2019-03-19
Fix flag parse error by calling flag.Parse first

nak3:
This patch adds flag.Parse() to avoid `logging before flag.Parse` error.

c.f. -  other repository also solved the issue by calling `flag.Parse()` first.

https://github.com/knative/eventing/blob/8ba0b39f5150f8fae7184ab189f4086cbc98bf42/contrib/gcppubsub/pkg/dispatcher/cmd/main.go#L46

Fixes https://github.com/knative/client/issues/37

### DIFF
--- a/cmd/kn/main.go
+++ b/cmd/kn/main.go
@@ -15,7 +15,6 @@
 package main
 
 import (
-	"flag"
 	"fmt"
 	"os"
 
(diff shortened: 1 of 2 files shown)


cppforlife:
@nak3 does this not affect cobra in any way? (im somewhat surprised)

nak3:
@cppforlife Thank you. Updated it now. (`kn -h` had an issue in previous patch as you guessed.)

bbrowning:
Kicking these tests since a fix for these CI failures has since been merged.

/retest

Issue #38, opened 2019-03-06
Control (default?) help usage printing by cobra upon CLI operational errors


navidshaikh:
If there is an error in an operation directed via CLI, CLI prints the error and prints the help usage as well.
This makes reading the error about operation a bit more difficult. One has to locate what went wrong in the text printed on stdout. Apparently, it seems its a default behavior of cobra. This issue is to track controlling the default stdout text printing behavior.

PR #43, opened 2019-03-22
Add SilenceUsage option to cobra command

nak3:
This patch adds `cmd.SilenceUsage` to command to stop usage output
after command validation.

Fixes https://github.com/knative/client/issues/38

### DIFF
--- a/pkg/kn/commands/revision_describe.go
+++ b/pkg/kn/commands/revision_describe.go
@@ -32,7 +32,6 @@ func NewRevisionDescribeCommand(p *KnParams) *cobra.Command {
 			if len(args) < 1 {
 				return errors.New("requires the revision name.")
 			}
-			cmd.SilenceUsage = true
 
 			client, err := p.ServingFactory()
 			if err != nil {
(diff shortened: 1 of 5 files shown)


nak3:
Perhaps the command func should have `Validate()` part in the future.

cppforlife:
@nak3 do we have to do it after validation?

nak3:
@cppforlife sorry, I think I did not understand your question. But do you mean `cmd.SilenceUsage` should be `before` the validation rather than `after`?

cppforlife:
@nak3 in knctl for example i set `SilenceUsage: true` on the root cmd (https://github.com/cppforlife/knctl/blob/3bcefbe5af3f809ecf5c1b04db9d5129b80e5984/pkg/knctl/cmd/knctl.go#L75). i m wondering if we can consolidate this configuration instead of placing it in all commands?

nak3:
@cppforlife I see. Yes, agree. Consolidating it insead of placing it in all commands is better.

PR #47, opened 2019-03-29
`service create` beginning implementation

sixolet:
This implements `kn service create` supporting only the `--image` and `--env` flags.

It also provides a structure for collecting flags that will be in common between `create`, `update`, and `replace`, and using them to modify a Configuration (so they can also be reused for `configuration create` etc later)

### DIFF
--- a/pkg/kn/commands/configuration_edit_flags.go
+++ b/pkg/kn/commands/configuration_edit_flags.go
@@ -14,10 +14,31 @@
 
 package commands
 
-import "github.com/spf13/cobra"
+import (
+	serving_lib "github.com/knative/client/pkg/serving"
+	servingv1alpha1 "github.com/knative/serving/pkg/apis/serving/v1alpha1"
+	"github.com/spf13/cobra"
+)
 
-func AddConfigurationEditFlags(command *cobra.Command, image *string, env *map[string]string) {
-	command.Flags().StringVar(image, "image", "", "Image to run.")
-	command.Flags().StringToStringVar(env, "env", map[string]string{},
+type ConfigurationEditFlags struct {
+	Image string
+	Env   map[string]string
+}
+
+func (p *ConfigurationEditFlags) AddFlags(command *cobra.Command) {
+	command.Flags().StringVar(&p.Image, "image", "", "Image to run.")
+	command.Flags().StringToStringVar(&p.Env, "env", map[string]string{},
 		"Environment, comma-separated NAME=value.")
+}
+
+func (p *ConfigurationEditFlags) Apply(config *servingv1alpha1.ConfigurationSpec) (err error) {
+	err = serving_lib.UpdateEnvVars(config, p.Env)
+	if err != nil {
+		return err
+	}
+	err = serving_lib.UpdateImage(config, p.Image)
+	if err != nil {
+		return err
+	}
+	return nil
 }
--- a/pkg/kn/commands/configuration_edit_flags.go
+++ b/pkg/kn/commands/configuration_edit_flags.go
@@ -18,7 +18,7 @@ import (
 	"fmt"
 	"strings"
 
-	serving_lib "github.com/knative/client/pkg/serving"
+	servinglib "github.com/knative/client/pkg/serving"
 	servingv1alpha1 "github.com/knative/serving/pkg/apis/serving/v1alpha1"
 	"github.com/spf13/cobra"
 )
@@ -35,7 +35,7 @@ func (p *ConfigurationEditFlags) AddFlags(command *cobra.Command) {
 			"any number of times to set multiple environment variables.")
 }
 
-func (p *ConfigurationEditFlags) Apply(config *servingv1alpha1.ConfigurationSpec) (err error) {
+func (p *ConfigurationEditFlags) Apply(config *servingv1alpha1.ConfigurationSpec) error {
 	envMap := map[string]string{}
 	for _, pairStr := range p.Env {
 		pairSlice := strings.SplitN(pairStr, "=", 2)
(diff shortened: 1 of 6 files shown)


pkg/kn/commands/configuration_edit_flags.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+type ConfigurationEditFlags struct {
+	Image string
+	Env   map[string]string
+}
+
+func (p *ConfigurationEditFlags) AddFlags(command *cobra.Command) {
+	command.Flags().StringVar(&p.Image, "image", "", "Image to run.")
+	command.Flags().StringToStringVar(&p.Env, "env", map[string]string{},
```
cppforlife: comma can be present in the values: https://github.com/cppforlife/knctl/issues/58

pkg/kn/commands/configuration_edit_flags.go
```diff
@@ -0,0 +1,44 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package commands
+
+import (
+	serving_lib "github.com/knative/client/pkg/serving"
```
cppforlife: it's not typical to have underscores in pkg aliases in go

pkg/kn/commands/configuration_edit_flags.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func (p *ConfigurationEditFlags) AddFlags(command *cobra.Command) {
+	command.Flags().StringVar(&p.Image, "image", "", "Image to run.")
+	command.Flags().StringToStringVar(&p.Env, "env", map[string]string{},
+		"Environment, comma-separated NAME=value.")
+}
+
+func (p *ConfigurationEditFlags) Apply(config *servingv1alpha1.ConfigurationSpec) (err error) {
+	err = serving_lib.UpdateEnvVars(config, p.Env)
```
cppforlife: can throw in `err :=` and kill off named return

pkg/kn/commands/service_create.go
```diff
@@ -0,0 +1,78 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	servingv1alpha1 "github.com/knative/serving/pkg/apis/serving/v1alpha1"
+
+	serving_lib "github.com/knative/client/pkg/serving"
+	"github.com/spf13/cobra"
+)
+
+func NewServiceCreateCommand(p *KnParams) *cobra.Command {
+
```
cppforlife: extra newline.

pkg/kn/commands/service_create.go
```diff
@@ -0,0 +1,78 @@
...
+			namespace := cmd.Flag("namespace").Value.String()
+
+			service := servingv1alpha1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      args[0],
+					Namespace: namespace,
+				},
+			}
+			service.Spec.RunLatest = &servingv1alpha1.RunLatestType{}
+			service.GetObjectKind().SetGroupVersionKind(schema.GroupVersionKind{
```
cppforlife: dont think this is (`SetGroupVersionKind(...)`) necessary as object is typed as servingv1alpha1.Service

pkg/serving/config_changes.go
```diff
@@ -0,0 +1,69 @@
...
+	servingv1alpha1 "github.com/knative/serving/pkg/apis/serving/v1alpha1"
+)
+
+// Give the configuration all the env var values listed in the given map of
+// vars.  Does not touch any environment variables not mentioned, but it can add
+// new env vars and change the values of existing ones.
+func UpdateEnvVars(config *servingv1alpha1.ConfigurationSpec, vars map[string]string) error {
+	set := make(map[string]bool)
+	for i, _ := range config.RevisionTemplate.Spec.Container.Env {
+		env_var := &config.RevisionTemplate.Spec.Container.Env[i]
```
cppforlife: camel case is preferred

sixolet: oops. Code switching too fast from py.

pkg/serving/config_changes.go
```diff
@@ -0,0 +1,69 @@
...
+		value, present := vars[env_var.Name]
+		if present {
+			env_var.Value = value
+			set[env_var.Name] = true
```
cppforlife: TODO assumes that env var name is not empty

pkg/serving/config_changes.go
```diff
@@ -0,0 +1,69 @@
...
+				})
+		}
+	}
+	return nil
+
+}
+
+// Utility function to translate between the API list form of env vars, and the
+// more convenient map form.
+func EnvToMap(vars []corev1.EnvVar) (map[string]string, error) {
```
cppforlife: seems like this could be moved into test file.

pkg/kn/commands/configuration_edit_flags.go
sixolet: Oh, damn. Somehow I expected this very-well-used flags lib to support some kind of quoting, but I guess not. Ugh, I'll use the StringArray version.

pkg/serving/config_changes.go
sixolet: It's useful enough I kinda want to leave it lib for now.

pkg/kn/commands/service_create.go
```diff
@@ -0,0 +1,78 @@
...
+
+func NewServiceCreateCommand(p *KnParams) *cobra.Command {
+
+	var editFlags ConfigurationEditFlags
+
+	serviceCreateCommand := &cobra.Command{
+		Use:   "create NAME",
+		Short: "Create a service.",
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			if len(args) < 1 {
```
cppforlife: `!=` instead of `<` ?

pkg/kn/commands/configuration_edit_flags.go
sixolet: Ok.

pkg/kn/commands/service_create.go
sixolet: Oh hey you're right.

pkg/serving/config_changes.go
sixolet: Can you explain? I think it works even with env var name empty, no?

Issue #40, opened 2019-03-13
Add more details in the output of `service list` command


navidshaikh:
`kn service list` lists only the names of services. It could add new columns in the output to print
additional information about the service.

navidshaikh:
```
./kn service list
NAME            DOMAIN                              LATESTCREATED         LATESTREADY
helloservice    helloservice.default.example.com    helloservice-vs6bm    helloservice-vs6bm
```

does this set of columns look okay ? Do we want to add more there ?

navidshaikh:
Added another column in the output of `service list` output

```
./kn service list
NAME            DOMAIN                              LATESTCREATED         LATESTREADY           AGE
helloservice    helloservice.default.example.com    helloservice-vs6bm    helloservice-vs6bm    1d
```

PR #56, opened 2019-04-04
Adds more columns in service get command output

navidshaikh:
This changeset adds basic details about service in
 service get command output.
 text/tabwriter is used to print columns on stdout.

 Fixes #https://github.com/knative/client/issues/40


pkg/kn/commands/service_list_test.go
```diff
@@ -79,22 +95,37 @@ func TestListDefaultOutput(t *testing.T) {
...
 	})
 	if err != nil {
 		t.Fatal(err)
 	}
-	expected := []string{"foo", "bar", ""}
+	// each line's tab/spaces are replaced by comma
+	expected := []string{"NAME,DOMAIN,LATESTCREATED,LATESTREADY,AGE",
+		"foo,foo.default.example.com,foo-abcde,foo-abcde,",
+		"bar,bar.default.example.com,bar-abcde,bar-abcde,"}
+	expected_lines := strings.Split(tabbedOutput(expected), "\n")
```
cppforlife: should use camel case (expectedLines)

pkg/util/printers/tabwriter.go
```diff
@@ -0,0 +1,34 @@
...
+const (
+	tabwriterMinWidth = 6
+	tabwriterWidth    = 4
+	tabwriterPadding  = 3
+	tabwriterPadChar  = ' '
+	tabwriterFlags    = tabwriter.TabIndent
+)
+
+// GetNewTabWriter returns a tabwriter that translates tabbed columns in input into properly aligned text.
+func GetNewTabWriter(output io.Writer) *tabwriter.Writer {
```
rhuss: Just would call it "NewTabWriter" in accordance with go conventions.

pkg/kn/commands/service_list.go
```diff
@@ -44,22 +45,42 @@ func NewServiceListCommand(p *KnParams) *cobra.Command {
...
 }
+
+func printServiceList(printer *tabwriter.Writer, services servingv1alpha1.ServiceList) error {
+	// case where no services are present
+	if len(services.Items) < 1 {
+		fmt.Fprintln(printer, "No resources found.")
+		return nil
+	}
+	columnNames := []string{"NAME", "DOMAIN", "LATESTCREATED", "LATESTREADY", "AGE"}
+	if _, err := fmt.Fprintf(printer, "%s\n", strings.Join(columnNames, "\t")); err != nil {
```
rhuss: What's about just having a printf format with customized columns widths? 

e.g. `"-20.20s %-10.10s %-10.10s %-10.10s %s\n"`

pkg/kn/commands/service_list.go
```diff
@@ -44,22 +45,42 @@ func NewServiceListCommand(p *KnParams) *cobra.Command {
...
+	}
+	for _, ksvc := range services.Items {
+		_, err := fmt.Fprintf(printer, "%s\n", strings.Join([]string{ksvc.Name, ksvc.Status.Domain,
+			ksvc.Status.LatestCreatedRevisionName, ksvc.Status.LatestReadyRevisionName,
+			printers.CalculateAge(ksvc.CreationTimestamp.Time)}, "\t"))
+		if err != nil {
+			return err
+		}
+	}
+	return nil
```
rhuss: You shouldn't forget to `printer.Flush()` to get the output in any case at that point.

navidshaikh: Right, `printer.Flush()` is invoked at https://github.com/knative/client/pull/56/files#diff-633efd289eb54521b1cdd9b31c7025c9R58 in calling function before final return.

navidshaikh:
@rhuss : Thanks for your review, each column width is re-adjusted based on the cell's max width (of that column). Hard-coding the padding and column width wont scale IMO.
For `printers.Flush()`, lets use `defer` to make sure it's called before returning. Sending an update to the PR.

pkg/kn/commands/service_list.go
navidshaikh: now using `defer`

pkg/util/printers/tabwriter.go
navidshaikh: fixed

rhuss:
> @rhuss : Thanks for your review, each column width is re-adjusted based on the cell's max width (of that column). Hard-coding the padding and column width wont scale IMO.

Yeah right. My fault, I've misread the documentation here. I think you are right, we should stay with this approach.

sixolet:
/hold
The previous approach for getting a printer enabled a lot of output flags shared with kubectl. This one doesn't seem to, if I read it correctly?

I think we should not implement our own table-writing code.

rhuss:
> The previous approach for getting a printer enabled a lot of output flags shared with kubectl. This one doesn't seem to, if I read it correctly?

I agree, that we should reuse the formatting machinery provided by cli-runtime for creating various output formats when specific options are provided.

> I think we should not implement our own table-writing code.

Sorry for my ignorance (just ramping up with the k8s codebase), but doesn't kubectl itself uses custom code for _human readable_ output ? I.e. in https://github.com/kubernetes/kubernetes/blob/c082ace102e217846833baace01d80966369ecd8/pkg/kubectl/cmd/get/get.go#L513 if `o.IsHumanReadablePrinter == true` then a tabWriter is used like in this PR.

I think we should combine both, but I'm afraid we have to write our own formatting code for the default console output, not covered by CLI options.

rhuss:
in kubectl, the tab-writer is used when this combination of options is true:

```golang
// human readable printers have special conversion rules, so we determine if we're using one.
	if (len(*o.PrintFlags.OutputFormat) == 0 && len(templateArg) == 0) || *o.PrintFlags.OutputFormat == "wide" {
		o.IsHumanReadablePrinter = true
	}
```

rhuss:
I'm a bit confused:  Both `kubernetes` itself and `cli-runtime` have a `PrintFlags` abstraction for obtaining a suitable Printer depending on context. But only Kubernetes itself has a [HumanReadableFlags](https://github.com/kubernetes/kubernetes/blob/6d7a9048b67d81e57923892650a23636f1afba42/pkg/kubectl/cmd/get/get_flags.go#L37) field for obtaining the printer for printing columnwise on the console (as known to `kubectl get`). `kubectl` doesn't use `cli-runtime` for the flags.

I'm pretty sure that the supported printers (JSON, YAML, Name, Template) are not suitable for direct console output.

@sixolet Any suggestion how to proceed? We could either implement a HumanReadableWriter on our own or just use a tabbed writer for the beginning (if no other options have been selected).

navidshaikh:
@rhuss : Thanks for the feedback on human readable output format and enabling the generic output format flags, I have updated the PR respectively.

Now the command supports printing  with generic output format flags. Excerpts
```
./kn service list --help
List available services.

Usage:
  kn service list [flags]

Flags:
      --all-namespaces                If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even if specified with --namespace.
  -h, --help                          help for list
  -o, --output string                 Output format. One of: json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-file.
[..]
```

and the `service list`
```
./kn service list
NAME   DOMAIN                   LATESTCREATED   LATESTREADY   AGE
s1     s1.default.example.com   s1-t4mbr        s1-t4mbr      5h
s2     s2.default.example.com   s2-8xrvv        s2-8xrvv      36m
```
and the service list with jsonpath
```
/kn service list -o "jsonpath={range .items[*]}{.metadata.name}{\"\\n\"}{end}"
s1
s2
```

pkg/kn/commands/service_list_test.go
navidshaikh: fixed

navidshaikh:
@sixolet : The previous approach for getting printer from cli-runtime is now restored. If no output flags are specified, the tabwriter is used to print the human-readable output.
PTAL
cc @cppforlife

navidshaikh:
Based on the recent discussions, renamed `service list` command to `service get`. `service get` with an argument should be a separate PR.

navidshaikh:
> I think we should not implement our own table-writing code.

@sixolet @cppforlife  What are other recommendations ?

I was checking https://github.com/kubernetes/kubernetes/tree/master/pkg/printers to see if we can re-use it, however this would add dependency on kube-versioned packages. I think we don't want that ?

pkg/util/age.go
```diff
@@ -0,0 +1,30 @@
...
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implie
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package printers
```
sixolet: pkg/util or pkg/util/printers?

pkg/util/age.go
```diff
@@ -0,0 +1,30 @@
+// Copyright © 2018 The Knative Authors
```
sixolet: I'm not in love with "util" as a name of things. it seems like the thing people name a bin where they stick things they don't know where else to put.

pkg/util/printers/tabwriter.go
```diff
@@ -0,0 +1,34 @@
+// Copyright © 2018 The Knative Authors
```
sixolet: I'd love pkg/output or pkg/printing or something.

pkg/kn/commands/service_get.go
```diff
@@ -0,0 +1,105 @@
...
+					Version: "v1alpha1",
+					Kind:    "Service"})
+				err = printer.PrintObj(service, cmd.OutOrStdout())
+				if err != nil {
+					return err
+				}
+				return nil
+			}
+			// if no output format flag is set, lets print the human readable output
+			printer := printers.NewTabWriter(cmd.OutOrStdout())
```
sixolet: This needs to be a `printer`, implementing the interface here: https://github.com/kubernetes/cli-runtime/blob/master/pkg/printers/interface.go#L34

navidshaikh: This is a solved problem for kubectl and the human-friendly printer implementation is in place already  https://github.com/kubernetes/kubernetes/blob/master/pkg/printers/tableprinter.go#L49 ,
however its not importable (or we don't want k8s dep) as genericclioptions.
A bunch of utilities can be re-used though from `pkg/printers` (re-use = borrow the code?), and we'll just need to add the column-definitions and their printing logic as done here
https://github.com/kubernetes/kubernetes/blob/master/pkg/printers/internalversion/printers.go#L79  for resources operable via kn.

WDYT? 
Cc: @bbrowning @rhuss

rhuss: @navidshaikh If there are no efforts underway to port that part to cli-runtime anyway, I would try to move that tableprinter stuff to `kn` in an isolated way and if this is successful, create a PR for cli-runtime so that we could use it from there eventually. It depends on the size of the code to move around and how easy it is to isolate it, but I think this could be a strategy from which everyone could benefit.

We could do it the other way round (first move it to cli-runtime), but this is (a) doesn't help for `kn` in the short term and (b) if we fail at cli-runtime, the effort is wasted. That's not the case with the first approach where the only danger is, that we would have to maintain the ported code on our own.

navidshaikh: @rhuss : makes sense, lets move ahead with first approach as you mentioned.

navidshaikh:
Closing in favor of #90

PR #57, opened 2019-04-04
 Update test-infra to use the new cluster_setup function

bbrowning:
This updates the version of test-infra to use the new cluster_setup function added in https://github.com/knative/test-infra/pull/667.
    
It simplifies our e2e setup logic into one place instead of being split over presubmit-tests.sh and e2e-tests.sh.

This builds on top of #55 and I'll rebase this to remove the duplicated commits and take it off hold after that one gets merged. I opened this as a separate PR just in case the updating of test-infra causes other issues to not hold up getting the integration tests in.

/hold


bbrowning:
CLosing this and merging into #55 since we can't actually have a successful E2E without an updated test-infra.

PR #59, opened 2019-04-04
Actually, vendor everything

sixolet:
We were getting errors when trying to build in CI when a backing repository 502'd, so let's vendor all our deps.


bbrowning:
With this change, it means go modules are no longer technically required to build the client repo. And, we should probably add `hack/update-deps.sh` and `hack/verify-codegen.sh` to both update the deps and let CI verify vendor doesn't drift from go.mod/go.sum. And we can likely remove our license checking hack from `test/presubmit-tests.sh`. So, a couple of scripts and an update to `DEVELOPMENT.md` indicating how to use them.

Adding a lgtm with a hold. I'm ok to merge this now and address the related changes in subsequent PRs, but punting that decision to you with the hold.

bbrowning:
Helps if I actually do the prow commands as well...

/lgtm
/hold

sixolet:
How much would this help you get e2e tests up and running if we landed it?

bbrowning:
This PR should solve the e2e flake issues as-is. After it lands I'll kick off tests on the e2e PR a few times to confirm.

sixolet:
According to https://github.com/golang/go/issues/27348 there is no good way to actually verify the contents of the vendor directory. Do you happen to know if I'm missing something?

I added a trivial script to go mod vendor things, and I think that might be the best compromise for now. What do you think?

bbrowning:
Thi seems like a good compromise. I'll poke at how we actually verify vendor in other Knative repos - I think they do some special trick with copying files to a tmp directory, vendoring, and comparing. But that can wait.

/lgtm

adrcunha:
> I'll poke at how we actually verify vendor in other Knative repos - I think they do some special trick with copying files to a tmp directory, vendoring, and comparing.

That's correct:

https://github.com/knative/serving/blob/master/hack/verify-codegen.sh

PR #64, opened 2019-04-05
Add hack/verify-codegen.sh so CI verifies vendor/ doesn't drift

bbrowning:
Yes, the file is called verify-codegen.sh, and yes in the other
Knative projects it both verifies generated code and verifies vendored
dependencies are up-to-date. For client, it just verifies vendored
dependencies don't drift.

The test-infra scripts specifically look for a file with this name. If
we'd prefer to call this something else, we'll also need to update
test-infra to look for that something else.

### DIFF
--- a/hack/update-deps.sh
+++ b/hack/update-deps.sh
@@ -1,4 +1,23 @@
+#!/usr/bin/env bash
 
-#!/bin/bash
+# Copyright 2019 The Knative Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


bbrowning:
The new verify-codegen.sh calls update-deps.sh and I noticed while doing that the license on update-deps.sh was missing. So, this adds that missing license as well.

bbrowning:
New output looks like this in build tests:
```
I0405 19:33:21.020] ---------------------------------------------------
I0405 19:33:21.020] ---- Checking autogenerated code is up-to-date ----
I0405 19:33:21.021] ---------------------------------------------------
...
I0405 19:33:22.625] Diffing /go/src/github.com/knative/client against freshly update dependencies
I0405 19:33:23.030] /go/src/github.com/knative/client up to date.
```

adrcunha:
Sweet
/lgtm

PR #69, opened 2019-04-11
adding hack/build* to make life easy and added version command

maximilien:
Getting my feet wet...

### DIFF
--- a/pkg/kn/commands/version.go
+++ b/pkg/kn/commands/version.go
@@ -1,0 +1,34 @@
+// Copyright © 2019 The Knative Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


pkg/kn/commands/version.go
```diff
@@ -0,0 +1,36 @@
...
+
+package commands
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+)
+
+const (
+	Version = "0.0.0"
```
rhuss: How is this version set ? Imo this should go into the build script via `-ldflags "-X commands.Version=0.1.0"` to `go build`.

maximilien: Yeah, we can do that. I'll add it. 

FYI... Still trying to get the CLA thingie done.

maximilien:
I signed it... I think

sixolet:
/approve
Waiting on lgtm for response to rhuss

maximilien:
Hi, @sixolet, thanks. The `-ldflags "-X commands.Version=0.1.0` did not work for me. Silently works but no version information is passed so the Version variable is set to default "" (empty string).

I investigated a bit and trying new Golang versions to see if that makes a difference. Will try a couple more things, if still not working I'll push the code so you all can see what I mean.

PR #74, opened 2019-04-15
Basic service update & a test for it.

sixolet:
Note that this only tests updating the image, since updating the env vars but _not_ the image is likely to want in the future the client convention of locking it to the image-by-digest currently running. That's for a future PR though.  This version of update is dead simple — fetch, modify, put. Rely on resourceVersion to keep us from clobbering someone else's edit.


cppforlife:
@sixolet looks good as a starting step, but i think we should add retry to update and re-apply modification on failure. ie `retry { get + modify + update }`

/lgtm
/approve

sixolet:
Yep. Just wanted to get the basic thing there :)

Issue #48, opened 2019-04-02
service get


csantanapr:
implement `kn service get <service_name>`

navidshaikh:
@csantanapr : what information about `<service_name>` should be printed in this command ?

How does this defer from `kn service list`?

If we can list the data points to be provided by CLI for these two commands, it should help.

rhuss:
I would like to suggest to combine `kn service list` and `kn service get` to a single command `kn service get` for the following reasons:

* It reduces the UX surface by having one subcommand less. This is IMO always a good thing. `kn service get` without argument can just list all services.
* It mimics the behaviour of `kubectl get` where `kubectl get` without argument just list all resources. So having only `kn service get` also follows the [_Principle of least astonishment_](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) which is a good thing.

Wdyt ?

If there are no objections, I'd like to start working on `kn service get`

rhuss:
Another argument for having a single 'kn service get' for listing and showing is a typical ux flow:

* List all service available
* Then: Select a single service to get details.

So you can call 'kn service get', check the output, cursor up for one step back in shell history and then just add the service of interest to the end of the command.

cppforlife:
@rhuss downside to combining list and get is that i imagine output format would be quite different. list i imagine is more compact whereas get would show more detailed information.

sixolet:
I like having different list and single-object describe commands. Get vs describe, no particluarly strong opinion about the single-object one, but I do like describe.

rhuss:
@sixolet +1, I think too that we should distinguish between `get` and `describe` a la kubectl. 

I think we could just rename `kn service list` to `kn service get` and allow an argument for filtering on the list. That would be similar to `kubectl get` (and hence less surprising).

`kn service describe` would be reserved then for a detail view of a single service.

Wdyt ?

Of course it depends how close we want to follow a UX like for kubectl. I think it would be good to stay close to its UX (`get` for list/summary like info, `describe` for details) because people know this already. On the other hand when changing to something new, then I would do it for all verbs. I.e. change also `describe` to something else, like `show` or `info`(so, `list` for list/summary and `show` for details, `new` for create).

rhuss:
> @rhuss downside to combining list and get is that i imagine output format would be quite different. list i imagine is more compact whereas get would show more detailed information.

@cppforlife I'm not sure whether the output format should be different, if we want to keep the semantics of `get` in kubectl (which actually is a list command there).

[...]

PR #75, opened 2019-04-17
feature(service describe): Output of service details [vanilla, color, hipster]

rhuss:
As discussed in #48 this `kn service describe` as changed in this commit
mimics `kubectl describe <sth>` as that it output detail information
in human readable output.

Summary information of a list of services or single services as well
as exporting the service in various formats (JSON, YAML) is reserved to
`kn service list` (or `kn service get` as suggested to align with
kubectl conventions).

For this reason, the generic Printer handling from cli-runtime has been
removed as `kn service describe` only outputs human readable format
(that is btw also true for `kubectl` describe which does not allow for
`-o` kind of options).

The presented information is just a start and the full content
should be discussed.

For the moment three modes for presenting information are implemented, which can be combined:

* `--all`    : Print more information. By default only are shorter summary is shown
* `--color`  : Use a colorful output (but only when on a tty)
* `--hipster`: Experimental output mode for a very compact representation using emojis

Below are some screenshot for different examples, totally open to discuss the format, and it's also ok if the consense is that the emoji thing goes over the top.

### `kn service describe random`

grafik

### `kn service describe random --all`

grafik

### `kn service describe random --color`

grafik

### `kn service describe random --all --color`

grafik

### `kn service describe random --hipster`

grafik

### `kn service describe random --hipster --all`

grafik

### `kn service describe random --hipster --all --color`

grafik


pkg/kn/commands/service_describe.go
```diff
@@ -16,53 +16,120 @@ package commands
...
+
+// TODO: Should be moved to util class to be reused across all human readable output ?
+func newDescribeWriter(output io.Writer) *describeWriter {
+	return &describeWriter{
+		writer: tabwriter.NewWriter(output, 6, 4, 3, ' ', tabwriter.TabIndent),
+	}
+}
+
+func (tw *describeWriter) sCol(label string, val string) {
+	fmt.Fprintf(tw.writer, "%s:\t%s\n", label, val)
```
navidshaikh: should use `tw.printf` here as we have defined it?

pkg/kn/commands/service_describe.go
```diff
@@ -16,53 +16,120 @@ package commands
...
+}
+
+// TODO: Should be moved to util class to be reused across all human readable output ?
+func newDescribeWriter(output io.Writer) *describeWriter {
+	return &describeWriter{
+		writer: tabwriter.NewWriter(output, 6, 4, 3, ' ', tabwriter.TabIndent),
+	}
+}
+
+func (tw *describeWriter) sCol(label string, val string) {
```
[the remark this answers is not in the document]
rhuss: good point. 'sCol' was meant to be 'singleColumn' (which actually is not true as there are two cols ;0)

let's call it `printlnWithLabel` ?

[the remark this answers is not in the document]
navidshaikh: `printlnWithLabel` is fine too, though I think soon we'll need a way to print tables, that should constitute a library for `kn`.
That library should re-use  `k8s.io/apimachinery/pkg/apis/meta/v1beta1.TableColumnDefinition` and re-write parts of `https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/get/table_printer.go`.

That way, describe <resource> output becomes a two column table with no-headers bool set.

pkg/kn/commands/service_describe.go
```diff
@@ -16,53 +16,120 @@ package commands
...
+	   - Traffic distribution
+	   - Timeout seconds
+	   - Conditions
+	   - Relevant Events (if any)
+	   - Annotations / Labels
+	   - Reference to configuration
+	*/
+}
+
+func getService(namespace string, name string, p *KnParams) (*v1alpha1.Service, error) {
```
navidshaikh: We'll be using GetService in multiple commands, this is also a target for including in some package. Fine keeping it here for now.

pkg/kn/commands/service_describe.go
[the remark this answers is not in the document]
rhuss: Sounds good to me. Maybe we can delay the switch until we decided *what* we actually want to show for `kn service describe`. Some suggestions are mentioned in the description of this PR.

If there are no objections, I would work on this list. We can change the fields presented any time anyway.

> That library should re-use `k8s.io/apimachinery/pkg/apis/meta/v1beta1.TableColumnDefinition` and re-write parts of [kubernetes/kubernetes:pkg/kubectl/cmd/get/table_printer.go@`master`](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/get/table_printer.go) .

I wonder whether the proper spot for such a library would be `cli-runtime` itself, where the other, machine-readable printers are living ?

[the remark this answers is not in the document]
navidshaikh: > Maybe we can delay the switch until we decided what we actually want to show for kn service describe

Yes! I just wanted pitch and get feedback. 

> I wonder whether the proper spot for such a library would be cli-runtime

Correct, however doing that would eventually increase the timeline for being able to re-use it via package in `kn`. I think we should first target it for `kn`, which needs it sooner and then propose for inclusion in cli-runtime.

navidshaikh:
> I could imagine to add the following additional fields in the output:

@rhuss : Lets proceed with the info you mentioned in the issue description.

cppforlife:
@rhuss any plans to address @navidshaikh comments?

rhuss:
@sixolet I'm going to pick up work this now again

@cppforlife yes, I will address @navidshaikh comments

As I'm currently on the road (conference), I think I will have an update here until the beginning of next week.

pkg/serving/client.go
```diff
@@ -0,0 +1,78 @@
...
+	revisionSlice := make([]*v1alpha1.Revision, len(revisions.Items))
+	for idx, _ := range revisions.Items {
+		revision := revisions.Items[idx].DeepCopy()
+		updateRevisionWithGVK(revision)
+		revisionSlice[idx] = revision
+	}
+	return revisionSlice, nil
+}
+
+// ============================================================================
```
maximilien: What's the meaning of these `//====`? To separate private functions/methods? If yes, I've been using simple `// Private funcs` or similar...

pkg/serving/client.go
```diff
@@ -0,0 +1,78 @@
...
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+type NamespacedClient struct {
+	client    client_v1alpha1.ServingV1alpha1Interface
+	namespace string
+}
+
+func NewNamespacedClient(client client_v1alpha1.ServingV1alpha1Interface, namespace string) *NamespacedClient {
```
maximilien: Might want to include single line comments for `/lint` so something like:

```go
// NewNamespacedClient creates a new client bound to namespace
func NewNamespacedClient(client client_v1alpha1.ServingV1alpha1Interface, namespace string) *NamespacedClient {
...
}
```

etc.

pkg/serving/client.go
rhuss: Yes, that's an old habit by me (to make it more visible). But happy to remove if this is too distracting.

pkg/serving/client.go
rhuss: Good point, I will add those. BTW, I plan to split up this monster PR into multiple PRs (based on each other), and extract eg. the part with this "NamespacedClient" in a separate PR and combine it with #134

pkg/serving/client.go
maximilien: No I am cool, though we should agree on one approach and use that consistently. Perhaps good to see other opinions on this...

duglin:
Rebase needed:

Overall I like the output of the default. I'm trying to think of a different word for "all" because in other contexts "all" typically means "more than one", but here you really mean "expanded view" or "non-compact".  Perhaps `--details`? not sure but let's bikeshed a bit on it  :-)
I'm not a fan of the hipster format - just seems a bit too cutsy and somewhat more confusing because people can easily interpret icons to mean something different than what was originally intended.

cppforlife:
i would suggest changing this a bit to something like this:

- have one view (instead of compact vs all) with a flag --revisions=x (x = 3 by default)
- kill hipster (that's my personal pref)
- figure out how to get --color to be global level configuration

rhuss:
Agreed that --all is probably not the right name. I like `--details` (or maybe `--verbose` if this is not already occupied ?)

Having an option for restricting the number of revisions is also a good idea, however, it would be cool to get all revisions listed without trial and error to find the proper number of `--revisions`.

I see the fruit salad format has not many fans, so happy to remove the --hipster option if this is the common consensus.

I also agree that `--color` should be a global option influencing all output.

If there are no objections, I will remove colour and emoji options, so that we can think on a more general level how to introduce colour.

duglin:
"fruit salad"  : https://www.youtube.com/watch?v=iRu1Yp2br00&t=33

rhuss:
Closing it for now, might serve as a reference if colour or emojis should ever come back.

bye, bye hipster 🤓

PR #76, opened 2019-04-17
chore(version): Add buildtime and git revision to version command

rhuss:
For that, also updated `hack/build.sh` and `hack/build-binaries.sh` to
set the values from the outside.

The file `hack/NEXT_VERSION` contains the version number of the next
release. A regular build will create version numbers in the format
"0.1.0-20190417" which according to semver.org is a version less
than 0.1.0 (i.e. a pre-release).

A release script should pick up NEXT_VERSION, make the release with
the number found there and then update to the next release, possibly
a minor a release.


hack/build-binaries.sh
```diff
@@ -14,10 +14,19 @@
...
-GOOS=darwin GOARCH=amd64 go build -o kn-darwin-amd64 ./cmd/...
-GOOS=linux GOARCH=amd64 go build -o kn-linux-amd64 ./cmd/...
-GOOS=windows GOARCH=amd64 go build -o kn-windows-amd64.exe ./cmd/...
+dir=$(dirname "${BASH_SOURCE[0]}")
+base=$(cd "$dir/.." && pwd)
+source ${base}/hack/util/flags.sh
 
-shasum -a 256 ./kn-*-amd64*
+flags=$(build_flags ${base}/hack)
+echo "🚧 🐧 Building for Linux"
```
matzew: :heart_eyes_cat:

pkg/kn/commands/version.go
```diff
@@ -20,13 +20,17 @@ import (
 )
 
 var Version string
+var Buildtime string
```
cppforlife: BuildTime?

hack/build-binaries.sh
```diff
@@ -14,10 +14,28 @@
...
+dir=$(dirname "${BASH_SOURCE[0]}")
+base=$(cd "$dir/.." && pwd)
+source ${base}/hack/util/flags.sh
+
+ld_flags="$(ld_flags ${base}/hack)"
+export GO111MODULE=on
+
+echo "🚧 🐧 Building for Linux"
+GOOS=darwin GOARCH=amd64 go build -mod=vendor -ldflags "${ld_flags}" -o ${base}/kn-darwin-amd64 ${base}/cmd/...
+echo "🚧 🍏 Building for macOs"
```
navidshaikh: probably macOS ?

PR #79, opened 2019-04-26
Adds --force flag to service create command (service replace)

### LINKED ISSUE #13: service replace
Write the command to replace a service! It should take all new arguments just like create!

navidshaikh:
Fixes #13 

Note the `--force` flag
```
Examples:

  # Create a service 'mysvc' using image at dev.local/ns/image:latest
  kn service create mysvc --image dev.local/ns/image:latest

  # Create a service with multiple environment variables
  kn service create mysvc --env KEY1=VALUE1 --env KEY2=VALUE2 --image dev.local/ns/image:latest

  # Replace a service 's1' with image dev.local/ns/image:v2 using --force flag
  kn service create --force s1 --image dev.local/ns/image:v2

  # Replace environment variables of service 's1' using --force flag

  # Reset resources to default ones of a service 's1' using --force flag
  # (earlier configured resource requests and limits will be replaced with default)
  # (earlier configured environment variables will be cleared too if any)
  kn service create --force s1 --image dev.local/ns/image:v1
```

### DIFF
--- a/pkg/kn/commands/configuration_edit_flags.go
+++ b/pkg/kn/commands/configuration_edit_flags.go
@@ -29,6 +29,7 @@ type ConfigurationEditFlags struct {
 	Image                      string
 	Env                        []string
 	RequestsFlags, LimitsFlags ResourceFlags
+	ForceCreate                bool
 }
 
 type ResourceFlags struct {
@@ -49,6 +50,7 @@ func (p *ConfigurationEditFlags) AddUpdateFlags(command *cobra.Command) {
 
 func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
 	p.AddUpdateFlags(command)
+	command.Flags().BoolVar(&p.ForceCreate, "force", false, "Create service forcefully, replaces existing service if any.")
 	command.MarkFlagRequired("image")
 }
 
--- a/pkg/kn/commands/service_create.go
+++ b/pkg/kn/commands/service_create.go
@@ -38,13 +38,14 @@ func NewServiceCreateCommand(p *KnParams) *cobra.Command {
   # Create a service with multiple environment variables
   kn service create mysvc --env KEY1=VALUE1 --env KEY2=VALUE2 --image dev.local/ns/image:latest
 
-  # Replace a service 's1' with image dev.local/ns/image:v2 using --force flag
+  # Create or replace a service 's1' with image dev.local/ns/image:v2 using --force flag
+  # if service 's1' doesn't exist, it's just a normal create operation
   kn service create --force s1 --image dev.local/ns/image:v2
 
-  # Replace environment variables of service 's1' using --force flag
+  # Create or replace environment variables of service 's1' using --force flag
   kn service create --force s1 --env KEY1=NEW_VALUE1 --env NEW_KEY2=NEW_VALUE2 --image dev.local/ns/image:v1
 
-  # Reset resources to default ones of a service 's1' using --force flag
+  # Create or replace default resources of a service 's1' using --force flag
   # (earlier configured resource requests and limits will be replaced with default)
   # (earlier configured environment variables will be cleared too if any)
   kn service create --force s1 --image dev.local/ns/image:v1`,
@@ -82,17 +83,31 @@ func NewServiceCreateCommand(p *KnParams) *cobra.Command {
 			if err != nil {
 				return err
 			}
+			var serviceExists bool = false
 			// check if --force flag is given
-			if force, err := cmd.Flags().GetBool("force"); err != nil {
-				return err
-			} else if force {
-				client.Services(namespace).Delete(args[0], &v1.DeleteOptions{})
+			if editFlags.ForceCreate {
+				// --force flag is given, lets check if the service exists
+				existingService, err := client.Services(namespace).Get(args[0], v1.GetOptions{})
+				if err == nil {
+					serviceExists = true
+					// copy over the resource version
+					service.ResourceVersion = existingService.ResourceVersion
+					// lets update the service
+					_, err = client.Services(namespace).Update(&service)
+					if err != nil {
+						return err
+					}
+					fmt.Fprintf(cmd.OutOrStdout(), "Service '%s' successfully replaced in namespace '%s'.\n", args[0], namespace)
+				}
 			}
-			_, err = client.Services(namespace).Create(&service)
-			if err != nil {
-				return err
+			// if service doesn't exist, perform a normal create operation
+			if !serviceExists {
+				_, err = client.Services(namespace).Create(&service)
+				if err != nil {
+					return err
+				}
+				fmt.Fprintf(cmd.OutOrStdout(), "Service '%s' successfully created in namespace '%s'.\n", args[0], namespace)
 			}
-			fmt.Fprintf(cmd.OutOrStdout(), "Service '%s' successfully created in namespace '%s'.\n", args[0], namespace)
 			return nil
 		},
 	}
--- a/pkg/kn/commands/service_create.go
+++ b/pkg/kn/commands/service_create.go
@@ -84,15 +84,11 @@ func NewServiceCreateCommand(p *KnParams) *cobra.Command {
 				return err
 			}
 			var serviceExists bool = false
-			// check if --force flag is given
 			if editFlags.ForceCreate {
-				// --force flag is given, lets check if the service exists
 				existingService, err := client.Services(namespace).Get(args[0], v1.GetOptions{})
 				if err == nil {
 					serviceExists = true
-					// copy over the resource version
 					service.ResourceVersion = existingService.ResourceVersion
-					// lets update the service
 					_, err = client.Services(namespace).Update(&service)
 					if err != nil {
 						return err
@@ -100,7 +96,6 @@ func NewServiceCreateCommand(p *KnParams) *cobra.Command {
 					fmt.Fprintf(cmd.OutOrStdout(), "Service '%s' successfully replaced in namespace '%s'.\n", args[0], namespace)
 				}
 			}
-			// if service doesn't exist, perform a normal create operation
 			if !serviceExists {
 				_, err = client.Services(namespace).Create(&service)
 				if err != nil {
--- a/pkg/kn/commands/service_create.go
+++ b/pkg/kn/commands/service_create.go
@@ -22,6 +22,7 @@ import (
 	servingv1alpha1 "github.com/knative/serving/pkg/apis/serving/v1alpha1"
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func NewServiceCreateCommand(p *KnParams) *cobra.Command {
@@ -35,7 +36,18 @@ func NewServiceCreateCommand(p *KnParams) *cobra.Command {
   kn service create mysvc --image dev.local/ns/image:latest
 
   # Create a service with multiple environment variables
-  kn service create mysvc --env KEY1=VALUE1 --env KEY2=VALUE2 --image dev.local/ns/image:latest`,
+  kn service create mysvc --env KEY1=VALUE1 --env KEY2=VALUE2 --image dev.local/ns/image:latest
+
+  # Replace a service 's1' with image dev.local/ns/image:v2 using --force flag
+  kn service create --force s1 --image dev.local/ns/image:v2
+
+  # Replace environment variables of service 's1' using --force flag
+  kn service create --force s1 --env KEY1=NEW_VALUE1 --env NEW_KEY2=NEW_VALUE2 --image dev.local/ns/image:v1
+
+  # Reset resources to default ones of a service 's1' using --force flag
+  # (earlier configured resource requests and limits will be replaced with default)
+  # (earlier configured environment variables will be cleared too if any)
+  kn service create --force s1 --image dev.local/ns/image:v1`,
 
 		RunE: func(cmd *cobra.Command, args []string) (err error) {
 			if len(args) != 1 {
@@ -69,6 +81,12 @@ func NewServiceCreateCommand(p *KnParams) *cobra.Command {
 			client, err := p.ServingFactory()
 			if err != nil {
 				return err
+			}
+			// check if --force flag is given
+			if force, err := cmd.Flags().GetBool("force"); err != nil {
+				return err
+			} else if force {
+				client.Services(namespace).Delete(args[0], &v1.DeleteOptions{})
 			}
 			_, err = client.Services(namespace).Create(&service)
 			if err != nil {
--- a/pkg/kn/commands/service_create_test.go
+++ b/pkg/kn/commands/service_create_test.go
@@ -290,3 +290,66 @@ func parseQuantity(t *testing.T, quantityString string) resource.Quantity {
 	}
 	return quantity
 }
+
+func TestServiceCreateImageForce(t *testing.T) {
+	_, _, _, err := fakeServiceCreate([]string{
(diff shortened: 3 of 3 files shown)


pkg/kn/commands/configuration_edit_flags.go
```diff
@@ -82,6 +80,38 @@ func (p *ConfigurationEditFlags) Apply(config *servingv1alpha1.ConfigurationSpec
 	return nil
 }
 
+// Replace resets previous ConfigurationSpec and configures the provided one
+func (p *ConfigurationEditFlags) Replace(config *servingv1alpha1.ConfigurationSpec) error {
```
sixolet: I think the replace command should work more like `create` — don't bother blanking things and setting them to new things, just create a new service resource as if you were creating it anew.

pkg/kn/commands/configuration_edit_flags.go
```diff
@@ -45,21 +45,19 @@ func (p *ConfigurationEditFlags) AddFlags(command *cobra.Command) {
...
-	for _, pairStr := range p.Env {
-		pairSlice := strings.SplitN(pairStr, "=", 2)
-		if len(pairSlice) <= 1 {
-			return fmt.Errorf(
-				"--env argument requires a value that contains the '=' character; got %s",
-				pairStr)
-		}
-		envMap[pairSlice[0]] = pairSlice[1]
+	// lets get the map of env vars provided
+	envMap, err := p.envStrToMap(p.Env)
```
sixolet: I like this refactor

pkg/kn/commands/service_replace.go
```diff
@@ -0,0 +1,77 @@
...
+			}
+			namespace, err := GetNamespace(cmd)
+			if err != nil {
+				return err
+			}
+			client, err := p.ServingFactory()
+			if err != nil {
+				return err
+			}
+			service, err := client.Services(namespace).Get(args[0], v1.GetOptions{})
```
sixolet: Yeah, no need to get it here.

pkg/kn/commands/service_replace.go
```diff
@@ -0,0 +1,77 @@
...
+			}
+			config, err := knserving.GetConfiguration(service)
+			if err != nil {
+				return err
+			}
+			err = editFlags.Replace(config)
+			if err != nil {
+				return err
+			}
+			_, err = client.Services(namespace).Update(service)
```
sixolet: ... unless it's to get the resourceVersion and blat that onto the *new* Service object we use to update.

pkg/serving/config_changes.go
```diff
@@ -48,6 +48,21 @@ func UpdateEnvVars(config *servingv1alpha1.ConfigurationSpec, vars map[string]st
 
 }
 
+// ReplaceEnvVars clears any env vars configured in ConfigurationSpec and configures
+// the provided env vars
```
sixolet: I think the changes I'm suggesting make this file not really need to change at all.

pkg/kn/commands/configuration_edit_flags.go
navidshaikh: @sixolet : I think I didn't get this. 

>  just create a new service resource as if you were creating it anew.

How does this is different than `service create` then?
Or
What are the scenarios  a dev would use `service replace` over `service create` ?

cppforlife: create doesnt allow to "update" resource if it doesnt exist. though i could see extending create to allow that functionality via a flag instead of having a dedicated command.

rhuss: So `kn service replace` is similar to `kubectl apply` when using a resource file, right ?  (except maybe that `kubectl apply` also allows a create)

* It correlates over the resource `name`
* It replaces the existing resource but only if there are not immutable fields which can't be removed

In this simple case, it's just a kind of combination of `delete` followed by a `create`. This works nicely for top-level resources. I think we should follow the same semantic to not allow a "replace" (aks "apply") if there is an immutable field set (though I currently can't think about one in the `service`).

For possibly auto-created objects like revisions or configuration, we should check that this by verifying the `ownerReference` and forbid a replace is that case (only for non-auto generated objects I would allow a replace).

------

@sixolet Having a second look at the example in the issue description, I'm now not sure anymore whether I get it, too, when you say

> I think the replace command should work more like create — don't bother blanking things and setting them to new things, just create a new service resource as if you were creating it anew.

and how it applies to e.g.

```
kn service replace s1 --env KEY1=NEW_VALUE1 --env NEW_KEY2=NEW_VALUE2
```

In this case, where should the `image` come from then ? Shouldn't it be just taken from the original service? And if so, this is really not different to a`service update` ? Or asked otherwise, which fields would you expect to be blank / resetted after this operation ? This question applies to the other usage examples in the issue description, too.

@cppforlife I agree that if this command all about a "set" of a resource with all fields to set given as as parameters, its probably a better choice to use a `kn service create --force` to allow an overwrite of an existing resource (with the minor caveat that this would not validate that a resource already exists like a `kn service replace` probably would do)

sixolet: Replace should require `--image`, and should always remove all currently-set env vars, along with limits, etc, anything not metnioned explicitly.

Feel free to impl a --force flag on create instead, I'm not super picky about the wording.

rhuss: Cool, I think in order to keep the UX surface small (and to avoid confusions between `create`, `replace` and `update`) I opt for a `--force` option to create as `create` already has to check for all mandatory parameters for creating a service (like the image).

navidshaikh: @rhuss @cppforlife @sixolet : Okay, I have added `--force` flag to `service create` command instead, PTAL.

pkg/kn/commands/service_create.go
```diff
@@ -69,11 +81,17 @@ func NewServiceCreateCommand(p *KnParams) *cobra.Command {
 			if err != nil {
 				return err
 			}
+			// check if --force flag is given
+			if force, err := cmd.Flags().GetBool("force"); err != nil {
+				return err
+			} else if force {
+				client.Services(namespace).Delete(args[0], &v1.DeleteOptions{})
```
rhuss: What happens if the service does not exist ? In this case, no delete should be attempted.

pkg/kn/commands/service_create.go
```diff
@@ -34,7 +35,18 @@ func NewServiceCreateCommand(p *KnParams) *cobra.Command {
   kn service create mysvc --image dev.local/ns/image:latest
 
   # Create a service with multiple environment variables
-  kn service create mysvc --env KEY1=VALUE1 --env KEY2=VALUE2 --image dev.local/ns/image:latest`,
+  kn service create mysvc --env KEY1=VALUE1 --env KEY2=VALUE2 --image dev.local/ns/image:latest
+
+  # Replace a service 's1' with image dev.local/ns/image:v2 using --force flag
```
rhuss: Maybe also mention that if "s1" does not exists, its just a normal create:

```
# Create or replace a service 's1' ....
```

pkg/kn/commands/configuration_edit_flags.go
```diff
@@ -49,6 +50,7 @@ func (p *ConfigurationEditFlags) AddUpdateFlags(command *cobra.Command) {
 
 func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
 	p.AddUpdateFlags(command)
+	command.Flags().BoolVar(&p.ForceCreate, "force", false, "Create service forcefully, replaces existing service if any.")
```
rhuss: "Overwrite an existing service with the same name instead of throwing an error"

Something liket that, because I think it's important to mention, what a "existing service" means (i.e. an existing service with the _same_ name)

pkg/kn/commands/service_create.go
```diff
@@ -69,11 +81,17 @@ func NewServiceCreateCommand(p *KnParams) *cobra.Command {
...
+				return err
+			} else if force {
+				client.Services(namespace).Delete(args[0], &v1.DeleteOptions{})
+			}
 			_, err = client.Services(namespace).Create(&service)
 			if err != nil {
 				return err
 			}
-
+			fmt.Fprintf(cmd.OutOrStdout(), "Service '%s' is successfully created in namespace '%s'.\n", args[0], namespace)
```
rhuss: `Service '%s' successfully created in namespace '%s'` (i.e. without "is")

pkg/kn/commands/service_create_test.go
```diff
@@ -287,3 +287,66 @@ func parseQuantity(t *testing.T, quantityString string) resource.Quantity {
...
+	expectedOutput := "Service 'foo' is successfully created in namespace 'default'.\n"
+	if err != nil {
+		t.Fatal(err)
+	} else if conf.RevisionTemplate.Spec.Container.Image != "gcr.io/foo/bar:v2" {
+		t.Fatalf("wrong image set: %v", conf.RevisionTemplate.Spec.Container.Image)
+	} else if !reflect.DeepEqual(
+		actualEnvVars,
+		expectedEnvVars) {
+		t.Fatalf("wrong env vars:%v", conf.RevisionTemplate.Spec.Container.Env)
+	} else if output != expectedOutput {
```
rhuss: More a personal opinion: I never would match on the exact string (with punctuation and what not). Instead its IMO much better to check whether all important context information is contained.

This makes the test more robust and more valuable as it checks the important parts (especially for error message a much context should be contained as possible).

In this case, I suggest to check that the "foo" and "default" is included and that it contains "created" and "service" somewhere.

pkg/kn/commands/service_create.go
navidshaikh: We can check that we `Get` call, here I am not error checking for for `Delete` though, since error is not handled, it creates the service.
@rhuss : Do you have some tweak in mind for this operation ?

sixolet: Don't delete the service here at all. Instead, do a GET, if it exists copy the resourceVersion to &service, and issue an update.

Why? This preserves revision history.

pkg/kn/commands/service_create.go
sixolet: If we replaced it, say so in the string too

pkg/kn/commands/service_create_test.go
sixolet: +1

pkg/kn/commands/service_create.go
rhuss: Using an update instead of a delete/create would also help with the transactional issue I mentioned (if "delete" works but not "create"), because when the "update" fails you still have the situation as before (e.g. the service running).

The only issue I then have is, that this is not really a `create`, as I think people would expect a clean history for a create. That would again a plus for a 'replace' command or 'update --set' or something. 

For my understanding we should implement:

* `kn service create --force` to create a fresh service without history by deleting any old service of this name. No error if the service already exists.
* `kn service create --replace` for replacing a service, keeping the history, but erroring if the service does not exist.

An alternative for the "replace with keeping history" would be `kn service replace` or `kn update --set`.

Sorry for fueling that discussion again, but I really think that 'create afresh by potentially overwriting an existing service' (good for development) and 'replace everything, potentially with default values, but keep the history' (good for getting back to a defined and known state) are looking similar but are subtly different.

pkg/kn/commands/service_create.go
```diff
@@ -69,11 +82,31 @@ func NewServiceCreateCommand(p *KnParams) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			_, err = client.Services(namespace).Create(&service)
-			if err != nil {
-				return err
+			var serviceExists bool = false
+			// check if --force flag is given
```
cppforlife: unnecessary comment.

pkg/kn/commands/service_create.go
```diff
@@ -69,11 +82,31 @@ func NewServiceCreateCommand(p *KnParams) *cobra.Command {
...
+			if editFlags.ForceCreate {
+				// --force flag is given, lets check if the service exists
+				existingService, err := client.Services(namespace).Get(args[0], v1.GetOptions{})
+				if err == nil {
+					serviceExists = true
+					// copy over the resource version
+					service.ResourceVersion = existingService.ResourceVersion
+					// lets update the service
```
cppforlife: unnecessary comment.

pkg/kn/commands/service_create.go
```diff
@@ -69,11 +82,31 @@ func NewServiceCreateCommand(p *KnParams) *cobra.Command {
...
-			if err != nil {
-				return err
+			var serviceExists bool = false
+			// check if --force flag is given
+			if editFlags.ForceCreate {
+				// --force flag is given, lets check if the service exists
+				existingService, err := client.Services(namespace).Get(args[0], v1.GetOptions{})
+				if err == nil {
+					serviceExists = true
+					// copy over the resource version
```
cppforlife: unnecessary comment.

pkg/kn/commands/service_create.go
```diff
@@ -69,11 +82,31 @@ func NewServiceCreateCommand(p *KnParams) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			_, err = client.Services(namespace).Create(&service)
-			if err != nil {
-				return err
+			var serviceExists bool = false
+			// check if --force flag is given
+			if editFlags.ForceCreate {
+				// --force flag is given, lets check if the service exists
```
cppforlife: unnecessary comment.

pkg/kn/commands/service_create.go
```diff
@@ -69,11 +82,31 @@ func NewServiceCreateCommand(p *KnParams) *cobra.Command {
...
+					service.ResourceVersion = existingService.ResourceVersion
+					// lets update the service
+					_, err = client.Services(namespace).Update(&service)
+					if err != nil {
+						return err
+					}
+					fmt.Fprintf(cmd.OutOrStdout(), "Service '%s' successfully replaced in namespace '%s'.\n", args[0], namespace)
+				}
+			}
+			// if service doesn't exist, perform a normal create operation
```
cppforlife: unnecessary comment.

pkg/kn/commands/configuration_edit_flags.go
```diff
@@ -29,6 +29,7 @@ type ConfigurationEditFlags struct {
 	Image                      string
 	Env                        []string
 	RequestsFlags, LimitsFlags ResourceFlags
+	ForceCreate                bool
```
cppforlife: i wonder if at some point it would make sense to have 

```
type ConfigurationCreateFlags struct {
  ConfigurationEditFlags
  ForceCreate bool
  ...
}
```

to better separate create from update at the compile time. wdyt?

pkg/kn/commands/service_create.go
navidshaikh: > kn service create --replace for replacing a service, keeping the history, but erroring if the service does not exist.

@rhuss ; should it really bother for erroring out ? as in, IMO the `create` verb/command shouldn't block on existence of service, though we could put a proper info on stdout if it existed or not.

pkg/kn/commands/configuration_edit_flags.go
navidshaikh: +1
@cppforlife : Do you want me to update the same PR? or log an issue and address is post v0.1 ?

cppforlife: @navidshaikh up to you on the timeline. i think current implementation is fine.

cppforlife:
/approve
/lgtm

there is potential race between other tools running create, but for now this should be fine since cli will just error.

pkg/kn/commands/configuration_edit_flags.go
navidshaikh: Logged the suggestion to be addressed later https://github.com/knative/client/issues/110 , thanks @cppforlife 

Also rebased the PR to resolve the conflicts.

PR #90, opened 2019-05-03
Adds service get command

### LINKED ISSUE #40: Add more details in the output of `service list` command

navidshaikh:
Fixes #40 

```
./kn service get --all-namespaces
NAME   DOMAIN                   GENERATION   AGE     CONDITIONS   READY       REASON
s1     s1.default.example.com   4            2d19h   3 OK / 3     True        
s2     s2.default.example.com   1            17h     0 OK / 3     False       RevisionMissing : Configuration "s2" does not have any ready Revision.
```


pkg/printers/printers.go
```diff
@@ -0,0 +1,131 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1beta1 "k8s.io/apimachinery/pkg/apis/meta/v1beta1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/duration"
+	"time"
+)
+
+// AddHandlers adds print handlers for knative resources
+func AddHandlers(h PrintHandler) {
+	kServiceColumnDefinitions := []metav1beta1.TableColumnDefinition{
```
cppforlife: can we "place" kServiceColumnDefinitions closer to the service get command. it seems that pkg/printers package is meant to be generic and it shouldnt have specific knowledge of knative service columns.

navidshaikh: @cppforlife: I think yes, we can place it in `service_get_flags.go`. Another option is to create `internalversion` pkg in `pkg/printers/` and have `printers.go` file with columns and their printer functions (this is how its done in for kubectl). 
I think moving the columns closer to service get command make sense, we'll need to add the handlers for individual commands in `ToPrinter` method though.

The number of files in `pkg/kn/commands` is increasing, we'll soon be requiring some refactoring to have individual command's sub-packages (as its done in knctl).

pkg/kn/commands/service_get_flags.go
```diff
@@ -0,0 +1,182 @@
...
+	var ok int
+	for _, condition := range conditions {
+		if condition.Status == "True" {
+			ok++
+		}
+	}
+	return fmt.Sprintf("%d OK / %d", ok, len(conditions))
+}
+
+// readyCondition returns status of resource's Ready type condition
```
navidshaikh: Following few functions should be grouped together in some core lib to re-used in other resource listing. I plan to regroup them post v0.1 part of some refactoring.

navidshaikh:
@cppforlife : Addressed your review comments PTAL, if the structure looks good, I will add some unit tests.

pkg/kn/commands/humanreadable_flags.go
```diff
@@ -0,0 +1,54 @@
...
+	// TODO: Add more formats eg: wide
+	return []string{""}
+}
+
+// ToPrinter receives returns a printer capable of
+// handling human-readable output.
+func (f *HumanPrintFlags) ToPrinter() (hprinters.ResourcePrinter, error) {
+	p := hprinters.NewTablePrinter(hprinters.PrintOptions{})
+	// Add the column definitions and respective functions
+	ServiceGetHandlers(p)
```
cppforlife: would be just keep on adding *Handlers here for other command. will they conflict?

navidshaikh: @cppforlife: Nope, I don't think so. To confirm, wrote `revision get` command too and added the handler https://github.com/knative/client/pull/97/files#diff-293a37e783b8135a790583552133ca27R42

pkg/kn/commands/humanreadable_flags.go
```diff
@@ -0,0 +1,54 @@
+// Copyright © 2019 The Knative Authors
```
sixolet: Can we make this file called human_readable_flags.go, or humanreadableflags.go?

pkg/kn/commands/service_get_flags.go
sixolet: +1

rhuss:
After analysing the test errors, I think we have a more serious issue here when we want to use the `pkg/printers` from `cli-runtime`:

* `pkg/printers` were only introduced in `cli-runtime` since kubernetes-1.14.0
* `knative/serving` depends on kubernetes-1.12.6
* `client-go` (used by both, `serving` and `cli-runtime`) is incompatible between Kubernetes 1.12 and 1.14 (that's indicated in the CI test error, where client-go `latest` is used).

The two options, to try cli-runtime with client-go 1.12 or to try knative-serving with client-go 1.14 both failed.

Afais, we have two options if we want to reuse the functionality of the printers package in cli-runtime:

* Convince knative-serving to upgrade to Kubernetes 1.14. This is probably not something which can be done quickly.
* Copy over `printers` package from cli-runtime here temporarily, and switch over to cli-runtim when knative-serving has upgraded.

But maybe there are other solutions ? 

Maybe we should also reconsider the re-usage of `cli-runtime` as this shared dependency (both cli-runtime and knative are depending on client-go) will likely cause issues in the future again, considering the different lifecycles between knative and kubernetes (and hence cli-runtime).

@cppforlife @sixolet wdyt ?

pkg/kn/commands/humanreadable_flags.go
navidshaikh: done

navidshaikh:
@rhuss : I think there was a mess up in for vendors files due to multiple rebases (this PR was rebased onto another one, which is now merged). I reset hard to point to master and added / updated the files in this PR to create fresh commits, the failure is gone!

rhuss:
@navidshaikh I think so, too. I was a bit confused about the reference to `cli-runtime/printers` as I thought this was a hard dependency for the knative client code itself. However, this doesn't seems to be the case. Also it seems that the `cli-runtime/genericclioptions/printers` in 1.12 has been moved to `cli-runtime/printers` in 1.14 (I overlooked that).

So, if there is no direct usage of 1.14 `printers`, then there is also no two client-go vesions required.

Looks all good to me now, and sorry for the noise ...

pkg/kn/commands/service_get_flags.go
```diff
@@ -0,0 +1,181 @@
...
+// AllowedFormats is the list of formats in which data can be displayed
+func (f *ServiceGetFlags) AllowedFormats() []string {
+	formats := f.GenericPrintFlags.AllowedFormats()
+	formats = append(formats, f.HumanReadableFlags.AllowedFormats()...)
+	return formats
+}
+
+// ToPrinter attempts to find a composed set of ServiceGetFlags suitable for
+// returning a printer based on current flag values.
+func (f *ServiceGetFlags) ToPrinter() (hprinters.ResourcePrinter, error) {
```
navidshaikh: @rhuss : There is one thing that I have updated, changed the reference for `ResourcePrinter` from `printers.ResourcePrinter` defined [here](https://github.com/kubernetes/cli-runtime/blob/master/pkg/printers/interface.go#L34) to `hprinters.ResourcePrinter` defined in client [here](https://github.com/knative/client/blob/master/pkg/printers/interface.go#L26).

rhuss: Yes, that was it. As long as you don't have a reference into 'cli-runtime/pkg/printers' anymore, there is no need for 'go modules' to add a newer version of cli-runtime (i.e. 1.14) which conflicts with serving 0.5.2 (and k8s 1.12).

So this is actually the important change and of course a better fix then copying over stuff.

PR #95, opened 2019-05-08
Adds info message after service create and delete operations

### LINKED ISSUE #87: Add a message on stdout after service create
`kn service create` returns without any hint, lets add a message about service creation status.
For eg:
Created <SERVICE_NAME> in <NAMESPACE> namespace.

navidshaikh:
Fixes #87

### DIFF
--- a/pkg/kn/commands/service_create.go
+++ b/pkg/kn/commands/service_create.go
@@ -16,7 +16,7 @@ package commands
 
 import (
 	"errors"
-
+	"fmt"
 	serving_lib "github.com/knative/client/pkg/serving"
 	servingv1alpha1 "github.com/knative/serving/pkg/apis/serving/v1alpha1"
 	"github.com/spf13/cobra"
@@ -73,7 +73,7 @@ func NewServiceCreateCommand(p *KnParams) *cobra.Command {
 			if err != nil {
 				return err
 			}
-
+			fmt.Fprintf(cmd.OutOrStdout(), "Service '%s' successfully created in namespace '%s'.\n", args[0], namespace)
 			return nil
 		},
 	}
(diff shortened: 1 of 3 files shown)


pkg/kn/commands/service_create.go
```diff
@@ -73,7 +73,7 @@ func NewServiceCreateCommand(p *KnParams) *cobra.Command {
 			if err != nil {
 				return err
 			}
-
+			fmt.Fprintf(cmd.OutOrStdout(), "Service %s is created in %s namespace.\n", args[0], namespace)
```
cppforlife: i would recommend adding single quotes around names. also consistently putting name after resource type. something like this:

```
Service '%s' was successfully created in namespace '%s'.\n
```

navidshaikh: Ok, updated the PR, also updated the info message for `service delete` too.

pkg/kn/commands/service_create.go
```diff
@@ -16,7 +16,7 @@ package commands
 
 import (
 	"errors"
-
+	"fmt"
```
cppforlife: i think we keeping newline after std imports.

pkg/kn/commands/service_create_test.go
```diff
@@ -79,6 +79,9 @@ func TestServiceCreateImage(t *testing.T) {
 		t.Fatal(err)
 	} else if conf.RevisionTemplate.Spec.Container.Image != "gcr.io/foo/bar:baz" {
 		t.Fatalf("wrong image set: %v", conf.RevisionTemplate.Spec.Container.Image)
+	} else if !strings.Contains(output, "foo") || !strings.Contains(output, "default") || !strings.
```
cppforlife: pretty awkward to keep `strings.` on one line. lets end this line at `||`

Issue #89, opened 2019-05-03
Every error is printed twice


navidshaikh:
```
kn service list
Error: Get https://192.168.39.212:8443/apis/serving.knative.dev/v1alpha1/services: dial tcp 192.168.39.212:8443: connect: no route to host
Get https://192.168.39.212:8443/apis/serving.knative.dev/v1alpha1/services: dial tcp 192.168.39.212:8443: connect: no route to host
```

maximilien:
This sucks and I think has to do with the fact that we don't wrap errors. My hunch. Did you plan to fix this @navidshaikh? Otherwise I can take a crack at it. LMK.

maximilien:
Sweet @navidshaikh, will do. I'll update here if there are items to discuss before PR. Cheers 🍻

maximilien:
Actually it's a simpler change. Submitting PR now.

navidshaikh:
Fixed by #96 , thanks @maximilien !

PR #96, opened 2019-05-08
Set `SilenceErrors: true` on `rootCmd` so that we can deal with errors ourselves

maximilien:
Fixes issue #89

### DIFF
--- a/pkg/kn/commands/root.go
+++ b/pkg/kn/commands/root.go
@@ -68,6 +68,9 @@ Eventing: Manage event subscriptions and channels. Connect up event sources.`,
 
 		// Affects children as well
 		SilenceUsage: true,
+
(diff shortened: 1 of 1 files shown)


cppforlife:
where is the printing coming from? i think we should keep consistent top level error printing and remove dup from somewhere else.

maximilien:
Actually the issue is that Cobra is automatically dealing with errors by default. See latest commit. I silence errors from Cobra and deal with them globally again in `main.go`.

maximilien:
Tests failures seem to come from tests servers giving `404` for Istio. See sample output:

```
I0509 21:00:41.340] error: unable to read URL "https://storage.googleapis.com/knative-nightly/serving/latest/istio-crds.yaml", server reported 404 Not Found, status code=404
I0509 21:00:41.346] ERROR: Knative setup failed
```

Issue #102, opened 2019-05-10
`service delete` output should be nicer


maximilien:
Change:

`Service hello in namespace default deleted`

to 

`Service 'hello' in namespace 'default' deleted`

maximilien:
Fixed in PR #103

rhuss:
I also think this issue is fixed in #95

maximilien:
OK. Sorry about that.

PR #103, opened 2019-05-10
Addresses issue #102

maximilien:
Changes `service delete` output to be nicer to something like:

`Service 'hello' in namespace 'default' deleted`

### DIFF
--- a/pkg/kn/commands/service_delete.go
+++ b/pkg/kn/commands/service_delete.go
@@ -53,7 +53,7 @@ func NewServiceDeleteCommand(p *KnParams) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			fmt.Fprintf(cmd.OutOrStdout(), "Deleted %s in %s namespace.\n", args[0], namespace)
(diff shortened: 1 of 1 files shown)


pkg/kn/commands/service_delete.go
```diff
@@ -53,7 +53,7 @@ func NewServiceDeleteCommand(p *KnParams) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			fmt.Fprintf(cmd.OutOrStdout(), "Deleted %s in %s namespace.\n", args[0], namespace)
+			fmt.Fprintf(cmd.OutOrStdout(), "Deleted '%s' in '%s' namespace.\n", args[0], namespace)
```
rhuss: Could we please add 'service' as extra context info (as this is easier to detect e.g. when grepping in logs when this output is redirected) ? Also, would prefer to have "namespace" before the actual namespace (as this is the usual order: <resource kind> <name> in message.

I.e. `Deleted service '%s' in namespace '%s'`

Whether to add punctuation or not should be follow a common convention over all messages (not sure how it is in the client).

navidshaikh:
This is addressed in PR #95.

maximilien:
Closing since already addressed in #95

PR #108, opened 2019-05-14
Updates generated Cobra docs for commands and adds new

maximilien:
* service get
* revision get


docs/cmd/kn.md
```diff
@@ -25,4 +25,4 @@ Eventing: Manage event subscriptions and channels. Connect up event sources.
 * [kn service](kn_service.md)	 - Service command group
 * [kn version](kn_version.md)	 - Prints the client version
 
-###### Auto generated by spf13/cobra on 3-May-2019
```
cppforlife: lets remove autogen tag similar to this: https://github.com/cppforlife/knctl/blob/master/pkg/knctl/cmd/knctl.go#L78

PR #109, opened 2019-05-14
Updates docs/workflows.md

maximilien:
* checked each command and add any ouput
* removes all `service list` since that's no longer a command
* adds `service get` and output


maximilien:
/assign @cppforlife 

Updated with `service get` and various others. See commend of PR.

PR #115, opened 2019-05-16
Setup release infrastructure for Knative Client

adrcunha:
Just like the other Knative modules, and ready for automation.

Requires updating test-infra to the latest version.

Part of knative/test-infra#745


hack/release.sh
```diff
@@ -0,0 +1,41 @@
...
+  local version="$(cat hack/NEXT_VERSION)-$(date -u +%Y%m%d)"
+  local now="$(date -u '+%Y-%m-%d %H:%M:%S')"
+  local rev="$(git rev-parse --short HEAD)"
+  local pkg="github.com/knative/client/pkg/kn/commands"
+  local ld_flags="-X '${pkg}.BuildDate=${now}' -X ${pkg}.Version=${version} -X ${pkg}.GitRevision=${rev}"
+
+  export GO111MODULE=on
+  export CGO_ENABLED=0
+  echo "🚧 🐧 Building for Linux"
+  GOOS=darwin GOARCH=amd64 go build -mod=vendor -ldflags "${ld_flags}" -o ./kn-darwin-amd64 ./cmd/...
```
navidshaikh: ./kn-linux-amd64

hack/release.sh
```diff
@@ -0,0 +1,41 @@
...
+  echo "🚧 🍏 Building for macOS"
+  GOOS=linux GOARCH=amd64 go build -mod=vendor -ldflags "${ld_flags}" -o ./kn-linux-amd64 ./cmd/...
```
navidshaikh: ./kn-darwin-amd64

hack/release.sh
adrcunha: Funny, the original code was swapped and we never realized it. :) Fixed, thanks.

PR #116, opened 2019-05-16
Updates docs for service create with --force flag and examples

navidshaikh:
Docs and examples for `kn service create --force`.


sixolet:
This got overridden when I generated docs a little later.

PR #123, opened 2019-05-17
Fix build.sh

### LINKED ISSUE #120: ./hack/build.sh failing
```bash
client git:(master) ./hack/build.sh
./hack/build.sh: line 22: /Users/maximilien/Desktop/github/knative-client/src/github.com/knative/client/hack/util/flags.sh: No such file or directory
```

Looks liken the file was deleted in #115 but still present in [line 22 of build.sh](https://github.com/knative/client/blob/9ec9671b2ebf8f2094ee3435e009a0f64d9f98db/hack/build.sh#L22)

Commenting that out results in:

```bash
➜  client git:(master) ✗ ./hack/build.sh
📋 Formatting
🚧 Building
./hack/build.sh: line 29: ld_flags: command not found
```

So not clear we can remove it.

adrcunha:
This was missed in #115. Delegate the build process to `release.sh`, so we have a single, consistent way of building `kn`.

Fixes #120.

/assign @sixolet

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -14,21 +14,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+source $(dirname $0)/../vendor/github.com/knative/test-infra/scripts/library.sh
+cd ${REPO_ROOT_DIR}
+
 set -o pipefail
 set -eu
-
-dir=$(dirname "${BASH_SOURCE[0]}")
-base=$(cd "$dir/.." && pwd)
-source ${base}/hack/util/flags.sh
 
 export GO111MODULE=on
 
 echo "📋 Formatting"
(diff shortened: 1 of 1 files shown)


adrcunha:
> So creating binaries every time for all platform is IMO overkill and increases the build time
> considerably. THis should be at least switchable.
> 
> Also, we should run the unit tests from build.sh, which should be also configurable.
> 
> This as also the reason why there were two script: build.sh and build-binaries.sh

Adding a switch to build only the current platform or running the unit tests (even though it's a new feature) from build.sh is not that hard. I can add the switches, or this can be closed in favor of 122. Let me know what you prefer.

adrcunha:
Naomi approved 122, so I'm closing this one.

PR #131, opened 2019-05-20
Adds IBM's IKS to the list of services

maximilien:


docs/cmd/kn_service_get.md
```diff
@@ -17,7 +17,7 @@ kn service get [flags]
       --allow-missing-template-keys   If true, ignore any errors in templates when a field or map key is missing in the template. Only applies to golang and jsonpath output formats. (default true)
   -h, --help                          help for get
   -n, --namespace string              List the requested object(s) in given namespace.
-  -o, --output string                 Output format. One of: json|yaml|name|template|go-template|go-template-file|templatefile|jsonpath|jsonpath-file.
+  -o, --output string                 Output format. One of: json|yaml|name|templatefile|template|go-template|go-template-file|jsonpath|jsonpath-file.
```
rhuss: This part is for sure not relevant for this PR, but very likely a left-over of a PR which has missed to regenerate the documentation.

I wonder whether we could add a check to the CI as we do for 'update dependencies', i.e. failing the CI, if the generated docs during the CI is not the same as the committed docs, which implies that it has been forgotten to recreate the docs for the given PR.

This would avoid this kind of unrelated commits.

navidshaikh: @rhuss : +1

also, I have been this inconsistency of ordering in supported output format, for eg: we can run the generate-docs.sh and the diff might not be there at my machine.

maximilien: Sure, I can remove that.

PR #134, opened 2019-05-20
refactor(serving): KnClient interface for single point of cluster access

### LINKED ISSUE #133: GVK is hard-coded to wrong coordinates
E.g. in 

* https://github.com/knative/client/blob/bd7d57fa0928d605ec60ae47268994cac85ceb33/pkg/kn/commands/service_get.go#L50-L51
* https://github.com/knative/client/blob/a264f060c19723d1a00a678f106da2955ca36fb4/pkg/kn/commands/service_describe.go#L54-L57

the `GroupVersionKind` is hardcoded.

In fact, it should be obtained via from the client library and not hardcoded.

rhuss:
Fixes #133.

Introducing a KnClient interface through which all access to the k8s cluster should happen. This facade takes care that GVK are properly set, simplifies the usage and provide domain-specific methods (like `GetRevisionsForService(string)`. 

For testing, it uses `https://godoc.org/gotest.tools` but otherwise plain `testing`. gotest.tools is reasonably lightweight and is already used in other knative projects (like pkg iirc).

Having this kind of clear separation and single-point of cluster contact also will make the transition easier when switching to newer API versions.

You find the test code in https://github.com/rhuss/client/blob/pr/add-schema-handling/pkg/serving/v1alpha1/client_test.go

Follow the discussion [overthere](https://github.com/knative/client/pull/185#issuecomment-506984415) whether to pick up this PR or #214


pkg/kn/commands/revision_describe.go
```diff
@@ -16,10 +16,11 @@ package commands
 
 import (
 	"errors"
+	"github.com/knative/client/pkg/serving"
```
navidshaikh: in order to have distinction between kn serving and actual serving lib, lets streamline the alias for one of them?

For eg: we alias `servinglib` for kn serving pkg, while we use the knative/serving as is.

WDYT ?

pkg/kn/commands/revision_describe.go
```diff
@@ -16,10 +16,11 @@ package commands
 
 import (
 	"errors"
```
navidshaikh: separator line between std imports and rest

pkg/serving/schema_handling.go
```diff
@@ -0,0 +1,31 @@
+package serving
+
+import (
+	"errors"
+	"fmt"
```
navidshaikh: separator line between std imports and rest

pkg/serving/schema_handling.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/knative/serving/pkg/client/clientset/versioned/scheme"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func UpdateGroupVersionKind(obj runtime.Object, gv schema.GroupVersion) error {
```
navidshaikh: exported functions with a comment is helpful, `/lint` prow command will pick this up.

pkg/serving/schema_handling.go
```diff
@@ -0,0 +1,31 @@
...
+	gvk, err := getGroupVersionKind(obj, gv)
+	if err != nil {
+		return err
+	}
+	obj.GetObjectKind().SetGroupVersionKind(*gvk)
+	return nil
+}
+
+func getGroupVersionKind(obj runtime.Object, gv schema.GroupVersion) (*schema.GroupVersionKind, error) {
```
navidshaikh: thinking if we'll need it somewhere else too, to make it public :thinking:

pkg/serving/schema_handling_test.go
```diff
@@ -0,0 +1,29 @@
...
+func TestGVKUpdate(t *testing.T) {
+	service := v1alpha1.Service{}
+	err := UpdateGroupVersionKind(&service, v1alpha1.SchemeGroupVersion)
+	if err != nil {
+		t.Fatalf("cannot update GVK to a service %v", err)
+	}
+	if service.Kind != "Service" {
+		t.Fatalf("wrong kind '%s'", service.Kind)
+	}
+	if service.APIVersion != v1alpha1.SchemeGroupVersion.Group+"/"+v1alpha1.SchemeGroupVersion.Version {
```
navidshaikh: :+1:

pkg/serving/schema_handling.go
rhuss: It's specific to the Knative scheme (so not for other schemes like the core schema), and I don't yet see any other use case except for setting the GVK on an object.

I suggest to keep it private and only open it up if there is a concrete use case which requires this method.

cppforlife:
@rhuss i forget why do we even set this on retrieved objects, dont they already have this on them?

rhuss:
> @rhuss i forget why do we even set this on retrieved objects, dont they already have this on them?

No, the recieved objects from the RestClient don't have the GVK set by intention so that the field is empty.

In memory, these fields are empty and will be filled only during marshalling to JSON/protobuf (using the same mechanism as here). That is automatically done by client-go when reaching out to the server.

That seems to be a global convention throughout the go-client.

(BTW, I have this info from first-hand experience where I struggled quite a bit with this, but it is also described in the forthcoming "Programming Kubernetes" book by @stts . Excellent book, I learned a lot already from the early access version.)

rhuss:
@cppforlife @sixolet I addressed all review comments. Can we please get this merged asap, as this affects all other API calls (like the one I'm just working on for `kn service describe`) ?

cppforlife:
@rhuss hmm. i was looking around kubernetes issues/source and found a few issues related to these fields being cleared out just to avoid duplicating data from the struct type itself. kind of ugly...

it seems that we need this info there just for printing. i think we should create a new printer (implements ResourcePrinter interface) that would backfill this info before passing on obj to underlying printer. im not a big fan of having public UpdateGroupVersionKind method added all over the place.

rhuss:
> it seems that we need this info there just for printing. i think we should create a new printer (implements ResourcePrinter interface) that would backfill this info before passing on obj to underlying printer. im not a big fan of having public UpdateGroupVersionKind method added all over the place.

I see your point, but I'm afraid that we have to do this all the time when we read resources to present it to the user in any way (so everything except for create/update/delete).  I wouldn't put this into the printer directly, as there might be occasions where print resource without printers. But yes, we could go this way (as I think its really only needed for machine-readable output, but this is delegated to cli-runtime).

An alternative approach is to collect all server-side access into a thin layer, which is suited for the client (and which could be also used as a kn-client library in other client, too).

An example for such layer is part of #75 ,  in https://github.com/knative/client/blob/a44413cab56cc4fc568e5c02c09f6f2aa1ebc958/pkg/serving/client.go

It's kind of a DAO layer (you see where I'm coming from :), which encapsulates nicely client specific context information and provides specialized method which suit the client's use case, like https://github.com/knative/client/blob/a44413cab56cc4fc568e5c02c09f6f2aa1ebc958/pkg/serving/client.go#L56 returns all revisions for a specific service (you can't do this directly with standard serving API).

If all server access would go through this specialized `client`, then there is a single place to update the GVK (in this PR still hardcoded but should be eventually updated to the dynamic lookup) in https://github.com/knative/client/blob/a44413cab56cc4fc568e5c02c09f6f2aa1ebc958/pkg/serving/client.go#L73-L77

rhuss:
Said that, let me try to put this into the printer layer and looks how it feels.

sixolet:
Other than that, I like this approach more than #153 ; it doesn't seem like a ResourcePrinter's concern to mutate the resources in question.

rhuss:
@sixolet thanks for the review. I will update this according to your suggestion.

> I've now had a chance to look at #153 and this PR in detail, and come to the conclusion that the GVK should come from `ServingFactory`, along with the client object. That'll preserve the flexibility we need to add a `--v1beta1` flag later. Can we do that?

I extended the client returned by ServiceFactory() with a method `UpdateGroupVersionKind()`, which can be called with just the object to update (and without scheme). This means that the command doesnt need to know about the api version. I would expect when going to a v1beta1 version (and we need to support both in the client) to have a second ServiceFactory() method for v1beta1 (`ServiceFactoryV1Beta1()` or so) and the commands have to decide which to use, depending on which version of Knative is installed on the serverside.

Updating the tests was a bit involved, I think the tests could be streamlined, too, to remove quite some duplicated code.

sixolet:
A method on the client is good.

As for the `ServingFactory`, you may be right, but I'm hoping to make the right abstraction cut for some clean code then. It'll be a fun refactor; ideally only the lib code will care, not the command code.

sixolet:
A method on the client is good.

As for the `ServingFactory`, you may be right, but I'm hoping to make the right abstraction cut for some clean code then. It'll be a fun refactor; ideally only the lib code will care, not the command code.

Anyway, it is a problem for later.

Fix conflicts and we can get this merged?

rhuss:
@sixolet Conflicts resolved and test green.

rhuss:
The conflicts are a bit more involved, will fix this over the weekend.

rhuss:
@sixolet conflicts resolved, after refactorings of @maximilien this PR looks even simpler wrt/ test code changes.

lgtm label was removed, so please lgtm again (if ok)

pkg/kn/commands/types.go
```diff
@@ -27,10 +31,28 @@ var CfgFile string
 // KubeCfgFile is the path for the Kubernetes config
 var KubeCfgFile string
 
+// Extend v1alpha1 serving client by a method for updating GVK
+type ServingV1alpha1InterfaceWithGvkUpdater interface {
```
cppforlife: if we are wrapping ServingV1alpha1Interface why not set GVK transparently in its methods?

rhuss: Yes, we could do that, too (IMO depends on how many methods to wrap, though, whether that is suitable. Haven't checked yet). 

For the alternative is to have a more custom client library like approach suited for kn as provided in https://github.com/rhuss/client/blob/pr/get/pkg/serving/client.go (especially for cross-cutting methods like `RevisionsForService` as in https://github.com/rhuss/client/blob/744872f9350f29f1cdf09ba21a7b732a53cece53/pkg/serving/client.go#L56-L69)
which I'm going to propose towards the end of the week.

pepperbot: @cppforlife @sixolet I reworked that PR to make that update much more transparent and frees the caller from doing that update on its own.

Therefor a serving.v1alpha1.KnClient interface introduced which is an ambassador/proxy to the serving API. Such a client is specific to a namespace (comes in as construction parameter) and holds also kn specific methods (like 'RevisionsForService' which are not found the serving API.

The benefits of this approach:

* Simpler usage for the commands (see diff). E.g. error checking, no GVK update required by the commands, shorter calling chain.
* Fully controlled access layer to the service for every client. Mangling like updating GVK can be performed for this single spot.
* Only a reduced set of APIs which is used by kn is offered
* Already in a separate v1alpha1 package, semi-prepared for switch to 0.7.0 (or v1beta1), coming in two weeks.
* Testing of commands can be made much simpler as only this KnClient interface needs to be mocked, not the full serving API

Please let now whether this third attempt is now heading in the right direction. If so, I would add the missing tests.

pkg/serving/v1alpha1/client_test.go
```diff
@@ -0,0 +1,320 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	client_testing "k8s.io/client-go/testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/knative/client/pkg/serving"
+)
+
+func TestClient(t *testing.T) {
```
maximilien: This can move to a common `*_suite.go` file so as to avoid having to repeat it in other tests in that package. Ginkgo provides an easy way to generate these:

```bash
cd package-dir
ginkgo bootstrap
```

pkg/serving/v1alpha1/client_test.go
```diff
@@ -0,0 +1,320 @@
...
+		serving  fake.FakeServingV1alpha1
+	)
+
+	BeforeEach(func() {
+		serving = fake.FakeServingV1alpha1{Fake: &client_testing.Fake{}}
+		knClient = NewKnServingClient(&serving, namespace)
+	})
+
+	// -----------------------------------------------------------------------------------------
+	Describe("Getting a service", func() {
```
maximilien: One thing I like to do is to have a `Describe("Function")` for every public function or method that is being tested. This is especially useful is the test file matches a struct with lots of methods.

But as you can imagine. There are lots of way to organize and I tend not to be pedantic about things like that :)

pkg/serving/v1alpha1/client_test.go
```diff
@@ -0,0 +1,320 @@
...
+					name := a.(client_testing.CreateAction).GetObject().(metav1.Object).GetName()
+					if name == serviceNew.Name {
+						serviceNew.Generation = 2
+						return true, serviceNew, nil
+					}
+					return true, nil, fmt.Errorf("error while creating service %s", name)
+				})
+		})
+
+		Context("without error", func() {
```
maximilien: Nice use of context overall, that's precisely where Ginkgo shines. When you run the tests and there are failures makes it easy to understand what's expected and where the failures are occurring.

vendor/golang.org/x/text/encoding/charmap/maketables.go
```diff
@@ -0,0 +1,556 @@
+// Copyright 2013 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build ignore
+
+package main
+
```
maximilien: Hmm, what's this dependency for? Just curious.

rhuss: That just comes in as a transitive dependency of ginkgo and/or gomega. Should be in your version test PR, too.

pkg/serving/v1alpha1/client_test.go
rhuss: Yeah, just tried to be a bit more 'behavioral' to use some natural service. We could as well use 'GetService', as description name. I don't have a strong opinion here.

rhuss:
@sixolet @cppforlife What's missing here to get that PR merged ? If requested, I can remove the testing part.

From my POV there are no open issues, would be awesome to get that done soonish, as constantly rebasing this PR is not a pleasant work.

rhuss:
@navidshaikh I rebased to the latest revision list PR, but I changed "serving.knative.dev/configuration" to "serving.knative.dev/service" for the label selector. Not 100% sure which one is the better selector, but I think connecting it to the service label seems to be more appropriate.

rhuss:
@adrcunha we have an issue with the license_check in the presubmit checks. 

These fail with 

```
go: finding github.com/google/licenseclassifier latest
go: finding github.com/stretchr/testify v1.3.0
go: finding github.com/google/go-cmp v0.2.0
go: finding github.com/sergi/go-diff v1.0.0
go: finding github.com/davecgh/go-spew v1.1.0
go: finding github.com/stretchr/objx v0.1.0
go: finding github.com/stretchr/objx v0.2.0
go: downloading github.com/google/licenseclassifier v0.0.0-20190501212618-47b603fe1b8c
go: extracting github.com/google/licenseclassifier v0.0.0-20190501212618-47b603fe1b8c
go: downloading github.com/sergi/go-diff v1.0.0
go: extracting github.com/sergi/go-diff v1.0.0
2019/06/21 08:24:20 Error creating license classifier: cannot register licenses: EOF
```

on all our PRs. However, running locally the equivalent checks

```
go install github.com/knative/test-infra/tools/dep-collector
dep-collector -check $(go list ./...)

2019/06/21 10:34:11 No errors found.
```

works without issue. Any idea what that could be ?

adrcunha:
> @adrcunha we have an issue with the license_check in the presubmit checks.
> on all our PRs. However, running locally the equivalent checks

I haven't seen this in other repos. observability also uses modules and there's no PR failing build tests for the same reason. Was there any recent change to the deps or modules? The error seems to be coming from https://github.com/google/licenseclassifier/blob/master/classifier.go#L110 but the message is slightly different.

adrcunha:
The error message seems to be from an old version of license classifier:

https://github.com/google/licenseclassifier/commit/57ecb7c44031f8d66cc9316e6e51e9303c65d9e5#diff-98a095ba54afd336cbb1d8223d41f7a5

pkg/kn/commands/types.go
rhuss: (sorry that this comment was written under another account as I was logged in accidentally with my bots GitHub account at that time)

rhuss:
@adrcunha the reason for the failure is, that an old, installed dep-collector is picked up from `$GOPATH/bin` without rebuilding is with a new license dependency:

```
+ go get -u github.com/google/licenseclassifier
go: finding github.com/google/licenseclassifier latest
go: downloading github.com/google/licenseclassifier v0.0.0-20190501212618-47b603fe1b8c
go: extracting github.com/google/licenseclassifier v0.0.0-20190501212618-47b603fe1b8c
go: downloading github.com/sergi/go-diff v1.0.0
go: extracting github.com/sergi/go-diff v1.0.0
+ find /home/prow/go -name licenseclassifier
/home/prow/go/pkg/mod/cache/download/github.com/google/licenseclassifier
+ run_go_tool ./vendor/github.com/knative/test-infra/tools/dep-collector dep-collector -check github.com/knative/client/cmd/kn github.com/knative/client/hack github.com/knative/client/pkg github.com/knative/client/pkg/kn/commands github.com/knative/client/pkg/kn/commands/revision github.com/knative/client/pkg/kn/commands/service github.com/knative/client/pkg/kn/core github.com/knative/client/pkg/printers github.com/knative/client/pkg/serving github.com/knative/client/pkg/serving/v1alpha1 github.com/knative/client/test/e2e
+ local tool=dep-collector
++ which dep-collector
+ [[ -z /go/bin/dep-collector ]]
+ shift 2
+ dep-collector -check github.com/knative/client/cmd/kn github.com/knative/client/hack github.com/knative/client/pkg github.com/knative/client/pkg/kn/commands github.com/knative/client/pkg/kn/commands/revision github.com/knative/client/pkg/kn/commands/service github.com/knative/client/pkg/kn/core github.com/knative/client/pkg/printers github.com/knative/client/pkg/serving github.com/knative/client/pkg/serving/v1alpha1 github.com/knative/client/test/e2e
2019/06/22 06:53:47 Error creating license classifier: cannot register licenses: EOF
+ set +x
```

I'm trying to change now to unconditionally rebuild dep-collector for the run, so that it does not depend on the binary installed by a previous run (or even different repository). I consider this to be dangerous anyway, so if this works here in my test, going to submit a PR over there in test-infra.

adrcunha:
> @adrcunha the reason for the failure is, that an old, installed dep-collector is picked up from `$GOPATH/bin` without rebuilding is with a new license dependency:

Indeed. However, I already generated a new prow-tests image but Prow insists on using the previous image. :/ I'm still trying to fix this.

adrcunha:
Tracking the licence checking issue in https://github.com/knative/test-infra/issues/1018

adrcunha:
@rhuss the license checking error is gone, now you just need to run update-deps.sh.

rhuss:
I think we are good now, now that we have decided on the testing framework. thanks to everyone ;-)

'would be happy when this old-timer (on which implementation we agreed at some point in time) could be merged asap (rebasing gets harder and harder). Happy to help to adapt outstanding PRs for using the client layer or fix any issue and refinements afterwards. // @sixolet @cppforlife @maximilien

rhuss:
hmm, next conflict to resolve (not the modules one, but I have to refactor routes/list.go. I make three crosses when this PR is merged ;-)

rhuss:
ok, here we go. // @sixolet @cppforlife anything still missing ? 

To recpaitulized, the core part is the refactoring leading to the following `KnClient` interface:

```golang
type KnClient interface {


	// Get a service by its unique name
	GetService(name string) (*v1alpha1.Service, error)


	// List services
	ListServices(opts ...ListConfig) (*v1alpha1.ServiceList, error)


	// Create a new service
	CreateService(service *v1alpha1.Service) error


	// Update the given service
	UpdateService(service *v1alpha1.Service) error


	// Delete a service by name
	DeleteService(name string) error


	// Wait for a service to become ready, but not longer than provided timeout
	WaitForService(name string, timeout time.Duration, out io.Writer) error


	// Get a revision by name
	GetRevision(name string) (*v1alpha1.Revision, error)


	// List revisions
	ListRevisions(opts ...ListConfig) (*v1alpha1.RevisionList, error)


	// Delete a revision
	DeleteRevision(name string) error


	// List routes
	ListRoutes(opts ...ListConfig) (*v1alpha1.RouteList, error)
}
```

in https://github.com/rhuss/client/blob/f264ae6f1ff6b6b816d30274f7cab096dd29066e/pkg/serving/v1alpha1/client.go#L38-L69

This is specific to v1alhpa1 but we could easily provide a similar interface for v1beta1. Then the callers can dispatch on the current cluster version to decide which version to use. The WaitForService could be a bit nicer without an io.Writer argument (i.e. with just a `ProgressHandler` callback function which in turn can print out th e progress). I would address this in a subsequent PR.

The interface is deliberately kept easy (e.g without namespace params as this is the context stored during creation of the implementation, and also no return of the created service object) so that it could easily serve as an API for other users (like e.g. plugins which want to combine knative functionality with something else).

For `list` methods a fluid config model is implemented to easily define a filter (e.g. `ListRevisions(WithService("myservice"))`)

Really I would be very happy if we get that merged soon and I'm totally comitted to make it better and improvements after this. But now, for every PR coming in, I have again to refactor during a rebase to use this model (e.g. like I just did for `ListRoutes()`). It would be much better if people already could jump on this new API right now.

Ah yes, testing will becomes much easier, too soon (as only this client needs to deal with a fake service) whereas the command test can jump on a test implementation of this simple interface.

Please.

PR #136, opened 2019-05-20
chore: Streamlined build.sh

rhuss:
* Merged generate-docs.sh into build.sh
* Split up into smaller functions
* Added serving dependency version to 'kn version' command

Not sure whether the latter is really useful (showing the serving dependency)
as this is might get confused with the knative version
installed on the cluster.


hack/generate-docs.go
```diff
@@ -18,14 +18,18 @@ import (
...
 
 func main() {
 	rootCmd := commands.NewKnCommand()
 
-	err := doc.GenMarkdownTree(rootCmd, "./docs/cmd/")
+	dir := "."
+	if len(os.Args) > 1 {
+		dir = os.Args[1]
+	}
+		err := doc.GenMarkdownTree(rootCmd, dir + "/docs/cmd/")
```
cppforlife: extra indent

Issue #143, opened 2019-05-27
kn service get <service-name> will list all the services


savitaashture:
Get all the services in particular namespace
```
kn service get -n test
NAME           DOMAIN                          GENERATION   AGE    CONDITIONS   READY     REASON
autoscale-go   autoscale-go.test.example.com   1            124m   0 OK / 3     Unknown   RevisionMissing : Configuration "autoscale-go" is waiting for a Revision to become ready.
test1          test1.test.example.com          1            118m   0 OK / 3     False     RevisionMissing : Configuration "test1" does not have any ready Revision.
```

get a particular service by specifying service name

**current:**
```
kn service get -n test autoscale-go
NAME           DOMAIN                          GENERATION   AGE    CONDITIONS   READY     REASON
autoscale-go   autoscale-go.test.example.com   1            125m   0 OK / 3     Unknown   RevisionMissing : Configuration "autoscale-go" is waiting for a Revision to become ready.
test1          test1.test.example.com          1            119m   0 OK / 3     False     RevisionMissing : Configuration "test1" does not have any ready Revision.
```

**Expected:** Should list only that perticular service.
something like below
```
kn service get -n test autoscale-go
NAME           DOMAIN                          GENERATION   AGE    CONDITIONS   READY     REASON
autoscale-go   autoscale-go.test.example.com   1            125m   0 OK / 3     Unknown   RevisionMissing : Configuration "autoscale-go" is waiting for a Revision to become ready.
```

Because of above behavior when we do 

```
kn service get -n test autoscale-go -o yaml
```

It will give the detailed yaml info for all the services.

**Applicable for all kn commands**

navidshaikh:
Resource name as argument to `get` command is not yet implemented, `get` command's default behavior is to list all resources in requested namespace. Single resource's `get` is actually `describe` command in human readable format which should support different output formats with `-o` option. @rhuss is on it check https://github.com/knative/client/pull/75

rhuss:
@navidshaikh I think `get` should be able to list a single resource as well (in short form), much like `kubectl get` does, too. This is especially useful for machine-readable output (`-o json`). We could use the name as literal or as a prefix file (which would be nice for revisions to catch all similar revisions).

`kn service describe` then is for detailed human-readable output only, but `kn get service myservice -o yaml` is important so that `kn` can easily be used for scripting, too.

navidshaikh:
@rhuss : sounds good, +1

[...]

PR #150, opened 2019-05-30
get a particular service and revision based on name parameter

savitaashture:
https://github.com/knative/client/issues/143

**current:**
```
kn service list -n test autoscale-go
NAME           DOMAIN                          GENERATION   AGE    CONDITIONS   READY     REASON
autoscale-go   autoscale-go.test.example.com   1            125m   0 OK / 3     Unknown   RevisionMissing : Configuration "autoscale-go" is waiting for a Revision to become ready.
test1          test1.test.example.com          1            119m   0 OK / 3     False     RevisionMissing : Configuration "test1" does not have any ready Revision.
```

**Expected:** Should list only that perticular service.
something like below
```
kn service list -n test autoscale-go
NAME           DOMAIN                          GENERATION   AGE    CONDITIONS   READY     REASON
autoscale-go   autoscale-go.test.example.com   1            125m   0 OK / 3     Unknown   RevisionMissing : Configuration "autoscale-go" is waiting for a Revision to become ready.
```

### DIFF
--- a/pkg/kn/commands/revision/revision_list.go
+++ b/pkg/kn/commands/revision/revision_list.go
@@ -29,7 +29,7 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
 	revisionListFlags := NewRevisionListFlags()
 
 	revisionListCommand := &cobra.Command{
-		Use:   "list",
+		Use:   "list [name]",
 		Short: "List available revisions.",
 		Long:  "List revisions for a given service.",
 		Example: `
@@ -37,7 +37,13 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
   kn revision list
 
   # List revisions for a service 'svc1' in namespace 'myapp'
-  kn revision list -s svc1 -n myapp`,
+  kn revision list -s svc1 -n myapp
+
+  # List all revisions in JSON output format
+  kn revision list -o json
+  
+  # List revision 'web'
+  kn revision list web`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			namespace, err := p.GetNamespace(cmd)
 			if err != nil {
@@ -65,7 +71,7 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
 					return nil
 				}
 			} else {
-				revisionList, err = client.ListRevisions()
+				revisionList, err = getRevisionInfo(args, client)
 				if err != nil {
 					return err
 				}
@@ -74,7 +80,6 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
 					return nil
 				}
 			}
-
 			printer, err := revisionListFlags.ToPrinter()
 			if err != nil {
 				return err
@@ -89,4 +94,20 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
 	commands.AddNamespaceFlags(revisionListCommand.Flags(), true)
 	revisionListFlags.AddFlags(revisionListCommand)
 	return revisionListCommand
+}
+
+func getRevisionInfo(args []string, client v1alpha12.KnClient) (*v1alpha1.RevisionList, error) {
+	var (
+		revisionList *v1alpha1.RevisionList
+		err          error
+	)
+	switch len(args) {
+	case 0:
+		revisionList, err = client.ListRevisions()
+	case 1:
+		revisionList, err = client.ListRevisions(v1alpha12.WithName(args[0]))
+	default:
+		return nil, fmt.Errorf("'kn revision list' accepts maximum 1 argument")
+	}
+	return revisionList, err
 }
--- a/pkg/kn/commands/revision/revision_list_test.go
+++ b/pkg/kn/commands/revision/revision_list_test.go
@@ -61,6 +61,19 @@ func TestRevisionListEmpty(t *testing.T) {
 	}
 }
 
+func TestRevisionListEmptyByName(t *testing.T) {
+	action, _, err := fakeRevisionList([]string{"revision", "list", "name"}, &v1alpha1.RevisionList{})
+	if err != nil {
+		t.Error(err)
+		return
+	}
+	if action == nil {
+		t.Errorf("No action")
+	} else if !action.Matches("list", "revisions") {
+		t.Errorf("Bad action %v", action)
+	}
+}
+
 func TestRevisionListDefaultOutput(t *testing.T) {
 	revision1 := createMockRevisionWithParams("foo-abcd", "foo")
 	revision2 := createMockRevisionWithParams("bar-wxyz", "bar")
@@ -121,6 +134,30 @@ func TestRevisionListForService(t *testing.T) {
 		t.Errorf("Bad action %v", action)
 	}
 	assert.Assert(t, util.ContainsAll(output[0], "No", "revisions", "svc3"), "no revisions")
+}
+
+func TestRevisionListOneOutput(t *testing.T) {
+	revision := createMockRevisionWithParams("foo-abcd", "foo")
+	RevisionList := &v1alpha1.RevisionList{Items: []v1alpha1.Revision{*revision}}
+	action, output, err := fakeRevisionList([]string{"revision", "list", "foo-abcd"}, RevisionList)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if action == nil {
+		t.Errorf("No action")
+	} else if !action.Matches("list", "revisions") {
+		t.Errorf("Bad action %v", action)
+	}
+
+	assert.Assert(t, util.ContainsAll(output[0], "NAME", "SERVICE", "AGE", "CONDITIONS", "READY", "REASON"))
+	assert.Assert(t, util.ContainsAll(output[1], "foo", "foo-abcd"))
+}
+
+func TestRevisionListOutputWithTwoRevName(t *testing.T) {
+	t.Log("verify by passing two revision name to get command")
+	RevisionList := &v1alpha1.RevisionList{Items: []v1alpha1.Revision{}}
+	_, _, err := fakeRevisionList([]string{"revision", "list", "foo-abcd", "bar-abcd"}, RevisionList)
+	assert.ErrorContains(t, err, "'kn revision list' accepts maximum 1 argument")
 }
 
 func createMockRevisionWithParams(name, svcName string) *v1alpha1.Revision {
--- a/pkg/kn/commands/service/service_list.go
+++ b/pkg/kn/commands/service/service_list.go
@@ -18,6 +18,8 @@ import (
 	"fmt"
 
 	"github.com/knative/client/pkg/kn/commands"
+	v1alpha12 "github.com/knative/client/pkg/serving/v1alpha1"
+	"github.com/knative/serving/pkg/apis/serving/v1alpha1"
 	"github.com/spf13/cobra"
 )
 
@@ -26,8 +28,17 @@ func NewServiceListCommand(p *commands.KnParams) *cobra.Command {
 	serviceListFlags := NewServiceListFlags()
 
 	serviceListCommand := &cobra.Command{
-		Use:   "list",
+		Use:   "list [name]",
 		Short: "List available services.",
+		Example: `
+  # List all services
+  kn service list
+
+  # List all services in JSON output format
+  kn service list -o json
+
+  # List service 'web'
+  kn service list web`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			namespace, err := p.GetNamespace(cmd)
 			if err != nil {
@@ -37,7 +48,7 @@ func NewServiceListCommand(p *commands.KnParams) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			serviceList, err := client.ListServices()
+			serviceList, err := getServiceInfo(args, client)
 			if err != nil {
 				return err
 			}
@@ -60,4 +71,20 @@ func NewServiceListCommand(p *commands.KnParams) *cobra.Command {
 	commands.AddNamespaceFlags(serviceListCommand.Flags(), true)
 	serviceListFlags.AddFlags(serviceListCommand)
 	return serviceListCommand
+}
+
+func getServiceInfo(args []string, client v1alpha12.KnClient) (*v1alpha1.ServiceList, error) {
+	var (
+		serviceList *v1alpha1.ServiceList
+		err         error
+	)
+	switch len(args) {
+	case 0:
+		serviceList, err = client.ListServices()
+	case 1:
+		serviceList, err = client.ListServices(v1alpha12.WithName(args[0]))
+	default:
+		return nil, fmt.Errorf("'kn service list' accepts maximum 1 argument")
+	}
+	return serviceList, err
 }
--- a/pkg/kn/commands/service/service_list_test.go
+++ b/pkg/kn/commands/service/service_list_test.go
@@ -60,6 +60,19 @@ func TestListEmpty(t *testing.T) {
 	}
 }
 
+func TestGetEmpty(t *testing.T) {
+	action, _, err := fakeServiceList([]string{"service", "list", "name"}, &v1alpha1.ServiceList{})
+	if err != nil {
+		t.Error(err)
+		return
+	}
+	if action == nil {
+		t.Errorf("No action")
+	} else if !action.Matches("list", "services") {
+		t.Errorf("Bad action %v", action)
+	}
+}
+
 func TestServiceListDefaultOutput(t *testing.T) {
 	service1 := createMockServiceWithParams("foo", "foo.default.example.com", 1)
 	service2 := createMockServiceWithParams("bar", "bar.default.example.com", 2)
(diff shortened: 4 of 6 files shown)


pkg/kn/commands/revision_get.go
```diff
@@ -65,3 +76,38 @@ func NewRevisionGetCommand(p *KnParams) *cobra.Command {
...
+		revision, err := client.Revisions(namespace).List(v1.ListOptions{})
+		if err != nil {
+			return nil, err
+		}
+		if len(revision.Items) == 0 {
+			fmt.Fprintf(cmd.OutOrStdout(), "No resources found.\n")
+			return nil, nil
+		}
+		revision.GetObjectKind().SetGroupVersionKind(schema.GroupVersionKind{
+			Group:   "knative.dev",
```
rhuss: The coordinates are wrong (whould be `serving.knative.dev` for the  Group), but we should get #134 merged as soon as possible which avoids this kind of hardcoding of the GKV. @sixolet @cppforlife can we please get #134 merged soon to avoid piling up on this technical debt ?)

pkg/kn/commands/revision_get.go
```diff
@@ -65,3 +76,38 @@ func NewRevisionGetCommand(p *KnParams) *cobra.Command {
...
+		}
+
+		revision.GetObjectKind().SetGroupVersionKind(schema.GroupVersionKind{
+			Group:   "serving.knative.dev",
+			Version: "v1alpha1",
+			Kind:    "revision"})
+
+		return revision, nil
+	default:
+		return nil, errors.New("revision get does not accept more than one NAME\nUsage: kn revision get NAME")
```
rhuss: I don't think errors should contain any formatting characters. It's not about the inner command code which knows in which context it is used. Imaging that this command is used in a kind of library, then printing out the usage information here is for sure the wrong place. Just return the error without any implicit knowledge where this function could be used, but add as much context information that you have available. Like the revisions names provided and detected.

rhuss: Maybe something like

```
fmt.Errorf("'revision get only accepts no arguments or a single revision name, not %d arguments as given",len(args))
```

pkg/kn/commands/revision_get_test.go
```diff
@@ -28,7 +29,21 @@ import (
 	client_testing "k8s.io/client-go/testing"
 )
 
-func fakeRevisionGet(args []string, response *v1alpha1.RevisionList) (action client_testing.Action, output []string, err error) {
+func fakeRevisionGet(args []string, response interface{}, listOrsingle bool) (action client_testing.Action, output []string, err error) {
+	var (
+		resp runtime.Object
+		ok   bool
+	)
+	if listOrsingle {
```
rhuss: I don't get this conditional. if and else branch are the same, aren't they ?

pkg/kn/commands/revision_get_test.go
```diff
@@ -82,11 +111,69 @@ func TestRevisionGetDefaultOutput(t *testing.T) {
...
+	t.Log("verify by passing two revision name to get command")
+	_, output, err = fakeRevisionGet([]string{"revision", "get", "foo-abcd", "bar-abcd"}, revision, true)
+	assert.NotNil(t, err)
+	assert.Empty(t, output)
+
+}
+
+func createMockRevisionWithParams(name, svcName string, version bool) *v1alpha1.Revision {
+	var apiVersion string
+	if version {
```
rhuss: Why do you have different api versions ?

pkg/kn/commands/service_get.go
```diff
@@ -67,3 +77,38 @@ func NewServiceGetCommand(p *KnParams) *cobra.Command {
...
+		service, err := client.Services(namespace).List(v1.ListOptions{})
+		if err != nil {
+			return nil, err
+		}
+		if len(service.Items) == 0 {
+			fmt.Fprintf(cmd.OutOrStdout(), "No resources found.\n")
+			return nil, nil
+		}
+		service.GetObjectKind().SetGroupVersionKind(schema.GroupVersionKind{
+			Group:   "knative.dev",
```
rhuss: see comment for `revision get`

pkg/kn/commands/service_get_test.go
```diff
@@ -28,7 +29,21 @@ import (
 	client_testing "k8s.io/client-go/testing"
 )
 
-func fakeServiceGet(args []string, response *v1alpha1.ServiceList) (action client_testing.Action, output []string, err error) {
+func fakeServiceGet(args []string, response interface{}, listOrsingle bool) (action client_testing.Action, output []string, err error) {
+	var (
+		resp runtime.Object
+		ok   bool
+	)
+	if listOrsingle {
```
rhuss: both branches are equal, why do you need this `listOrsingle` variable ? (btw, would call it in all camelCase)

pkg/kn/commands/service_get_test.go
```diff
@@ -90,11 +165,17 @@ func testContains(t *testing.T, output string, sub []string, element string) {
 	}
 }
 
-func createMockServiceWithParams(name, domain string, generation int64) *v1alpha1.Service {
+func createMockServiceListWithParams(name, domain string, generation int64, version bool) *v1alpha1.Service {
+	var apiVersion string
+	if version {
```
rhuss: see comment in revision get

pkg/kn/commands/revision_get.go
savitaashture: Sure...Once that is merged i will rechange this

pkg/kn/commands/revision_get.go
savitaashture: I will change the message thanks

pkg/kn/commands/revision_get_test.go
savitaashture: This condition i have added to check the **apiVersion**
Because in the base code
when we do 
```
kn revision get -o yaml

apiVersion: knative.dev/v1alpha1
items:
- apiVersion: serving.knative.dev/v1alpha1
  kind: Revision
  metadata:
    annotations:
      autoscaling.knative.dev/target: "10"
      serving.knative.dev/lastPinned: "1559284138"
    creationTimestamp: "2019-05-31T06:28:47Z"
    generateName: autoscale-go-
    generation: 1
```
If its a list then it is giving apiversion as "knative.dev"

pkg/kn/commands/revision_get_test.go
savitaashture: Yes they are same..
Its my fault i was thinking some other scenario to handle and wrote that.
I will change and update

pkg/kn/commands/service_get_test.go
savitaashture: Yes they are same..
Its my fault i was thinking some other scenario to handle and wrote that.
I will change and update

pkg/kn/commands/service_get_test.go
savitaashture: https://github.com/knative/client/pull/150#discussion_r289274186

pkg/kn/commands/service_get.go
savitaashture: https://github.com/knative/client/pull/150#discussion_r289272069

savitaashture:
> it make sense to unify

I found **fakeRevisionGet** and **fakeServiceGet** duplicate code modified and updated.

rhuss:
@savitaashture thanks a lot for your updates ! Could you please rebase on latest mastrer, there has been some structural changes. Nothing big, probbaly just import conflicts.

savitaashture:
> @savitaashture thanks a lot for your updates ! Could you please rebase on latest mastrer, there has been some structural changes. Nothing big, probbaly just import conflicts.

@rhuss I have rebased with latest master changes.
Thank you.

navidshaikh:
@savitaashture : `get` is not renamed to `list`, please rebase and update the PR.

savitaashture:
> @savitaashture : `get` is not renamed to `list`, please rebase and update the PR.

@navidshaikh @rhuss Rebased

pkg/kn/commands/revision/revision_list.go
```diff
@@ -66,3 +86,26 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
 	revisionListFlags.AddFlags(revisionListCommand)
 	return revisionListCommand
 }
+
+func getRevisionInfo(args []string, namespace string, client serving.ServingV1alpha1Interface) (*v1alpha1.RevisionList, error) {
+
+	switch len(args) {
+	case 0:
+		revision, err := client.Revisions(namespace).List(v1.ListOptions{})
```
cppforlife: case 0 and 1 are the same except list options. seems like it would be better to have single client.Revisions(...) call and have listOpts variable be set in switch statement.

pkg/kn/commands/revision/revision_list.go
```diff
@@ -66,3 +86,26 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
...
+		if err != nil {
+			return nil, err
+		}
+
+		return revision, nil
+
+	case 1:
+		revision, err := client.Revisions(namespace).List(v1.ListOptions{FieldSelector: "metadata.name=" + args[0]})
```
cppforlife: is there any escaping that needs to be done for the name? i think there was some packages in apimachinery around generating field and label selectors.

navidshaikh:
```
  var listOptions v1.ListOptions  
  switch len(args) {
  case 0:
    listOptions = v1.ListOptions{}
  case 1:
    listOptions.FieldSelector = fields.Set(map[string]string{"metadata.name": args[0]}).String()
  default:
    return nil, fmt.Errorf("'revision list' only accepts no arguments or a single revision name, not %d arguments as given", len(args))
  }
  revision, err := client.Revisions(namespace).List(listOptions)
  if err != nil {
    return nil, err
  }
  return revision, nil
```
@savitaashture

pkg/kn/commands/revision/revision_list_test.go
```diff
@@ -71,17 +86,64 @@ func TestRevisionListDefaultOutput(t *testing.T) {
...
+	if action == nil {
+		t.Errorf("No action")
+	} else if !action.Matches("list", "revisions") {
+		t.Errorf("Bad action %v", action)
 	}
+	commands.TestContains(t, output[0], []string{"SERVICE", "NAME", "AGE", "CONDITIONS", "READY", "REASON"}, "column header")
+	commands.TestContains(t, output[1], []string{"bar", "bar-abcd"}, "value")
+
+	t.Log("verify by passing two revision name to get command")
+	_, output, err = fakeRevisionList([]string{"revision", "list", "foo-abcd", "bar-abcd"}, RevisionList)
```
navidshaikh: above test lines in this test are redundant, thats already being tested in `TestRevisionGetOneOutput`, if we are checking for error message here, no need to populate the `RevisionList`, it can be empty.

savitaashture: Updated

pkg/kn/commands/revision/revision_list.go
```diff
@@ -30,15 +33,24 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
...
 		Long:  "List revisions for a given service.",
 		Example: `
   # List all revisions
   kn revision list
 
   # List revisions for a service 'svc1' in namespace 'myapp'
-  kn revision list -s svc1 -n myapp`,
+  kn revision list -s svc1 -n myapp
+
+  # List a all revisions in JSON output format.
```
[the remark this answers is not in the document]
navidshaikh: @savitaashture : the latest update still needs to fix these review comments, ptal

pkg/kn/commands/revision/revision_list.go
```diff
@@ -30,15 +33,24 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
...
+  kn revision list -o json
+  
+  # List a single revisions with specified web.
+  kn revision list web
+  
+  # List a single revisions in YAML output format.
```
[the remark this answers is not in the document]
savitaashture: Updated

pkg/kn/commands/revision/revision_list.go
```diff
@@ -59,22 +71,27 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
...
 			revision.GetObjectKind().SetGroupVersionKind(schema.GroupVersionKind{
 				Group:   "knative.dev",
 				Version: "v1alpha1",
 				Kind:    "revision"})
+
 			printer, err := revisionListFlags.ToPrinter()
 			if err != nil {
 				return err
 			}
+
```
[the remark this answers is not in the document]
savitaashture: Removed extra lines and updated

pkg/kn/commands/service/service_list.go
```diff
@@ -28,8 +32,20 @@ func NewServiceListCommand(p *commands.KnParams) *cobra.Command {
...
+  kn service list
+
+  # List a all services in JSON output format.
+  kn service list -o json
+
+  # List a single service with specified web.
+  kn service list web
+
+  # List a single service in YAML output format.
+  kn service list web -o yaml`,
```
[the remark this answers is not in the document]
savitaashture: Removed

pkg/kn/commands/revision/revision_list.go
savitaashture: Updated

pkg/kn/commands/revision/revision_list.go
savitaashture: Updated

pkg/kn/commands/revision/revision_list.go
```diff
@@ -59,22 +71,27 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
 				listOptions.LabelSelector = labels.Set(
 					map[string]string{serving.ConfigurationLabelKey: service}).String()
 			}
-			revision, err := client.Revisions(namespace).List(listOptions)
+
+			revision, err := getRevisionInfo(args, namespace, client, listOptions)
 			if err != nil {
 				return err
 			}
+
```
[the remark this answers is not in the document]
savitaashture: Removed

pkg/kn/commands/service/service_list.go
```diff
@@ -28,8 +32,20 @@ func NewServiceListCommand(p *commands.KnParams) *cobra.Command {
 	serviceListFlags := NewServiceListFlags()
 
 	serviceListCommand := &cobra.Command{
-		Use:   "list",
+		Use:   "list NAME",
 		Short: "List available services.",
+		Example: `
+  # List a all services.
```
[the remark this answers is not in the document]
savitaashture: Updated

rhuss:
@savitaashture could you please rebase on `master` ? This will fix the integration tests error as the test now will wait until a service is ready.

rhuss:
I think the default timeout of 60 seconds for waiting for a service to become ready is too less. That's why the tests fails. You can see that the pod is still about to spin up:

```
kne2esmoketests   pod/svc1-dnf76-deployment-56b464cdf8-gk2mz                           0/2     ContainerCreating   0          25s
```

Maybe we should generally increase the default timeout to something like 2 mins ?

rhuss:
@savitaashture can we wait with this PR until #134 to be merged, as your change gets much easier with this in place (and we would have to refactor anyway) ?

savitaashture:
> @savitaashture can we wait with this PR until #134 to be merged, as your change gets much easier with this in place (and we would have to refactor anyway) ?

@rhuss sure

navidshaikh:
> Maybe we should generally increase the default timeout to something like 2 mins ?

One of the candidate to land in kn config, but yes the default timeout can be increased. Another option, we can set `--wait-timeout` in `create` calls for CI (as it seems the success rate of e2e tests > flakiness).

rhuss:
@savitaashture #134 has been merged. Could you please rebase ?

savitaashture:
> @savitaashture #134 has been merged. Could you please rebase ?

Done

pkg/kn/commands/service/service_list_test.go
```diff
@@ -78,6 +91,55 @@ func TestServiceListDefaultOutput(t *testing.T) {
...
+	action, output, err := fakeServiceList([]string{"service", "list", "foo"}, serviceList)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if action == nil {
+		t.Errorf("No action")
+	} else if !action.Matches("list", "services") {
+		t.Errorf("Bad action %v", action)
+	}
+	assert.Check(t, util.ContainsAll(output[0], "NAME", "DOMAIN", "GENERATION", "AGE", "CONDITIONS", "READY", "REASON"))
```
navidshaikh: The DOMAIN field is being replaced in #247 with URL, one of these two PRs need to rebase and update based on which gets merged first. 

This PR is around since a while, probably we can consider updating #247 once this is in.
WDYT @rhuss ?

pkg/kn/commands/service/service_list_test.go
```diff
@@ -78,6 +91,55 @@ func TestServiceListDefaultOutput(t *testing.T) {
...
+	assert.Check(t, util.ContainsAll(output[1], "foo", "foo.default.example.com", "1"))
+}
+
+func TestServiceGetWithTwoSrvName(t *testing.T) {
```
navidshaikh: IIUC, this is just repeating test in `TestServiceGetOneOutput` and not covering additional / different scenario, we can remove this test.

savitaashture: ```TestServiceGetWithTwoSrvName``` will validate functionality if we pass 2 names.
i have refactored function.

pkg/kn/commands/service/service_list_test.go
```diff
@@ -78,6 +91,31 @@ func TestServiceListDefaultOutput(t *testing.T) {
...
+	service := createMockServiceWithParams("foo", "foo.default.example.com", 1)
+	serviceList := &v1alpha1.ServiceList{Items: []v1alpha1.Service{*service}}
+	_, _, err := fakeServiceList([]string{"service", "list", "foo", "bar"}, serviceList)
+	if err != nil {
```
navidshaikh: I think we should use https://godoc.org/gotest.tools/assert#ErrorContains without any conditional as error is expected in this case.

pkg/kn/commands/service/service_list.go
```diff
@@ -26,8 +28,17 @@ func NewServiceListCommand(p *commands.KnParams) *cobra.Command {
 	serviceListFlags := NewServiceListFlags()
 
 	serviceListCommand := &cobra.Command{
-		Use:   "list",
+		Use:   "list NAME",
```
rhuss: Please change the name to indicate an optional parameter like `[name]`

pkg/kn/commands/service/service_list_test.go
rhuss: I'm fine with either. However I suggest to merge this PR first, and then update #247 based on this.

rhuss:
@savitaashture if you update the docs a bit, could you also please squash the PR to a single commit and force push it again ? I hope that we then get rid of the CLA issue so that we can merge that PR. Thanks !

pkg/kn/commands/revision/revision_list.go
```diff
@@ -29,15 +29,21 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
 	revisionListFlags := NewRevisionListFlags()
 
 	revisionListCommand := &cobra.Command{
-		Use:   "list",
+		Use:   "list NAME",
```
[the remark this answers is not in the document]
savitaashture: [NAME] or [name] ?

PR #153, opened 2019-05-30
fix(serving): Update GVK on objects within a ResourcePrinter.

rhuss:
This is an alternate solution to #134 for fixing #133. In this commit
the update of the GVK coordinates are moved to the resource printer.

### DIFF
--- a/pkg/kn/commands/revision_describe.go
+++ b/pkg/kn/commands/revision_describe.go
@@ -16,10 +16,10 @@ package commands
 
 import (
 	"errors"
+	"github.com/knative/client/pkg/printers"
 
 	"github.com/spf13/cobra"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 )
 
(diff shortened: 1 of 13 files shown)


rhuss:
@cppforlife this is an alternative solution to #133 which moves the update of the GVK to the printer itself on every print. Tbh, I don't see much benefit over #134, one now has now instantiated a specialized ResourcePrinter around the printer obtained from cli-runtime, instead of updating the GVK before printing out the object. It has no less code and adds one layer more over the printer. 

Also, you would need now create different printers based on which scheme (v1alpha1 or v1beta1) the objects are that you want to print, instead of reusing a generic printer usable for every object.

pkg/kn/commands/revision_describe.go
```diff
@@ -51,10 +51,7 @@ func NewRevisionDescribeCommand(p *KnParams) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			revision.GetObjectKind().SetGroupVersionKind(schema.GroupVersionKind{
-				Group:   "knative.dev",
-				Version: "v1alpha1",
-				Kind:    "Revision"})
+			printer = printers.NewGvkUpdatePrinter(printer)
```
rhuss: In #134 you would call out to `serving.UpdateGroupVersionKind()` instead of wrapping the printer, which calls this method on print. Still, you have a specialized handling.

cppforlife: can we hide `printers.NewGvkUpdatePrinter(printer)` in revisionDescribePrintFlags.ToPrinter()?

rhuss: sure, let me align it with the service handling.

rhuss:
Using #134, closing this one.

PR #155, opened 2019-05-31
fix(service update): Only update fields which have been sent by server.

### LINKED ISSUE #144: Return unnnested fields when server provides unnested fields
This is the first step in [the lemonade doc](https://docs.google.com/presentation/d/1mOhnhy8kA4-K9Necct-NeIwysxze_FUule-8u5ZHmwA/edit?pli=1#slide=id.g55ca79ba88_0_58) about how to switch to the new unnested form of the API.

rhuss:
This reflects the lemonade process step1. Tests have been adapted to
verify this behaviours.

The only situation when we update field coming from the server is for
"kn service update" for envs, image and requests/limits.

All other operation are either create (here, we always have to send the
old fields), or read (get/describe).

Fixes #144.

### DIFF
--- a/pkg/kn/commands/configuration_edit_flags.go
+++ b/pkg/kn/commands/configuration_edit_flags.go
@@ -54,7 +54,13 @@ func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
 	command.MarkFlagRequired("image")
 }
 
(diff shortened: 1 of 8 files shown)


sixolet:
/lgtm
/approve
I love when I read a largeish PR and it just looks right...

PR #156, opened 2019-06-02
feat(service create): Wait for a service to be ready when its created

### LINKED ISSUE #54: Wait for the service to be Ready by default
Use `--async` to not wait.

rhuss:
* By default wait for a service to be in condition "Ready"
* Use "--async" for returning immediately (which is the current behaviour)
* --wait-timeout to specify how long to wait for Ready condition (in seconds, default: 60s)
* In sync mode, print out the service URL as a last line (so that it can be used together with `tail -1`) to extract the service URL after the service is created.
* Refactored service create code to use smaller methods which are easier to understand

Current tests has been adapted to work with async behaviour but no
new tests has been added for the new sync mode.
This is coming in a second commit (and put into an extra test file).

Fixes #54


rhuss:
Some implementation notes:

* This implementation uses a plain `watch` on the Services endpoint instead of a fully fledged Informer for simplicities reasons and because I don't consider the breakage of that watch connection as fatal. The service has been created anyway, its just a convenient method to wait until its ready.
* Printing out the URL at the end might be or might be not a good idea. On the one hand is nice to allow something like `curl $(kn service create --image bla | tail -1)` but in the case of an error, this usage pattern probably doesn't make so much sense. I'm undecided here, I probably would like a `kn service show --url` (or `kn service describe --url`) better.
* Blocking for a `kn service create --force` doesn't work probably now when the service already exists as the first status for the Ready condition will be `true-true-unknown-true` (as the first events are still about the update of the existing service). So because the code currently sees the `true` first, it already thinks that the service is ready, not knowing that it becomes unknown two events later. Not sure how to solve this problem easily. Any ideas ?

rhuss:
hmm, the integration test failure doesn't seem to be related to this PR.

rhuss:
> hmm, the integration test failure doesn't seem to be related to this PR.

Nope, it's an issue with this PR. Going to fix it now.

rhuss:
It looks like that something changes on `latest` in Knative serving. With 0.6.0 locally the e2etest runs smoothly. 

I can fix that, but this, in general, raises the question against which version of knative serving we should test. And whether we should strive for testing against multiple Knative (0.5, 0.6, ...) versions. This becomes especially important when the internal structure changes.

I'm going to fix this for latest now and test it locally against 0.6.0, too (which is not covered then by the integration test).

rhuss:
Integration test should be fixed if #155 is merged and this PR rebased on top of it

rhuss:
@sixolet rebased and resolved conflicts. Please let me know whether I'm on the right track because then I would add more tests for the sync/async behaviour and also adapt the e2e test cases for the sync creation of the service.

maximilien:
> * Printing out the URL at the end might be or might be not a good idea. On the one hand is nice to allow something like `curl $(kn service create --image bla | tail -1)` but in the case of an error, this usage pattern probably doesn't make so much sense. I'm undecided here, I probably would like a `kn service show --url` (or `kn service describe --url`) better.

What would you print out if success? Same as before “Service Xyz created”? Then adding the URL on the next line for me is a great info to the user. This also does not prevent adding: `kn service describe --url` which is a great programmatic access to the URL.

> * Blocking for a `kn service create --force` doesn't work probably now when the service already exists as the first status for the Ready condition will be `true-true-unknown-true` (as the first events are still about the update of the existing service). So because the code currently sees the `true` first, it already thinks that the service is ready, not knowing that it becomes unknown two events later. Not sure how to solve this problem easily. Any ideas ?

One idea if I understand the problem correctly is some kind of a wait limit/timeout should be applied at each Ready stage so that when `unknown` is reached it would have to timeout but hopefully a long enough that the next `Ready == true` stage is reached.

maximilien:
> I can fix that, but this, in general, raises the question against which version of knative serving we should test. And whether we should strive for testing against multiple Knative (0.5, 0.6, ...) versions. This becomes especially important when the internal structure changes.

I’d vote to stay with latest since we are still in pre-release mode.

pkg/kn/commands/service/service_create.go
```diff
@@ -53,66 +60,185 @@ func NewServiceCreateCommand(p *commands.KnParams) *cobra.Command {
...
+
+func waitForReadyCondition(watcher watch.Interface, timeout int) error {
+	defer watcher.Stop()
+	for {
+		select {
+		case <-time.After(time.Duration(timeout) * time.Second):
+			return fmt.Errorf("timeout: Service not ready after %d seconds", timeout)
+		case event, ok := <-watcher.ResultChan():
+			if !ok || event.Object == nil {
+				return errors.New("timeout while waiting for service to become ready")
```
maximilien: I guess we to deal with the `unknown` status of `Ready` the idea would be to not timeout here but instead reset the timeout clock, giving a chance to the service to become ready from a `unknown` state. This would be for every case of `true` to `unknown` transition. Thoughts?

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -51,6 +53,8 @@ func (p *ConfigurationEditFlags) AddUpdateFlags(command *cobra.Command) {
 func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
 	p.AddUpdateFlags(command)
 	command.Flags().BoolVar(&p.ForceCreate, "force", false, "Create service forcefully, replaces existing service if any.")
+	command.Flags().BoolVar(&p.Async, "async", false, "Don't wait for the service to become Ready.")
+	command.Flags().IntVar(&p.WaitTimeout, "wait-timeout", 60, "Wait for that many seconds before giving up when creating a service.")
```
sixolet: "Seconds to wait before giving up on creating the service"

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -51,6 +53,8 @@ func (p *ConfigurationEditFlags) AddUpdateFlags(command *cobra.Command) {
 func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
 	p.AddUpdateFlags(command)
 	command.Flags().BoolVar(&p.ForceCreate, "force", false, "Create service forcefully, replaces existing service if any.")
+	command.Flags().BoolVar(&p.Async, "async", false, "Don't wait for the service to become Ready.")
```
sixolet: Can we have a separate "AddConditionWaitFlags" method, in the common dir, so the code is more reusable?

pkg/kn/commands/service/service_create.go
```diff
@@ -53,66 +60,185 @@ func NewServiceCreateCommand(p *commands.KnParams) *cobra.Command {
...
+
+	if err := waitForReadyCondition(watcher, timeout); err != nil {
+		fmt.Fprintln(out)
+		return err
+	}
+	fmt.Fprintln(out, "OK")
+	return nil
+}
+
+func waitForReadyCondition(watcher watch.Interface, timeout int) error {
```
sixolet: Let's make this reusable too.

pkg/kn/commands/service/service_create.go
sixolet: I think it's ok to complicate the code in that way later and not now.

rhuss: Good idea, @maximilien but I agree with @sixolet to keep simple for now, refactor to be reusable for other resources and the open an issue for this particular feature (i.e. don't timeout as long as in `Unknown` state)

rhuss:
> What would you print out if success? Same as before “Service Xyz created”? Then adding the URL on the next line for me is a great info to the user. This also does not prevent adding: `kn service describe --url` which is a great programmatic access to the URL.

The output looks like

```
./kn service create greeter --image quay.io/rhdevelopers/knative-tutorial-greeter:quarkus
Service 'greeter' successfully created in namespace 'default'.
Waiting for service greeter to become ready ... OK

Service URL:
http://greeter.default.apps.rhuss-dev.devcluster.openshift.com
```

I.e. the URL as a _separate last line_ which mean that you can use it like in `url=$(kn service create --image ... | tail -1)`

The question is, whether we should put that as a kind of API contract (i..e taking care that it will be the same in future versions to not break scripts using that schema), or whether we say that's too much and people should use a `kn service show <name> --url` after a `create` (btw the more I type it the more I would like `show` over `describe`. It's just shorter ;-)

rhuss:
> > * Blocking for a `kn service create --force` doesn't work probably now when the service already exists as the first status for the Ready condition will be `true-true-unknown-true` (as the first events are still about the update of the existing service). So because the code currently sees the `true` first, it already thinks that the service is ready, not knowing that it becomes unknown two events later. Not sure how to solve this problem easily. Any ideas ?
> 
> One idea if I understand the problem correctly is some kind of a wait limit/timeout should be applied at each Ready stage so that when `unknown` is reached it would have to timeout but hopefully a long enough that the next `Ready == true` stage is reached.

The flow of events for an existing service and `kn service create --force` of the same service is something like

```
Ok
Ok
Unknown
Ok
```

While we could introduce an initial wait, it's really hard to exactly hit the interval of the `Unknown` phase predictably for every service. I really would like to avoid to rely on timing behaviour. I have to look a bit close into these first two events with status == ok, so maybe they are unique for --force and could be filtered out.

rhuss:
> > I can fix that, but this, in general, raises the question against which version of knative serving we should test. And whether we should strive for testing against multiple Knative (0.5, 0.6, ...) versions. This becomes especially important when the internal structure changes.
> 
> I’d vote to stay with latest since we are still in pre-release mode.

It's already fixed, it was an issue with old and new v1alpha1 fields which were solved by the PR solving #144

pkg/kn/commands/service/configuration_edit_flags.go
rhuss: The only issue is that all the other help texts are in Imperative format ("Create ..." / "Don't wait ...") and not nominal style. So should we change all to nominal style, like in

```
--force                            Forceful service creation,  replacing an existing service
--async                           Service creation in the background
--wait-timeout <secs> Seconds to wait before giving up on creating the service
```

or we keep imperative style for `--wait-timeout` but reword to e.g.

```
--wait-timeout <60> Set maximal seconds to wait when creating a service
```

" Before giving up on creating the service" is also not technically correct as the service still gets possibly created, it's just that the CLI command returns. The timeout happens most of the time if there is a large image to pull on first service creation.

pkg/kn/commands/service/service_create.go
maximilien: OK fair enough. Cool.

maximilien:
> The question is, whether we should put that as a kind of API contract (i..e taking care that it will be the same in future versions to not break scripts using that schema), or whether we say that's too much and people should use a `kn service show <name> --url` after a `create` (btw the more I type it the more I would like `show` over `describe`. It's just shorter ;-)

+1 for `show` over `describe` which I always misspell for some reason.

rhuss:
Another question: When `kn service create` fails because of an unknown image, which you will see for a synchronous create:

```
kn service create random-generator --image k8spatterns/random-generator-quarkus:1.0
Service 'random-generator' successfully created in namespace 'default'.
Waiting for service random-generator to become ready ...
RevisionFailed: Revision "random-generator-7vn8w" failed with message: Unable to fetch image "k8spatterns/random-generator-quarkus:1.0": MANIFEST_UNKNOWN: manifest unknown; map[Tag:1.0].
```

then the service is still existent, but in error condition. If you retry with a corrected image (without --force) you get an error, so that you have to delete that service first before your retry (or use --force).

@sixolet whdyt, should we _remove_ the service again when it enters an error condition like here ? Or is it better to keep it so that a post mortem could be done.

Regardless what, we should add to the error message what happened to the service (still there or deleted).

pkg/kn/commands/service/configuration_edit_flags.go
sixolet: ... right "before giving up on waiting for the Service to be ready"

pkg/kn/commands/service/service_create.go
```diff
@@ -53,66 +60,185 @@ func NewServiceCreateCommand(p *commands.KnParams) *cobra.Command {
...
+		select {
+		case <-time.After(time.Duration(timeout) * time.Second):
+			return fmt.Errorf("timeout: Service not ready after %d seconds", timeout)
+		case event, ok := <-watcher.ResultChan():
+			if !ok || event.Object == nil {
+				return errors.New("timeout while waiting for service to become ready")
+			}
+			service := event.Object.(*serving_v1alpha1_api.Service)
+			for _, cond := range service.Status.Conditions {
+				if cond.Type == apis.ConditionReady {
```
sixolet: Ah, we have one more thing to make sure we do: consider any situation where Status.ObservedGeneration != Metadata.Generation to be equivalent to Ready still being Uknown.

rhuss:
I extracted now the wait logic in a service-independent and also cobra-independent package (top-level `wait`).

The option is now renamed to `--no-wait` to fit nicely to the common pattern for gnu getopts where `--no-` prefix is used for negating binary options.

Also, one can switch the default mode, so when the default mode would be async, then an option `--wait` is added. This default behaviour could be a candidate to be configured in a ~/.kn/config.yml.

For now only `--no-wait` and `--wait-timeout` is enabled.

------

@sixolet I added the tested of configured generation vs. observed generation, but I never managed to get an event where those two values differ. I tried to find out, in which case these values can be different, and whether these values are already consolidated before the first event is sent. Do you have a pointer for me where I could learn a bit more about the usage of these generation values ?

Also, for extracting the generation values I used `unstructured` to work typeless on any resource. If this is not acceptable I can change that to an extractor callback based method like I do for getting to the conditions in a general fashion.

Otherwise, I added quite some tests, but I haven't touched the e2e tests yet, which at least could be optimized to remove the sleep values (which are not needed in because waiting is the default)

maximilien:
> Otherwise, I added quite some tests, but I haven't touched the e2e tests yet, which at least could be optimized to remove the sleep values (which are not needed in because waiting is the default)

And soon enough you should be able to do it in Golang :)

docs/cmd/kn_service_create.md
```diff
@@ -47,8 +47,10 @@ kn service create NAME --image IMAGE [flags]
       --max-scale int            Maximal number of replicas.
       --min-scale int            Minimal number of replicas.
   -n, --namespace string         List the requested object(s) in given namespace.
+      --no-wait                  Don't wait for service to become ready.
```
cppforlife: i would prefer if we change it to --wait=bool with a desired default.

pkg/wait/wait_for_ready.go
```diff
@@ -0,0 +1,158 @@
...
+
+func (w *waitForReadyConfig) waitForReadyCondition(watcher watch.Interface, name string, timeout int) error {
+	defer watcher.Stop()
+	for {
+		select {
+		case <-time.After(time.Duration(timeout) * time.Second):
+			return fmt.Errorf("timeout: %s '%s' not ready after %d seconds", w.kind, name, timeout)
+		case event, ok := <-watcher.ResultChan():
+			if !ok || event.Object == nil {
+				return fmt.Errorf("timeout while waiting for %s '%s' to become ready", w.kind, name)
```
cppforlife: this case is not an error but rather it's necessary to retry watching. k8s will expire watchers after a few minutes consistently. see example here: https://github.com/k14s/kapp/blob/0c91e8a2208ff9a8f9d92f8c01d20dc9dd21256b/pkg/kapp/resources/pod_watcher.go#L64

pkg/kn/commands/wait_flags.go
```diff
@@ -0,0 +1,43 @@
...
+
+	// Don't access directly, use IsWait()
+	noWaitFlag bool
+	waitFlag   bool
+}
+
+// Add flags which influence the sync/async behaviour when creating or updating
+// resources. Set `waitDefault` argument if the default behaviour is synchronous.
+// Use `what` for describing what is waited for.
+func (p *WaitFlags) AddWaitFlags(command *cobra.Command, waitDefault bool, waitTimeoutDefault int, what string) {
```
cppforlife: i dont think we need waitDefault

docs/cmd/kn_service_create.md
rhuss: I change this. This `--no-wait` style was a reminiscence to GNU getopts where long boolean options are used to have `--with-` and `--no-` prefixes. But guess that's not so common anymore.

pkg/wait/wait_for_ready.go
rhuss: I see the point. Let's change that from an error to an informal message and add a retry mechanism later (as mentioned in https://github.com/knative/client/pull/156#discussion_r291419076), then with a real timeout.

pkg/kn/commands/wait_flags.go
rhuss: I probably went a bit too far here. The idea was that the default behaviour (sync/async) could be configured in configuration file, but on second thought that is not a good idea. A command line should be completely self-container without referring to any hidden state (so that e.g. c&p works always).

I remove that.

docs/cmd/kn_service_create.md
rhuss: @cppforlife @sixolet Just thought about the naming of the option again. If we make waiting the default value, then a `--wait=true` doesn't make much sense and doesn't provide any extra value. Also, having an option with a string parameter allows for additional errors like providing a non-standard value (like `no`, `off`, ....). So I'd really prefer a boolean option here.

My suggestion would be to fall back to the original suggestion of `--async`, or any other boolean toggle like `--no-wait`, `--background`, `--apply`, ...

pkg/wait/wait_for_ready.go
rhuss: I added now retry mechanism when the watch closes. Not really sure whether this will happen often as I use a timeout for the list watch which is 30s longer than the user provided timeout. But as a watch can stop anytime anyway, its indeed saver to retry.

docs/cmd/kn_service_create.md
navidshaikh: IMO, `--no-wait` reads simple and should be easy to use with default value of false.

rhuss:
@cppforlife @sixolet I think we are ready to go and that I adressed all open change requests:

* Fixed the help message
* Added a retry mechanism when the list watch stops
* Changed option from --no-wait to --async (and removed --wait option)
* Removed the switchable default mode (e.g. that one could have used also an async behaviour as default)

One minor issue which I would adress after a merge of this PR is, whether and how we should show a service URL. I think its need that you see that URL when commands returns (and it can show the URL now as the route is then alreaday in place). 

However, I'm not so sure about the how. I'd like that it would be easy to use the URL also in scripting so I though about a combination of `--url` and `--quiet` (from #182 )

* `kn service create ...` would not show the URL, just a success message.
* `kn service create ... --url` would should the URL as it is now
* `kn service create ... --quiet` would show nothing, but return with an error code
* `kn service create ... --quiet --url` would *only* show the URL.

The last command would make it super easy to use the URL in scripts so that you can do things like `curl $(kn service create greeter --image greeter --quiet --url)`

Alternative to this suggestion could be 

* `--url` shows *only* the URL
* `--quiet` shows allways nothing
* Default would showing the URL as it is now

I think I like the second option better, but as mentioned I would address this in a seperate issue/PR as this PR already has grown quite a bit.

PR #157, opened 2019-06-02
feat(service create/update): Add support of minScale/maxScale/target

### LINKED ISSUE #151: add support for setting containerConcurency
It would be nice to be able to set the `containerConcurrency` field via `kn`:
```
kind: Service
metadata:
  name: helloworld
spec:
  runLatest:
    configuration:
      revisionTemplate:
        spec:
          container:
            image: ${APP_IMAGE}
          containerConcurrency: 1
```
See last line.

Something like `--max-concurrency=1`

rhuss:
Autoscaler annotations are added to the revision template if
--min-scale / --max-scale / --target are provided to `kn service create` or
`kn service update`

Fixes #151

~This PR builds on top of #156 but can be back-ported on master if desired.
For review I recommend to look at commit efb23b778cf2104337fc6e4dc7d7e2b67d084465 first, which contains the delta.~

The PR is now based on upstream master only.

### DIFF
--- a/docs/cmd/kn_service_create.md
+++ b/docs/cmd/kn_service_create.md
@@ -36,12 +36,16 @@ kn service create NAME --image IMAGE [flags]
 ### Options
 
 ```
+      --concurrency-limit int    Hard Limit of concurrent requests to be processed by a single replica.
+      --concurrency-target int   Recommendation for when to scale up based on the concurrent number of incoming request. Defaults to --concurrency-limit when given.
   -e, --env stringArray          Environment variable to set. NAME=value; you may provide this flag any number of times to set multiple environment variables.
       --force                    Create service forcefully, replaces existing service if any.
   -h, --help                     help for create
       --image string             Image to run.
       --limits-cpu string        The limits on the requested CPU (e.g., 1000m).
       --limits-memory string     The limits on the requested CPU (e.g., 1024Mi).
+      --max-scale int            Maximal number of replicas.
+      --min-scale int            Minimal number of replicas.
   -n, --namespace string         List the requested object(s) in given namespace.
       --requests-cpu string      The requested CPU (e.g., 250m).
       --requests-memory string   The requested CPU (e.g., 64Mi).
--- a/docs/cmd/kn_service_update.md
+++ b/docs/cmd/kn_service_update.md
@@ -24,11 +24,15 @@ kn service update NAME [flags]
 ### Options
 
 ```
+      --concurrency-limit int    Hard Limit of concurrent requests to be processed by a single replica.
+      --concurrency-target int   Recommendation for when to scale up based on the concurrent number of incoming request. Defaults to --concurrency-limit when given.
   -e, --env stringArray          Environment variable to set. NAME=value; you may provide this flag any number of times to set multiple environment variables.
   -h, --help                     help for update
       --image string             Image to run.
       --limits-cpu string        The limits on the requested CPU (e.g., 1000m).
       --limits-memory string     The limits on the requested CPU (e.g., 1024Mi).
+      --max-scale int            Maximal number of replicas.
+      --min-scale int            Minimal number of replicas.
   -n, --namespace string         List the requested object(s) in given namespace.
       --requests-cpu string      The requested CPU (e.g., 250m).
       --requests-memory string   The requested CPU (e.g., 64Mi).
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -30,6 +30,10 @@ type ConfigurationEditFlags struct {
 	Env                        []string
 	RequestsFlags, LimitsFlags ResourceFlags
 	ForceCreate                bool
+	MinScale                   int
+	MaxScale                   int
+	ConcurrencyTarget          int
+	ConcurrencyLimit           int
 }
 
 type ResourceFlags struct {
@@ -46,6 +50,10 @@ func (p *ConfigurationEditFlags) AddUpdateFlags(command *cobra.Command) {
 	command.Flags().StringVar(&p.RequestsFlags.Memory, "requests-memory", "", "The requested CPU (e.g., 64Mi).")
 	command.Flags().StringVar(&p.LimitsFlags.CPU, "limits-cpu", "", "The limits on the requested CPU (e.g., 1000m).")
 	command.Flags().StringVar(&p.LimitsFlags.Memory, "limits-memory", "", "The limits on the requested CPU (e.g., 1024Mi).")
+	command.Flags().IntVar(&p.MinScale, "min-scale", 0, "Minimal number of replicas.")
+	command.Flags().IntVar(&p.MaxScale, "max-scale", 0, "Maximal number of replicas.")
+	command.Flags().IntVar(&p.ConcurrencyTarget, "concurrency-target", 0, "Recommendation for when to scale up based on the concurrent number of incoming request. Defaults to --concurrency-limit when given.")
+	command.Flags().IntVar(&p.ConcurrencyLimit, "concurrency-limit", 0, "Hard Limit of concurrent requests to be processed by a single replica.")
 }
 
 func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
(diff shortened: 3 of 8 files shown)


rhuss:
Integration test should be fixed if #155 is merged and this PR rebased on top of it

pkg/kn/commands/service_create.go
```diff
@@ -53,70 +60,193 @@ func NewServiceCreateCommand(p *KnParams) *cobra.Command {
...
+	if err != nil {
+		return nil, err
+	}
+	return &service, nil
+}
+
+func showUrl(client serving_v1alpha1_client.ServingV1alpha1Interface, serviceName string, namespace string, out io.Writer) error {
+	service, err := client.Services(namespace).Get(serviceName, v1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("cannot fetch service %s in namespace %s for extracting the URL", serviceName, namespace)
```
maximilien: Might read better with: "... to extract its URL" instead of "...for extracting the URL"

docs/cmd/kn_service_create.md
```diff
@@ -36,15 +36,20 @@ kn service create NAME --image IMAGE [flags]
...
   -h, --help                     help for create
       --image string             Image to run.
       --limits-cpu string        The limits on the requested CPU (e.g., 1000m).
       --limits-memory string     The limits on the requested CPU (e.g., 1024Mi).
+      --max-scale int            Maximal number of replicas.
+      --min-scale int            Minimal number of replicas.
   -n, --namespace string         List the requested object(s) in given namespace.
       --requests-cpu string      The requested CPU (e.g., 250m).
       --requests-memory string   The requested CPU (e.g., 64Mi).
+      --target int               Number of in-flight requests per pod.
```
duglin: The issues talks about containerConcurrency, so I think this should be named something like that. The "target" config option is different - we can add a flag for that too if people want it, but my original request is for CC.

rhuss: tbh, I think we can add all autoscaling processing annotations in one go. But I'm fine of course to remove `target` if we agree that this is not something we want to expose via the CLI (althouhg I believe every part of the public Knative API which is put into annotation *should* be available as CLI option as this is hard to find out elsewhere as it is not part of the CRD schema and nobody memorizes the full name of annotation keys)

duglin: I'm +1 for exposing all of them too - I just didn't want this PR to go thru w/o addressing what the issue was asking for :-)

rhuss: got you, valid point.

maximilien: The name `target` (unless I am missing something or has historical meaning) is vague and not meaningful? Why can't this be `no-in-flights` or similar?

rhuss: Agreed, its a bit vague and just copier over literally from the annotation (`autoscaling.knative.dev/target`).  "no-in-flights" is probably also not ideal (I immediately, ah that's the opposite of `allow-in-flights` ;-). Let's find a better name ...

duglin: I would suggest that we be consistent (even if not a 100% direct match) with what people will see in the yaml.  So, for now I would stick with `target` but suggest a new name for the yaml & serving code.

Having said that.... do I think `concurrency` might be good enough for the flag I was asking for - instead of `containerConcurrency`.  I think the word `container` is kind of redundant. Maybe serving can change this too at the same time  :-)

duglin: were you going to add that flag?

rhuss: > were you going to add that flag?

These concurrency flags are added to `kn service create` and `kn service update`. There is the general problem that for changing these annotations on update, this will trigger a new revision, and don't touch the old revision. This means if you have used `minScale` then this revision will always spin up that many pods even when its not the latest revision anymore (revisions are immutable). The only thing to get this back to 0 is to fiddle around with the autoscaler CRD. There is some discussions over there to make metadata like annotations mutable on revision instances, but as long as this doesn't happen we have to deal with that (i.e. don't modify the annotation on the running Revision but on the Service/Configuration's Revision template). See https://github.com/knative/serving/issues/3547#issuecomment-498206979 for the discussion over there in serving.

rhuss: For the naming: 

I like `--min-scale` and `--max-scale` as it's concise and pretty clear what it means. Alternatively we could call it `--min-replicas` & `--max-replicas` to reflect the Deployment fields affected (or even `--min-pods` and `--max-pods`). However this suggestion again connects to K8s which we shouldn't do ("replicas" and even more "pods" is a concept occupied by K8s). So I'm fine with `--min-scale` & `--max-scale`.

Another option would be `--scale-max` and `--scale-min` which would put the options nicely together, e.g. in an alphabetically ordered help options list or documentation. This is important when your option list gets long (ever tried `jx create cluster minikube --help` ;-P), and I think for create and update it will still grow.

For `--target` I agree that it is to unspecific as an option name. 

My suggestions would be (with preference in that order): 

* `--scale-target 10` (would fit to a `--scale-min` and `--scale-max` (but could be misread that it's a target to reach 10, but afaiu its just the number of concurrent requests after that a scale up event happen)
* `--scale-after-requests 10` (which reads nicely like in `scale after 10 request`) 
* `--scale-after 10`
* `--concurrency-target 10`
* `--scale-in-flight 10`(not sure whether this clear enough)
* `--scale-after-concurrent-requests 10`

markusthoemmes: For clarity, `containerConcurrency` and `target` are two different things. The first is an enforced limit, the second is a "soft" target for the autoscaler. In the absence of the second, the second is inferred from the first.

In that sense I'd opt for:

- `--concurrency-limit` to define the former
- `--concurrency-target` to define the latter. We should encode `concurrency` into here as for the KPA the default will be `concurrency`. If we also want to expose the HPA as an autoscaler, things will get more tricky here though.

I think that nicely encodes the difference in the name. The CLI docs could say that `--concurrency-target` defaults to `--concurrency-limit` if not passed.

rhuss: Ah, I see. There can be different types of autoscaling triggers, which currently is based on concurrent requests (but could be e.g. CPU consumption when using the HPA). But maxScale and minScale are more a global scaling (boundary) option which applies regardless of the type of scaling.

So my suggestion now would be:

* `--max-scale` and `--min-scale` (as in this PR)
* `--concurrency-target` (renaming the option)
* `--concurrency-limit` (adding to this PR, by setting the proper field in the spec)

rhuss: @markusthoemmes Just saw that `containerConcurrency` is part of `PodAutoscaler` which is not part of the public API. So is it correct, that this datum can't be changed from a client like `kn` ? Or is there a corresponding field on `Service` which the get put into the `PodAutoscaler` resource when a service is created ?

For now, I think for now we have to ommit `--concurrency-limit`

rhuss:
@cppforlife @sixolet From my POV we are good for review and eventually merge with this PR.

* Rebased to latest refactoring
* Introduced `--min-scale`, `--max-scale`, `--concurrency-target` for `service create` and `service update`
* Cleaned up and backported this feature on master only (was based on another PR before)

The introduction of `--concurrency-limit` is postponed until we know how to set this on a `PodAutoscaler` resource (if this possible at all).

duglin:
containerConcurrency is part of a KnService so kn should expose the option

rhuss:
@duglin Ah, I see. It's on the RevisionSpec, not ServiceSpec. Thanks for the heads up. So I will add `--concurrency-limit` now, too.

duglin:
> @duglin Ah, I see. It's on the RevisionSpec, not ServiceSpec. Thanks for the heads up. So I will add `--concurrency-limit` now, too.

Thanks!!

pkg/kn/commands/service/service_update.go
```diff
@@ -36,8 +37,9 @@ func NewServiceUpdateCommand(p *commands.KnParams) *cobra.Command {
   kn service update mysvc --requests-cpu 500m --limits-memory 1024Mi`,
 		RunE: func(cmd *cobra.Command, args []string) (err error) {
 			if len(args) != 1 {
-				return errors.New("requires the service name.")
+				return errors.New("'kn update' requires the service name")
```
sixolet: This isn't an accurate rendition of the actual command, and these changes seem unrelated to what you're doing in the rest of the PR. Can you make a separate PR?

rhuss: fair point, will do

rhuss: cleaned up and rebased

PR #160, opened 2019-06-03
feat(build.sh): Add a watch mode for automatic recompilation

rhuss:
This features uses and requires https://github.com/AgentCosmic/xnotify
and should work on any platform

Use it with

* Watch & compile: build.sh --watch
* Watch, compile, test: build.sh --watch --test
* Watch, compile, test & verbose output: build.sh --watch --test --verbose

(shortcuts -w && -t can be used, too)


rhuss:
@cppforlife @sixolet Could I please have a quick review and potential lgtm ? It's really a small change, for developer convenience. It's totally harmless as it doesn't have to be used and doesn't impact the compilation or release process. In a next step I would add a tempfs for watching which increase compile times ~40% as reported in https://daltontan.com/how-to-compile-go-code-faster-with-ram-disk/24

PR #162, opened 2019-06-03
Update go.mod and go.sum by test/presubmit-tests.sh

nak3:
This changes are updated by running `test/presubmit-tests.sh` against HEAD.

Fixes https://github.com/knative/client/issues/163


go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/google/go-cmp v0.3.0 // indirect
 	github.com/google/go-containerregistry v0.0.0-20190503220729-1c6c7f61e8a5 // indirect
 	github.com/google/gofuzz v1.0.0 // indirect
+	github.com/google/licenseclassifier v0.0.0-20190501212618-47b603fe1b8c // indirect
```
rhuss: I wonder from where these indirect dependencies come from, as we don't change direct dependencies in this PR. Any ideas ?

nak3: I'm not exactly sure about the `go mod`'s mechanism. But I am thinking that following code updates the dependencies.

https://github.com/knative/client/blob/b6a8fa9213e40c1059d39928ca09a016dffc2eec/vendor/github.com/knative/test-infra/scripts/library.sh#L370

When I ran `test/presubmit-tests.sh`, the code is updated after this message.

```
go: finding github.com/google/licenseclassifier latest
```

rhuss: Ah, ok. That's probably also why there is an `// indirect` comment which indicates that the dependencies don't come directly by being used in the source. I thought it was only about transitive dependencies but looks like the same mechanism is used for dependencies coming from other sources (like this `go get`). 

I really wished `go mod` would clearly separate the manual declared dependencies (or those suggested by examing the source code) from transitive dependencies, as Maven does.

Could we please remove the `// indirect` comment as often during the update I'm removing all `// indirects` and let `go mod` figure them out a fresh ? At the end, it's a direct dependency anyway for us.

nak3: Hmm... it looks like it is not possible. Even if we remove the `// indirect` comment manually from `go.mod`, it is added automatically whenever we run `test/presubmit-tests.sh`.

rhuss: ok, then may it be like this and I'm fine with the inevitable ;-) 

/lgtm

PR #169, opened 2019-06-07
fix(service update): Print success message after successful update

rhuss:


rhuss:
Hmm, I really don't understand how test coverage could decrease with this PR:

* I added 1 line and removed 2 in code
* The new line is always teste also by existing tests
* I added a new assertion (but that doesn't count to coverage, I know)

But no idea how @knative-metrics-robots measure coverage. Nevertheless, I question the result.

rhuss:
> But no idea how @knative-metrics-robots measure coverage. Nevertheless, I question the result.

Ok, I see. Because I _removed_ also two lines which were already covered, therefore the coverage dropped

```
c        c-2
--  >  ----
t         t-2
```

is true as along `t > c`.

pkg/kn/commands/service/service_update.go
```diff
@@ -55,19 +56,19 @@ func NewServiceUpdateCommand(p *commands.KnParams) *cobra.Command {
 			}
 			service = service.DeepCopy()
 
-			err = editFlags.Apply(service, cmd)
-			if err != nil {
+			if err = editFlags.Apply(service, cmd); err != nil {
```
cppforlife: dont think we are switching to the collapsed style judging by the rest of the code.

pkg/kn/commands/service/service_update_test.go
```diff
@@ -108,6 +109,10 @@ func TestServiceUpdateImage(t *testing.T) {
 	} else if template.Spec.DeprecatedContainer.Image != "gcr.io/foo/quux:xyzzy" {
 		t.Fatalf("wrong image set: %v", template.Spec.DeprecatedContainer.Image)
 	}
+
+	if !strings.Contains(output, "update") {
```
cppforlife: might as well test fuller message.

pkg/kn/commands/service/service_update.go
rhuss: fair point, shouldn't touch existing code for good (aka better) reasons.

pkg/kn/commands/service/service_update_test.go
rhuss: will anchor on some more keywords here.

cppforlife: @rhuss why no just expect what is being printed "Service '...' updated in namespace ..."?

rhuss: Because I want to focus on the important context provided by an error message, not on the exact wording. That's why I don't do == assert on error/success messages. That also helps me in focussing on what should be *contained* in good messages:

* All context available at that point in code (i.e. the input parameters of the method)
* A suggestion how to get over an error (if its an error message)
* Keyword of what has happened, reflecting the action performed ("update").

Side effect: Your tests become more robust as they don't depend on the wording and whether there is a comma or not (except you insist on that). In TDD world you would first write the test with the things you would expect to be included in the message, and then write the message. Not write the message first and just blindly check for it (that test doesn't hasn't much expressiveness anyway)

rhuss:
@cppforlife Anything missing for this PR to get merged ?

pkg/kn/commands/service/service_update_test.go
cppforlife: @rhuss one of the important aspects of a test is readability. personally i dont find following code readable because its intent is not clear. it also ends up more fragile as it may produce false positives.

```
	if !strings.Contains(strings.ToLower(output), "update") ||
		!strings.Contains(output, "foo") ||
		!strings.Contains(strings.ToLower(output), "service") ||
		!strings.Contains(strings.ToLower(output), "namespace") ||
		!strings.Contains(output, "bar") {
		t.Fatalf("wrong or no success message: %s", output)
	}
```

however preferred style of these types of assertions is quite subjective so i wont dwell on it too much.

Issue #168, opened 2019-06-07
Test against most recent Serving release && HEAD


sixolet:
Right now we e2e test against one version of serving. We should test against the most recent release AND head, so we can make sure to preserve support for Serving as it advances.

PR #170, opened 2019-06-07
Add the support to test against specified release of Knative serving

houshengbo:
This PR adds an environment variable $KNATIVE_VERSION for both e2e-tests.sh and presubmit-tests.sh, in order to specify the version of Knative serving to be installed.

Closes: #168

### DIFF
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -24,6 +24,10 @@
 # project $PROJECT_ID, start Knative serving, run the tests and delete
 # the cluster.
 
+# If you call this script after configuring the environment variable
+# $KNATIVE_VERSION with a valid release, e.g. 0.6.0, Knative serving
+# of this specified version will be installed in the Kubernetes cluster, and
+# all the tests will run against Knative serving of this specific version.
 source $(dirname $0)/../vendor/github.com/knative/test-infra/scripts/e2e-tests.sh
 
 # Helper functions.
@@ -35,17 +39,25 @@ function cluster_setup() {
 }
 
 function knative_setup() {
-  start_latest_knative_serving
+  local version=${KNATIVE_VERSION:-latest}
+  header "Installing Knative serving (${version})"
+
+  if [ "${version}" = "latest" ]; then
+    start_latest_knative_serving
+  else
+    start_release_knative_serving "${version}"
+  fi
 }
 
 # Add local dir to have access to built kn
 export PATH=$PATH:${REPO_ROOT_DIR}
+export KNATIVE_VERSION=${KNATIVE_VERSION:-latest}
 
 # Script entry point.
 
 initialize $@
 
-header "Running tests"
+header "Running tests for Knative serving $KNATIVE_VERSION"
 
 go_test_e2e ./test/e2e || fail_test
 
--- a/test/presubmit-integration-tests-latest-release.sh
+++ b/test/presubmit-integration-tests-latest-release.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+# Copyright 2018 The Knative Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script is used in Knative/test-infra as a custom prow job to run the
+# integration tests against Knative serving of a specific version. We
+# currently take 0.6.0 as the latest release version.
+
+export KNATIVE_VERSION="0.6.0"
+$(dirname $0)/presubmit-tests.sh --integration-tests
--- a/test/presubmit-tests.sh
+++ b/test/presubmit-tests.sh
@@ -24,9 +24,15 @@
 # Markdown linting failures don't show up properly in Gubernator resulting
 # in a net-negative contributor experience.
 # Tracked by https://github.com/knative/test-infra/issues/428
+
+# If you call this script after configuring the environment variable
+# $KNATIVE_VERSION with a valid release, e.g. 0.6.0, Knative serving
+# of this specified version will be installed in the Kubernetes cluster, and
+# all the tests will run against Knative serving of this specific version.
 export DISABLE_MD_LINTING=1
 
 export GO111MODULE=on
+export KNATIVE_VERSION=${KNATIVE_VERSION:-latest}
 source $(dirname $0)/../vendor/github.com/knative/test-infra/scripts/presubmit-tests.sh
 
 # We use the default build, unit and integration test runners.


test/e2e-tests.sh
```diff
@@ -34,29 +36,8 @@ function cluster_setup() {
...
-./kn service create hello --force --image gcr.io/knative-samples/helloworld-go -e TARGET=Awesome || fail_test
-./kn service create foo --force --image gcr.io/knative-samples/helloworld-go -e TARGET=foo || fail_test
-sleep 5
-./kn revision get || fail_test
-./kn service get || fail_test
-./kn service describe hello || fail_test
-./kn service delete hello || fail_test
-./kn service delete foo || fail_test
-success
+export KNATIVE_SERVING_VERSION=$PREVIOUS_RELEASE
```
rhuss: To be more concrete, my suggestion would be something along those lines:

* e2e-tests.sh:

```bash
sourcce e2e-commons.sh
for version in latest $PREVIOUS_RELEASE; do
   echon "===== Starting E2E tests for $version
   run_tests $version
done

```

* e2e-common.sh:

```bash
run_tests() {
  local $version=${1}
  export KNATIVE_SERVING_VERSION=$version
  initialize

  header "[$version]: Running tests"

  kn_tests  
}


kn_tests() {
  ./kn service create hello --image gcr.io/knative-samples/helloworld-go -e TARGET=Knative ||   fail_test
  sleep 5
  ./kn service get || fail_test
  ./kn service update hello --env TARGET=kn || fail_test
  sleep 3
  ./kn revision get || fail_test
  ./kn service get || fail_test
  ./kn service create hello --force --image gcr.io/knative-samples/helloworld-go -e   TARGET=Awesome || fail_test
  ./kn service create foo --force --image gcr.io/knative-samples/helloworld-go -e TARGET=foo ||   fail_testa
  sleep 5
  ./kn revision get || fail_test
  ./kn service get || fail_test
  ./kn service describe hello || fail_test
  ./kn service delete hello || fail_test
  ./kn service delete foo || fail_test
  success
}
```

test/e2e-common.sh
```diff
@@ -0,0 +1,85 @@
...
+
+DEFAULT_BUILD=latest
+
+# This is the variable saving the link to the YAML file to create and delete knative serving.
+KNATIVE_SERVING_RELEASE_YAML=""
+
+# Helper functions.
+
+function knative_setup() {
+  start_release_knative_serving "$KNATIVE_SERVING_VERSION"
```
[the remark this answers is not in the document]
houshengbo: This function knative_setup
is called by source $(dirname $0)/../vendor/github.com/knative/test-infra/scripts/e2e-tests.sh.

It has to be defined after source $(dirname $0)/../vendor/github.com/knative/test-infra/scripts/e2e-tests.sh. Am I correct?

test/e2e-tests.sh
```diff
@@ -34,29 +36,8 @@ function cluster_setup() {
   go build -v -mod=vendor ./cmd/... || return 1
 }
 
-function knative_setup() {
-  start_latest_knative_serving
-}
+export KNATIVE_SERVING_VERSION=latest
+source $(dirname $0)/e2e-common.sh
```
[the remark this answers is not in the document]
houshengbo: Done.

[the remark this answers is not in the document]
houshengbo: Done.

test/e2e-common.sh
```diff
@@ -0,0 +1,83 @@
...
+DEFAULT_BUILD=latest
+
+# This is the variable saving the link to the YAML file to create and delete knative serving.
+KNATIVE_SERVING_RELEASE_YAML=""
+
+# Install the stable release Knative/serving in the current cluster.
+# This method is copied from github.com/knative/test-infra/scripts/e2e-tests.sh, since the original var
+# $KNATIVE_SERVING_RELEASE is readonly. In order to make it changeable, we rewrite the method with another
+# variable $KNATIVE_SERVING_RELEASE_YAML.
+function start_release_knative_serving() {
```
rhuss: I see your motivation but we shouldn't do this, because:

* We don't benefit from changes in this method upstream
* When the overridden function gets renamed the script fails silently (because this method is just not called).

Instead we should push a change upstream to make the version to install parameterizable either via an argument or an environment variable.

Tbh I think its even a fault that this method _can_ be overridden. This subtle coupling is really hard to debug.

Let me open an issue in the testing upstream repo.

houshengbo: Yep, it is good to see test-infra is able to support the installation of custom release number.

test/e2e-common.sh
```diff
@@ -0,0 +1,76 @@
...
+  # Add local dir to have access to built kn
+  export PATH=$PATH:${REPO_ROOT_DIR}
+  initialize $@
+  header "[Knative serving version of $KNATIVE_SERVING_VERSION]: Running tests"
+  go_test_e2e ./test/e2e || fail_test
+  success
+}
+
+echo "===== Starting E2E tests for Knative serving of the release $KNATIVE_SERVING_VERSION"
+run_tests $@
```
rhuss: This is the function call which will trigger everything. Up to this point, everything needs to be defined/sourced.

rhuss: As I expect e22-common.sh to be a kind of library it should also only contain function definitions but must not call or execute any command when being included.

test/e2e-tests.sh
```diff
@@ -24,29 +24,11 @@
...
-initialize $@
-
-header "Running tests"
-
-go_test_e2e ./test/e2e || fail_test
-
-success
+for version in "${listOfRelease[@]}"; do
+   export KNATIVE_SERVING_VERSION=$version
+   $(dirname $0)/e2e-common.sh
```
rhuss: Instead of calling out to e2e-commons.sh, call here `run_tests $@` and source all function definitions before.

test/e2e-common.sh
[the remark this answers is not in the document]
rhuss: Tbh, I don't think that e2e-tests.sh execute any codes, but only defines functions which can be called later. Otherwise `knative_setup` would have to be defined before sourcing this file

As a rule of thumb: Any shell script which is source must only define functions or global environment variables (if they have to).

rhuss:
The integration tests fails with a 

```
I0617 01:33:17.642] 2019/06/17 01:33:17 process.go:155: Step '/go/src/github.com/knative/client/test/e2e-common.sh --run-tests' finished in 1m17.206419171s
I0617 01:33:17.658] 2019/06/17 01:33:17 main.go:312: Something went wrong: encountered 1 errors: [error during /go/src/github.com/knative/client/test/e2e-common.sh --run-tests: signal: segmentation fault (core dumped)]
I0617 01:33:17.714] Test subprocess exited with code 1
```

(not sure where this comes from) and the client test fails because the deps have not been updated.

@houshengbo could you please run a `hack/build.sh -u` and commit the updated modules.txt to fix the client test ?

Let's hope that the integration test failure is just a flake.

houshengbo:
@rhuss Thanks for your comments. I am thinking of another way to fix this issue.
For the repo of knative client, I would like to add the support of a param, so that e2e-tests.sh can either run against the latest build or a previous release.
Then, we need to make the integration job of pull-knative-client-integration-tests to test both the latest and previous release. Instead of running test/e2e-tests.sh once, we can two builds, like 
./test/e2e-tests.sh latest and ./test/e2e-tests.sh 0.6.0.
It is not quite friendly to run tests against two knative installations on local machines. Folks are a sort of forced to run twice of the setup and integration tests.
What do you think?

rhuss:
> What do you think?

yes, sounds good to me. Also if we could configure that as two different GitHub checks then it would be evident which one is failing. Also it helps in parallelizing the tests which should it make faster.

rhuss:
@houshengbo There are some minor merge conflicts. Could you please fix them? I will review the PR in the meantime. Thanks !

test/e2e-tests.sh
```diff
@@ -35,17 +39,28 @@ function cluster_setup() {
 }
 
 function knative_setup() {
-  start_latest_knative_serving
+  header "Installing Knative serving version of $KNATIVE_VERSION"
```
rhuss: "Installing Knative serving $KNATIVE_VERSION"

test/presubmit-integration-tests-previous-release.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
rhuss: I don't really get this file (but could be also my ignorance ;)

* Shouldn't it be called '...latest-release.sh' as it refers to the latest release whereas "latest" refers to HEAD of master (unreleased code). A bit confusing. Or maybe call it really with the version number ('presubmit-integration-tests-0.6.0.sh') and update the configuration that calls this file
* Who is calling this file ?
* Why it is calling presubmit checks and not the e2e tests like in e2e-tests.sh  but with a environment variable set ?

test/e2e-tests.sh
```diff
@@ -35,17 +39,28 @@ function cluster_setup() {
...
+  start_knative_serving_env "$KNATIVE_VERSION"
+}
+
+function start_knative_serving_env() {
+  local VERSION=${1:-latest}
```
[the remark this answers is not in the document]
rhuss: (local variables are supposed to be lowercase, only exported variables should be UPPERCASE)

test/presubmit-integration-tests-previous-release.sh
rhuss: So I would have expected something like

```
export KNATIVE_VERSION="0.6.0"
$(dirname $0)/e2e-tests.sh
```

rhuss: Talked offline with @houshengbo . It's complicated (as the second job needs to be a custom job).

test/e2e-tests.sh
```diff
@@ -35,17 +39,28 @@ function cluster_setup() {
...
 # Add local dir to have access to built kn
 export PATH=$PATH:${REPO_ROOT_DIR}
+export KNATIVE_VERSION=${KNATIVE_VERSION:-latest}
 
 # Script entry point.
 
 initialize $@
 
-header "Running tests"
+header "Running tests for Knative serving version of $KNATIVE_VERSION"
```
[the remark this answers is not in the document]
houshengbo: Done.

test/e2e-tests.sh
houshengbo: Done.

test/e2e-tests.sh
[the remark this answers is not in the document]
houshengbo: Done.

test/presubmit-integration-tests-previous-release.sh
houshengbo: Thank @rhuss for his suggestion. We will use latest-release.
The repo test-infra defines four basic type of jobs, every PR can enable: integration-tests, build-tests, go-coverage, and unit-tests. We can only four regular jobs, with one for each of them respectively. It means even if we would like to have a new INTEGRATION TEST, we need to go through CUSTOM TEST JOB in test-infra. We cannot use the existing `presubmit-tests.sh --integration-tests` with more argument to specify the version or anything else. That was the reason we need to have a separate script like this one, with the correct env var setting for a specific version, to launch separately by the CUSTOM TEST JOB in test-infra.

rhuss:
Thanks !

PR #172, opened 2019-06-09
Set current namespace from kubeconfig by default

nak3:
`kn` command does not pick up namespace from kubeconfig but hardcorded
`default` namespace.

This patch fixes to get namespace from kubeconfig.

Fixes https://github.com/knative/client/issues/7

### DIFF
--- a/pkg/kn/commands/namespaced_test.go
+++ b/pkg/kn/commands/namespaced_test.go
@@ -36,7 +36,7 @@ func TestGetNamespaceSample(t *testing.T) {
 	expectedNamespace := "test1"
 	testCmd.SetArgs([]string{"--namespace", expectedNamespace})
 	testCmd.Execute()
-	kp := &KnParams{}
+	kp := &KnParams{NamespaceFactory: func() (string, error) { return FakeNamespace, nil }}
 	actualNamespace, err := kp.GetNamespace(testCmd)
 	if err != nil {
 		t.Fatal(err)
@@ -69,7 +69,7 @@ func TestGetNamespaceAllNamespacesSet(t *testing.T) {
 	sampleNamespace := "test1"
 	testCmd.SetArgs([]string{"--namespace", sampleNamespace, "--all-namespaces"})
 	testCmd.Execute()
-	kp := &KnParams{}
+	kp := &KnParams{NamespaceFactory: func() (string, error) { return FakeNamespace, nil }}
 	actualNamespace, err := kp.GetNamespace(testCmd)
 	if err != nil {
 		t.Fatal(err)
@@ -86,7 +86,7 @@ func TestGetNamespaceDefaultAllNamespacesUnset(t *testing.T) {
 	expectedNamespace := ""
 	testCmd.SetArgs([]string{"--all-namespaces"})
 	testCmd.Execute()
-	kp := &KnParams{}
+	kp := &KnParams{NamespaceFactory: func() (string, error) { return FakeNamespace, nil }}
 	actualNamespace, err := kp.GetNamespace(testCmd)
 	if err != nil {
 		t.Fatal(err)
@@ -102,7 +102,7 @@ func TestGetNamespaceAllNamespacesNotDefined(t *testing.T) {
 	expectedNamespace := "test1"
 	testCmd.SetArgs([]string{"--namespace", expectedNamespace})
 	testCmd.Execute()
-	kp := &KnParams{}
+	kp := &KnParams{NamespaceFactory: func() (string, error) { return FakeNamespace, nil }}
 	actualNamespace, err := kp.GetNamespace(testCmd)
 	if err != nil {
 		t.Fatal(err)
--- a/pkg/kn/commands/namespaced_test.go
+++ b/pkg/kn/commands/namespaced_test.go
@@ -36,7 +36,8 @@ func TestGetNamespaceSample(t *testing.T) {
 	expectedNamespace := "test1"
 	testCmd.SetArgs([]string{"--namespace", expectedNamespace})
 	testCmd.Execute()
-	actualNamespace, err := GetNamespace(testCmd)
+	kp := &KnParams{}
+	actualNamespace, err := kp.GetNamespace(testCmd)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -45,12 +46,13 @@ func TestGetNamespaceSample(t *testing.T) {
 	}
 }
 
-// test without setting any namespace
+// test current namespace without setting any namespace
 func TestGetNamespaceDefault(t *testing.T) {
 	testCmd := testCommandGenerator(true)
-	expectedNamespace := "default"
+	expectedNamespace := "current"
 	testCmd.Execute()
-	actualNamespace, err := GetNamespace(testCmd)
+	kp := &KnParams{CurrentNamespace: "current"}
+	actualNamespace, err := kp.GetNamespace(testCmd)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -67,7 +69,8 @@ func TestGetNamespaceAllNamespacesSet(t *testing.T) {
 	sampleNamespace := "test1"
 	testCmd.SetArgs([]string{"--namespace", sampleNamespace, "--all-namespaces"})
 	testCmd.Execute()
-	actualNamespace, err := GetNamespace(testCmd)
+	kp := &KnParams{}
+	actualNamespace, err := kp.GetNamespace(testCmd)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -83,7 +86,8 @@ func TestGetNamespaceDefaultAllNamespacesUnset(t *testing.T) {
 	expectedNamespace := ""
 	testCmd.SetArgs([]string{"--all-namespaces"})
 	testCmd.Execute()
-	actualNamespace, err := GetNamespace(testCmd)
+	kp := &KnParams{}
+	actualNamespace, err := kp.GetNamespace(testCmd)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -98,21 +102,8 @@ func TestGetNamespaceAllNamespacesNotDefined(t *testing.T) {
 	expectedNamespace := "test1"
 	testCmd.SetArgs([]string{"--namespace", expectedNamespace})
 	testCmd.Execute()
-	actualNamespace, err := GetNamespace(testCmd)
-	if err != nil {
-		t.Fatal(err)
-	}
-	if actualNamespace != expectedNamespace {
-		t.Fatalf("Incorrect namespace retrieved: %v, expected: %v", actualNamespace, expectedNamespace)
-	}
-}
-
-// test with all-namespace flag not defined and no namespace given
-func TestGetNamespaceDefaultAllNamespacesNotDefined(t *testing.T) {
-	testCmd := testCommandGenerator(false)
-	expectedNamespace := "default"
-	testCmd.Execute()
-	actualNamespace, err := GetNamespace(testCmd)
+	kp := &KnParams{}
+	actualNamespace, err := kp.GetNamespace(testCmd)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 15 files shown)


pkg/kn/commands/namespaced_test.go
```diff
@@ -98,21 +103,8 @@ func TestGetNamespaceAllNamespacesNotDefined(t *testing.T) {
...
-	if err != nil {
-		t.Fatal(err)
-	}
-	if actualNamespace != expectedNamespace {
-		t.Fatalf("Incorrect namespace retrieved: %v, expected: %v", actualNamespace, expectedNamespace)
-	}
-}
-
-// test with all-namespace flag not defined and no namespace given
-func TestGetNamespaceDefaultAllNamespacesNotDefined(t *testing.T) {
```
nak3: I have removed this test case because of ambiguous purpose.

- `TestGetNamespaceDefaultAllNamespacesNotDefined` is very similar to `TestGetNamespaceDefault`.
- The only difference is whether `allNamespaceFlag` is enabled or not via `testCommandGenerator`.
- The removed `TestGetNamespaceDefaultAllNamespacesNotDefined` tests default namespace when `allNamespaceFlag` is enabled. (I think that this scenario is not happen and not necessary for test.)

rhuss:
Coolio, thanks a lot ! 'was just about to fix that on my own, but even better so :-)

I will have a look later today ...

pkg/kn/commands/types.go
```diff
@@ -31,12 +31,17 @@ var KubeCfgFile string
 type KnParams struct {
 	Output         io.Writer
 	ServingFactory func() (serving.ServingV1alpha1Interface, error)
+	// CurrentNamespace is the namespace set in the current context.
```
rhuss: I think it's clear what CurrentNamespace is. So please either remove the comment (recommended) or comment on the other parameters, too.

pkg/kn/commands/types.go
```diff
@@ -31,12 +31,17 @@ var KubeCfgFile string
...
+	CurrentNamespace string
 }
 
-func (c *KnParams) Initialize() {
+func (c *KnParams) Initialize() error {
 	if c.ServingFactory == nil {
 		c.ServingFactory = GetConfig
 	}
+	var err error
+	c.CurrentNamespace, err = GetCurrentNamespace(KubeCfgFile)
```
rhuss: I wonder whether we could get rid of sharing that global variable and use a function instead which could cache a lookup. This would free us for knowing the order when this variable was initialized (which I just checked, looks good, but you don't know what happens when people refactor code).

I.e. introducing a `GetKubeCfgFile()` function which checks a private variable kubeCfgFile and, if empty, does the actual lookup as it's done in `InitializeConfig`:


```go
func GetKubeCfgFile() string {
   if kubeCfgFile == "" {
       kubeCfgFile = initKubeConfig()
   }
   return kubeCfgFile
}
```

(and maybe improve the error handling over there, as initKubeConfig() must not call `os.Exit()` but should return an error which should be propagated here, too.

pkg/kn/commands/types.go
```diff
@@ -50,3 +55,11 @@ func GetConfig() (serving.ServingV1alpha1Interface, error) {
 	}
 	return client, nil
 }
+
+// GetCurrentNamespace returns current namespace from kubeconfig.
+func GetCurrentNamespace(kubeconfig string) (string, error) {
```
rhuss: can't really comment on the logic here as I don't know anything about how to extract a current context from the configuration. Looks complicated though ;-)

nak3:
Thank you @rhuss Updated (and moved `KubeCfgFile` into `KnParams` struct.)

pkg/kn/core/root.go
```diff
@@ -39,7 +36,10 @@ func NewKnCommand(params ...commands.KnParams) *cobra.Command {
 	} else {
 		panic("Too many params objects to NewKnCommand")
 	}
-	p.Initialize()
+	err := p.Initialize()
+	if err != nil {
+		panic(fmt.Sprintf("Failed to initialize command %v", err))
```
maximilien: Wondering if `panic` is the right thing to do here? I can understand the rational but the error to user will be kind of ugly. Returning error with message will be handled like all errors in `main.go`... 

Though I can also see the argument that this is not a common error and panicking is warranted. Conflicted a bit.

pkg/kn/commands/types.go
maximilien: Agree, I made it public and hated that. Better to go back to old Java-style getters for this one and hide the var. Thanks for suggesting this.

pkg/kn/commands/types.go
maximilien: Isn’t it one line in the `KUBECONFIG` file? Maybe my clusters are super simple...

pkg/kn/core/root.go
rhuss: +1 for not panicing but exiting with a "normal" error. I think it's not impossible that initialization fails, e.g. when some .kube config file is not readable. We always strive for nice error messages.

nak3:
Thank you @maximilien and @rhuss. Updated.

I realized that `Initalize()` should __not__ cause panic nor catch error as current code does. If we handle error, users must set `kubeconfig` even when we just want to check some help usage like `kn -h`.

pkg/kn/commands/types.go
```diff
@@ -47,3 +57,33 @@ func GetConfig() (serving.ServingV1alpha1Interface, error) {
...
+// GetCurrentNamespace returns current namespace from kubeconfig.
+func (c *KnParams) GetCurrentNamespace() (string, error) {
+	// Make sure that KubeCfgFile is initialized.
+	if c.KubeCfgFile == "" {
+		if err := c.initKubeConfig(); err != nil {
+			return "", err
+		}
+	}
+	rules := clientcmd.NewDefaultClientConfigLoadingRules()
+	rules.ExplicitPath = c.KubeCfgFile
```
cppforlife: see https://github.com/k14s/kapp/blob/0c91e8a2208ff9a8f9d92f8c01d20dc9dd21256b/pkg/kapp/cmd/core/config_factory.go#L60 for example. we dont need to specify ExplicitPath unless it's given as an explicit option. also KUBECONFIG env var will be picked up by default.

nak3:
Thank you @cppforlife Updated.

pkg/kn/commands/types.go
```diff
@@ -16,28 +16,38 @@ package commands
...
 
-func GetConfig() (serving.ServingV1alpha1Interface, error) {
-	config, err := clientcmd.BuildConfigFromFlags("", KubeCfgFile)
+func (c *KnParams) GetConfig() (serving.ServingV1alpha1Interface, error) {
+	if c.KubeCfgFile == "" {
+		if err := c.initKubeConfig(); err != nil {
+			return nil, err
+		}
+	}
+	config, err := clientcmd.BuildConfigFromFlags("", c.KubeCfgFile)
```
cppforlife: we should also get `ClientConfig()` from `clientcmd.NewNonInteractiveDeferredLoadingClientConfig()` instead of doing `initKubeConfig()`. i dont think we actually need initKubeConfig at all.

nak3:
@cppforlife Thank you! updated.

pkg/kn/commands/types.go
```diff
@@ -21,23 +21,34 @@ import (
...
-func GetConfig() (serving.ServingV1alpha1Interface, error) {
-	config, err := clientcmd.BuildConfigFromFlags("", KubeCfgFile)
+func (c *KnParams) GetConfig() (serving.ServingV1alpha1Interface, error) {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	if len(c.KubeCfgPath) > 0 {
+		loadingRules.ExplicitPath = c.KubeCfgPath
+	}
+	c.ClientConfig = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{})
+	var err error
+	c.CurrentNamespace, _, err = c.ClientConfig.Namespace()
```
cppforlife: wouldnt this clobber CurrentNamespace set from the flag?

cppforlife: oh nm read wrong code

pkg/kn/commands/namespaced_test.go
```diff
@@ -45,12 +46,13 @@ func TestGetNamespaceSample(t *testing.T) {
...
 
-// test without setting any namespace
+// test current namespace without setting any namespace
 func TestGetNamespaceDefault(t *testing.T) {
 	testCmd := testCommandGenerator(true)
-	expectedNamespace := "default"
+	expectedNamespace := "current"
 	testCmd.Execute()
-	actualNamespace, err := GetNamespace(testCmd)
+	kp := &KnParams{CurrentNamespace: "current"}
```
cppforlife: i think you should add this line to every test case since it's implied that CurrentNamespace() from clientcmd is set. additionally is there an easy way to test GetConfig(). if not in unit test land, can it be done in e2e?

nak3:
Thank you @cppforlife ! Updated.

navidshaikh:
@nak3 : oops
```
➜  client git:(current-ns) ./kn version
Version:      v20190620-local-0ce1486
Build Date:   2019-06-20 12:10:25
Git Revision: 0ce1486
Dependencies:
- serving:    v0.6.0

➜  client git:(current-ns) ./kn service create current-ns --image gcr.io/knative-samples/helloworld-go                                                                               
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0xf3124a]

goroutine 1 [running]:
github.com/knative/client/pkg/kn/commands.(*KnParams).CurrentNamespace(0xc0002c1740, 0xc0002d0d50, 0x120eadc, 0xe, 0x1d00b60)
	/home/nshaikh/go/src/github.com/knative/client/pkg/kn/commands/types.go:47 +0x2a
github.com/knative/client/pkg/kn/commands.(*KnParams).CurrentNamespace-fm(0xc0000f6f00, 0x120eadc, 0xe, 0x0)
	/home/nshaikh/go/src/github.com/knative/client/pkg/kn/commands/types.go:42 +0x2a
github.com/knative/client/pkg/kn/commands.(*KnParams).GetNamespace(0xc0002c1740, 0xc0002e9180, 0x32, 0x1d00b60, 0x0, 0x0)
	/home/nshaikh/go/src/github.com/knative/client/pkg/kn/commands/namespaced.go:58 +0xe1
github.com/knative/client/pkg/kn/commands/service.NewServiceCreateCommand.func1(0xc0002e9180, 0xc0002d1140, 0x1, 0x3, 0x0, 0x0)
	/home/nshaikh/go/src/github.com/knative/client/pkg/kn/commands/service/service_create.go:62 +0xde
github.com/spf13/cobra.(*Command).execute(0xc0002e9180, 0xc0002d10e0, 0x3, 0x3, 0xc0002e9180, 0xc0002d10e0)
	/home/nshaikh/go/src/github.com/knative/client/vendor/github.com/spf13/cobra/command.go:762 +0x473
github.com/spf13/cobra.(*Command).ExecuteC(0xc0002e8780, 0xc0002f6280, 0xc0002e9900, 0xc0002e8a00)
	/home/nshaikh/go/src/github.com/knative/client/vendor/github.com/spf13/cobra/command.go:852 +0x2fd
github.com/spf13/cobra.(*Command).Execute(0xc0002e8780, 0x0, 0x0)
	/home/nshaikh/go/src/github.com/knative/client/vendor/github.com/spf13/cobra/command.go:800 +0x2b
main.main()
	/home/nshaikh/go/src/github.com/knative/client/cmd/kn/main.go:25 +0x44

➜  client git:(current-ns) git checkout master && hack/build.sh -f && ./kn service create current-ns --image gcr.io/knative-samples/helloworld-go
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
🚧 Compile
────────────────────────────────────────────
Version:      v20190620-local-32ccc33
Build Date:   2019-06-20 12:14:09
Git Revision: 32ccc33
Dependencies:
- serving:    v0.6.0
Service 'current-ns' successfully created in namespace 'default'.
```

navidshaikh:
Well! `pull-knative-client-integration-tests` [dashboard](https://gubernator.knative.dev/build/knative-prow/pr-logs/pull/knative_client/172/pull-knative-client-integration-tests/1141683635548590080/) shows SUCCESS but here its Job failed!

Issue #161, opened 2019-06-03
Improve unit tests


maximilien:
Based on various discussions on different PRs (pasted below). But the gist is to improve unit tests for commands.

From @rhuss comment on PR #66 

> * Refactor tests to become more unit tests. E.g. commands should be tested only that they propagate the proper info to the library calls (e.g. in serving), the library calls should then use the serving mock library. Only a suggestion, but tests currently are not easy to maintain.

100% agree. I'd use [Ginkgo](https://onsi.github.io/ginkgo/)/[Gomega](https://onsi.github.io/gomega) and [Counterfeiter](https://github.com/maxbrunsfeld/counterfeiter).... Or similar tooling

maximilien:
Any opposition to the the tools I mentioned above?

rhuss:
> Any opposition to the the tools I mentioned above?

Nope as I don't know any of them yet so don't have any preferences ;-) Happy to learn though ...

maximilien:
Sweet. I'll cook up something on a simple command so we can all see the resulting shape and impact on imports etc.

maximilien:
OK so I submitted PR #185 with an example UT for the `kn version` command. I know it's a simple one, but I think it illustrates the power (I think) of BDD-style UTs, and of the Ginkgo/Gomega framework. Let me re-iterate a few points and give two examples of what we gain if we start doing UTs this way. There are more but perhaps best to highlight some of the key ones first.

First things first:

1. Ginkgo/Gomega is a very mature BDD-style test framework. You can write any tests with it. Unit, integration, etc.
2. It's been used for over 5 years in Cloud Foundry (CF) and other places. And CF is highly tests-driven community. So tons of tests and optimization already done with this framework. It's solid
3. It's compatible with `go test` so this means that we can mix and match `go test` tests with Ginkgo/Gomega. So with this new test for version, the other tests still run.
4. I did no change to current tests scripts, e.g., `./hack/build.sh` and `./test/presubmit-tests.sh` or e2e tests. All run fine.
5. We can gradually transition as needed.

Second, what are three quick main advantages of doing UTs with Ginkgo/Gomega:

1. Better test structure. The version command test is super clear, once you understand basic structure and focus on the tests. See PR and this simplified snippet to see what I mean:

```go
package commands_test

import (
   ...
)

var ...
var _ = Describe("Version", func() {
	const (
   ...
	)

	var (
		versionCmd *cobra.Command
		knParams   *KnParams
	)

	BeforeEach(func() {
[...]

PR #177, opened 2019-06-11
[WIP] Kn plugins implementation

maximilien:
Implements Kn plugins re-using some code from kubectl plugins.
This is WIP to allow concrete discussions and testing and
hopefully also allow sub tasks to be created and worked on
in parallel.

This version contains the following:

1. wraps the main root Kn command to support plugin
2. plugins are any executable in PATH with name kn-*
3. 'kn plugins list' to list found kn plugins
4. skips any kn plugins found with name that match core
   commands, e.g., kn-service would be ignored
5. execute any valid kn plugins found

And is missing:

1. unit and integration tests
2. plugin install command
3. plugin repository command
4. plugin / Knative server version negotiation
5. anything else we agree on in plugin req doc


maximilien:
@sixolet and @cppforlife this is WIP so as to avoid one HUGE PR once we have complete agreement on the plugins docs. However, for those who want to try, this can be used and is rebased on current master:HEAD.

To test this PR. After building `kn` using `./hack/build.sh`, do the following:

```bash
echo '#!/bin/bash

echo "Hello Knative, I am a Kn plugin"' > kn-hello
"
chmod +x kn-hello
mv kn-hello /usr/local/bin

# you can now list plugins
./kn plugin list
The following compatible plugins are available:

/usr/local/bin/kn-hello

# you can execute the 'kn-hello' plugin with
./kn hello
Hello Knative, I'm a Kn plugin
```

I will be adding unit tests and integration tests next.

docs/cmd/kn_plugin_list.md
```diff
@@ -0,0 +1,34 @@
...
+
+list all visible plugin executables on a user's PATH
+
+### Synopsis
+
+List all available plugin files on a user's PATH.
+
+		Available plugin files are those that are:
+		- executable
+		- anywhere on the user's PATH
```
duglin: Should mention what we do if we find duplicates - I'm assuming first one wins.

docs/cmd/kn_plugin_list.md
```diff
@@ -0,0 +1,34 @@
...
+
+```
+kn plugin list [flags]
+```
+
+### Options
+
+```
+  -h, --help        help for list
+      --name-only   If true, display only the binary name of each plugin, rather than its full path
```
duglin: I think this should be reversed. In most cases I don't think people will care where the exe lives and when they do care it's probably because they're debugging to see why things are not acting correctly.

pkg/kn/commands/plugin/plugin_handler.go
```diff
@@ -0,0 +1,109 @@
...
+	"strings"
+	"syscall"
+)
+
+// PluginHandler is capable of parsing command line arguments
+// and performing executable filename lookups to search
+// for valid plugin files, and execute found plugins.
+type PluginHandler interface {
+	// exists at the given filename, or a boolean false.
+	// Lookup will iterate over a list of given prefixes
```
duglin: "given prefixes"? Where is this list? none of the args to Lookup are slices

pkg/kn/commands/plugin/plugin_list.go
```diff
@@ -0,0 +1,268 @@
...
+		for _, f := range files {
+			if f.IsDir() {
+				continue
+			}
+			if !hasValidPrefix(f.Name(), ValidPluginFilenamePrefixes) {
+				continue
+			}
+
+			if isFirstFile {
+				fmt.Fprintf(o.ErrOut, "The following compatible plugins are available:\n\n")
```
duglin: this should go to stdout, not stderr

pkg/kn/commands/plugin/plugin_list.go
```diff
@@ -0,0 +1,268 @@
...
+				for _, err := range errs {
+					fmt.Fprintf(o.ErrOut, "  - %s\n", err)
+					pluginWarnings++
+				}
+			}
+		}
+	}
+
+	if !pluginsFound {
+		pluginErrors = append(pluginErrors, fmt.Errorf("error: unable to find any kubectl plugins in your PATH"))
```
duglin: No plugins shouldn't be considered an error - it's a valid state.

pkg/kn/commands/plugin/plugin_list.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	if pluginWarnings > 0 {
```
duglin: Why go thru all of this complexity just to print errors? Why not just print them as you encounter them? What's the point of saving them and printing them later?

pkg/kn/commands/plugin/plugin_list.go
```diff
@@ -0,0 +1,268 @@
...
+		return []error{fmt.Errorf("unable to verify path with nil root")}
+	}
+
+	// extract the plugin binary name
+	segs := strings.Split(path, "/")
+	binName := segs[len(segs)-1]
+
+	cmdPath := strings.Split(binName, "-")
+	if len(cmdPath) > 1 {
+		// the first argument is always "kubectl" for a plugin binary
```
duglin: kubectl?

pkg/kn/commands/plugin/plugin_list.go
```diff
@@ -0,0 +1,268 @@
...
+		cmdPath = cmdPath[1:]
+	}
+
+	errors := []error{}
+
+	if isExec, err := isExecutable(path); err == nil && !isExec {
+		errors = append(errors, fmt.Errorf("warning: %s identified as a kubectl plugin, but it is not executable", path))
```
duglin: kubectl?
By default I'm not sure there's value in telling people about this condition. What if they really do need to have some random text file on disk that starts with `kn-` that just happens to be in their PATH... they're going to continually see this error message when it's not an error condition.   We shouldn't force them to rename it.

pkg/kn/commands/plugin/plugin_list.go
```diff
@@ -0,0 +1,268 @@
...
+	} else if err != nil {
+		errors = append(errors, fmt.Errorf("error: unable to identify %s as an executable file: %v", path, err))
+	}
+
+	if existingPath, ok := v.seenPlugins[binName]; ok {
+		errors = append(errors, fmt.Errorf("warning: %s is overshadowed by a similarly named plugin: %s", path, existingPath))
```
duglin: I'd only show this when some special flag. Thinking about how normal Linux stuff works, I've never seen Linux give a warning like this so I'm not sure we should either.

sixolet:
To prepare for the world where we consider helping the user install plugins suggested by their server, I think I'm a lot more interested in there being a configurable plugin dir (remember when we removed `--config`? we might need it back.), and every executable in the plugin dir can be a plugin.

If we go with "anywhere on $PATH", then we have no place to install to, and also the user doesn't have independent control of what's on their PATH and what is a plugin, which bothers me somewhere in my aesthetics brain.

maximilien:
> To prepare for the world where we consider helping the user install plugins suggested by their server, I think I'm a lot more interested in there being a configurable plugin dir (remember when we removed `--config`? we might need it back.), and every executable in the plugin dir can be a plugin.

Yup. Thought of that too.
 
> If we go with "anywhere on $PATH", then we have no place to install to, and also the user doesn't have independent control of what's on their PATH and what is a plugin, which bothers me somewhere in my aesthetics brain.

I personally don’t care. Product folks might have strong opinions. `$PATH` has two benefits: 1) compatible with `kubectl` and 2) many options and easy to determine any way. However, I can also see people wanting to have tight control over where plugins go. So I could see having both.

Anyhow, let’s decide next call and move forward. Best.

maximilien:
@duglin, first off, thanks for feedback. However, I think it’s worth re-iterating two points globally before I address each of your comments:

1. This is [WIP] == [Work In Progress]. Says so in title.
2. First line of description: “Implements Kn plugins re-using some code from kubectl plugins.”

So you should expect this is not end game and that it looks and feels like `kubectl` plugins by design.

I wanted this to be a baseline so when we decide to deviate from `kubectl` plugins it’s as clear as possible.

pkg/kn/commands/plugin/plugin_list.go
maximilien: See my general comment.

pkg/kn/commands/plugin/plugin_list.go
maximilien: Again see my global comment. By design.

pkg/kn/commands/plugin/plugin_list.go
maximilien: See global comment. By design.

pkg/kn/commands/plugin/plugin_list.go
maximilien: Could be interesting. However, once you print an error you’d stop? I think collecting and summarizing is just as good an approach. Remember this command’s output is listing the plugins found.

However, I may revisit this. Especially as we decide on MVP features next week.

pkg/kn/commands/plugin/plugin_list.go
maximilien: Fair.

pkg/kn/commands/plugin/plugin_list.go
maximilien: Fair.

pkg/kn/commands/plugin/plugin_handler.go
maximilien: Hardcoded. There is a var for it

docs/cmd/kn_plugin_list.md
maximilien: OK, up for debate. Especially if we decide to have a fix location.

docs/cmd/kn_plugin_list.md
maximilien: Yup, but first one in `PATH` list to be more specific.

duglin:
yeah, but WIP or not, I think the suggested edits are valid

pkg/kn/commands/plugin/plugin_list.go
duglin: stopping on error vs waiting until the end to stop is independent to me from buffering of messages. You can just set a bool and then exit if true at the end. But yes, in general I do think stopping on first error is fine since this isn't a compiler that is expected to have lots of error messages for people to process.

pkg/kn/commands/plugin/plugin_handler.go
duglin: I should have been more clear. I think the comment needs to change because the phrase "given prefixes" (to me) implies "passed in prefixes" , meaning as an arg.  if it said "..iterate of the list in ValidPrefixes" then it would have been more clear.

pkg/kn/commands/plugin/plugin_list.go
duglin: I think the design is wrong :-) 
Put it this way, I think a file in the PATH should only be considered a Kn plugin if:
- it's name starts with `kn-`
- it's executable bit is set

if a file in the PATH doesn't match this then it's not an error condition, it's the user's way of saying "skip it".

pkg/kn/commands/plugin/plugin_list.go
duglin: Aligning with `kubectl` is an interesting goal but I think aligning with Linux is more important

rhuss: side-note: 'guess you mean Unix CLI conventions, not only Linux. Many of us are on macOs or Windows out there, so imo we should strive for a good CLI experience in general, with Unix conventions as a guideline. Not sure whether we should write this down somewhere in a kind of spec/vision document ?

duglin: I think MacOS and Windows work the same way as well.

pkg/kn/core/root.go
```diff
@@ -62,10 +95,17 @@ Eventing: Manage event subscriptions and channels. Connect up event sources.`,
 	if p.Output != nil {
 		rootCmd.SetOutput(p.Output)
 	}
+	rootCmd.PersistentFlags().StringVar(&commands.CfgFile, "config", "", "config file (default is $HOME/.kn/config.yaml)")
+	rootCmd.PersistentFlags().StringVar(&commands.PluginDir, "plugin-dir", "", "kn plugin directory (default is value in kn config or $PATH)")
 	rootCmd.PersistentFlags().StringVar(&commands.KubeCfgFile, "kubeconfig", "", "kubectl config file (default is $HOME/.kube/config)")
```
maximilien: Hi, @sixolet, see this version of `root.go` 

Still need to rebase and clean up, so still WIP. However, this is where I am adding the `plugin-dir` persistent flag and reading from kn's config (if exist) and defaulting to `$PATH`.

I am in middle of now using this in the plugin commands. Will finish this tomorrow. But at least let's make sure we are in sync. Let me know your thoughts if you get to read this otherwise, look for next update.

Cheers 🍻

maximilien:
OK I need to rebase. Will try to find time to do so today.

maximilien:
OK finally getting a chance to go back to this as I am wrapping up my trip in Beijing 🇨🇳 

Will work on rebasing with latest updates and also move all tests to `gotest.tools` as per agreement on issue #161 on yesterday's WG.

maximilien:
Closing this one since #249 is the first non-WIP version. Please review that one.

Issue #128, opened 2019-05-19
Change "get" to "list"


duglin:
Proposal: change `get` to `list`

When people want to list the set of services running I think `kn service list` is more natural for people. I know this is more of a personal preference kind of thing, especially when you consider that `kubectl` actually uses `get` for both (lists and singular), but if you think about how all of the `kn service` verbs act on a single instance of a service, `get` would therefore imply "get me info about A service".  While, the word `list` more naturally implies you want a "list" of something back.... in this case services.

Net: change it to: `kn service list` to see all services.

Same should apply to the other nouns as well.

rhuss:
FYI, we already had the reverse discussion starting at https://github.com/knative/client/issues/48#issuecomment-483597428 and decided in the WG meeting on [2019-04-23](https://docs.google.com/document/d/1Uh7jTWQruBBmic-WmTvtc9cMF95kQrKb5lsqWhNuikM/edit) to go with `get` over `list` (to be in alignment to kubernetes and following the principle of least surprise even when we do not 100% like the way kubectl does it).

We can revisit this decision, but at the moment `get` is mostly for an overview and machine-readable output (like with `-o yaml`) and `describe` for human-readable detail output of a single service (like in kubectl).

duglin:
I wouldn't compare things to kubectl since kubectl is kind of odd in this respect. Sometimes it GETs a single thing, sometimes it GETs multiple - then add in the optional "s" at the end of the resource type - and I wouldn't call it the most UX-friendly.  I view `kn` (like knative) as a better UX thank vanilla Kube. So I was hoping we could do better than the CRUD-like interface of kubectl.

rhuss:
I went now back and forth with this UX issue, and I tend to agree with @duglin Although I still think that the 'principle of least suprise' is still a good thing, I'm by no means sure anymore whether it applies to kubectl vs. kn:

* kn and kubectl are already called differently: <verb> <noun> (kubectl) vs <noun> <verb> (kn). Adding a similar behaviour afterwards for "get" is probably much more confusing than to make a clear cut.
* Also some code is shared (i.e. the one for machine readable output), the target audience is different enough to justify a different UX style, too.

As I had now the chance to play around quite a bit with kn, I'd like to make the following suggestion for a kind of compromise, but still with a clear theme:

[...]

PR #179, opened 2019-06-12
Renames service 'get' to 'list'

navidshaikh:
As per title.


duglin:
thanks. Can you add  `Closes #128` in the initial comment so it'll automagically close the issue?

navidshaikh:
> thanks. Can you add Closes #128 in the initial comment so it'll automagically close the issue?

I tend to do that, but we have another PR #180 as well solving same, I will update the later on.

PR #178, opened 2019-06-12
Chore: Fix typos in docs

rohanKanojia:
I was reading your docs, I found some mistakes. Hope you find them useful!

PR #181, opened 2019-06-12
Fix typo in README.md

aliok:


duglin:
while LGTM, dup of #178 which should probably go in first

PR #186, opened 2019-06-14
feat(build.sh): Enable goimports + cleanup imports in source files

rhuss:
Much like 'go fmt' for cleaning up go source, 'goimports' can help
in properly formatting and grouping imports. See https://goinbigdata.com/goimports-vs-gofmt/
for a quick overview.

This commit also cleansup imports on existing files.

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -152,7 +152,7 @@ check_license() {
 
 
 update_deps() {
-  echo "🕸️ Update"
+  echo "🕸️ ${S}Update"
   go mod vendor
 }
 
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -54,20 +54,10 @@ run() {
     update_deps
   fi
 
-  # Run build
-  go_build
-
-  # Run tests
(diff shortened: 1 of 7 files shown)


hack/build.sh
```diff
@@ -61,6 +61,9 @@ run() {
     # Format source code
     go_fmt
 
+    # Cleanup imports
+    go_imports
```
maximilien: Minor... same as comment to `go_fmt`. I'd move anything that changes the code before compile and test.

rhuss: I will change the order in another PR, just let those get merged first.

hack/build.sh
```diff
@@ -61,6 +61,10 @@ run() {
     # Format source code
     go_fmt
```
navidshaikh: should we remove the `go_fmt` now? since, in addition to fixing imports, goimports also formats code in the same style as gofmt.

rhuss: ah, ok. I thought they do different things. Let me check.

PR #187, opened 2019-06-14
feat(build.sh) - Add mandatory license check

rhuss:
We could decide to make that optional, too, if this is coming into the way.

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -61,7 +61,10 @@ run() {
     # Format source code
     go_fmt
 
-    # Generate docs
+    # Check for license headers
+    check_license
+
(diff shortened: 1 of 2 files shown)


hack/build.sh
```diff
@@ -61,7 +61,10 @@ run() {
     # Format source code
     go_fmt
```
maximilien: Minor point...

BTW, should we move `go_fmt` before compile and test? I hate the idea of the compiled and tested code being changed after.

rhuss: Good catch. We can do this, and as the --fast mode skips formatting anyway it doesn't matter whether it is skipped before or after compile ;-)

PR #190, opened 2019-06-17
WIP: Add the support to test against the previous release and the latest source

houshengbo:


rhuss:
@houshengbo what the difference of this PR to #170 ?

houshengbo:
@rhuss I am just trying another way to fix this issue with a param to see 0.6.0 can really work with the pr build of integration test. Ignore this one.

PR #191, opened 2019-06-18
Added changes to support container port information while creating/updating service

savitaashture:
Issue: https://github.com/knative/client/issues/142

Description:
When someone writes application which listens on port other than 8080(default port in knative-serving) then deploy using **cli(kn)** then there is no way to specify the port.

**Current:** 
Supported flags are
```
Flags:
      --concurrency-limit int    Hard Limit of concurrent requests to be processed by a single replica.
      --concurrency-target int   Recommendation for when to scale up based on the concurrent number of incoming request. Defaults to --concurrency-limit when given.
  -e, --env stringArray          Environment variable to set. NAME=value; you may provide this flag any number of times to set multiple environment variables.
      --force                    Create service forcefully, replaces existing service if any.
  -h, --help                     help for create
      --image string             Image to run.
      --limits-cpu string        The limits on the requested CPU (e.g., 1000m).
      --limits-memory string     The limits on the requested CPU (e.g., 1024Mi).
      --requests-memory string   The requested CPU (e.g., 64Mi).
```
**Expected:** 
Along with current flags it should support **port** flag so that user can specify port at which application is running.

ex: something like below
```
Flags:
   ...
  -p, --port int32               The port that the application should run. (default 8080)
   ...
```

**Usage:**
```
kn service create mysvc --port 80 --image dev.local/ns/image:latest
```
```
kn service update mysvc --port 80
```

### DIFF
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -34,6 +34,7 @@ type ConfigurationEditFlags struct {
 	MaxScale                   int
 	ConcurrencyTarget          int
 	ConcurrencyLimit           int
+	Port                       int32
 }
 
 type ResourceFlags struct {
@@ -54,6 +55,7 @@ func (p *ConfigurationEditFlags) AddUpdateFlags(command *cobra.Command) {
 	command.Flags().IntVar(&p.MaxScale, "max-scale", 0, "Maximal number of replicas.")
 	command.Flags().IntVar(&p.ConcurrencyTarget, "concurrency-target", 0, "Recommendation for when to scale up based on the concurrent number of incoming request. Defaults to --concurrency-limit when given.")
 	command.Flags().IntVar(&p.ConcurrencyLimit, "concurrency-limit", 0, "Hard Limit of concurrent requests to be processed by a single replica.")
+	command.Flags().Int32VarP(&p.Port, "port", "p", 0, "The port where application listens on.")
 }
 
 func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
@@ -100,6 +102,13 @@ func (p *ConfigurationEditFlags) Apply(service *servingv1alpha1.Service, cmd *co
 	err = servinglib.UpdateResources(template, requestsResources, limitsResources)
 	if err != nil {
 		return err
+	}
+
+	if cmd.Flags().Changed("port") {
+		err = servinglib.UpdateContainerPort(template, p.Port)
+		if err != nil {
+			return err
+		}
 	}
 
 	servinglib.UpdateConcurrencyConfiguration(template, p.MinScale, p.MaxScale, p.ConcurrencyTarget, p.ConcurrencyLimit)
--- a/pkg/kn/commands/service/service_create.go
+++ b/pkg/kn/commands/service/service_create.go
@@ -46,6 +46,9 @@ func NewServiceCreateCommand(p *commands.KnParams) *cobra.Command {
   # Create or replace environment variables of service 's1' using --force flag
   kn service create --force s1 --env KEY1=NEW_VALUE1 --env NEW_KEY2=NEW_VALUE2 --image dev.local/ns/image:v1
 
+  # Create service 'mysvc' with port 80
+  kn service create mysvc --port 80 --image dev.local/ns/image:latest
+
   # Create or replace default resources of a service 's1' using --force flag
   # (earlier configured resource requests and limits will be replaced with default)
   # (earlier configured environment variables will be cleared too if any)
--- a/pkg/kn/commands/service/service_update.go
+++ b/pkg/kn/commands/service/service_update.go
@@ -33,6 +33,9 @@ func NewServiceUpdateCommand(p *commands.KnParams) *cobra.Command {
   # Updates a service 'mysvc' with new environment variables
   kn service update mysvc --env KEY1=VALUE1 --env KEY2=VALUE2
 
+  # Update a service 'mysvc' with new port
+  kn service update mysvc --port 80
+
   # Updates a service 'mysvc' with new requests and limits parameters
   kn service update mysvc --requests-cpu 500m --limits-memory 1024Mi`,
 		RunE: func(cmd *cobra.Command, args []string) (err error) {
--- a/pkg/serving/config_changes.go
+++ b/pkg/serving/config_changes.go
@@ -86,6 +86,18 @@ func UpdateImage(template *servingv1alpha1.RevisionTemplateSpec, image string) e
 	return nil
 }
 
+// UpdateContainerPort updates container with a give port
+func UpdateContainerPort(template *servingv1alpha1.RevisionTemplateSpec, port int32) error {
+	container, err := extractContainer(template)
+	if err != nil {
+		return err
+	}
+	container.Ports = []corev1.ContainerPort{{
+		ContainerPort: port,
+	}}
+	return nil
+}
+
 func UpdateResources(template *servingv1alpha1.RevisionTemplateSpec, requestsResourceList corev1.ResourceList, limitsResourceList corev1.ResourceList) error {
 	container, err := extractContainer(template)
 	if err != nil {
--- a/pkg/serving/config_changes_test.go
+++ b/pkg/serving/config_changes_test.go
@@ -139,6 +139,30 @@ func testUpdateEnvVarsModify(t *testing.T, revision *servingv1alpha1.RevisionTem
 	}
 }
 
+func TestUpdateContainerPort(t *testing.T) {
+	template, _ := getV1alpha1Config()
+	err := UpdateContainerPort(template, 8888)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Verify update is successful or not
+	checkPortUpdate(t, template, 8888)
+	// update template with container port info
+	template.Spec.Containers[0].Ports[0].ContainerPort = 9090
+	err = UpdateContainerPort(template, 80)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Verify that given port overrides the existing container port
+	checkPortUpdate(t, template, 80)
+}
+
+func checkPortUpdate(t *testing.T, template *servingv1alpha1.RevisionTemplateSpec, port int32) {
+	if len(template.Spec.Containers) != 1 || template.Spec.Containers[0].Ports[0].ContainerPort != port {
+		t.Error("Failed to update the container port")
+	}
+}
+
 func TestUpdateEnvVarsBoth(t *testing.T) {
 	template, container := getV1alpha1RevisionTemplateWithOldFields()
 	testUpdateEnvVarsBoth(t, template, container)
--- a/docs/cmd/kn_service_create.md
+++ b/docs/cmd/kn_service_create.md
@@ -27,6 +27,9 @@ kn service create NAME --image IMAGE [flags]
   # Create or replace environment variables of service 's1' using --force flag
   kn service create --force s1 --env KEY1=NEW_VALUE1 --env NEW_KEY2=NEW_VALUE2 --image dev.local/ns/image:v1
 
+  # Create service 'mysvc' with port 80
+  kn service create mysvc --port 80 --image dev.local/ns/image:latest
+
   # Create or replace default resources of a service 's1' using --force flag
   # (earlier configured resource requests and limits will be replaced with default)
   # (earlier configured environment variables will be cleared too if any)
@@ -47,6 +50,7 @@ kn service create NAME --image IMAGE [flags]
       --max-scale int            Maximal number of replicas.
       --min-scale int            Minimal number of replicas.
   -n, --namespace string         List the requested object(s) in given namespace.
+  -p, --port int32               The port where application listens on.
       --requests-cpu string      The requested CPU (e.g., 250m).
       --requests-memory string   The requested CPU (e.g., 64Mi).
 ```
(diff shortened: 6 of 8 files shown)


pkg/kn/commands/service/service_update.go
```diff
@@ -33,6 +33,9 @@ func NewServiceUpdateCommand(p *commands.KnParams) *cobra.Command {
   # Updates a service 'mysvc' with new environment variables
   kn service update mysvc --env KEY1=VALUE1 --env KEY2=VALUE2
 
+  # Update a service with given port
```
rhuss:
```
# Update service "mysvc" to expose port 80
```

Be more specific, it's an example.

pkg/kn/commands/service/service_create.go
```diff
@@ -46,6 +46,10 @@ func NewServiceCreateCommand(p *commands.KnParams) *cobra.Command {
   # Create or replace environment variables of service 's1' using --force flag
   kn service create --force s1 --env KEY1=NEW_VALUE1 --env NEW_KEY2=NEW_VALUE2 --image dev.local/ns/image:v1
 
+  # Create a service with given port
+  # If port is not specified it will take 8080 as a default port
```
rhuss:
```
# Create service 'mysvc' with port 80
# If no --port is given the default port 8080 is used
```

I would be more specific and more concise in the documentation.

hack/generate-docs.go
```diff
@@ -29,9 +29,9 @@ func main() {
 	if len(os.Args) > 1 {
 		dir = os.Args[1]
 	}
-	err := doc.GenMarkdownTree(rootCmd, dir + "/docs/cmd/")
+	err := doc.GenMarkdownTree(rootCmd, dir+"/docs/cmd/")
```
rhuss: would revert that as it's not related to this PR.

pkg/serving/config_changes.go
```diff
@@ -86,6 +86,27 @@ func UpdateImage(template *servingv1alpha1.RevisionTemplateSpec, image string) e
...
+		return err
+	}
+
+	if len(container.Ports) == 0 {
+		container.Ports = []corev1.ContainerPort{{
+			ContainerPort: port,
+		}}
+
+	} else {
+		for i, _ := range container.Ports {
```
rhuss: There should be only one. Please either assert that there is only one port (that is what Knative serving allows) and return an error or truncate to take only the first.

Tbh, I would just overwrite unconditionally any existing port.

pkg/serving/config_changes_test.go
```diff
@@ -139,6 +139,43 @@ func testUpdateEnvVarsModify(t *testing.T, revision *servingv1alpha1.RevisionTem
...
 }
 
+func TestUpdateContainerPort(t *testing.T) {
+	template, _ := getV1alpha1Config()
+	err := UpdateContainerPort(template, 8080)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Verify update is successful or not
+	for p, _ := range template.Spec.Containers {
```
rhuss: assert that there are only 1 container and 1 port. Then just assert on `template.Spec.Containers[0].Ports[0]`

pkg/serving/config_changes_test.go
```diff
@@ -139,6 +139,43 @@ func testUpdateEnvVarsModify(t *testing.T, revision *servingv1alpha1.RevisionTem
 	}
 }
 
+func TestUpdateContainerPort(t *testing.T) {
+	template, _ := getV1alpha1Config()
+	err := UpdateContainerPort(template, 8080)
```
rhuss: Why do you use the default port for the update ? I think for clarities sake its better to use something different.

pkg/serving/config_changes_test.go
```diff
@@ -139,6 +139,43 @@ func testUpdateEnvVarsModify(t *testing.T, revision *servingv1alpha1.RevisionTem
...
+			if template.Spec.Containers[p].Ports[q].ContainerPort != 8080 {
+				t.Error("Failed to update the container port")
+			}
+		}
+	}
+
+	// update template with container port info
+	for i, _ := range template.Spec.Containers {
+		for j, _ := range template.Spec.Containers[i].Ports {
+			template.Spec.Containers[i].Ports[j].ContainerPort = 9090
```
rhuss: there is only 1 container and 1 port. If there are more, than that's an error.

pkg/serving/config_changes_test.go
```diff
@@ -139,6 +139,43 @@ func testUpdateEnvVarsModify(t *testing.T, revision *servingv1alpha1.RevisionTem
...
+		}
+	}
+
+	err = UpdateContainerPort(template, 80)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Verify that given port overrides the existing container port
+	for p, _ := range template.Spec.Containers {
```
rhuss: assert again, that there is only 1 container and 1 port after the update (you could put that into an extrac func as its needed twice)

pkg/kn/commands/service/service_update.go
savitaashture: Modified with proper comments

pkg/kn/commands/service/service_create.go
savitaashture: Modified with proper comments

hack/generate-docs.go
savitaashture: When i ran go fmt for complete client code its modified  i have reverted back now.

pkg/serving/config_changes.go
savitaashture: Overwritten the existing port

pkg/serving/config_changes_test.go
savitaashture: Modified

pkg/serving/config_changes_test.go
savitaashture: Modified

pkg/serving/config_changes.go
```diff
@@ -86,6 +86,26 @@ func UpdateImage(template *servingv1alpha1.RevisionTemplateSpec, image string) e
...
 }
 
+// UpdateContainerPort updates container with a give port
+func UpdateContainerPort(template *servingv1alpha1.RevisionTemplateSpec, port int32) error {
+	container, err := extractContainer(template)
+	if err != nil {
+		return err
+	}
+
+	if len(container.Ports) == 0 {
```
navidshaikh: @savitaashture : I think @rhuss mentioned to worry about the len of container ports, just replace it with `container.Ports[0].ContainerPort = port`

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -102,6 +105,11 @@ func (p *ConfigurationEditFlags) Apply(service *servingv1alpha1.Service, cmd *co
 		return err
 	}
 
+	err = servinglib.UpdateContainerPort(template, p.Port)
```
navidshaikh: Say I have configured my service to run on port 9090, now I want to update my service with an environment variable, this `Apply` gets called, irrespective of user intent to update the port, this will update the container port to default one (8080).

invoke this operation by flag check 
`if cmd.Flag().Changed('port')` (as done for `image` above)

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -54,12 +55,14 @@ func (p *ConfigurationEditFlags) AddUpdateFlags(command *cobra.Command) {
 	command.Flags().IntVar(&p.MaxScale, "max-scale", 0, "Maximal number of replicas.")
 	command.Flags().IntVar(&p.ConcurrencyTarget, "concurrency-target", 0, "Recommendation for when to scale up based on the concurrent number of incoming request. Defaults to --concurrency-limit when given.")
 	command.Flags().IntVar(&p.ConcurrencyLimit, "concurrency-limit", 0, "Hard Limit of concurrent requests to be processed by a single replica.")
+	command.Flags().Int32VarP(&p.Port, "port", "p", 8080, "The port that the application should run.")
```
navidshaikh: *The port where application listens on.

pkg/serving/config_changes.go
savitaashture: @navidshaikh this condition i have added because while executing
```kn service create --image imagename --port 80```
during that time contanier.Ports is nil so iam adding that condition to check and initialize the container.Ports.

File: pkg/kn/commands/service/configuration_edit_flags.go (hunk 10 above)
Comment (savitaashture) [reply pkg/kn/commands/service/configuration_edit_flags.go]: Updated

File: pkg/kn/commands/service/configuration_edit_flags.go (hunk 11 above)
Comment (savitaashture) [reply pkg/kn/commands/service/configuration_edit_flags.go]: modified

File: pkg/kn/commands/service/configuration_edit_flags.go [hunk 12]
```diff
@@ -54,12 +55,14 @@ func (p *ConfigurationEditFlags) AddUpdateFlags(command *cobra.Command) {
...
 	command.Flags().IntVar(&p.ConcurrencyTarget, "concurrency-target", 0, "Recommendation for when to scale up based on the concurrent number of incoming request. Defaults to --concurrency-limit when given.")
 	command.Flags().IntVar(&p.ConcurrencyLimit, "concurrency-limit", 0, "Hard Limit of concurrent requests to be processed by a single replica.")
+	command.Flags().Int32VarP(&p.Port, "port", "p", 8080, "The port where application listens on.")
 }
 
 func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
 	p.AddUpdateFlags(command)
 	command.Flags().BoolVar(&p.ForceCreate, "force", false, "Create service forcefully, replaces existing service if any.")
 	command.MarkFlagRequired("image")
+
```
maximilien: why this extra line?

savitaashture: Added by mistake :) removed now.
Thanku

savitaashture:
> Nice job overall. Needed and welcome addition to `kn`.
> 
> Did you consider also adding to the e2e test suite? Would be easy to change one of the created services update its port. Thanks.

@maximilien Thank you
I will add e2e testcases.

rhuss:
Build error is the same as in https://github.com/knative/client/pull/192#issuecomment-504343382 so let's wait until this if fixed and retest here again.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -102,6 +104,13 @@ func (p *ConfigurationEditFlags) Apply(service *servingv1alpha1.Service, cmd *co
 		return err
 	}
 
+	if cmd.Flags().Changed("port") {
```
cppforlife: i think it's a bit weird for us to not set container port _if_ we are going to specify it as a default in the above flag definition.

csantanapr:
Glad to see that this is being added. I just hit this problem where the container I want to use doesn't listen on port 8080

pkg/kn/commands/service/configuration_edit_flags.go
navidshaikh: Right! `Apply` gets called for `create` as well `update` both.

@savitaashture: We'll probably need to identify the command in `Apply` definition. 
`cmd.Name()` returns the command name, we can update this check to not set the port when `cmd.Name() == "update" && !cmd.Flags().Changed("port") ` and set for all other cases.

WDYT ?

savitaashture: @navidshaikh yes current logic will be called while ```create``` and ```update```

savitaashture: @navidshaikh there is one more scenario
consider 
1. I have created a service ```test``` with ```imagename helloworld which runs on 8888```
2. want to update the same service ```test``` with ```imagename httpproxy which runs on 9999```

then in 2nd case update will be successful for image but port will not be updated.

So we need to consider when there is update in image and if there is change in port then also need to execute logic in ```update``` scenario

navidshaikh: @cppforlife: 
> i think it's a bit weird for us to not set container port if we are going to specify it as a default in the above flag definition.

Do you mean, we should update the port value to default one for `service update`, even if user doesnt specify `-p` port for `service update` command ?

navidshaikh: @savitaashture : or we just remove the default value of port and the mentioned conditional should suffice.

savitaashture: yes we can just remove default port value 8080 from the flag and ```if cmd.Flags().Changed("port") {``` condition will work for all the scenario.

cppforlife: i think it's best to rely on the server default and not specify default in the cli (shows up in -h). if user does specify the port configuration then it should be applied.

savitaashture: yes i have removed saying default value from (shows up in -h) and if user specify it will be applied.

pkg/serving/config_changes.go
rhuss: I would just then unconditionally set

```go
container.Ports = []corev1.ContainerPort{{
			ContainerPort: port,
		}}
```

in any case. This makes the code simple to read, and we are sure that only 1 port is set (although as mentioned the slice should have only one entry)

savitaashture: Updated

rhuss:
@savitaashture Thanks a lot, looks good to me. I'm on the integration-test issue, and if this is resolved (it unrelated to this PR), lets merge it.

savitaashture:
@rhuss Thank you :)

PR #194, opened 2019-06-19
List revisions for a given service

### LINKED ISSUE #127: List revisions for given service
Presently, `kn revision get` will return all available revisions. `kn` should be able to filter the revisions for a given service, for eg: `kn revision get --service svc1`.

navidshaikh:
Fixes #127 

```
kn revision list
SERVICE   NAME          AGE     CONDITIONS   READY   REASON
hello     hello-dfc7p   109m    4 OK / 5     True    
hello     hello-h9qdr   20h     4 OK / 5     True    
svc1      svc1-ksds5    7m11s   4 OK / 5     True    

kn revision list -s svc1
SERVICE   NAME         AGE     CONDITIONS   READY   REASON
svc1      svc1-ksds5   7m15s   4 OK / 5     True    

kn revision list -s hello
SERVICE   NAME          AGE    CONDITIONS   READY   REASON
hello     hello-dfc7p   109m   4 OK / 5     True    
hello     hello-h9qdr   20h    4 OK / 5     True    
```

### DIFF
--- a/pkg/kn/commands/revision/revision_list.go
+++ b/pkg/kn/commands/revision/revision_list.go
@@ -18,8 +18,10 @@ import (
 	"fmt"
 
 	"github.com/knative/client/pkg/kn/commands"
+	"github.com/knative/serving/pkg/apis/serving"
 	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/kn/commands/revision/revision_list.go
```diff
@@ -39,7 +48,18 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
...
+			listOptions := v1.ListOptions{}
+			if cmd.Flags().Changed("service") {
+				service := cmd.Flag("service").Value.String()
+				// Ensure requested service exist
+				_, err := client.Services(namespace).Get(service, v1.GetOptions{})
+				if err != nil {
+					return err
+				}
+				listOptions.LabelSelector = labels.Set(
+					map[string]string{serving.ConfigurationLabelKey: service}).String()
```
rhuss: Shouldn't this be `serving.ServiceLabelKey` ? Don't really see how this is tested in the tests, too ...

navidshaikh: Hmm, I checked an actual revision and both the keys in labels are holding service name
```
                "labels": {
                    "serving.knative.dev/configuration": "svc1",
                    "serving.knative.dev/configurationGeneration": "1",
                    "serving.knative.dev/service": "svc1"
                },
```

@rhuss : I will need to check among 2 labels, which one we should filter against.

in tests [here](https://github.com/knative/client/pull/194/files#diff-482bb3cf4550490e7cb84ffc5ddc760fR140).

rhuss: But isn't the service label the authorative answer for which revision belongs to which service and the configuration is just (by accident) named like the service because it has been created implicitely ? Or does the configuration alway be named like the service ? (I don't think so)

PR #197, opened 2019-06-21
[DO NOT MERGE] test presubmit checks

adrcunha:


vendor/github.com/knative/test-infra/scripts/library.sh
```diff
@@ -368,7 +368,13 @@ function update_licenses() {
 function check_licenses() {
   # Fetch the google/licenseclassifier for its license db
   go get -u github.com/google/licenseclassifier
+  echo @@@@@
+  ls /home/prow/go/github.com/google/licenseclassifier
+  echo @@@@@
+  cat /home/prow/go/github.com/google/licenseclassifier/classifier.go
+  echo @@@@@
   # Check that we don't have any forbidden licenses in our images.
+  echo run_go_tool ./vendor/github.com/knative/test-infra/tools/dep-collector dep-collector -check $@
```
rhuss: I think ./vendor/github.com/knative/test-infra/tools/dep-collector is just missing because it was missing as import in pkg/dummy_test.go. Let me try to add this in the other PR and rerun.

rhuss: Sorry, misread the error log, it alteady breaks with running the licenseclassifier.

@adrcunha Please note that we have go modules enabled, so the licenseclassifier should be installed in the vendor dir, or ? (still fishing in the dark)

rhuss:
@adrcunha are you able to remove `/go/bin/dep-collect` on the Prow build machine ? I think if it could be rebuilt with the latest licenseclassifier included then the tests should work again. 

See also #134 for an analysis that an old, existing dep-collect binary has precedence over rebuilding it.

PR #199, opened 2019-06-22
fix(build.sh): Cosmetic spacing fix

rhuss:
Iterm has an issue with rendering multi byte unicode chars. This
commit adds an extra space to certain emojis on iterm only.

This fixes current rendering issues on Linux term because that extra
space was added previously unconditionally.

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -25,6 +25,11 @@ fi
 
 set -eu
 
+# Temporary fix for iTerm issue https://gitlab.com/gnachman/iterm2/issues/7901
+S=""
+if [ -n "$ITERM_PROFILE" ]; then
+  S=" "
+fi
 # Run build
 run() {
   # Switch on modules unconditionally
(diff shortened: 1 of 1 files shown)


rhuss:
Alternatively the icons could be exchanged to emoji which work every where (smaller unicode points)

adrcunha:
```
=========================
==== BUILDING CLIENT ====
=========================
/home/prow/go/src/github.com/knative/client/hack/build.sh: line 30: ITERM_PROFILE: unbound variable
ERROR: cluster setup failed
```

hack/build.sh
```diff
@@ -25,6 +25,11 @@ fi
 
 set -eu
 
+# Temporary fix for iTerm issue https://gitlab.com/gnachman/iterm2/issues/7901
```
adrcunha: I suggest replacing the 4 lines with

```
[[ -n "${ITERM_PROFILE:-}" ]] S=" " || S=""
```

rhuss: Updated but kept the if-syntax although a bit more terse but easier to understand for non-shell gurus.

PR #200, opened 2019-06-23
Fix the periodic CI flow error in client

Fredy-Z:
See https://testgrid.knative.dev/client#continuous.

/cc @adrcunha

### DIFF
--- a/test/e2e/e2e_flags.go
+++ b/test/e2e/e2e_flags.go
@@ -1,0 +1,39 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


adrcunha:
I think this is overkill: 316 files and 50K lines just for a flag. `-emitmetrics` is not used in the e2e tests, so just define it as a no-op flag and add an appropriate comment.

adrcunha:
Ugh, looks like the build job is affected by https://github.com/knative/test-infra/issues/1018

test/e2e/e2e_flags.go
```diff
@@ -0,0 +1,39 @@
...
+*/
+
+package e2e
+
+import (
+	"flag"
+)
+
+// ClientFlags saves the environment flags for client.
+var ClientFlags = initializeFlags()
```
maximilien: How does doing this fix the CI flow errors? Is there something missing in this PR? Or perhaps elaborate the problem and fix. Thx

Fredy-Z: `emitmetrics` is a flag used by our testing framework, which is also *required* for the running of periodic jobs. But since `Knative client` is not using the framework, it does not have the flag, so adding this `flag` as a no-op can solve the error.
I have made it simpler and added more meaningful comments.

maximilien: Not obvious but thanks for the explanation. Hopefully fixes things. Thanks.

PR #207, opened 2019-06-25
Add kn revision delete command

### LINKED ISSUE #203: Implement kn revision delete command
Just like `kn service delete`

navidshaikh:
Fixes #203 

```
➜  client git:(fix-203)  ./kn revision list
SERVICE   NAME          AGE     CONDITIONS   READY     REASON
hello     hello-2krsg   4s      1 OK / 5     Unknown   Deploying
hello     hello-7w8z7   47m     4 OK / 5     True      
kn1       kn1-kcwt7     9m23s   4 OK / 5     True      
svc1      svc1-j7hqr    23h     4 OK / 5     True      

➜  client git:(fix-203)  ./kn revision delete hello-7w8z7
Revision 'hello-7w8z7' successfully deleted in namespace 'default'.

➜  client git:(fix-203)  ./kn revision list              
SERVICE   NAME          AGE     CONDITIONS   READY   REASON
hello     hello-2krsg   16s     5 OK / 5     True    
kn1       kn1-kcwt7     9m35s   4 OK / 5     True    
svc1      svc1-j7hqr    23h     4 OK / 5     True    
```

### DIFF
--- a/pkg/kn/commands/revision/delete.go
+++ b/pkg/kn/commands/revision/delete.go
@@ -33,7 +33,7 @@ func NewRevisionDeleteCommand(p *commands.KnParams) *cobra.Command {
   kn revision delete svc1-abcde`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) != 1 {
-				return errors.New("requires the revision name.")
+				return errors.New("'revision delete' requires the revision name given as single argument")
 			}
 			client, err := p.ServingFactory()
 			if err != nil {
--- a/pkg/kn/commands/revision/delete.go
+++ b/pkg/kn/commands/revision/delete.go
@@ -20,7 +20,7 @@ import (
 
 	"github.com/knative/client/pkg/kn/commands"
 	"github.com/spf13/cobra"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // NewRevisionDeleteCommand represent 'revision delete' command
@@ -43,10 +43,9 @@ func NewRevisionDeleteCommand(p *commands.KnParams) *cobra.Command {
 			if err != nil {
 				return err
 			}
-
 			err = client.Revisions(namespace).Delete(
 				args[0],
-				&v1.DeleteOptions{},
+				&metav1.DeleteOptions{},
 			)
 			if err != nil {
 				return err
--- a/pkg/kn/commands/revision/delete.go
+++ b/pkg/kn/commands/revision/delete.go
@@ -1,0 +1,60 @@
+// Copyright © 2019 The Knative Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package revision
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/knative/client/pkg/kn/commands"
+	"github.com/spf13/cobra"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NewRevisionDeleteCommand represent 'revision delete' command
+func NewRevisionDeleteCommand(p *commands.KnParams) *cobra.Command {
+	RevisionDeleteCommand := &cobra.Command{
+		Use:   "delete NAME",
+		Short: "Delete a revision.",
+		Example: `
+  # Delete a revision 'svc1-abcde' in default namespace
+  kn revision delete svc1-abcde`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
+				return errors.New("requires the revision name.")
+			}
+			client, err := p.ServingFactory()
+			if err != nil {
+				return err
+			}
+			namespace, err := p.GetNamespace(cmd)
+			if err != nil {
+				return err
+			}
+
+			err = client.Revisions(namespace).Delete(
+				args[0],
+				&v1.DeleteOptions{},
+			)
+			if err != nil {
+				return err
+			}
+			fmt.Fprintf(cmd.OutOrStdout(), "Revision '%s' successfully deleted in namespace '%s'.\n", args[0], namespace)
+			return nil
+		},
+	}
+	commands.AddNamespaceFlags(RevisionDeleteCommand.Flags(), false)
+	return RevisionDeleteCommand
+}
--- a/pkg/kn/commands/revision/delete_test.go
+++ b/pkg/kn/commands/revision/delete_test.go
@@ -1,0 +1,66 @@
+// Copyright © 2019 The Knative Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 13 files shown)


rhuss:
Is think there's a race in the test

```
Running 'kn service update hello --env TARGET=kn -n kne2etests'...
Failed to successfully execute 'kn service update hello --env TARGET=kn -n kne2etests': Execution error: stderr: 'Operation cannot be fulfilled on services.serving.knative.dev "hello": the object has been modified; please apply your changes to the latest version and try again
' error: 'exit status 1'FAIL	github.com/knative/client/test/e2e	0.528s
```

rhuss:
Or update is broken.

pkg/kn/commands/revision/delete.go
```diff
@@ -0,0 +1,60 @@
...
+			client, err := p.ServingFactory()
+			if err != nil {
+				return err
+			}
+			namespace, err := p.GetNamespace(cmd)
+			if err != nil {
+				return err
+			}
+
+			err = client.Revisions(namespace).Delete(
```
maximilien: Nit picking... I am all for skipping lines between groups of go statements but inconsistency is worst. So I’d remove the extra line in 46...

navidshaikh:
@maximilien : Thanks for review, I've updated the PR with unit and integration tests (added TestRevisionWorkflow, as IMO this isn't basic workflow).

navidshaikh:
> Is think there's a race in the test

@rhuss : Yup, in [this e2e run](https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_client/207/pull-knative-client-integration-tests/1143832944389918722/) TestBasicWorkflow passed.
`

rhuss:
Oops, test failures ;-(

pkg/kn/commands/revision/delete.go
```diff
@@ -0,0 +1,59 @@
...
+func NewRevisionDeleteCommand(p *commands.KnParams) *cobra.Command {
+	RevisionDeleteCommand := &cobra.Command{
+		Use:   "delete NAME",
+		Short: "Delete a revision.",
+		Example: `
+  # Delete a revision 'svc1-abcde' in default namespace
+  kn revision delete svc1-abcde`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
+				return errors.New("requires the revision name.")
```
rhuss: In "service create" we have 

`'service create' requires the service name given as single argument`

as error message. I suggest to adapt this for now, but we should have a broader discussion around naming labelling. E.g. having a message like "successfully created" can be shortened to just "created" (as this already implies its a success, there's nothing like "unsuccessfully created")

pkg/kn/commands/revision/delete.go
```diff
@@ -0,0 +1,59 @@
...
+				return err
+			}
+			err = client.Revisions(namespace).Delete(
+				args[0],
+				&metav1.DeleteOptions{},
+			)
+			if err != nil {
+				return err
+			}
+			fmt.Fprintf(cmd.OutOrStdout(), "Revision '%s' successfully deleted in namespace '%s'.\n", args[0], namespace)
```
rhuss: As mentioned above, I would always skip "successfully" as this is redundant, but for consistencies sake I recommend to keep it for now.

pkg/kn/commands/revision/delete_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/knative/client/pkg/kn/commands"
+	"k8s.io/apimachinery/pkg/runtime"
+	client_testing "k8s.io/client-go/testing"
+)
+
+func fakeRevisionDelete(args []string) (action client_testing.Action, name string, output []string, err error) {
+	knParams := &commands.KnParams{}
+	cmd, fakeServing, buf := commands.CreateTestKnCommand(NewRevisionCommand(knParams), knParams)
+	fakeServing.AddReactor("delete", "*",
```
rhuss: I would replaces "*" with "revisions" as * is to broad IMO.

pkg/kn/commands/revision/delete_test.go
```diff
@@ -0,0 +1,66 @@
...
+		func(a client_testing.Action) (bool, runtime.Object, error) {
+			deleteAction, ok := a.(client_testing.DeleteAction)
+			action = deleteAction
+			if !ok {
```
rhuss: I think this check on ok can be skipped, as it is supposed to be called always with a delete action, if not its a framework error. And then its ok when the code panics here.

pkg/kn/commands/revision/delete_test.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		t.Error(err)
+		return
+	}
+	expectedOutput := fmt.Sprintf("Revision '%s' successfully deleted in namespace '%s'.", revName, commands.FakeNamespace)
+	if action == nil {
+		t.Errorf("No action")
+	} else if !action.Matches("delete", "revisions") {
+		t.Errorf("Bad action %v", action)
+	} else if output[0] != expectedOutput {
```
rhuss: As mentioned elsewhere some times, why not just match on keywords like that output contains "revisions", "deleteded", the name of the revision, the name of the namespace *somewhere*. This would make the test much more stable. E.g. in case we decide to remove "successfully" then you have adapt this test as it is now, but not if you just check that the context is all provided.

Also this clearly would emphasis, what do you think what the context is (namely the asserted keywords). So its just a copy & paste test which IMO is not so strong.

Another benefit of having a 'contains' check is, that you don't have to split the output into a slice but just can take the output as is. Also, checking that this is in the *first* line is also fragile.

pkg/kn/commands/revision/delete_test.go
```diff
@@ -0,0 +1,66 @@
...
+		t.Errorf("Bad output %s\nExpected output %s", output[0], expectedOutput)
+	} else if name != revName {
```
rhuss: You should also check that the name and namespace is contained in the output But see comment above.

pkg/wait/wait_for_ready.go
```diff
@@ -16,14 +16,15 @@ package wait
 
 import (
 	"fmt"
-	"github.com/knative/pkg/apis"
 	"io"
+	"time"
+
```
rhuss: Also not related to this PR, I see where this comes from. As this PR was rebased on #186 which introduces automatic import reformatting, you can't avoid this here.

This kind of bogus changes will/can happen on all existing PRs but should be gone afterwards.

I'm fine with that.

test/e2e/revision_workflow_test.go
```diff
@@ -0,0 +1,50 @@
...
+		t.Errorf("Error executing 'revision list -o' command. Error: %s", err.Error())
+	}
+	if strings.Contains(revName, "No resources found.") {
+		t.Errorf("Could not find revision name.")
+	}
+	out, err := k.RunWithOpts([]string{"revision", "delete", revName}, runOpts{})
+	if err != nil {
+		t.Errorf("Error executing 'revision delete %s' command. Error: %s", revName, err.Error())
+	}
+	expectedOutput := fmt.Sprintf("Revision '%s' successfully deleted in namespace '%s'.\n", revName, defaultKnE2ENamespace)
```
rhuss: See my comment above why exact matching is not a good idea.

navidshaikh:
@rhuss : Thanks for the review, updated the PR, PTAL.

rhuss:
```
# github.com/knative/client/test/e2e [github.com/knative/client/test/e2e.test]
test/e2e/revision_workflow_test.go:20:2: imported and not used: "fmt"
```

pkg/kn/commands/revision/delete_test.go
```diff
@@ -0,0 +1,59 @@
...
+		return
+	}
+	if action == nil {
+		t.Errorf("No action")
+	} else if !action.Matches("delete", "revisions") {
+		t.Errorf("Bad action %v", action)
+	} else if name != revName {
+		t.Errorf("Bad revision name returned after delete.")
+	}
+	commands.TestContains(t, output, []string{"Revision", revName, "deleted", "namespace", commands.FakeNamespace}, "word in output")
```
rhuss: Nice 👍

PR #214, opened 2019-06-29
refactor(serving): KnClient interface for single point of cluster access [ginkgo version]

rhuss:
This PR exactly the same as #134 except that the test is implemented with Gingko.

The purpose of this PR is to get a feeling for the difference in writing tests in Ginkgo and/or a testframework baset on Golang's intrinsic `testing` with some sugar added.

The relevant tests to compare are:

* Ginkgo --> https://github.com/rhuss/client/blob/pr/add-schema-handling-ginko-test/pkg/serving/v1alpha1/client_test.go
* gotest.tools --> https://github.com/rhuss/client/blob/pr/add-schema-handling/pkg/serving/v1alpha1/client_test.go

Follow the discussion [overthere](https://github.com/knative/client/pull/185#issuecomment-506984415) whether to pick up this PR or #134


maximilien:
Thanks for doing this work @rhuss!
Nice of you going the extra mile to show concrete non-trivial example.
I am ambivalent as to which wins. The great thing is either will be better than what we have now.
So happy whichever decision we (as a team makes): Ginkgo or gotest.tools.

:beers:

rhuss:
> I am ambivalent as to which wins. The great thing is either will be better than what we have now.
> So happy whichever decision we (as a team makes): Ginkgo or gotest.tools.

cool, thanks for your feedback. Let's see how this discussion goes tomorrow, but I expect a definitive decision then.

Then, when we merge either of the two PR, we have also the benefit that in the command unit tests we soon don't have to bother with faking the knative API directly (this is done in the test for KnClient as shown), but with an own mock implementation of the KnClient interface. I will work on this soon after one of these PRs are merged.

rhuss:
closing as the gotest.tools version from #134 goes in.

PR #216, opened 2019-06-30
fix(serving): Fix logic 'service create --force' and some bogus tests

rhuss:
Now a more meaningful error is returned when the user tries to create
a service that already exists. Until know the service is tried
to be created even if the client already knows that it exists.

Also some broken (broken in the sense that the test was nonsense) tests has been fixed.

### DIFF
--- a/pkg/kn/commands/service/service_create_test.go
+++ b/pkg/kn/commands/service/service_create_test.go
@@ -405,7 +405,7 @@ func TestServiceCreateImageExistsAndNoForce(t *testing.T) {
 		t.Fatal(err)
 	}
 	if !strings.Contains(output, "foo") ||
-		!strings.Contains(output, "current") ||
+		!strings.Contains(output, commands.FakeNamespace) ||
 		!strings.Contains(output, "create") ||
 		!strings.Contains(output, "--force") {
 		t.Errorf("Invalid error output: '%s'", output)
--- a/pkg/kn/commands/service/service_create_test.go
+++ b/pkg/kn/commands/service/service_create_test.go
@@ -38,14 +38,16 @@ import (
 
 func fakeServiceCreate(args []string, withExistingService bool, sync bool) (
 	action client_testing.Action,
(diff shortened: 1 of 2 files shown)


pkg/kn/commands/service/service_create_test.go
```diff
@@ -382,17 +398,27 @@ func parseQuantity(t *testing.T, quantityString string) resource.Quantity {
...
-		"service", "create", "foo", "--image", "gcr.io/foo/bar:v1", "--async"}, false, false)
-	if err != nil {
+func TestServiceCreateImageExistsAndNoForce(t *testing.T) {
+	_, _, output, err := fakeServiceCreate([]string{
+		"service", "create", "foo", "--image", "gcr.io/foo/bar:v2", "--async"}, true, false)
+	if err == nil {
 		t.Fatal(err)
 	}
+	if !strings.Contains(output, "foo") ||
+		!strings.Contains(output, "current") ||
```
navidshaikh: should replace `"currrent"` with `commands.FakeNamespace` here ?

pkg/kn/commands/service/service_create_test.go
```diff
@@ -118,7 +134,7 @@ func TestServiceCreateImage(t *testing.T) {
 
 func TestServiceCreateImageSync(t *testing.T) {
 	action, created, output, err := fakeServiceCreate([]string{
-		"service", "create", "foo", "--image", "gcr.io/foo/bar:baz"}, true, true)
+		"service", "create", "foo", "--image", "gcr.io/foo/bar:baz"}, false, true)
```
navidshaikh: @rhuss : can you please help me explain why `nrGetCalled` is used ?

For test case `	[]string{"service", "create", "foo", "--image", "gcr.io/foo/bar:baz"}, false, true)`, shouldn't following conditional fail? (as nrGetCalled should be 1)
```
			if withExistingService || (sync && nrGetCalled > 1) {
```

Issue #111, opened 2019-05-15
Unkown subcommand doesn't return any errors


navidshaikh:
For example, we've renamed `kn service list` -> `kn service get`, however if ran `list`, it just shows the `help` text for `service` group and doesn't return with exit status 1 or prints that `list` is unkown command.
```
➜ kn service list
Service command group

Usage:
  kn service [command]

Available Commands:
  create      Create a service.
  delete      Delete a service.
  describe    Describe available services.
  get         Get available services.
  update      Update a service.

Flags:
  -h, --help   help for service

Global Flags:
      --config string       config file (default is $HOME/.kn.yaml)
      --kubeconfig string   kubectl config file (default is $HOME/.kube/config)

Use "kn service [command] --help" for more information about a command.
➜ echo $?
0
```

This seems to be a known inconsistency with only sub-commands and not with commands. 

However, we've a false positive https://github.com/knative/client/blob/master/test/e2e-tests.sh#L50 .

navidshaikh:
Issue at https://github.com/spf13/cobra/issues/706

maximilien:
We can easily fix this for all command grouped. I'll submit a PR and we can discuss details there.

/assign @maximilien

navidshaikh:
Fixed by https://github.com/knative/client/pull/218

PR #218, opened 2019-07-01
fix(core): generic solution for issue #111 for all command groups

maximilien:
BFS exploration of all sub-commands from `rootCmd` and adds a `RunE` for all commands that are command groups with child sub-commands and without a `RunE`.

The added `RunE` will return the correct errors when the command is called with empty sub-command or with an unknown sub-command. It will also print the command help message first.

----
I did not add a test for this since the we need to decide on #185 and #214 and #134. Once we do I will add the gotest.tools or Ginkgo test for it.

### DIFF
--- a/pkg/kn/core/root.go
+++ b/pkg/kn/core/root.go
@@ -15,6 +15,7 @@
 package core
 
 import (
+	"errors"
 	"flag"
 	"fmt"
 	"os"
@@ -30,7 +31,7 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
 )
 
-// rootCmd represents the base command when called without any subcommands
+// NewKnCommand creates new rootCmd represents the base command when called without any subcommands
 func NewKnCommand(params ...commands.KnParams) *cobra.Command {
 	var p *commands.KnParams
 	if len(params) == 0 {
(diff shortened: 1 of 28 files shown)


maximilien:
/assign @cppforlife

cc: @rhuss

pkg/kn/core/root.go
```diff
@@ -71,11 +72,33 @@ Eventing: Manage event subscriptions and channels. Connect up event sources.`,
...
+// deal with errors when called with empty or unknown sub command
+func emptyAndUnknownSubcommands(cmd *cobra.Command) {
+	for _, childCmd := range cmd.Commands() {
+		if childCmd.HasSubCommands() && childCmd.RunE == nil {
+			childCmd.RunE = func(aCmd *cobra.Command, args []string) error {
+				aCmd.Help()
+				if len(args) == 0 {
+					return errors.New(fmt.Sprintf("please provide a valid sub-command for \"kn %s\"", aCmd.Name()))
+				} else {
+					return errors.New(fmt.Sprintf("unknown command \"%s\" for \"kn %s\"", args[0], aCmd.Name()))
```
navidshaikh: *unkown sub-command

rhuss:
>     3\. Can we add some tests please?

I think @maximilien wants just to wait with tests until with which testing setup we want to continue. So let's wait until tomorrows WG meeting.

pkg/kn/core/root.go
maximilien: Hmm, did not see the changes to the docs. I’ll try again tomorrow.

Yes, once we agree I will add the tests as we agree on the route for tests.

maximilien: OK addressed this in latest push. Resolving. Re-open if I did not. Thx.

maximilien:
OK @rhuss, I think I am convinved `gotests.tools` is better for this project. Added a test for `root.go` that also tests the functionality for issue #111 and I like the result. See latest update here.

cc: @navidshaikh and @cppforlife

rhuss:
@maximilien I think you missed to check in the vendored files (use a `build.sh -u` to populate the vendor dir)

The PR looks good to me, let's merge it tomorrow if there are no objections against gotest.tools in today's WG call.

maximilien:
> @maximilien I think you missed to check in the vendored files (use a `build.sh -u` to populate the vendor dir)
> 
> The PR looks good to me, let's merge it tomorrow if there are no objections against gotest.tools in today's WG call.

Yes, thanks @rhuss, one day I will remember to do this before pushing :) Converting all ginkgo tests now. Best.

rhuss:
@maximilien cool, thanks ! 

Now that gotest.tools is in, we have to rebase our other PRs which introduce them (but should be hopefully trivial)

PR #219, opened 2019-07-01
Exchange SERVICE and NAME column in kn revision list output

### LINKED ISSUE #195: Exchange SERVICE and NAME column in kn revision list output
The NAME should be always first, and only then the service name to which a revision belongs. This not only for consistencies sake, also to make the first column to be the id (and since a revision name typically is prefixed with the service name, you have both info already in the first column)

navidshaikh:
Fixes #195

### DIFF
--- a/pkg/kn/commands/revision/human_readable_flags.go
+++ b/pkg/kn/commands/revision/human_readable_flags.go
@@ -26,10 +26,10 @@ import (
 // RevisionListHandlers adds print handlers for revision list command
 func RevisionListHandlers(h hprinters.PrintHandler) {
 	RevisionColumnDefinitions := []metav1beta1.TableColumnDefinition{
-		{Name: "Service", Type: "string", Description: "Name of the knative service."},
 		{Name: "Name", Type: "string", Description: "Name of the revision."},
+		{Name: "Service", Type: "string", Description: "Name of the Knative service."},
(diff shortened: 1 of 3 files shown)


pkg/kn/commands/revision/human_readable_flags.go
```diff
@@ -52,10 +52,10 @@ func printRevisionList(revisionList *servingv1alpha1.RevisionList, options hprin
 	return rows, nil
 }
 
-// printRevision populates the knative revision table rows
+// printRevision populates the Knative revision table rows
 func printRevision(revision *servingv1alpha1.Revision, options hprinters.PrintOptions) ([]metav1beta1.TableRow, error) {
-	service := revision.Labels[serving.ConfigurationLabelKey]
 	name := revision.Name
+	service := revision.Labels[serving.ConfigurationLabelKey]
```
rhuss: I'm still not sold that it should be the `ConfigurationLabelKey`. It's just by accident that the Configuration is named like the service. You should use `ServiceLabelKey` instead.

navidshaikh:
```
  // ServiceLabelKey is the label key attached to a Route and Configuration indicating by
  // which Service they are created.
  ServiceLabelKey = GroupName + "/service"
```
@rhuss : yup, updating..

Issue #196, opened 2019-06-21
kn fails when KUBECONFIG multi config files


csantanapr:
kn doesn't support when the env variable `KUBECONFIG` has multiple files

As reference: https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#append-home-kube-config-to-your-kubeconfig-environment-variable

For example in my setup I use IKS and have multiple cluster I deal with, so I load `KUBECONFIG` variable with multiple KUBECONFIG paths including the default one.

```
echo $KUBECONFIG
```
value:
```
/Users/csantanapr/.kube/config:/Users/csantanapr/.bluemix/plugins/container-service/clusters/cluster1/kube-config-wdc04-cluster1.yml:/Users/csantanapr/.bluemix/plugins/container-service/clusters/cluster2/kube-config-dal13-cluster2.yml
```
This allows to use things like `kubetcx`
```
$ kubectx
cluster1
cluster2
docker-desktop
```
The CLI `kubectl` works fine current context is `cluster1`
```
kubectl get ns | grep knative
knative-build        Active   98m
knative-eventing     Active   98m
knative-monitoring   Active   98m
knative-serving      Active   98m
knative-sources      Active   98m
```

The problem is `kn` doesn't work and recognize the current context from the merged files like `kubectl`
```
kn service list              
stat /Users/csantanapr/.kube/config:/Users/csantanapr/.bluemix/plugins/container-service/clusters/cluster1/kube-config-wdc04-cluster1.yml:/Users/csantanapr/.bluemix/plugins/container-service/clusters/cluster2/kube-config-dal13-cluster2.yml: no such file or directory
```

navidshaikh:
This should be fixed by PR #36.

csantanapr:
@navidshaikh

But PR #36 has no activity since 3 months ago and it’s closed 🤷🏻‍♂️

navidshaikh:
@csantanapr 
My bad, selected wrong PR from dropdown, this is correct one https://github.com/knative/client/pull/172

csantanapr:
Thank you 🙏🏻

edwardstudy:
Still didn't work if:
```
$ ./kn service list --kubeconfig=$KUBECONFIG
[...]

PR #222, opened 2019-07-02
Support multi config files

edwardstudy:
Fixed: #196 

```
./kn service list --kubeconfig=$KUBECONFIG
the server could not find the requested resource (get services.serving.knative.dev)
```

### DIFF
--- a/pkg/kn/commands/types.go
+++ b/pkg/kn/commands/types.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"path/filepath"
 
 	serving_kn_v1alpha1 "github.com/knative/client/pkg/serving/v1alpha1"
@@ -75,13 +76,25 @@ func (params *KnParams) GetConfig() (serving_v1alpha1_client.ServingV1alpha1Inte
 // GetClientConfig gets ClientConfig from KubeCfgPath
 func (params *KnParams) GetClientConfig() (clientcmd.ClientConfig, error) {
 	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
-	if len(params.KubeCfgPath) > 0 {
-		paths := filepath.SplitList(params.KubeCfgPath)
-		if len(paths) == 1 {
-			loadingRules.ExplicitPath = paths[0]
-		} else {
-			return nil, errors.New(fmt.Sprintf("Config file '%s' could not be found. For configuration lookup path please use the env variable KUBECONFIG", params.KubeCfgPath))
-		}
+	if len(params.KubeCfgPath) == 0 {
+		return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{}), nil
 	}
-	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{}), nil
+
+	_, err := os.Stat(params.KubeCfgPath)
+	if err == nil {
+		loadingRules.ExplicitPath = params.KubeCfgPath
+		return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{}), nil
+	}
+
+	if !os.IsNotExist(err) {
+		return nil, err
+	}
+
+	paths := filepath.SplitList(params.KubeCfgPath)
+	if len(paths) > 1 {
+		return nil, errors.New(fmt.Sprintf("Can not find config file. '%s' looks like a path. "+
+			"Please use the env var KUBECONFIG if you want to check for multiple configuration files", params.KubeCfgPath))
+	} else {
+		return nil, errors.New(fmt.Sprintf("Config file '%s' can not be found", params.KubeCfgPath))
+	}
 }
--- a/pkg/kn/commands/types.go
+++ b/pkg/kn/commands/types.go
@@ -15,7 +15,10 @@
 package commands
 
 import (
+	"errors"
+	"fmt"
 	"io"
+	"path/filepath"
 
 	serving_kn_v1alpha1 "github.com/knative/client/pkg/serving/v1alpha1"
 
@@ -51,11 +54,17 @@ func (params *KnParams) newClient(namespace string) (serving_kn_v1alpha1.KnClien
 	return serving_kn_v1alpha1.NewKnServingClient(client, namespace), nil
 }
 
+// GetConfig returns Serving Client
 func (params *KnParams) GetConfig() (serving_v1alpha1_client.ServingV1alpha1Interface, error) {
-	if params.ClientConfig == nil {
-		params.ClientConfig = params.GetClientConfig()
-	}
 	var err error
+
+	if params.ClientConfig == nil {
+		params.ClientConfig, err = params.GetClientConfig()
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	config, err := params.ClientConfig.ClientConfig()
 	if err != nil {
 		return nil, err
@@ -63,10 +72,16 @@ func (params *KnParams) GetConfig() (serving_v1alpha1_client.ServingV1alpha1Inte
 	return serving_v1alpha1_client.NewForConfig(config)
 }
 
-func (params *KnParams) GetClientConfig() clientcmd.ClientConfig {
+// GetClientConfig gets ClientConfig from KubeCfgPath
+func (params *KnParams) GetClientConfig() (clientcmd.ClientConfig, error) {
 	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
 	if len(params.KubeCfgPath) > 0 {
-		loadingRules.ExplicitPath = params.KubeCfgPath
+		paths := filepath.SplitList(params.KubeCfgPath)
+		if len(paths) == 1 {
+			loadingRules.ExplicitPath = paths[0]
+		} else {
+			return nil, errors.New(fmt.Sprintf("Config file '%s' could not be found. For configuration lookup path please use the env variable KUBECONFIG", params.KubeCfgPath))
+		}
 	}
-	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{})
+	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{}), nil
 }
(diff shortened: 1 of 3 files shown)


pkg/kn/commands/types.go
```diff
@@ -70,7 +71,8 @@ func (c *KnParams) GetConfig() (serving.ServingV1alpha1Interface, error) {
 func (c *KnParams) GetClientConfig() clientcmd.ClientConfig {
 	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
 	if len(c.KubeCfgPath) > 0 {
-		loadingRules.ExplicitPath = c.KubeCfgPath
+		paths := filepath.SplitList(c.KubeCfgPath)
```
rhuss: Can we check if the returned slice has 1 element only and then use `ExplicitPath` instead ? The benefit that this would result in an additional check for this file leading to an error if it does not exist ("fail fast").

edwardstudy: @rhuss Sure. I will add test cases following gotest firstly.

adrcunha:
> @sixolet @adrcunha Is there a way to "overrule" the code coverage check ? Or what is the recommendation for this kind of situation ?

I *think* the coverage check was marked as required by mistake, based on the numbers in https://testgrid.knative.dev/client#coverage

Unless you feel very confident that, at this point that no PR with coverage below 50% should get merged, I suggest that an admin simply uncheck the go coverage job as required to merge.

cppforlife:
i dont think kubectl supports following (someone verify:

```
./kn service list --kubeconfig=$KUBECONFIG
```

it does support KUBECONFIG having multiple paths set though which we support as well via kubectl's library.

original issue mentioned in this pr is fixed via https://github.com/knative/client/pull/172.

rhuss:
good catch, @cppforlife 

> someone verify:

correct, I get an error with a `--kubeconfig=$PATH_LIKE_VAR` : 

```
error: stat /Users/roland/.kube/config:/Users/roland/Development/openshift/ocp4/dev/auth/kubeconfig: no such file or directory
```

@edwardstudy sorry for bugging you with the test and change (really good work, thx again), but what was actually your use case ? Does setting `KUBECONFIG` as env var to a path does not work as expected ? Or why would you like to have a path like arg for `--kubeconfig` ?

pkg/kn/commands/types_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+type typeTestCase struct {
+	kubeCfgPath  string
+	precedence   []string
+	explicitPath string
+}
+
+func TestGetClientConfig(t *testing.T) {
```
maximilien: Nice work @edwardstudy. Consider also using the new `gotest.tools` which should make your tests cleaner, especially assertions and dividing into multiple parts. We agreed during yesterday's WG meeting to move all tests using this small addition to `gotest.tools`. See #134's tests for examples or let's chat today in Beijing, I am working on some. Cheers :beers:

edwardstudy:
@rhuss My case is simple, specify multi configs over `--kubeconfig` flag. When you specify `--kubeconfig`, you should make sure that kn works no matter what values passed. It should work: get current kube context or return a proper error message. What do you think?  😃

rhuss:
> @rhuss My case is simple, specify multi configs over `--kubeconfig` flag. When you specify `--kubeconfig`, you should make sure that kn works no matter what values passed. It should work: get current kube context or return a proper error message. What do you think? 😃

For multiple configs, I think that `kubectl` settle on using an env variable `KUBECONFIG` for this, so the kubectl is like this (in this order) :

* Setting a very specific configuration via `--kubeconfig`
* Setting the location of the configuration files in a path like manner via `KUBECONFIG`
* Fallback to `~/.kube/config

I think your use case can be simply be done with 

```
KUBECONFIG=dir1:dir2 kn service create ....
```

instead of using one `--kubeconfig` with a path or multiple kubeconfigs.

We should mimic the behaviour of `kubectl` here as I think it makes sense, too (no one expects a path with `--kubeconfig`).

I totally agree thought that the error message should be better, maybe even suggesting that `KUBECONFIG` env var should be used for multiple configs.

The current kubectl behaviour is:

```
kubectl get pods --kubeconfig=/Users/roland/.kube/config --kubeconfig=/Users/roland/Development/openshift/ocp4/dev/auth/kubeconfig

---> Take only the last --kubeconfig given

kubectl get pods --kubeconfig=/Users/roland/.kube/config:/Users/roland/Development/openshift/ocp4/dev/auth/kubeconfig

---> error: stat /Users/roland/.kube/config:/Users/roland/Development/openshift/ocp4/dev/auth/kubeconfig: no such file or directory
```

For the last case, I suggest an error message like

```
Configuration file "/Users/roland/.kube/config:/Users/roland/Development/openshift/ocp4/dev/auth/kubeconfig" could not be found. For configuration lookup path please use the environment variable KUBECONFIG.
```

The last sentence I would add when the path contains a `:` (this might be a valid filename part, too (egg on Windows), but printing out that extra error message wouldn't harm either then. Important is that it should be checked if the provided argument exists as a file first.

@edwardstudy does this make sense ?

edwardstudy:
@rhuss Hi, I just recoded and let `os.PathListSeparator` to handle OS-specific path separator.

rhuss:
Let me jump on the PR later (probably tomorrow, its already late here). Thanks for doing the rework !

rhuss:
/retest

(sorry, test failure is again because of a too short timeout in the it test. We will address this soonish).

pkg/kn/commands/types.go
```diff
@@ -34,6 +37,7 @@ type KnParams struct {
 	ClientConfig clientcmd.ClientConfig
 }
 
+// Initialize Initializes command params
```
rhuss: You might consider to remove this comment. It's not really meaningful.

pkg/kn/commands/types.go
```diff
@@ -43,19 +47,31 @@ func (c *KnParams) Initialize() {
 	}
 }
 
+// CurrentNamespace returns current namespace
```
rhuss: Maybe something like
// Return the current namespace which is either provided as option or picked up from kubeconfig

pkg/kn/commands/types.go
rhuss: What's about the following suggestion, which would also work if the filename contains a path seperator (not very likely but IMO totall valid to have a file which contains a ':' as part of the name):

* Check the fullname if it exists as file via `os.Stat`
  - If yes, pick that name
  - If not:
    * Check if the fullname contains a path seperator.
      - If yes: Return error "Can not find configuration file. %s looks like a combine path. Please use the env var KUBECONFIG if you want to check for multiple configuration files"
      - If no: "Configuration file %s can not be found"

Does this make sense ?

pkg/kn/commands/types_test.go
```diff
@@ -0,0 +1,149 @@
...
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+func TestInitialize(t *testing.T) {
+	p := &KnParams{}
+
+	p.Initialize()
+
+	if reflect.ValueOf(p.ServingFactory).Pointer() != reflect.ValueOf(p.GetConfig).Pointer() {
```
rhuss: Does it really make sense to check for implementation details like this ? This is not behaviour and how this is wired is really an implementation detail not visible from the outside.

pkg/kn/commands/types_test.go
```diff
@@ -0,0 +1,149 @@
...
+		t.Errorf("wrong function assignment detected: %v (expected) != %v (actual)", reflect.ValueOf(p.GetConfig).Pointer(), reflect.ValueOf(p.ServingFactory).Pointer())
+	}
+
+	if reflect.ValueOf(p.NamespaceFactory).Pointer() != reflect.ValueOf(p.CurrentNamespace).Pointer() {
```
rhuss: Same as above. Every test that use reflections to check for internal implementation details has a smell.

pkg/kn/commands/types_test.go
```diff
@@ -0,0 +1,149 @@
...
+			errors.New(fmt.Sprintf("Config file '%s' could not be found. For configuration lookup path please use the env variable KUBECONFIG", multiConfigs)),
+		},
+	} {
+		p := &KnParams{
+			KubeCfgPath: tc.kubeCfgPath,
+		}
+
+		clientConfig, err := p.GetClientConfig()
+
+		if !reflect.DeepEqual(err, tc.expectedError) {
```
rhuss: Checking on exact error strings can be quite fragile. Better often to check whether important information is contained like the name of the config file, "not", "found", "path"). As @maximilien mentioned, we are now using the gotest.tools assertion library, so you write the test like in

```golang
assert.Assert(util.ContainsAll(err.String(), "not", "found", multiConfigs)
```

pkg/kn/commands/types_test.go
maximilien: Can we make `Initialize()` private?

rhuss: I believe `Initialize()` is part of the cobra command contract/interface so visibility can't be really changed.

maximilien:
@edwardstudy you also need to rebase to resolve `pkg/kn/commands/types.go` conflict

pkg/kn/commands/types.go
navidshaikh: +1 on improving the error message returned `"Config file '%s' could not be found. For configuration lookup path please use the env variable KUBECONFIG`

pkg/kn/commands/types_test.go
maximilien: OK... thx

pkg/kn/commands/types.go
```diff
@@ -51,22 +55,46 @@ func (params *KnParams) newClient(namespace string) (serving_kn_v1alpha1.KnClien
...
+		return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{}), nil
+	}
+
+	if !os.IsNotExist(err) {
+		return nil, err
+	}
+
+	paths := filepath.SplitList(params.KubeCfgPath)
+	if len(paths) > 1 {
+		return nil, errors.New(fmt.Sprintf("Can not find config file. '%s' looks like a combine path. "+
```
rhuss:
```suggestion
		return nil, errors.New(fmt.Sprintf("Can not find config file. '%s' looks like a path. "+
```

edwardstudy:
Done. Sorry for the delay in getting a chance to respond to this PR.

Issue #188, opened 2019-06-14
Consolidate all `kn` output and use `KnParams.Output`


maximilien:
Indeed. Saw that. And an item to discuss. Why are not all output redirected in `KnParams.Output`. I will open an issue on this if it's not done. It should help also address #182 in a simpler fashion. Been wanting to have a central place for all communication to user. This is good time to do it.

_Originally posted by @maximilien in https://github.com/knative/client/pull/185/review_comment/create_

PR #223, opened 2019-07-03
test(commands): backfills gotest.tools tests for version & completion

maximilien:
`gotest.tools` tests for `kn version` and `kn completion` commands.

NOTE the additions to `test_helper.go` to be able to read `os.Stdout` from commands, i.e., `ReadStdout(t)`, `CaptureStdout(t)`, and `ReleaseStdout(t)`. Once a fix for #188 is solved then we can inject `io.Writer` object for commands and read output there and remove these.

Closes issue #161 and #185


pkg/kn/commands/test_helper.go
```diff
@@ -38,6 +52,51 @@ func CreateTestKnCommand(cmd *cobra.Command, knParams *KnParams) (*cobra.Command
 	return knCommand, fakeServing, buf
 }
 
+// TestContains is a test helper function, checking if a substring is present in given
+// output string
+func TestContains(t *testing.T, output string, sub []string, element string) {
```
rhuss: I don't think you should start your func with "Test" as this is a trigger for unit tests functions (in _test.go files, though). Maybe just `ContainsAll` ? 

Ah, I see, you just have moved that functions (not used by yourself). So lets keep it for now and clean it up in another PR.

rhuss:
The test coverage has an issue anyway IMO:

* Why are only the code marked as covered used by the changed files in this PR and not by every file ? I mean the test_helper.go tools are for sure in other tests still, just not from the changed code in this PR.
* How can we exclude helper code used in tests from regular compilation ? In Java that's easy as all test code is in a different directory hierachy. Here, we code-under-test and test code are sitting side by side, its not so easy. Or maybe just reame `test_helper.go` to `helper_test.go` (marking this a test file) ?

I think the latter suggestion could work. Could you please try that out, @maximilien ?

pkg/kn/commands/test_helper.go
maximilien: Cool, agree. I don't think I added this func.

maximilien:
> * How can we exclude helper code used in tests from regular compilation ? In Java that's easy as all test code is in a different directory hierarchy. Here, we code-under-test and test code are sitting side by side, its not so easy. Or maybe just rename `test_helper.go` to `helper_test.go` (marking this a test file) ?

Happy to help, however, when you do this then `go test` thinks `helper_test.go` has test code if there is any func with Test in it. So:

```bash
➜  client git:(issue161_v2) mv pkg/kn/commands/test_helper.go pkg/kn/commands/helper_test.go
➜  client git:(issue161_v2) ✗ ./hack/build.sh
🧽  Format
⚖️  License
📖 Docs
🚧 Compile
🧪  Test
🔥 Failure
# github.com/knative/client/pkg/kn/commands
/Users/maximilien/Desktop/github/knative-client/src/github.com/knative/client/pkg/kn/commands/helper_test.go:57:1: wrong signature for TestContains, must be: func TestContains(t *testing.T)
FAIL	github.com/knative/client/pkg/kn/commands [setup failed]
[...]
```
 
> I think the latter suggestion could work. Could you please try that out, @maximilien ?

Renaming to `testing_helper.go` which should work. See update.

rhuss:
Test still breaks ;-(

> Happy to help, however, when you do this then `go test` thinks `helper_test.go` has test code if there is any func with `Test` as prefix in it. So:

Maybe we could just rename it still to helper_test and just add a dummy test? (cool idea by @jcrossley3)

rhuss:
@maximilien 

> ```shell
> # github.com/knative/client/pkg/kn/commands
> /Users/maximilien/Desktop/github/knative-client/src/github.com/knative/client/pkg/kn/commands/helper_test.go:57:1: wrong signature for TestContains, must be: func TestContains(t *testing.T)
> FAIL	github.com/knative/client/pkg/kn/commands [setup failed]
> ```

Ok, the real reason is really that a _helper_ method is called `TestContains` and if the file is called `helper_test.go` then go considers to be a *Test* not a helper function and hence must confirm to the test contract.

Could you please rename `TestContains` into something else (like `ContainsAllOf`) as I mentioned in my PR review and then we should be fine.

rhuss:
I think this would be also a good occasion to get ones feet wet with the `Comparison` api as described in https://vincent.demeester.fr/posts/2018-08-16-gotest-tools-assertions/ (or just use `cmp.Contains` which also a comparison but only for one thing to check). @navidshaikh would this maybe something you wanna take (with an extra PR) ?

navidshaikh:
@rhuss : Ack!

navidshaikh:
@rhuss : PTAL https://github.com/knative/client/pull/227 and see if it makes sense.

rhuss:
@maximilien when #227 and this PR rebased, testing_helpers.go can be renamed to helpers_test.go (as the bogus `Test..` method was removed with #227) so that it doesn't count towards code coverage anymore and we can merge then.

rhuss:
As #227 is now merged, this PR could be updated, and the test_helper.go could be renamed to helper_test.go (the `TestContains` method has been removed in #227).

This will hopefully fix the client go coverage issue.

maximilien:
> As #227 is now merged, this PR could be updated, and the test_helper.go could be renamed to helper_test.go (the `TestContains` method has been removed in #227).
> This will hopefully fix the client go coverage issue.

OK. Will add this to my TODO items next. Trying to complete a few items that have been pending for a while.

maximilien:
OK rebased and pushed. Not 100% clear why the `pull-knative-client-go-coverage ` is still failing

maximilien:
Yup @rhuss only `pull-knative-client-go-coverage` is failing :( so maybe OK to merge and hopefully subsequent PRs won't still have this issue for me.

rhuss:
@maximilien I'm pretty sure when you rename `testing_helper.go` to `help_test.go` that the coverage test will pass (as there is no coverage measurement for test files). This didn't work before, because `testing_helper.go` contained a method starting with `Test...` and this method hadn't the expected signature. This is gone now, an nothing should prevent us anymore to rename the file (which is good, as this file only contains test util code anyway)

maximilien:
> @maximilien I'm pretty sure when you rename `testing_helper.go` to `help_test.go` that the coverage test will pass (as there is no coverage measurement for test files). This didn't work before, because `testing_helper.go` contained a method starting with `Test...` and this method hadn't the expected signature. This is gone now, an nothing should prevent us anymore to rename the file (which is good, as this file only contains test util code anyway)

Actually that causes errors since the `help_test.go` does not compile as go will ignore files with suffix `*_test.go`.

maximilien:
But removing the `TestContains` method might do the trick. Thanks. Let's see the test results.

maximilien:
OK finally @rhuss we are good 😀 whew

rhuss:
@maximilien nice. Tbh, I would have called it pkg/kn/commands/helper_test.go instead of pkg/kn/commands/help_testing.go so that it is considered a test file and wouldn't be even considered for test coverage (and would even compile only for regular tests).

But let's bring that forward and we can still fix this later (its not a serious issue imo). We need to fix the tests anyway as one of the first things after 0.2.0 is out.

PR #227, opened 2019-07-04
Add custom comparison for checking multiple substrings

navidshaikh:
This changeset introduces custom comparison for checking if a target
 string has given multiple substrings, namely `ContainsMultipleSubstrings`.

 Also removes `commands.TestContains` function and replaces all references
 of `commands.TestContains` with `gotest.tools/assert.Check` function and
 uses `ContainsMultipleSubstrings` as comparison function.

### DIFF
--- a/pkg/kn/commands/compare_utils.go
+++ b/pkg/kn/commands/compare_utils.go
@@ -1,0 +1,40 @@
+// Copyright © 2019 The Knative Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package commands
+
+import (
+	"fmt"
+	"strings"
+
+	"gotest.tools/assert/cmp"
+)
+
+// ContainsMultipleSubstring is a comparison utility, compares given substrings against
+// target string and returns the gotest.tools/assert/cmp.Comaprison function.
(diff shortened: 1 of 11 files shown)


pkg/kn/commands/compare_utils.go
```diff
@@ -0,0 +1,40 @@
...
+	"fmt"
+	"strings"
+
+	"gotest.tools/assert/cmp"
+)
+
+// ContainsMultipleSubstring is a comparison utility, compares given substrings against
+// target string and returns the gotest.tools/assert/cmp.Comaprison function.
+// Provide message to form an error message of format 'Missing $message: $missing_elements'
+func ContainsMultipleSubstrings(target string, substrings []string, message string) cmp.Comparison {
```
rhuss: I would change the signature to

```
ContainsMultipleSubstrings(message string, target string, substrings ...string)
```

so that the usage is easier (no slice creation required) and the message is more prominent in the beginning.

A usage then would look like

```
commands.ContainsMultipleSubstrings("value", output[2], "bar-wxyz", "svc1"))
```

The second argument will be always a variable name and the other probably mostly literals. So makes a nice separation, too.

Btw, not sure that the "message" is required. You could add that message to the assert.

So, tbh, I would like to see an assertion like:

```golang
assert.Check(t, commands.ContainsMultipleSubstrings(output[0], "NAME", "URL", "AGE", "CONDITIONS", "TRAFFIC"), "missing column header")
```

i.e. using the message slot from `Check()`, which is more natural anyway.

rhuss: And maybe call it `ContainsAll` (as its clear that it's about string from the signature, and `All` is more normative than just `Multiple`) (so kind of an equivalent to `string.Contains()` ?

navidshaikh: @rhuss : Renamed the function and also made it variadic, also removed the message arg altogether, the compare function now generates the generic error message. I think as long as it can show the diff about compare, we don't need custom message.
Example
```
        Actual output: foo-abcd   foo       <unknown>   0 OK / 0     <unknown>   <unknown>
        Missing strings: foo-abcd1, foo1
```

Issue #205, opened 2019-06-25
Implement kn route describe command


navidshaikh:
with `kn route describe` taking an argument for route name

zhangtbj:
Raise a PR #228 to fix this issue.

jchesterpivotal:
Following from the [4 Feb 2020 CLI WG meeting](https://docs.google.com/document/d/1Uh7jTWQruBBmic-WmTvtc9cMF95kQrKb5lsqWhNuikM/edit#heading=h.98s6rrkt5nbu), I think that we need to decide:

1. Whether we're keeping `route describe`, and
2. If we're keeping it, what format it should take.

This issue might need to be closed -- _technically_ there's a command and it does spit out information. But the UX is quite different from `service describe` or `revision describe`.

Issue #206, opened 2019-06-25
Implement kn revision describe command


navidshaikh:
with `kn revision describe` taking an argument for revision name

zhangtbj:
Hi @navidshaikh ,

I saw the `kn revision describe` is implemented here:
https://github.com/knative/client/blob/master/pkg/kn/commands/revision/revision_describe.go

I think it is already finished. correctly?

navidshaikh:
Hi @zhangtbj : Present `kn revision describe` prints the yaml output of the object, the goal is to present the revision object in better human readable format with some additional details.

zhangtbj:
Thx @navidshaikh , now I see the yaml output is native revision information output from kube. And kn client also provides other formats like json, name....

what is mean `better human readable format`? And is there any rule or standard that how to present?

zhangtbj:
Oh, I saw the format is discussed in this PR:
https://github.com/knative/client/pull/75

Let us also discuss there. Thanks

rhuss:
@zhangtbj I will split up the #75 PR soonish, focussing first on the non-color options. There won't be much activity on this PR anymore therefor.

At #235 I started a discussion about the general nature of `kn <noun> describe`, so its worth to watch and comment on this, too.

navidshaikh:
This is the only issue pending for v0.8.0 milestone, can we scope this for next milestone and cut v0.8.0?

@rhuss @sixolet @maximilien WDYT ?

rhuss:
I'm fine with postponing this feature for the next release, we already have made quite some progress since 0.2.0 which would justify a release on its own.

navidshaikh:
Implemented by #475 
/close

PR #228, opened 2019-07-05
Fix issue #205 to add route describe feature and test

zhangtbj:
1, Implement kn route describe command with required parameters:
```
✔ $ go run cmd/kn/main.go route describe --help
Describe available route.

Usage:
  kn route describe NAME [flags]

Flags:
  -o, --output string                 Output format. One of: json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-file. (default "yaml")

Global Flags:
```

2, Add `gotest.tools` tests for kn route describe commands:
 - test without route name
 - test route describe with the default output
 - test route describe with the YAML output
 - test route describe with the JSON output
 - test route describe with the name output

For issue: #205

### DIFF
--- a/docs/cmd/kn.md
+++ b/docs/cmd/kn.md
@@ -6,8 +6,9 @@ Knative client
 
 Manage your Knative building blocks:
 
-* [Serving](https://github.com/knative/serving/tree/master): Manage your services and release new software to them.
-* [Eventing](https://github.com/knative/eventing/tree/master): Manage event subscriptions and channels. Connect event sources.
+Serving: Manage your services and release new software to them.
+Build: Create builds and keep track of their results.
+Eventing: Manage event subscriptions and channels. Connect up event sources.
 
 ### Options
 
--- a/pkg/kn/commands/service/service_create.go
+++ b/pkg/kn/commands/service/service_create.go
@@ -104,10 +104,16 @@ func NewServiceCreateCommand(p *commands.KnParams) *cobra.Command {
 			}
 
 			if !waitFlags.Async {
-				err := client.WaitForService(name, time.Duration(waitFlags.TimeoutInSeconds)*time.Second, cmd.OutOrStdout())
+				out := cmd.OutOrStdout()
+				fmt.Fprintf(out, "Waiting for service '%s' to become ready ... ", name)
+				flush(out)
+
+				err := client.WaitForService(name, time.Duration(waitFlags.TimeoutInSeconds)*time.Second)
 				if err != nil {
+					fmt.Fprintln(out)
 					return err
 				}
+				fmt.Fprintln(out, "OK")
 				return showUrl(client, name, namespace, cmd.OutOrStdout())
 			}
 
@@ -118,6 +124,17 @@ func NewServiceCreateCommand(p *commands.KnParams) *cobra.Command {
 	editFlags.AddCreateFlags(serviceCreateCommand)
 	waitFlags.AddConditionWaitFlags(serviceCreateCommand, 60, "service")
 	return serviceCreateCommand
+}
+
+// Duck type for writers having a flush
+type flusher interface {
+	Flush() error
+}
+
+func flush(out io.Writer) {
+	if flusher, ok := out.(flusher); ok {
+		flusher.Flush()
+	}
 }
 
 func createService(client v1alpha1.KnClient, service *serving_v1alpha1_api.Service, namespace string, out io.Writer) error {
--- a/pkg/serving/v1alpha1/client.go
+++ b/pkg/serving/v1alpha1/client.go
@@ -16,7 +16,6 @@ package v1alpha1
 
 import (
 	"fmt"
-	"io"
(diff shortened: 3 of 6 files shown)


pkg/kn/commands/route/route_describe_test.go
```diff
@@ -0,0 +1,130 @@
...
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "foo",
+				Namespace: "default",
+			},
+		}
+	}
+
+	t.Run("requires the route name", func(t *testing.T) {
+		_, _, err := fakeRouteDescribe([]string{"route", "describe"}, &v1alpha1.Route{})
```
maximilien: Why is `setup(t)` not called here? Perhaps do it for consistency

pkg/kn/commands/route/route_describe_test.go
```diff
@@ -0,0 +1,130 @@
...
+	})
+
+	t.Run("describe a valid route with default output", func(t *testing.T) {
+		setup(t)
+
+		action, output, err := fakeRouteDescribe([]string{"route", "describe", "test-foo"}, &expectedRoute)
+		assert.Assert(t, err == nil)
+		assert.Assert(t, action != nil)
+		assert.Assert(t, action.Matches("get", "routes"))
+
```
maximilien: Not sure I see the value of JSON conversion here. Why not just parse the YAML returned (it’s the default output)

maximilien:
Nice job @zhangtbj left a few minor comments. Adding @navidshaikh as reviewer too.

pkg/kn/commands/route/route_describe_test.go
zhangtbj: Hi Max, because this test is only for `wihtout route name` test. So don't need any fake route to be initialized in setup. So I skip it. If we hope to be consistent. I can add it back. Thanks!

pkg/kn/commands/route/route_describe_test.go
zhangtbj: Yes, in this test, I parse to json format and Unmarshal it to a route object returnedRoute. Then compare it with expectedRoute

rhuss:
@zhangtbj coudl you please rebase on master ? Gotest.tools is already included on master but also part of this PR. This should not be the case.

pkg/kn/commands/route/route_describe.go
```diff
@@ -0,0 +1,68 @@
...
+			describeRoute, err := client.Routes(namespace).Get(args[0], v1.GetOptions{})
+			if err != nil {
+				return err
+			}
+
+			printer, err := routeDescribePrintFlags.ToPrinter()
+			if err != nil {
+				return err
+			}
+			describeRoute.GetObjectKind().SetGroupVersionKind(schema.GroupVersionKind{
```
rhuss: Can we please get #134 merged first as this directly is changed in this PR ?

I mark this PR as 'on-hold' if this ok, and remove it if we have #134 in place. 'hope this is ok.

pkg/kn/commands/route/route_describe.go
```diff
@@ -0,0 +1,68 @@
...
+				Group:   "knative.dev",
```
rhuss: its the wrong group anyway, but that gets fixed with #134.

pkg/kn/commands/route/route_describe.go
```diff
@@ -0,0 +1,68 @@
...
+				Version: "v1alpha1",
+				Kind:    "Route"})
+			err = printer.PrintObj(describeRoute, cmd.OutOrStdout())
```
rhuss: 'route describe' (or as I would love to rename it to 'route show', but that's another discussion ;)  should imo really be about human readable output. I think its ok to have -o name|json|yaml, too but the main focus is to have a detaile out of the the routes, what revisions they are targetting which with which split and which service the route is targetting.

See #75 for the idea (although this PR is too huge and as soon as #134 is merged it will be broken down).

navidshaikh:
@zhangtbj : Thanks for your PR.
`show` (or `describe` verb) for a resource should print better representation of a resource's detailed info as compare to printing yaml for same. This is the reason we have `revision describe` #206 open. 

Please check work done by @rhuss on #75 , we'd like to produce something like that. I'd say lets discuss info we'd like show for route `show` or `describe` on the issue.

rhuss:
I opened #235 for discussion the general direction we want to go with `kn <noun> show` (or `kn <noun> describe`)

maximilien:
@rhuss are we adding this to discussion tomorrow or do we already have too much?

In any case, let's make sure to document decision(s) in WG or here as @zhangtbj is in Beijing and the time for the meeting is 1:30a so it's not clear he can attend in person.

zhangtbj:
Sorry I closed this PR by mistake. I created a new PR https://github.com/knative/client/pull/251 which is rebased from master and change to use new knClient.  Thanks!

PR #234, opened 2019-07-05
refactor(service): Add a ProgressHandler for the Wait interface

rhuss:
This remove the io.Writer from the client interface and replaces it with
ProgressHandler interface.

A SimpleProgressHandler instance can be used to print out the wait progress
on a writer.

This commit relies on that being PR #134 merged first.


pkg/wait/progress_handler.go
```diff
@@ -0,0 +1,87 @@
...
+	if complete < 0 {
+		fmt.Fprint(w.out, w.ticMark)
+	} else {
+		fmt.Fprintf(w.out, " %d%%", complete)
+	}
+}
+
+// Printout ERROR label, but not the error.
+// The error will be printed out later anyway
+func (w *simpleProgressHandler) Fail(err error) {
```
sixolet: Unsure of this code. I feel like the contract "the progress handler will print an ERROR label but you-the-caller have to fill out the rest of the line" kind of splits responsibilities uncomfortably. Do we have any other reasonable options?

rhuss: Tbh, the handler is free to do what it wants to do with the error (so the documentation is misleading and explains the default implementation). 

The contract is, that this method will be called when error occurs and the WaitForServce() method will return this error.

Thinking about this, maybe we don't even need the ProgressHandler as success/error can be easily handled based on the return value of the method (it's a synchronous method). Only for the progress indication (which we currently don't really use), we would need the callback, but the start and end indicators can be completely handled outside the method.

Let me check, how I can make that simpler.

rhuss:
I'm closing this PR in favour of #248 which is much simpler and achieves the same: Removing `io.Writer` from the KnClient interface. 

In case we ever need a ProgressHandler to print out real progress (like watching and printing the various phases during startup), we could reintroduce this interface later again.

For now its just overkill. Thanks for the heads-up, @sixolet

PR #236, opened 2019-07-08
Update kn.md

joaedwar:
Components are not bulleted

### DIFF
--- a/docs/cmd/kn.md
+++ b/docs/cmd/kn.md
@@ -6,9 +6,9 @@ Knative client
 
 Manage your Knative building blocks:
 
-* Serving: Manage your services and release new software to them.
-* Build: Create builds and keep track of their results.
-* Eventing: Manage event subscriptions and channels. Connect event sources.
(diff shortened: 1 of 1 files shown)


docs/cmd/kn.md
```diff
@@ -6,9 +6,9 @@ Knative client
 
 Manage your Knative building blocks:
 
-Serving: Manage your services and release new software to them.
-Build: Create builds and keep track of their results.
-Eventing: Manage event subscriptions and channels. Connect up event sources.
+* [Serving](https://github.com/knative/serving/tree/master): Manage your services and release new software to them.
+* [Build](https://github.com/knative/build/tree/v0.7.0): Create builds and keep track of their results.
```
navidshaikh: Not sure if we want to keep references for Build.

rhuss: Right, we should remove altogether as we will never support this.

rhuss:
```suggestion
```

PR #239, opened 2019-07-08
Add gofmt simplify code option in hack/build.sh

navidshaikh:
-  Adds step to simplify code along with goimports
-  Runs goimports on all packages except vendor dir and filter on only *.go files
-  Runs gofmt on all dirs except vendor and filters on *.go files, also writes the results instead of printing diff
- Fixes typo s/insteat/insead
- Adds the changes after adding simply code step in hack/build.sh

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -16,6 +16,8 @@
 
 set -o pipefail
 
+source_dirs="cmd pkg test"
+
 # Store for later
 if [ -z "$1" ]; then
     ARGS=("")
@@ -81,7 +83,7 @@ run() {
 
 go_fmt() {
   echo "🧹 ${S}Format"
-  go fmt ./cmd/... ./pkg/...
+  find $(echo $source_dirs) -name "*.go" -print0 | xargs -0 gofmt -s -w
 }
 
 source_format() {
@@ -92,11 +94,12 @@ source_format() {
      echo "✋   go get golang.org/x/tools/cmd/goimports"
      echo "✋ to enable import cleanup. Import cleanup skipped."
 
-     # Run go fmt insteat
+     # Run go fmt instead
      go_fmt
   else
      echo "🧽 ${S}Format"
-     goimports -w cmd pkg
+     goimports -w $(echo $source_dirs)
+     find $(echo $source_dirs) -name "*.go" -print0 | xargs -0 gofmt -s -w
   fi
   set -e
 }
--- a/pkg/serving/config_changes.go
+++ b/pkg/serving/config_changes.go
@@ -146,7 +146,7 @@ func extractContainer(template *servingv1alpha1.RevisionTemplateSpec) (*corev1.C
 
 func updateEnvVarsFromMap(env []corev1.EnvVar, vars map[string]string) []corev1.EnvVar {
 	set := make(map[string]bool)
-	for i, _ := range env {
+	for i := range env {
 		envVar := &env[i]
 		value, present := vars[envVar.Name]
 		if present {
--- a/pkg/serving/config_changes_test.go
+++ b/pkg/serving/config_changes_test.go
@@ -118,7 +118,7 @@ func TestUpdateEnvVarsModify(t *testing.T) {
 
 func testUpdateEnvVarsModify(t *testing.T, revision *servingv1alpha1.RevisionTemplateSpec, container *corev1.Container) {
 	container.Env = []corev1.EnvVar{
-		corev1.EnvVar{Name: "a", Value: "foo"}}
+		{Name: "a", Value: "foo"}}
 	env := map[string]string{
 		"a": "fancy",
 	}
@@ -176,8 +176,8 @@ func TestUpdateEnvVarsBoth(t *testing.T) {
 
 func testUpdateEnvVarsBoth(t *testing.T, template *servingv1alpha1.RevisionTemplateSpec, container *corev1.Container) {
 	container.Env = []corev1.EnvVar{
-		corev1.EnvVar{Name: "a", Value: "foo"},
-		corev1.EnvVar{Name: "c", Value: "caroline"}}
+		{Name: "a", Value: "foo"},
+		{Name: "c", Value: "caroline"}}
 	env := map[string]string{
 		"a": "fancy",
 		"b": "boo",


navidshaikh:
I saw this in serving https://github.com/knative/serving/pull/4628/, lets add it in our hack/build.sh to ensure we push simplified version.

hack/build.sh
```diff
@@ -81,7 +81,9 @@ run() {
 
 go_fmt() {
   echo "🧹 ${S}Format"
-  go fmt ./cmd/... ./pkg/...
+  gofmt -w $(find -name '*.go' | grep -v vendor)
+  echo "🧹 ${S}Format: Simplify code.."
```
rhuss: I think we should skip that log message (as well as below). The message it intentionally kept terse to not clutter the CLI. If you need more details you could always use `--verbose`. The overall "Format" message is good enough (even when it encompasses  two commands)

hack/build.sh
```diff
@@ -81,7 +81,9 @@ run() {
 
 go_fmt() {
   echo "🧹 ${S}Format"
-  go fmt ./cmd/... ./pkg/...
+  gofmt -w $(find -name '*.go' | grep -v vendor)
```
rhuss: is 'gofmt' always available ? (go fmt is) If not we would need to check for it, too.

Also, I'd prefer a positive list (cmd, pgk, e2e) instead of a black list (-v vendor). To easily thing could slip through otherwise.

hack/build.sh
```diff
@@ -92,11 +94,13 @@ source_format() {
...
 
-     # Run go fmt insteat
+     # Run go fmt instead
      go_fmt
   else
      echo "🧽 ${S}Format"
-     goimports -w cmd pkg
+     goimports -w $(find -name '*.go' | grep -v vendor)
+     echo "🧽 ${S}Format: Simplify code.."
+     gofmt -s -w $(find -path './vendor' -prune -o -type f -name '*.go' -print)
```
rhuss: This find as well as above will break on filenames which contain spaces.

Also, you need to be careful as e.g. `-prune` is not available for macOs' find.

```
find -path './vendor' -prune -o -type f -name '*.go' -print
find: illegal option -- p
usage: find [-H | -L | -P] [-EXdsx] [-f path] path ... [expression]
       find [-H | -L | -P] [-EXdsx] -f path [path ...] [expression]
```

hack/build.sh
rhuss: I think its best to define a script variable like "source_dirs" which point to the top-level source directories and use that in every check/formatting command.

rhuss:
```suggestion
  find $(echo $source_dirs) -name "*.go" -print0 | xargs gofmt -w
```

rhuss:
```
source_dirs="cmd pkg test"
```

hack/build.sh
```diff
@@ -92,11 +94,13 @@ source_format() {
      echo "✋   go get golang.org/x/tools/cmd/goimports"
      echo "✋ to enable import cleanup. Import cleanup skipped."
 
-     # Run go fmt insteat
+     # Run go fmt instead
      go_fmt
   else
      echo "🧽 ${S}Format"
-     goimports -w cmd pkg
+     goimports -w $(find -name '*.go' | grep -v vendor)
```
rhuss:
```suggestion
     goimports -w $(echo $source_dirs)
```

hack/build.sh
rhuss:
```suggestion
     find $(echo $source_dir) -name "*.go" -print0 | xargs gofmt -s -w
```

rhuss: Why are you using `-s` here but not in the variation above ?

navidshaikh: I thought on that, but as I originally used separate Format message for simplifying code, I used a separate command altogether.

hack/build.sh
navidshaikh: resolved

hack/build.sh
navidshaikh: `gofmt` should be always available as it's packaged along with other golang binaries shipped part of golang installation, also go fmt calls `gofmt -w -l` under the cover

navidshaikh: explicit `source_dir` sounds good, updated.

hack/build.sh
navidshaikh: good catch!

Wonder if the clients should have provision in CI system to test against different platforms we ship the binaries for..

hack/build.sh
navidshaikh: thanks!

navidshaikh: added -null option to xargs to suppress following warning 
```
xargs: WARNING: a NUL character occurred in the input.  It cannot be passed through in the argument list.  Did you mean to use the --null option?
```

hack/build.sh
navidshaikh: With single message for formatting, added `-s` option.

hack/build.sh
```diff
@@ -81,7 +83,7 @@ run() {
 
 go_fmt() {
   echo "🧹 ${S}Format"
-  go fmt ./cmd/... ./pkg/...
+  find $(echo $source_dirs) -name "*.go" -print0 | xargs --null gofmt -s -w
```
rhuss:
```suggestion
  find $(echo $source_dirs) -name "*.go" -print0 | xargs -0 gofmt -s -w
```

`--null` is not known on macOs. `-0` should do the trick everywhere.

hack/build.sh
```diff
@@ -92,11 +94,12 @@ source_format() {
...
      echo "✋ to enable import cleanup. Import cleanup skipped."
 
-     # Run go fmt insteat
+     # Run go fmt instead
      go_fmt
   else
      echo "🧽 ${S}Format"
-     goimports -w cmd pkg
+     goimports -w $(echo $source_dirs)
+     find $(echo $source_dirs) -name "*.go" -print0 | xargs --null gofmt -s -w
```
rhuss:
```suggestion
     find $(echo $source_dirs) -name "*.go" -print0 | xargs -0 gofmt -s -w
```

hack/build.sh
```diff
@@ -126,7 +129,7 @@ go_test() {
 }
 
 check_license() {
-  echo "⚖️ ${S}License"
+  echo "⚖️ ${S} License"
```
rhuss:
```suggestion
  echo "⚖️ ${S}License"
```

Issue #224, opened 2019-07-03
Flaky smoke test


rhuss:
In https://storage.googleapis.com/knative-prow/pr-logs/pull/knative_client/221/pull-knative-client-integration-tests-latest-release/1146230064891826179/build-log.txt there is still a flaky test:

```
namespace/kne2esmoketests created
Service 'svc1' successfully created in namespace 'kne2esmoketests'.
Service 'hello' successfully created in namespace 'kne2esmoketests'.
Waiting for service 'hello' to become ready ... OK

Service URL:
http://hello.kne2esmoketests.example.com
NAME    DOMAIN   GENERATION   AGE   CONDITIONS   READY   REASON
hello            1            48s   3 OK / 3     True    
svc1             1            48s   3 OK / 3     True    
Service 'hello' updated in namespace 'kne2esmoketests'.
NAME          SERVICE   AGE   CONDITIONS   READY   REASON
hello-dm9gg   hello     29s   4 OK / 4     True    
svc1-z7kbh    svc1      29s   4 OK / 4     True    
NAME    DOMAIN   GENERATION   AGE   CONDITIONS   READY   REASON
hello            1            48s   3 OK / 3     True    
svc1             1            48s   3 OK / 3     True    
Operation cannot be fulfilled on services.serving.knative.dev "hello": the object has been modified; please apply your changes to the latest version and try again
```

This is exactly this test:

```
./kn service create hello --force --image gcr.io/knative-samples/helloworld-go -e TARGET=Awesome -n $KN_E2E_SMOKE_TESTS_NAMESPACE || fail_test
```

So I suspect that there is an issue with replacing an existing service like this in our replacement code:

```golang
func replaceService(client v1alpha1.KnClient, service *serving_v1alpha1_api.Service, namespace string, out io.Writer) error {
	existingService, err := client.GetService(service.Name)
	if err != nil {
		return err
	}
	service.ResourceVersion = existingService.ResourceVersion
	err = client.UpdateService(service)
	if err != nil {
		return err
	}
	fmt.Fprintf(out, "Service '%s' successfully replaced in namespace '%s'.\n", service.Name, namespace)
	return nil
}
```

rhuss:
Another flake:

```
Failed to successfully execute 'kn service create hello --image gcr.io/knative-samples/helloworld-go -n kne2etests': Execution error: stderr: 'RevisionMissing: Configuration "hello" does not have any ready Revision.
[...]

PR #240, opened 2019-07-08
fix(service update): Retry an update in case of a conflict.

rhuss:
Is related to #224 and should fix one flake.

### DIFF
--- a/pkg/kn/commands/service/service_create.go
+++ b/pkg/kn/commands/service/service_create.go
@@ -148,7 +148,7 @@ func createService(client v1alpha1.KnClient, service *serving_v1alpha1_api.Servi
 
 func replaceService(client v1alpha1.KnClient, service *serving_v1alpha1_api.Service, namespace string, out io.Writer) error {
 	var retries = 0
-	for true {
+	for {
 		existingService, err := client.GetService(service.Name)
 		if err != nil {
 			return err
@@ -166,8 +166,6 @@ func replaceService(client v1alpha1.KnClient, service *serving_v1alpha1_api.Serv
 		fmt.Fprintf(out, "Service '%s' successfully replaced in namespace '%s'.\n", service.Name, namespace)
 		return nil
 	}
-	// This line will be never reached, but go insists on having it
-	return fmt.Errorf("can not update service '%s' in namespace '%s'", service.Name, namespace)
 }
 
 func serviceExists(client v1alpha1.KnClient, name string, namespace string) (bool, error) {
--- a/pkg/kn/commands/service/service_create.go
+++ b/pkg/kn/commands/service/service_create.go
@@ -147,17 +147,27 @@ func createService(client v1alpha1.KnClient, service *serving_v1alpha1_api.Servi
 }
 
 func replaceService(client v1alpha1.KnClient, service *serving_v1alpha1_api.Service, namespace string, out io.Writer) error {
-	existingService, err := client.GetService(service.Name)
-	if err != nil {
-		return err
+	var retries = 0
+	for true {
+		existingService, err := client.GetService(service.Name)
+		if err != nil {
+			return err
+		}
+		service.ResourceVersion = existingService.ResourceVersion
+		err = client.UpdateService(service)
+		if err != nil {
+			// Retry to update when a resource version conflict exists
+			if api_errors.IsConflict(err) && retries < MaxUpdateRetries {
+				retries++
+				continue
+			}
+			return err
+		}
+		fmt.Fprintf(out, "Service '%s' successfully replaced in namespace '%s'.\n", service.Name, namespace)
+		return nil
 	}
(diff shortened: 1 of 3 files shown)


pkg/kn/commands/service/service_create.go
```diff
@@ -131,17 +131,27 @@ func createService(client serving_v1alpha1_client.ServingV1alpha1Interface, serv
...
-	_, err = client.Services(namespace).Update(service)
-	if err != nil {
-		return err
+	var retries = 0
+	for true {
+		existingService, err := client.Services(namespace).Get(service.Name, v1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		service.ResourceVersion = existingService.ResourceVersion
```
navidshaikh: @rhuss : should we take these 3 operations off for loop? (get, err != nil and assigning ResourceVersion)
as we should get the ResourceVersion value same in each retry?

rhuss: No, it will be a new resource version so need to be set for each retry. Actually, the conflict occurs because of its a different resource version on the server than that we have sent. So we need to pick up this new resource version.

navidshaikh: ah okay! thanks!

pkg/kn/commands/service/service_create.go
```diff
@@ -131,17 +131,27 @@ func createService(client serving_v1alpha1_client.ServingV1alpha1Interface, serv
...
+		_, err = client.Services(namespace).Update(service)
+		if err != nil {
+			// Retry to update when a resource version conflict exists
+			if api_errors.IsConflict(err) && retries < MaxUpdateRetries {
```
navidshaikh: should we add a message here about conflict occurred and we're retrying ? 
this is just to let user know about provision we have in client for retrying
WDYT @rhuss ?

rhuss: I don't think its needed as the outcome will be the same. For `kn service create --force` we are unconditionally overwriting the service anyway, regardless what it contains on the server side. So there is no visible difference for the user.

For `kn service update` we are also refetching the current object and only apply our changes also on second and third retry.

Doing a retry is mostly an implementation details a user should not worry about.

navidshaikh: alright.

pkg/kn/commands/service/service_create.go
```diff
@@ -131,17 +131,27 @@ func createService(client serving_v1alpha1_client.ServingV1alpha1Interface, serv
...
-	existingService, err := client.Services(namespace).Get(service.Name, v1.GetOptions{})
-	if err != nil {
-		return err
-	}
-	service.ResourceVersion = existingService.ResourceVersion
-	_, err = client.Services(namespace).Update(service)
-	if err != nil {
-		return err
+	var retries = 0
+	for true {
```
cppforlife: `for true` should just be `for`

pkg/kn/commands/service/service_create.go
```diff
@@ -131,17 +131,27 @@ func createService(client serving_v1alpha1_client.ServingV1alpha1Interface, serv
...
+			}
+			return err
+		}
+		fmt.Fprintf(out, "Service '%s' successfully replaced in namespace '%s'.\n", service.Name, namespace)
+		return nil
 	}
-	fmt.Fprintf(out, "Service '%s' successfully replaced in namespace '%s'.\n", service.Name, namespace)
-	return nil
+	// This line will be never reached, but go insists on having it
+	return fmt.Errorf("can not update service '%s' in namespace '%s'", service.Name, namespace)
```
cppforlife: common convention in go in such cases is to use `panic("Unreachable")`

pkg/kn/commands/service/service_update.go
```diff
@@ -53,24 +55,33 @@ func NewServiceUpdateCommand(p *commands.KnParams) *cobra.Command {
...
-				return err
-			}
-			service = service.DeepCopy()
+			var retries = 0
+			for true {
+				service, err := client.Services(namespace).Get(args[0], v1.GetOptions{})
+				if err != nil {
+					return err
+				}
+				service = service.DeepCopy()
```
cppforlife: i dont think we need DeepCopy

rhuss: Not introduced in this PR, but happy to remove if safe. it was just my general understanding to copy objects before working on them to not compromise shared caches and other data structures which hold a reference.

rhuss:
* Change to for { } loop, now the final statement is not necessary anymore
* Kept DeepCopy as I think it doesn't harm. Happy to remove in a subsequent PR (wasn't introduced by this PR)

// @cppforlife

pkg/kn/commands/service/service_update.go
sixolet: I'm with rhuss.

PR #241, opened 2019-07-08
chore(changelog): Add a changelog

rhuss:


cppforlife:
@rhuss what's the rationale for having this file? is it for devs or end users?

rhuss:
> @rhuss what's the rationale for having this file? is it for devs or end users?

end users, a curated list which is growing while going and not as an afterthought when doing a release.

rhuss:
the format of this changelog is kind of experimental, happy to revert it to a more conventional (bullet list like) format.

sixolet:
/lgtm
I like emoji

PR #243, opened 2019-07-08
Bulleted components in kn.md

joaedwar:


rhuss:
@joaedwar I think it does not work as expected, as the documentation is generated. I will send a PR which fixes that.

PR #244, opened 2019-07-08
fix: Update root docs for removing Build reference.

rhuss:


rhuss:
@joaedwar we have to use that fix, as the documentation is auto generated. As the description is also used as textual output for `kn --help`, we can't use markdown links here, as this doesnt render nicely on the console. But I added a bullet list and removed teh "build" reference.

pkg/kn/core/root.go
```diff
@@ -49,9 +49,8 @@ func NewKnCommand(params ...commands.KnParams) *cobra.Command {
 		Short: "Knative client",
 		Long: `Manage your Knative building blocks:
 
-Serving: Manage your services and release new software to them.
-Build: Create builds and keep track of their results.
```
navidshaikh: Lets remove it from https://github.com/knative/client/blob/master/pkg/kn/core/root.go#L53 , else it will come back.

rhuss: It is already removed, isn't it ? (you are refering to a deleted line)

navidshaikh: damn!

rhuss:
> @rhuss: you cannot LGTM your own PR.

thought it was small enough to fail. But yes, have to wait then.

rhuss:
@sixolet @cppforlife @maximilien Can we have a quick /lgtm for this mini doc fix ? It slipped in and now every new PR get the bogus generated docs as part of the PR. 

thanks

maximilien:
@rhuss it's likely that the unresolved conflict  in `docs/cmd/kn.md` is preventing the merge.

rhuss:
@cppforlife @maximilien conflict resolved, PTAL for an /lgtm

rhuss:
/retest

Flakiness in smoke test should be fixed with #250

PR #247, opened 2019-07-08
fix(service): Show URL (external) instead of Address (internal) when listing service

### LINKED ISSUE #246: No domain/address for `kn service list` shown for serving 0.7.0
The reason is that the old, deprecated DOMAIN field is not populated anymore. One should pick up the `status.address.url` field instead.

Also, the column "DOMAIN" should be renamed to "ADDRESS".

rhuss:
Starting with 0.7.0 old fields are not populated anymore. Let's switch to the new fields then.

This fix should work with 0.6.0, too as the new fields already have been populated back then.

So we can unconditionally pick status.address.url which is accordance with
our policy to support the latest release and the release before.

Fixes #246.

### DIFF
--- a/pkg/kn/commands/service/human_readable_flags.go
+++ b/pkg/kn/commands/service/human_readable_flags.go
@@ -26,7 +26,7 @@ import (
 func ServiceListHandlers(h hprinters.PrintHandler) {
 	kServiceColumnDefinitions := []metav1beta1.TableColumnDefinition{
 		{Name: "Name", Type: "string", Description: "Name of the Knative service."},
-		{Name: "Address", Type: "string", Description: "Address of the Knative service."},
+		{Name: "Url", Type: "string", Description: "URL of the Knative service."},
 		//{Name: "LastCreatedRevision", Type: "string", Description: "Name of last revision created."},
 		//{Name: "LastReadyRevision", Type: "string", Description: "Name of last ready revision."},
 		{Name: "Generation", Type: "integer", Description: "Sequence number of 'Generation' of the service that was last processed by the controller."},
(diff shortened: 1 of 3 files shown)


pkg/kn/commands/service/service_list_test.go
```diff
@@ -73,12 +76,13 @@ func TestServiceListDefaultOutput(t *testing.T) {
 	} else if !action.Matches("list", "services") {
 		t.Errorf("Bad action %v", action)
 	}
-	assert.Check(t, util.ContainsAll(output[0], "NAME", "DOMAIN", "GENERATION", "AGE", "CONDITIONS", "READY", "REASON"))
+	assert.Check(t, util.ContainsAll(output[0], "NAME", "ADDRESS", "GENERATION", "AGE", "CONDITIONS", "READY", "REASON"))
```
navidshaikh: How about using `URL` for column header ?
We are referring this field as `URL` in `kn route list`.
WDYT ?

rhuss: URL (`status.url`) and ADDRESS (`status.addess`) are different fields:

* `URL` is the externally routable endpoint for this service going over the ingress (http://hello.default.example.com)
* `ADDRESS` is the internal k8s service address (http://hello.default.svc.cluster.local)

In the list I think we should only show the externally routed URL, or ? So you are right, should be URL then (but have to check whether we are really picking up the right field)

rhuss: Changed to URL, PTAL.

pkg/kn/commands/service/human_readable_flags.go
```diff
@@ -25,8 +25,8 @@ import (
 // ServiceListHandlers adds print handlers for service list command
 func ServiceListHandlers(h hprinters.PrintHandler) {
 	kServiceColumnDefinitions := []metav1beta1.TableColumnDefinition{
-		{Name: "Name", Type: "string", Description: "Name of the knative service."},
-		{Name: "Domain", Type: "string", Description: "Domain name of the knative service."},
+		{Name: "Name", Type: "string", Description: "Name of the Knative service."},
```
navidshaikh: :+1:

Issue #182, opened 2019-06-12
Add a -q/quiet option


duglin:
While the current model of printing a "success" message can be useful for newbies, I actually think it kind of odd since IMO most Linux commands tend to be silent unless there are error - or there's really important info you need to get your next job done.  But things like creating a service really shouldn't need any output when things work.  I tend to think that messages like this are handy the first couple of times you run a command (as I said, for newbies), then after that it gets annoying because I've been conditioned to learn that "no output means things worked" and anything else is something I need to look at because odds are it means there are error messages.  Remember, most people don't do `echo $?` after they run a command to see the exit code, so asking them to scan the output for an error message is (IMO) asking a lot when people are flying thru their commands to get a job done.

However, I'm not going to suggest we drop those messages (yet, even though I do think it would be better to do so :-) ).  For now, I'd be ok with just adding a `-q/quiet` type of flag so that I can hide this verbosity. An env var would be nice too so I can set it once and forget about it.

What do people think?

rhuss:
I'm happy to support a quiet mode, it's a perfectly valid use case. However, I think there is valuable information which can be printed out, too,  which goes beyond a pure success:

* The URL how a service is reachable (coming with #156) so you save an extra call for looking this up
* The service name (or its 'id') used when after the service is created (you know that discussion ;). Not sure whether this comes, though.

So I'm all for supporting both: Using a human-friendly message for interactive CLI usage and a silent usage for script (but even then it might be useful e.g. when running in a CI environment).

maximilien:
+ 1, quiet option makes sense.

duglin:
Per our call today, some additional thoughts and things I've seen. Using `tar` as the example linux app:
```
$ tar -cf a.tar *   # will show no output because it worked

$ tar -vcf a.tar *   # will show list of files processed due to -v option, these go to stdout
file1
file2
```

And this makes sense to me because 1) there's no error, so nothing on stderr, and 2) the user asked for additional output, so it goes to stdout.

Now, there might be other flags (like `--debug` type of flags) that generate output and I do actually think it makes sense for that to go to stderr so they don't disrupt your normal workflow. I think "warnings" fit into that category but I'm not sure, I'm having trouble finding a good standard linux cmd that printing warnings.

rhuss:
I still believe that the default assumption should be that `kn` is used as an interactive tool, so having by default success and error messages for e.g. `kn service create` is much more user-friendly than having only an error code to detect whether an operation is successful or not.

[...]

PR #249, opened 2019-07-09
feature(kn plugins): Implements Kn plugins re-using some code from kubectl plugins.

### LINKED ISSUE #92: kn plugins
Add ability for the `kn` CLI to have plugins. This is particularly useful for three important scenarios:

1. allows users of `kn` to extend it with functionality
2. allows experimentation of commands that we don't want to include as a core command, e.g., `$kn doctor`
3. allows those wanting to package `kn` with their own offering of Knative to extend `kn` with additional commands that either:
  a. is not yet included in the current release
  b. are experimental
  c. are specific to a cloud environment, e.g., IBM's Cloud or Google Cloud

There could be additional reasons to have plugins to `kn` so please add them here. Also if you are against such a feature, please comment your reasoning here too.

maximilien:
This version contains the following:

1. wraps the main root Kn command to support plugin
2. plugins are any executable in kn's config new pluginDir
   variable which defaults to $PATH
3. plugins must have name kn-*
4. 'kn plugin list' sub-command to list found kn plugins
5. skips any kn plugins found with name that match core
   commands, e.g., kn-service would be ignored
6. can execute any valid kn plugins found, e.g.,
   `kn valid` where the plugin file `kn-valid` is in path
   specified in 2.
7. unit tests (using gotest.tools)

And is missing:

1. integration tests
2. plugin install command
3. plugin repository command
4. plugin / Knative server version negotiation
5. anything else we agree on in plugin req doc

I plan to create issues for the things missing so we don't
end up with an even bigger PR. It's already big as is but is a
good MVP as per plugins requirement doc.

Fixes #92


maximilien:
OK so `pull-knative-client-go-coverage` is failing since I don't have any unit tests for:

* `pkg/kn/commands/plugin/plugin_handler.go`
* `pkg/kn/commands/plugin/plugin_list.go`

Fair and good catch. I will add these tomorrow and update this PR. However, feel free to start your review and/or testing this. Thx.

maximilien:
OK added tests (lots) for `pkg/kn/commands/plugin/plugin_list.go`. 

I'll work on adding for `pkg/kn/commands/plugin/plugin_handler.go` next.

rhuss:
Maybe now it's a good time to move help_testing.go to helper_test.go ? ;-)

maximilien:
> Maybe now it's a good time to move help_testing.go to helper_test.go ? ;-)

OK, we are ping ponging this file, but for consistency, agree. when I push the `plugin_handler_test.go` I'll also rename. Cheers :beers:

rhuss:
Ok, I see. 

I really talked with some folks who know testing better than me, and they assured me its possible to have _test.go files without any tests which are used only during tests. That's why is so dogged ;)

Let's get the fixed because I really think that these helpers should be reachable only in test scope.

maximilien:
> Let's get the fixed because I really think that these helpers should be reachable only in test scope.

Let me push the `pkg/kn/commands/plugin/plugin_handler_test.go` today. Which completes this PR and we can see what makes sense. Thx.

maximilien:
OK @rhuss and others, this version has all the test but the coverage fails since I have a common `testing_helper.go` that the coverage tool needs to ignore. I cannot name it `helper_test.go` since doing so fails compilation.

I think we need to ask the `test_infra` team if they have a way to flag or tag a file to be skipped in the coverage test. Similar to tagging files for ignoring during build, e.g.:

```go
// +build !linux
...
```

Perhaps @adrcunha has some magic up his sleeves already... I know he's been super helpful in the past. Hoping.

cmd/kn/main.go
```diff
@@ -19,12 +19,22 @@ import (
...
+	defer cleanup()
+	err := core.NewDefaultKnCommand().Execute()
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
 }
+
+func cleanup() {
+	viper.WriteConfig()
```
duglin: Does this mean that any update to the config will only be written when the CLI exits? It may not be a big deal but if there are two CLIs active at the same time  then 1) they could over-write each other and while this will always be true, the longer they wait to write it the bigger chance it'll happen (instead of read, update, write immediately), 2) if a CLI command takes a while to run and a 2nd CLI command starts it won't see the updated config until the first one finishes, which might seem odd to the user.
Net: can we write after each update instead?

docs/cmd/kn.md
```diff
@@ -16,11 +16,13 @@ Eventing: Manage event subscriptions and channels. Connect up event sources.
       --config string       config file (default is $HOME/.kn/config.yaml)
```
duglin: just to make it clear what the diff is between this and the --kubeconfig value, can you make this `kn config file...`  - just add `kn` to the description

docs/cmd/kn.md
```diff
@@ -16,11 +16,13 @@ Eventing: Manage event subscriptions and channels. Connect up event sources.
...
   -h, --help                help for kn
       --kubeconfig string   kubectl config file (default is $HOME/.kube/config)
+      --plugin-dir string   kn plugin directory (default is value in kn config or $PATH) (default "$PATH")
```
duglin: Looks weird to have 2 defaults in the description. I'd remove the 2nd sentence.

docs/cmd/kn_completion.md
```diff
@@ -22,6 +22,7 @@ kn completion [flags]
 ```
+      --plugin-dir string   kn plugin directory (default is value in kn config or $PATH) (default "$PATH")
```
duglin: remove 2nd sentence

docs/cmd/kn_plugin.md
```diff
@@ -0,0 +1,34 @@
+## kn plugin
+
+Plugin command group
+
+### Synopsis
+
+Provides utilities for interacting with kn plugins.
+
+Plugins provide extended functionality that is not part of the major kn command-line distribution.
```
duglin: s/major/core/

adrcunha:
> OK @rhuss and others, this version has all the test but the coverage fails since I have a common `testing_helper.go` that the coverage tool needs to ignore. I cannot name it `helper_test.go` since doing so fails compilation.

Its current code  coverage is at 47%. 3pp more and the code coverage job will pass. If you click the file name in the coverage report above you'll see the methods that weren't tested.

I think I mentioned it in some other PR, but if the requirement of having at least 50% code coverage in PRs is currently too hard, an admin can mark the job optional so it won't block submissions.

docs/cmd/kn_plugin.md
```diff
@@ -0,0 +1,34 @@
...
+### Options
+
+```
+  -h, --help   help for plugin
+```
+
+### Options inherited from parent commands
+
+```
+      --config string       config file (default is $HOME/.kn/config.yaml)
```
duglin: add `kn` to help text

docs/cmd/kn_plugin.md
```diff
@@ -0,0 +1,34 @@
...
+      --kubeconfig string   kubectl config file (default is $HOME/.kube/config)
+      --plugin-dir string   kn plugin directory (default is value in kn config or $PATH) (default "$PATH")
```
duglin: remove 2nd ()

docs/cmd/kn_plugin_list.md
```diff
@@ -0,0 +1,38 @@
...
+
+```
+  -h, --help        help for list
+      --name-only   If true, display only the binary name of each plugin, rather than its full path
+```
+
+### Options inherited from parent commands
+
+```
+      --config string       config file (default is $HOME/.kn/config.yaml)
```
duglin: `kn`

pkg/kn/commands/plugin/path_verifier.go
```diff
@@ -0,0 +1,102 @@
...
+// Verify implements PathVerifier and determines if a given path
+// is valid depending on whether or not it overwrites an existing
+// kn command path, or a previously seen plugin.
+func (v *CommandOverrideVerifier) Verify(path string) []error {
+	if v.Root == nil {
+		return []error{fmt.Errorf("unable to verify path with nil root")}
+	}
+
+	// extract the plugin binary name
+	segs := strings.Split(path, "/")
```
duglin: should you error check for path being "" ?

duglin: I guess below you assume it starts with "kn-" so I guess not...  unless you want to error check for "kn-" as the prefix  :-)

pkg/kn/commands/plugin/path_verifier.go
```diff
@@ -0,0 +1,102 @@
...
+	cmdPath := strings.Split(binName, "-")
+	if len(cmdPath) > 1 {
+		// the first argument is always "kn" for a plugin binary
+		cmdPath = cmdPath[1:]
+	}
+
+	errors := []error{}
+	isExec, err := isExecutable(path)
+	if err == nil && !isExec {
+		errors = append(errors, fmt.Errorf("warning: %s identified as a kn plugin, but it is not executable", path))
```
duglin: Either make this an error , add a flag to hide warnings or remove this check - otherwise this will just continually annoy people seeing this each time for what is a valid situation.

I'd prefer to just remove it since IMO it's not an error, but 2nd vote is for making it an error.

pkg/kn/commands/plugin/path_verifier.go
```diff
@@ -0,0 +1,102 @@
...
+	} else if err != nil {
+		errors = append(errors, fmt.Errorf("error: unable to identify %s as an executable file: %v", path, err))
```
duglin: Then just skip it w/o saying anything. The problem with saying something is that if it's complaining about something in the PATH that the user has no control over then they're going to get this every time and can't do anything about it.  If we can't execute the file then we should just silently skip it since it doesn't meet the criteria for what a valid plugin is.

pkg/kn/commands/plugin/path_verifier.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	if existingPath, ok := v.SeenPlugins[binName]; ok {
+		errors = append(errors, fmt.Errorf("warning: %s is overshadowed by a similarly named plugin: %s", path, existingPath))
```
duglin: Same comments as above - this will just annoy people for what is a valid state. sh doesn't complain about dups on PATH so why should we?  If people really want to help here then let's add some kind of "plugin verify" action that can print all of these kinds of "warnings" for people. Otherwise let's just annoy folks.

pkg/kn/commands/plugin/path_verifier.go
```diff
@@ -0,0 +1,102 @@
...
+	} else {
+		v.SeenPlugins[binName] = path
+	}
+
+	cmd, _, err := v.Root.Find(cmdPath)
+	if err == nil {
+		errors = append(errors, fmt.Errorf("warning: %s overwrites existing command: %q", binName, cmd.CommandPath()))
```
duglin: ditto
Plus, once we remove all of these wanting we can make this func just return a single error instead of an array - which IMO is kind of weird to begin with.

Also, you're now merging warning and errors which basically makes warning treated as errors - so they're not warnings any more - they'll stop the flow. That's a bit inconsistent

pkg/kn/commands/plugin/path_verifier.go
```diff
@@ -0,0 +1,102 @@
...
+		fileExt := strings.ToLower(filepath.Ext(fullPath))
+
+		switch fileExt {
+		case ".bat", ".cmd", ".com", ".exe", ".ps1":
+			return true, nil
+		}
+		return false, nil
+	}
+
+	if m := info.Mode(); !m.IsDir() && m&0111 != 0 {
```
duglin: not sure what to do about it, but... technically it could be executable by this user even if it's not marked as executable by "group" or "other".  But I think that's a harder check.... just saying....

pkg/kn/commands/plugin/path_verifier_test.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+var cleanup = func(t *testing.T) {
+	if pluginPath != "" {
+		DeleteTestPlugin(t, pluginPath)
+	}
+}
+
+func TestCommandOverrideVerifier_Verify_with_nil_root_command(t *testing.T) {
+	t.Run("returns error verifying path", func(t *testing.T) {
```
duglin: Technically this level of indirection isn't needed - you can just do everything in this Func at the top level and get basically the same results

pkg/kn/commands/plugin/plugin.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"github.com/knative/client/pkg/kn/commands"
+	"github.com/spf13/cobra"
+)
+
+func NewPluginCommand(p *commands.KnParams) *cobra.Command {
+	pluginCmd := &cobra.Command{
+		Use:   "plugin",
+		Short: "Plugin command group",
+		Long: `Provides utilities for interacting with kn plugins.
```
duglin: I wonder if s/interacting with/managing/  would sounds better since we're not actually interacting/talking-to the plugins.

pkg/kn/commands/plugin/plugin_test.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"strings"
+	"testing"
+
+	"github.com/knative/client/pkg/kn/commands"
+	"github.com/spf13/cobra"
+	"gotest.tools/assert"
+)
+
+const PluginCommandUsage = `Provides utilities for interacting with kn plugins.
```
duglin: s/interacting with/managing/

pkg/kn/commands/plugin/plugin_test.go
```diff
@@ -0,0 +1,109 @@
...
+		assert.Assert(t, rootCmd != nil)
+	}
+
+	t.Run("creates a new cobra.Command", func(t *testing.T) {
+		setup(t)
+
+		assert.Assert(t, pluginCmd != nil)
+		assert.Assert(t, pluginCmd.Use == "plugin")
+		assert.Assert(t, pluginCmd.Short == "Plugin command group")
+		assert.Assert(t, strings.Contains(pluginCmd.Long, "Provides utilities for interacting with kn plugins."))
```
duglin: "managing"

docs/cmd/kn_plugin.md
```diff
@@ -0,0 +1,34 @@
+## kn plugin
+
+Plugin command group
+
+### Synopsis
+
+Provides utilities for interacting with kn plugins.
```
duglin: s/interacting with/managing/

pkg/kn/commands/plugin/plugin.go
```diff
@@ -0,0 +1,33 @@
...
+
+Plugins provide extended functionality that is not part of the major kn command-line distribution.
```
duglin: s/major/core/

cmd/kn/main.go
maximilien: Not sure that’s the write thing. Though I’ll consider it. One could argue that writing immediately is also not good since the command executing has not completed so having intermediary side effect is odd

docs/cmd/kn.md
maximilien: it’s obvious from the last part `(default ...)` but does not hurt to add

docs/cmd/kn.md
maximilien: Yup. Good catch

docs/cmd/kn_completion.md
maximilien: Same comment as above

maximilien:
> > OK @rhuss and others, this version has all the test but the coverage fails since I have a common `testing_helper.go` that the coverage tool needs to ignore. I cannot name it `helper_test.go` since doing so fails compilation.
> 
> Its current code coverage is at 47%. 3pp more and the code coverage job will pass. If you click the file name in the coverage report above you'll see the methods that weren't tested.
> 
> I think I mentioned it in some other PR, but if the requirement of having at least 50% code coverage in PRs is currently too hard, an admin can mark the job optional so it won't block submissions.

Yeah, that’s fair. I was thinking about that when I wrote my comment yesterday :) But I do think we need a way to add helper test go files that do not count towards coverage. Am I the only asking for this? :)

pkg/kn/commands/plugin/path_verifier_test.go
maximilien: I want to separate the different context w/o having to write multiple tests

pkg/kn/commands/plugin/path_verifier.go
maximilien: It will fail with some error to user. Let’s get people using this to have concrete feedback

pkg/kn/commands/plugin/path_verifier.go
maximilien: Dude, try it and you will see how it works.

pkg/kn/commands/plugin/path_verifier.go
maximilien: For me this is where a quiet option as in #182 would help. Once that’s in I will quiet these.

pkg/kn/commands/plugin/path_verifier.go
maximilien: See my comment about enabling quiet when we do #182

pkg/kn/commands/plugin/path_verifier.go
maximilien: Same comment as below. I think #182 when done will be the way to do the behavior you are asking for

pkg/kn/commands/plugin/path_verifier.go
maximilien: Precisely

pkg/kn/commands/plugin/plugin_test.go
maximilien: Well right now it’s interaction since there is not management features but I won’t dwell on what term to use... Let’s see if others want to chime in ... otherwise, I can do managing.

maximilien:
Thanks for the various feedback @duglin. I will wait 1/2 day to see if others want to chime in (especially on the descriptions changes). If nothing I’ll update and push tomorrow. Cheers 🍻

cmd/kn/main.go
duglin: the current code writes it out no matter what, error or success, due to the defer.

pkg/kn/commands/plugin/path_verifier.go
duglin: the problem with this is that using -q will, IMO, being used all the time and thus hiding everything just not this "valid" state.  I'd like to understand why we're bothering to even do this check at all. There are several conditions that a file had to meet before it'll be consider... why tell people the file doesn't meet them?  Imagine how annoying it would be if each time I can `cat` I got a warning that another `cat` was in my PATH but w/o the +x bit.

pkg/kn/commands/plugin/path_verifier.go
duglin: I can read the code - you're merging warning and errors into the same array.

pkg/kn/commands/plugin/plugin_list.go
```diff
@@ -0,0 +1,190 @@
...
+	isFirstFile := true
+	pluginErrors := []error{}
+	pluginWarnings := 0
+
+	for _, dir := range uniquePathsList(o.PluginPaths) {
+		if dir == "" {
+			continue
+		}
+		files, err := ioutil.ReadDir(dir)
+		if err != nil {
```
duglin: I would silently ignore any errors. Right now when I run it w/o any plugins I see this:
```
$ ./kn plugin list
Unable read directory "/root/src/k8s.io/kubernetes/_output/dockerized/bin/linux/amd64" from your pluginDir: open /root/src/k8s.io/kubernetes/_output/dockerized/bin/linux/amd64: no such file or directory. Skipping...
warning: unable to find any kn plugins in your pluginDir
```
I would expect no output instead. The error message does not help me at all and `kn` has not reason to care about dirs that I don't have access to but are on my PATH.
No plugins isn't an error or a warning-able thing.

pkg/kn/commands/plugin/path_verifier.go
duglin:
```
$ kn plugin list
The following compatible plugins are available:

kn-hi
/root/bin/kn-hi
  - warning: /root/bin/kn-hi is overshadowed by a similarly named plugin: kn-hi
```
the warning isn't helpful I can see two of the same name. It's just an annoyance at this point. Warning are for things that I should fix - there's no reason to "fix" standard PATH-search logic behavior.

pkg/kn/commands/plugin/path_verifier.go
duglin:
```
$ kn plugin list
The following compatible plugins are available:

kn-hi
  - warning: kn-hi identified as a kn plugin, but it is not executable
/root/bin/kn-hi
  - warning: /root/bin/kn-hi is overshadowed by a similarly named plugin: kn-hi
```
The first warning isn't right. It's not identified as a plugin because I turned off the +x bit - so showing it in the list is the error, not me having it my PATH.

The exit code on this (which shows no errors just warnings) returns "1", not "0", which means these "warnings" are acting as "errors" - warning do not normally cause a non-zero exit code.

duglin: Also, 
```
$ kn plugin list
The following compatible plugins are available:

kn-hi
  - warning: kn-hi identified as a kn plugin, but it is not executable
/root/bin/kn-hi
  - warning: /root/bin/kn-hi is overshadowed by a similarly named plugin: kn-hi
Unable read directory "/root/src/k8s.io/kubernetes/_output/dockerized/bin/linux/amd64" from your pluginDir: open /root/src/k8s.io/kubernetes/_output/dockerized/bin/linux/amd64: no such file or directory. Skipping...
error: 2 plugin warnings were found
```
The last line is funny :-)   "error" I found two "warnings".  Aside from that inconsistency in wording, there are actually 3 "warnings" printed, not 2.  IMO none of these are errors or warnings. My system is in a perfectly valid state and so there's no reason to pester me with these messages.

pkg/kn/commands/plugin/plugin_list.go
```diff
@@ -0,0 +1,190 @@
...
+			In:     os.Stdin,
+			Out:    os.Stdout,
+			ErrOut: os.Stderr,
+		},
+	}
+
+	pluginListCommand := &cobra.Command{
+		Use:   "list",
+		Short: "List all visible plugin executables",
+		Long: `List all visible plugin executables.
```
duglin: The indenting here is a bit off, it's too deep:
```
$ ./kn plugin  list --help
List all visible plugin executables.

		Available plugin files are those that are:
		- executable
		- begin with "kn-
		- anywhere on the path specfied in Kn's config pluginDir variable, which:
		  * defaults to $PATH if not specified
		  * can be overridden with the --plugin-dir flag

Usage:
  kn plugin list [flags]
...

The "Available..." block should be left aligned.
```

pkg/kn/commands/plugin/plugin_list.go
```diff
@@ -0,0 +1,190 @@
...
+				for _, err := range errs {
+					fmt.Fprintf(o.ErrOut, "  - %s\n", err)
+					pluginWarnings++
+				}
+			}
+		}
+	}
+
+	if !pluginsFound {
+		pluginErrors = append(pluginErrors, fmt.Errorf("warning: unable to find any kn plugins in your pluginDir"))
```
duglin: This is not an error

duglin:
overall it looks good - I just have some UX concerns

cmd/kn/main.go
maximilien: yes, but that will happen when it exits (or fails) so no intermediary changes.

pkg/kn/commands/plugin/plugin_list.go
maximilien: dude, it tells you that your pluginDir is not readable. I am happy to make the message clearer but I think silent ignore is not the correct approach.

pkg/kn/commands/plugin/plugin_list.go
maximilien: yes, there is no way to return a warning error so the message is "warning"

maximilien:
> overall it looks good - I just have some UX concerns

I suggest we get something in and start polishing the UX after.

pkg/kn/commands/plugin/plugin.go
maximilien: Currently only interacting since no management commands are present. We will however be adding `add` and `remove` and server side stuff. So interacting and managing is sensible.

pkg/kn/commands/plugin/plugin_list.go
duglin: look at my example, I'm going to get that message every single time for no good reason. That's not a good UX.

maximilien:
OK @rhuss this is ready for your review and lgtm. I have addressed all pending comments and added tests for pass coverage.

rhuss:
I've put in on my work list for Monday. Happy Weekend ....

duglin:
since my comment is now hidden, I wanted to add my concerns here so they're not lost:

w/o any plugins I see this:
```
$ ./kn plugin list
Unable read directory "/root/src/k8s.io/kubernetes/_output/dockerized/bin/linux/amd64" from your pluginDir: open /root/src/k8s.io/kubernetes/_output/dockerized/bin/linux/amd64: no such file or directory. Skipping...
warning: unable to find any kn plugins in your pluginDir

$ echo $?
1
```
- It should not complain about items in my PATH that it can not find or read - that is not an error or warning (or something I should fix)
- it should not complain about not finding any plugins - that is not an error or warning (or something I should fix)
- the exit code should not be 1 because nothing is wrong with my setup and an exit code of 1 will result in down-stream issues, such as a script running this that has `set -e` and stops on any non-zero exit codes. Warning are not errors, hence two different words. So, even if we do decide to keep some of these warning, the exit code for warning should still be zero.


when I create a text file called `kn-hi`, I get this
```
$ ./kn plugin list
The following compatible plugins are available:

kn-hi
  - warning: kn-hi identified as a kn plugin, but it is not executable
$ echo $?
1
```
- This is not an error or warning (or something I should fix). I purposely did not meet the contract of "what is a kn plugin" (not executable) so it should not complain about a text file


Have two versions of `kn-hi` in my path:
```
$ ./kn plugin list
The following compatible plugins are available:

kn-hi
/root/bin/kn-hi
  - 
warning: /root/bin/kn-hi is overshadowed by a similarly named plugin: kn-hi
```
- this is not an error, this is normal shell/PATH processing and allowed. I should not need to fix anything.

maximilien:
> since my comment is now hidden, I wanted to add my concerns here so they're not lost:

Good feedback and super easy to change. However, I think the way it's working now is reasonable since it gives information to user.

Returning an error when listed plugins would not be able to be executed or when execution would be potentially confusion (in the case of the overshadowed plugin) *is a good thing*. If I don't return an error to the user, or silently ignore, then the user would try to run plugin and not get the correct behavior. I believe that would be a worst UX than what is implemented now.

--------
I've opined on this enough, I'll let others chime in. IMO unless we have others here agreeing with your proposed changes, I suggest we get this current implementation in front of beta testers and plugin writers and we or they can create issues to address if the current behavior is incorrect. Especially since your requested changes are trivial to implement.

Finally, it's worth noting that the current behavior mimics the kubectl plugins UX. I think this is a good thing since you'd expect that a large percentage of kn plugin users / creators will have some experience with kubectl plugins. Keeping the UX the same (or close to the same) brings familiarity and cohesion to the two projects.

pkg/kn/commands/plugin/plugin_handler.go
```diff
@@ -0,0 +1,109 @@
...
+// HandlePluginCommand receives a pluginHandler and command-line arguments and attempts to find
+// a plugin executable on the PATH that satisfies the given arguments.
+func HandlePluginCommand(pluginHandler PluginHandler, cmdArgs []string) error {
+	remainingArgs := []string{} // all "non-flag" arguments
+
+	for idx := range cmdArgs {
+		if strings.HasPrefix(cmdArgs[idx], "-") {
+			break
+		}
+		remainingArgs = append(remainingArgs, strings.Replace(cmdArgs[idx], "-", "_", -1))
```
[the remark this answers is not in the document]
maximilien: Not sure I understand your question...

pkg/kn/commands/plugin/plugin_handler.go
```diff
@@ -0,0 +1,109 @@
...
+func NewDefaultPluginHandler(validPrefixes []string) *DefaultPluginHandler {
+	return &DefaultPluginHandler{
+		ValidPrefixes: validPrefixes,
+	}
+}
+
+// Lookup implements PluginHandler
+func (h *DefaultPluginHandler) Lookup(filename string) (string, bool) {
+	for _, prefix := range h.ValidPrefixes {
+		path, err := exec.LookPath(fmt.Sprintf("%s-%s", prefix, filename))
```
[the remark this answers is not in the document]
maximilien: Not sure I understand your question...

pkg/kn/commands/plugin/plugin_list.go
```diff
@@ -0,0 +1,190 @@
...
+// Private
+
+func (o *PluginFlags) complete(cmd *cobra.Command) error {
+	o.Verifier = &CommandOverrideVerifier{
+		Root:        cmd.Root(),
+		SeenPlugins: make(map[string]string, 0),
+	}
+
+	pluginPath := commands.PluginDir
+	if pluginPath == "$PATH" {
```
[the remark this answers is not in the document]
maximilien: This is the value in the config. That could be $PATH. Why not?

pkg/kn/commands/plugin/plugin_list.go
```diff
@@ -0,0 +1,190 @@
...
+
+// Private
+
+func (o *PluginFlags) complete(cmd *cobra.Command) error {
+	o.Verifier = &CommandOverrideVerifier{
+		Root:        cmd.Root(),
+		SeenPlugins: make(map[string]string, 0),
+	}
+
+	pluginPath := commands.PluginDir
```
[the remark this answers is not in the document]
maximilien: Ok fine.

pkg/kn/commands/plugin/plugin_list.go
```diff
@@ -0,0 +1,190 @@
...
+	pluginWarnings := 0
+
+	for _, dir := range uniquePathsList(o.PluginPaths) {
+		if dir == "" {
+			continue
+		}
+		files, err := ioutil.ReadDir(dir)
+		if err != nil {
+			if _, ok := err.(*os.PathError); ok {
+				fmt.Fprintf(o.ErrOut, "Unable read directory %q from your pluginDir: %v. Skipping...", dir, err)
```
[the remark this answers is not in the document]
maximilien: You could also argue that you’ve specified a directory for plugin that is not readable and important to tell you.

pkg/kn/commands/plugin/plugin_list.go
```diff
@@ -0,0 +1,190 @@
...
+			pluginPath := f.Name()
+			if !o.NameOnly {
+				pluginPath = filepath.Join(dir, pluginPath)
+			}
+
+			fmt.Fprintf(o.Out, "%s\n", pluginPath)
+			if errs := o.Verifier.Verify(filepath.Join(dir, f.Name())); len(errs) != 0 {
+				for _, err := range errs {
+					fmt.Fprintf(o.ErrOut, "  - %s\n", err)
+					pluginWarnings++
```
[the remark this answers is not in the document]
maximilien: Well, listing plugins requires you to give some info on each when there is a problem. Some are warnings and some are errors. I could omit warnings but that’s missing info to users. I think giving the user as much info as possible is important. With quiet #258 user will be able to chose to make things quiet if she needs.

pkg/kn/commands/plugin/plugin_handler.go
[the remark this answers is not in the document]
maximilien: Not sure I understand your point. Elaborate. Or better try it and see that it actually does what it’s suppose to./

pkg/kn/commands/plugin/path_verifier.go
```diff
@@ -0,0 +1,102 @@
...
+	binName := segs[len(segs)-1]
+
+	cmdPath := strings.Split(binName, "-")
+	if len(cmdPath) > 1 {
+		// the first argument is always "kn" for a plugin binary
+		cmdPath = cmdPath[1:]
+	}
+
+	errors := []error{}
+	isExec, err := isExecutable(path)
```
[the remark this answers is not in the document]
maximilien: That’s fair, though, the point of listing the plugins is to also give some “status” for each since if listed user should be able to run it. The question then becomes do we do more or less checks for user. I am always in favor of helping the user since a non-executable plugin for the current user will not work. So listing it w/o warning or error is misleading at best.

pkg/kn/commands/plugin/path_verifier.go
```diff
@@ -0,0 +1,102 @@
...
+func (v *CommandOverrideVerifier) Verify(path string) []error {
+	if v.Root == nil {
+		return []error{fmt.Errorf("unable to verify path with nil root")}
+	}
+
+	// extract the plugin binary name
+	segs := strings.Split(path, "/")
+	binName := segs[len(segs)-1]
+
+	cmdPath := strings.Split(binName, "-")
```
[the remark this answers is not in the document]
maximilien: Yup, you can and it becomes: `kn list types` where list is the command group.

Try it! :)

pkg/kn/commands/plugin/path_verifier.go
```diff
@@ -0,0 +1,102 @@
...
+	if len(cmdPath) > 1 {
+		// the first argument is always "kn" for a plugin binary
```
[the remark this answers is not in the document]
maximilien: “Verify this” what is this? Not clear.

maximilien:
> * `plugin-dir` is not used when executing a plugin (only for list)

I’ll double check that. If anything add a specific test.

> plugin-dir is not something you want to provide for every command when you have plugins installed but should be an option in the configuration file

You don’t have to. It’s implemented now as a persistent flag but should lookup from the config if that’s present and specified. I’ll see if I can be clearer about the search order.

> I highly would prefer that ~/.kn/plugins is the default 

Currently the default is anywhere in your $PATH to be compatible with `kubectl` But we can agree on a specific default. @sixolet want to chime in to this as well?

> Looking up the whole path is an error-prone process (e.g. think about network mounted directories, bogus pathes added, the sheer amount of files to check (>2k for my machine). 

Actually it’s fast since it’s mostly provided by Golang. See [`exec.LookPath(...)`](https://golang.org/pkg/os/exec/#LookPath)

> should add the plugin list just to kn help

That’s what I plan for in #266

pkg/kn/commands/plugin/path_verifier.go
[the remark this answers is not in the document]
sixolet: I *want* to be able to make `kn list-types`

pkg/kn/commands/plugin/plugin_handler.go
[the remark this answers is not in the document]
sixolet: https://golang.org/pkg/os/exec/#LookPath

"LookPath searches for an executable named file in the directories named by the PATH environment variable. If file contains a slash, it is tried directly and the PATH is not consulted. The result may be an absolute path or a path relative to the current directory."

@rhuss  is right. This code only looks in your PATH.

pkg/kn/commands/plugin/plugin_handler.go
```diff
@@ -0,0 +1,109 @@
...
+
+// PluginHandler is capable of parsing command line arguments
+// and performing executable filename lookups to search
+// for valid plugin files, and execute found plugins.
+type PluginHandler interface {
+	// exists at the given filename, or a boolean false.
+	// Lookup will iterate over a list of given prefixes
+	// in order to recognize valid plugin filenames.
+	// The first filepath to match a prefix is returned.
+	Lookup(filename string) (string, bool)
```
sixolet: This doesn't seem like it takes a filename, it takes a concatenated-with-dash-subcommand-aggregation. That's a messy interface, I think we can do better.

pkg/kn/core/root.go
```diff
@@ -68,11 +96,20 @@ Eventing: Manage event subscriptions and channels. Connect up event sources.`,
 	if p.Output != nil {
 		rootCmd.SetOutput(p.Output)
 	}
-	rootCmd.PersistentFlags().StringVar(&commands.CfgFile, "config", "", "config file (default is $HOME/.kn/config.yaml)")
+
+	// Persistent flags
+	rootCmd.PersistentFlags().StringVar(&commands.CfgFile, "config", "", "kn config file (default is $HOME/.kn/config.yaml)")
+	rootCmd.PersistentFlags().StringVar(&commands.PluginDir, "plugin-dir", "$PATH", "kn plugin directory also stored in kn config")
```
sixolet: "also stored in kn config" is not helpful right now. Let's get better help text.

sixolet: I think I don't like this as a persistent flag, I don't like it showing up in help for everything ever, and I think it's something you set per environment rather than every time you invoke the command. At the very lieast it should be a hidden flag. viper.BindEnvVar with something like KN_PLUGIN_DIR sounds a lot better.

Also viper.BindEnvVar KN_PLUGIN_PREFIX.

sixolet:
> > I highly would prefer that ~/.kn/plugins is the default
> Currently the default is $PATH (as a value) which translates to anywhere in your $PATH. Done to be > compatible with kubectl Plugins. That’s my vote.

I also prefer `~/.kn/plugins`. And no prefix. Makes later code to install plugins *much* easier.

pkg/kn/commands/plugin/path_verifier.go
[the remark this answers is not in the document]
rhuss: Ok, then please make the `isExecutable()` complete then and verify that the current user can actually really execute it and not only checking if _somebody_ can execute it.

I think the check that the go library does for is good enough (and complete).

pkg/kn/commands/plugin/plugin_handler.go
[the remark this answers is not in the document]
rhuss: @maximilien it's also clear that only the OS path could be consulted as you don't use `pluginDir` anywhere in this method here for executing the binary.

pkg/kn/commands/plugin/plugin_list.go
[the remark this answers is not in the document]
rhuss: What happens when you call ?

```
kn --pluginDir=$PATH my-plugin
```

`kn` won't see `$PATH` literally but the *content* of $PATH as its interpolated by the shell. 

To get this behaviour you have to use 

```
kn --pluginDir=\$PATH my-plugin
```

to get a literal $PATH. Isn't that confusing ? (beside that a path is not a directory). Also, why would one ever set pluginDir to the path if path is the default anyway ?

pkg/kn/commands/plugin/plugin_list.go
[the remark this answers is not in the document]
rhuss: I haven't specified anything. 

It's just that my PATH variable contains invalid directories. No shell complains about this when looking for executables, and I realized that just because kn told me.

So I have to fix path and my shell startup script for `kn` only to get rid of the warning. Sometimes this is not even possible for standard startup script which I can't write, so I have to live with a broken path with I can't fix. 

How would I get rid of this warning in such a situation ?

pkg/kn/commands/plugin/plugin_handler.go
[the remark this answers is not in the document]
rhuss: I refer to this validation in https://github.com/maximilien/client/blob/1819e52821d36202a2e070bdcf531f1fbfd74c93/pkg/kn/commands/plugin/path_verifier.go#L71-L74

Where do you this kind of conversion from `_` to `-` there before handing it over to cobra for verifying ?

pkg/kn/commands/plugin/plugin_handler.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// HandlePluginCommand receives a pluginHandler and command-line arguments and attempts to find
+// a plugin executable on the PATH that satisfies the given arguments.
+func HandlePluginCommand(pluginHandler PluginHandler, cmdArgs []string) error {
+	remainingArgs := []string{} // all "non-flag" arguments
+
+	for idx := range cmdArgs {
+		if strings.HasPrefix(cmdArgs[idx], "-") {
+			break
```
rhuss:
```suggestion
			continue
```

pkg/kn/commands/plugin/plugin_list.go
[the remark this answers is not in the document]
rhuss: I'm fine with that for this PR (its not blocking for me), we can refine it later if needed.

pkg/kn/commands/plugin/path_verifier.go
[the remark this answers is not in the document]
rhuss: Afaiu the `kubectl` logic (and also the implementation here when executing the plugin), it's that for a plugin `kn-service-list_types` you can call `kn service list-types` (so you can't have commands with underscores in it though). This is not reflected in the in this validation code here though when checking for shadowing existing commands).

I think, when using `kn plugin list` this file -> command translation should be shown there as well, otherwise when only the plain file name is shown, the user would need to know about this convention to find out how to call the plugin.

rhuss:
> > > I highly would prefer that ~/.kn/plugins is the default
> > > Currently the default is $PATH (as a value) which translates to anywhere in your $PATH. Done to be > compatible with kubectl Plugins. That’s my vote.
> 
> I also prefer `~/.kn/plugins`. And no prefix. Makes later code to install plugins _much_ easier.

Maybe let me rephrase my suggested compromise for covering those two use cases *out of the box without any extra configuration* :

* **Use Case A**: I want to store my plugins anywhere in path with a `kn-` prefix and I don't care about any plugin directory (like `kubectl` does)
* **Use Case B**: I want to store my plugins in a dedicated directory, so that I know where I can put it in and where kn could automatically put it when downloaded from a remote repository** (like [krew](https://github.com/kubernetes-sigs/krew))

The compromise algorithm for looking up a plugin looks like this:

1. Check for the plugin in `~/.kn/plugins` first. If this directory exists and plugin is in there, use that.
2. If not found in step 1, check the whole $PATH for the plugin name, prefixed with `kn-`. Pick it if found.
3. Return an error if not found.

As configuration variable use a `lookupPluginsInPath`. If set to `false` then skip step 2. The default value is `true`. Should be a config variable only (no command line flag)

Optionally, introduce a configuration option `pluginDirectory`, which would override the default `~/.kn/plugins`. But we could as well keep it hardcoded to `~/.kn /plugins` as well.

-------------------

This would support both use cases out of the box **without any configuration**. 

* **A**: Users who don't want a plugin directory, just don't create it or leave it empty.
* **B**: Users put all the plugin in to plugin directory. This works ootb, but they might consider to add add `lookupPluginsInPath` to false to be sure to not pick up a plugin accidentally.

--------------------

Would this something we could agree on ?

rhuss:
> I’ve not benchmarked it so I am going by hunch and I’d be curious if you tried it on a VM or machine that is as bad as you describe and time it. Send results. I’m guessing it’s fast :)

to be clear, performance is not my main issue at all (though I wonder why shells still use caches for this). Its about the open-world assumption, and this world outside can be cruel with some surprises which can't anticipate right now (as network file systems). But yeah, I might think to much here, so I would be happy with a solution which would support both sides (closed-world and open-world). See my suggested compromise solution above.

pkg/kn/commands/plugin/plugin_list.go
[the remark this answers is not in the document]
duglin: I raised this point too.  Giving a warning about something that I might not have any control over will just annoy people. Besides, it's not an error (or even warning) condition to be in this state.

duglin:
re: prefixes... if we ONLY support a well-defined plug-in dir then we can go w/o the `kn-` prefix. However, if we support looking for plugins on the PATH then I would suggest we keep the `kn-` prefix regardless of where the plugin is found (PATH or well-known dir). I say this because I don't think people should be forced to do a rename if they move a plugin between locations.

pkg/kn/commands/plugin/path_verifier.go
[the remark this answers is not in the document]
duglin: This goes to something I don't think I've said very clearly. There's a contract for what a plug-in is: 1) starts with `kn-`, 2) can be found via PATH or well-known dir, 3) is marked as executable.

All 3 conditions much be met. However, how you view these requires is very important. For example, I look at them as "the user must meet all 3 otherwise kn will ignore the file". I believe @maximilien is looking it as "any file that meets 1) and 2) must also meet 3) otherwise it's an error".  IOW, 1) and 2) are triggers for kn to believe that the file is meant to it and therefore the user made a mistake by not doing 3).  While to me, if the user chooses not to do 3) it's because they wanted to. Meaning, they purposely chose not to full-fill the contract because they don't want it to be considered.

Is the contract 1+2 with 3 being a check, or is the contract 1+2+3 ?   IMO, it's the latter.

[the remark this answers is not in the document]
rhuss: As mentioned I wouldn't make it that complicated and just let the error pop up when the plugin is executed. So just don't check for it. Its a rare case that a file in a directory within the $PATH is not executable. (except maybe when the user directly downloaded into the directory via curl)

But this here is not mission-critical, happy to accept it as is (when we got the other issues sorted out). We can refine later.

pkg/kn/commands/plugin/path_verifier.go
```diff
@@ -0,0 +1,102 @@
...
+		errors = append(errors, fmt.Errorf("error: unable to identify %s as an executable file: %v", path, err))
+	}
+
+	if existingPath, ok := v.SeenPlugins[binName]; ok {
+		errors = append(errors, fmt.Errorf("warning: %s is overshadowed by a similarly named plugin: %s", path, existingPath))
+	} else {
+		v.SeenPlugins[binName] = path
+	}
+
+	cmd, _, err := v.Root.Find(cmdPath)
```
rhuss: I've mentioned it elsewhere already why this check here is not sufficient. Let me try to explain it with an example:

* Assume we already have a command `kn importer list-types`, i.e. a command "importer" and a sub-type "list-types", built-in into kn.
* Now kn finds a plugin called `kn-importer-list_types`. In this case the array `cmdPath` will contain `("importer", "list_types")` (note, that `_` is not converted to a `-`). So `v.Root.Find` will fail and hence verify succeed.

Or did I miss something ?

pkg/kn/commands/plugin/plugin_handler.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// Execute implements PluginHandler
+func (h *DefaultPluginHandler) Execute(executablePath string, cmdArgs, environment []string) error {
+	return syscall.Exec(executablePath, cmdArgs, environment)
+}
+
+// HandlePluginCommand receives a pluginHandler and command-line arguments and attempts to find
+// a plugin executable on the PATH that satisfies the given arguments.
+func HandlePluginCommand(pluginHandler PluginHandler, cmdArgs []string) error {
```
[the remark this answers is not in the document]
rhuss: Not important now, feel free to resolve.

pkg/kn/commands/plugin/plugin_handler.go
```diff
@@ -0,0 +1,109 @@
...
+			continue
+		}
+		return path, true
+	}
+
+	return "", false
+}
+
+// Execute implements PluginHandler
+func (h *DefaultPluginHandler) Execute(executablePath string, cmdArgs, environment []string) error {
```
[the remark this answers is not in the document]
rhuss: Not important now, feel free to resolve.

pkg/kn/commands/plugin/plugin_handler.go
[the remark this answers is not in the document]
maximilien: OK, must have not pushed the latest code from my branch. But let me do this right and include tests for this specifically by next update. Thanks for catching this.

pkg/kn/commands/plugin/plugin_list.go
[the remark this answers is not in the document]
maximilien: OK thanks for clarifying. I can see your point now.

pkg/kn/commands/plugin/plugin_list.go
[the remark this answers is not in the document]
maximilien: Change your $PATH value.

pkg/kn/commands/plugin/plugin_handler.go
maximilien: That's what `kubectl` plugins do. I think we should be careful in not breaking existing agreed upon interfaces. We'd just be re-inventing something slightly different IMO.

pkg/kn/core/root.go
maximilien: I'll have a `knPluginPrefix` in config with default value `kn`

pkg/kn/commands/plugin/plugin_handler.go
maximilien: OK, why don't I have a failing test ... Let me look into this.

rhuss:
> However, if we support looking for plugins on the PATH then I would suggest we keep the `kn-` prefix regardless of where the plugin is found (PATH or well-known dir). I say this because I don't think people should be forced to do a rename if they move a plugin between locations.

I agree with that totally, that it should be possible to put a plugin with `kn-` prefix into the plugin dir, but maybe make the prefix just optional (for those using *only* the plugin-dir). But well, that's a detail which we can think about later. For now I just would go with a single, fixed, non-optional prefix: `kn-`

pkg/kn/commands/plugin/path_verifier.go
maximilien:
```bash
➜  client git:(kn-plugins4) cat /usr/local/bin/kn-foo-bar
#!/bin/bash

echo "My first command-line argument was '$1'"
➜  client git:(kn-plugins4) cat /usr/local/bin/kn-foo_bar2
#!/bin/bash

echo "My plugin got command-line argument was '$1'"
➜  client git:(kn-plugins4) ./kn plugin list
The following compatible plugins are available:

/usr/local/bin/kn-foo-bar
/usr/local/bin/kn-foo_bar2
➜  client git:(kn-plugins4) ./kn foo bar lol
My first command-line argument was 'lol'
➜  client git:(kn-plugins4) ./kn foo_bar2 lol
My plugin got command-line argument was 'lol'
```

rhuss: Apologies, I mis-interpreted that line (confused 'old' with 'new'):

https://github.com/maximilien/client/blob/1819e52821d36202a2e070bdcf531f1fbfd74c93/pkg/kn/commands/plugin/plugin_handler.go#L80

What is it supposed to do then when not escaping _ with a - for allowing dashed command? (I really thought that the replacement was vice versa which in my naive thinking makes more sense, therefore that brain slip probably)

Or let's pose the question around: How do you allow dashed commands?

pkg/kn/commands/plugin/plugin_handler.go
sixolet: The arg name at least shouldn't be misleading.

sixolet:
> This would support both use cases out of the box **without any configuration**.
> 
> * **A Users** who don't want a plugin directory, just don't create it or leave it empty. They put all their plugins somewhere into the path
> * **B Users** put all the plugins into plugin directory. This works ootb, but they might consider to add `lookupPluginsInPath` to false to be sure to not pick up a plugin accidentally.
> 
> Would this something we could agree on ?

I'd be ok with this. I think it's better than having a prefix for the plugin directory, or not having one.

pkg/kn/commands/plugin/path_verifier.go
maximilien: Yup, that's a known problem. Not allowed now. 

But if we must have I guess we'd need some convention, naively: `/usr/local/bin/kn-foo--bar` should be interpreted as plugin named: `kn foo-bar`

Thoughts?

Let me open a separate issue on this since we can discuss there and implement once resolved.

rhuss: @maximilien if you look to [kubectl documentation](https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/#names-with-dashes-and-underscores) they allow `-` in the command as well and treat it on the same footprint as `_`:

> Although the kubectl plugin mechanism uses the dash (-) in plugin filenames to separate the sequence of sub-commands processed by the plugin, it is still possible to create a plugin command containing dashes in its commandline invocation by using underscores (_) in its filename.
> ....
> Note that the introduction of underscores to a plugin filename does not prevent us from having commands such as kubectl foo_bar. The command from the above example, can be invoked using either a dash (-) or an underscore (_)

I think we can do the same and shouldn't be that hard. @maximilien in order to copy the similarity to kubectl that you already emphasized, we should provide the same semantics and allow dashes (i.e. treat a `-` and `_` in the file name both as `-` and `_`). We should not introduce a new convention. 'hope that aligns with your thinking to stay close to kubectl in that respect (plugin feature).

BTW, this diff in behaviour was really what confused me. Apologies for that.

maximilien: Resolving since issue #272 captures and should be used for future discussion on this point.

pkg/kn/commands/plugin/plugin_handler.go
rhuss: Maybe because you put the options always last ?

pkg/kn/core/root.go
rhuss: I don't think this should be configurable.

pkg/kn/commands/plugin/path_verifier.go
maximilien: All for not introducing new convention and tailing on `kubectl` plugins. Hence why initial PR is specifically trying to be like `kubectl` plugins. Heck, in the name of this PR :)

pkg/kn/commands/plugin/plugin_list.go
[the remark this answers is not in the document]
rhuss: Why would should `kn` force it if no shell does when evaluating the path ? Are we really that dogmatic ?

pkg/kn/commands/plugin/path_verifier.go
rhuss: Ok, if we then agree that we are currently not like kubectl PRs. 

Btw, I think it would be honest if the spots where code has been copied over would be marked with a comment or so. Not only for credits, but also to have a backlink in cases things are changes upstream.

maximilien:
OK @rhuss and @sixolet. Pushed latest version. It's close but not quite ready as I can see three remaining issues:

1. using `lookupPluginsInPath` and `pluginsDir` options will list correct plugins (for the 4 combinations). However, executing the plugins is trickier. Currently only plugins in `pluginsDir` is executing. I think I need to remove `--lookup-plugins-in-path` flag if present to allow current code to work. But I need to test this before committing the change.

2. Viper loading and saving of config seems broken. Likely how I am using Viper. Essentially, when I change the config and run `./kn ...` they are not loaded, so the default is used.

3. I have some integration tests that I need to also include. However, I need #274 to be merged first and simply had not had a chance to rebase 274 so it could be merged. Will do so tonight or tomorrow AM.

---
That said, one approach if the current code looks good to you is to merge so I can address these issues separately. Otherwise, I'll keep churning tomorrow and get this PR with fixes for above.

Let me know if you also see any additional issues. Thanks.

pkg/kn/commands/plugin/plugin_handler.go
```diff
@@ -0,0 +1,109 @@
...
+	"os"
+	"os/exec"
+	"strings"
+	"syscall"
+)
+
+// PluginHandler is capable of parsing command line arguments
+// and performing executable filename lookups to search
+// for valid plugin files, and execute found plugins.
+type PluginHandler interface {
```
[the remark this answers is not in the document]
maximilien: OK let me update that in next push

pkg/kn/commands/plugin/plugin_handler.go
```diff
@@ -0,0 +1,109 @@
...
+	Lookup(filename string) (string, bool)
+	// Execute receives an executable's filepath, a slice
+	// of arguments, and a slice of environment variables
+	// to relay to the executable.
+	Execute(executablePath string, cmdArgs, environment []string) error
+}
+
+// DefaultPluginHandler implements PluginHandler
+type DefaultPluginHandler struct {
+	ValidPrefixes []string
```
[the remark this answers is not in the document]
maximilien: Keeping since that's how we'd be able to specify different prefixes

rhuss:
I will have a look tomorrow morning our time (but from my understanding I'm really not sure that (a) we need both options and (b) whether it should be not only configuration variables.

maximilien:
OK @rhuss and @sixolet this is ready to merge. Assuming I pass CI.

Please note:

1. I have not added my integration tests since not done, however, manual tests worked. I have lots of plugins on my system in different directories and have a small script (converting to IT) to check.
2. While your requested changes appear easy, they are not :) but it's done. Pay attention to what I had to do in `root.go` for instance. Needed since the `cobra.Command` is actually not executed for a plugin this the flags are not loaded. Was fun to discover!
3. I am sure we can improve things. But I suggest we get this in (assuming no major comments) and we can refine after. 

Ping me on Slack if you need my urgent attention on something. Would be good to have this in soon. Thx.

pkg/kn/commands/plugin/plugin_list.go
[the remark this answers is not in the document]
maximilien: I don't know if I changed this. But I will in subsequent releases. Just did not get to it.

pkg/kn/commands/plugin/plugin_list.go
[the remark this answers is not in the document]
maximilien: OK see what I have now is good enough.

pkg/kn/commands/plugin/plugin_list.go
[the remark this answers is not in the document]
maximilien: Current should work. Note, however, that it's `--plugin-dir` not `--pluginDir`

pkg/kn/commands/plugin/plugin_handler.go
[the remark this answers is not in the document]
maximilien: Done.

pkg/kn/commands/plugin/path_verifier.go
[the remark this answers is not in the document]
maximilien: See now.

pkg/kn/core/root.go
maximilien: Let's open an issue for the prefix. I can add it after merge.

pkg/kn/commands/plugin/plugin_handler.go
maximilien: @rhuss and @sixolet here's how it's working:

```bash
➜  client git:(kn-plugins4) cat /usr/local/bin/kn-foo_bar
#!/bin/bash

echo "Hello Knative, I'm a Kn plugin"
➜  client git:(kn-plugins4) cat /usr/local/bin/kn-foo-bar
#!/bin/bash

echo "Hello Knative, I'm a Kn plugin"
➜  client git:(kn-plugins4) ./kn --lookup-plugins-in-path=true plugin list
The following compatible plugins are available, using options:
  - plugins dir: '~/.kn/plugins'
  - lookup plugins in path: 'true'

/Users/maximilien/.kn/plugins/kn-hello
/usr/local/bin/kn-foo-bar
/usr/local/bin/kn-foo_bar
/usr/local/bin/kn-hello
  - warning: /usr/local/bin/kn-hello is overshadowed by a similarly named plugin: /Users/maximilien/.kn/plugins/kn-hello
/usr/local/bin/kn-kn
/usr/local/bin/kn-knctl

error: one plugin warning was found

➜  client git:(kn-plugins4) ./kn --lookup-plugins-in-path=true foo-bar
Hello Knative, I'm a Kn plugin
➜  client git:(kn-plugins4) ./kn --lookup-plugins-in-path=true foo bar
Hello Knative, I'm a Kn plugin
```

pkg/kn/commands/plugin/path_verifier.go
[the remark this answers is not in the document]
maximilien:
```bash
➜  client git:(kn-plugins4) cat /usr/local/bin/kn-foo_bar
#!/bin/bash

echo "Hello Knative, I'm a Kn plugin"
➜  client git:(kn-plugins4) cat /usr/local/bin/kn-foo-bar
#!/bin/bash

echo "Hello Knative, I'm a Kn plugin"
➜  client git:(kn-plugins4) ./kn --lookup-plugins-in-path=true plugin list
The following compatible plugins are available, using options:
  - plugins dir: '~/.kn/plugins'
  - lookup plugins in path: 'true'

/Users/maximilien/.kn/plugins/kn-hello
/usr/local/bin/kn-foo-bar
/usr/local/bin/kn-foo_bar
/usr/local/bin/kn-hello
  - warning: /usr/local/bin/kn-hello is overshadowed by a similarly named plugin: /Users/maximilien/.kn/plugins/kn-hello
/usr/local/bin/kn-kn
/usr/local/bin/kn-knctl

error: one plugin warning was found

➜  client git:(kn-plugins4) ./kn --lookup-plugins-in-path=true foo-bar
Hello Knative, I'm a Kn plugin
➜  client git:(kn-plugins4) ./kn --lookup-plugins-in-path=true foo bar
Hello Knative, I'm a Kn plugin
```

maximilien:
@sixolet see issue #288 for the modifiable prefix for plugins

pkg/kn/commands/plugin/path_verifier.go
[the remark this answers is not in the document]
maximilien: Let's fix this later as a bug/feature request

pkg/kn/commands/plugin/plugin_flags.go
```diff
@@ -0,0 +1,35 @@
...
+
+	Verifier    PathVerifier
+	PluginPaths []string
+
+	genericclioptions.IOStreams
+}
+
+// AddPluginFlags adds the various flags to plugin command
+func (p *PluginFlags) AddPluginFlags(command *cobra.Command) {
+	command.Flags().BoolVar(&p.NameOnly, "name-only", false, "If true, display only the binary name of each plugin, rather than its full path")
```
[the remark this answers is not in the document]
maximilien: Let's fix this after as a bug or feature request.

pkg/kn/commands/plugin/plugin.go
```diff
@@ -0,0 +1,33 @@
...
+func NewPluginCommand(p *commands.KnParams) *cobra.Command {
+	pluginCmd := &cobra.Command{
+		Use:   "plugin",
+		Short: "Plugin command group",
+		Long: `Provides utilities for interacting and managing with kn plugins.
+
+Plugins provide extended functionality that is not part of the core kn command-line distribution.
+Please refer to the documentation and examples for more information about how write your own plugins.`,
+	}
+	pluginCmd.AddCommand(NewPluginListCommand(p))
```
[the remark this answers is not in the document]
maximilien: Yes, that's how you know where your plugins are and which will be executed when you call it. Plus compatible with `kubectl` plugins.

pkg/kn/commands/plugin/plugin_handler.go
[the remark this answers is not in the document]
maximilien: yes, needed in the `plugin_list_test.go` for mocking. Possible to do it differently I guess.

pkg/kn/commands/plugin/plugin_handler.go
```diff
@@ -0,0 +1,109 @@
...
+	return syscall.Exec(executablePath, cmdArgs, environment)
+}
+
+// HandlePluginCommand receives a pluginHandler and command-line arguments and attempts to find
+// a plugin executable on the PATH that satisfies the given arguments.
+func HandlePluginCommand(pluginHandler PluginHandler, cmdArgs []string) error {
+	remainingArgs := []string{} // all "non-flag" arguments
+
+	for idx := range cmdArgs {
+		if strings.HasPrefix(cmdArgs[idx], "-") {
```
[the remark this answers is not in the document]
maximilien: actually they are passed, except for the plugins flags: `--plugins-dir` and `--lookup-plugins-in-path`. I'll remove comment next time.

maximilien:
:) CI green @rhuss @sixolet

rhuss:
> @sixolet see issue #288 for the modifiable prefix for plugins

I don't think it was about a configurable prefix what @sixolet meant (but to avoid a prefix in ~/.kn/plugins as its redundant). I added my opinion on this in https://github.com/knative/client/issues/288#issuecomment-514498411 and why I think this is not a good idea.

rhuss:
@maximilien I'm going to make my review later today. sorry for the delay, just tons on my plate right now :(

maximilien:
> @maximilien I'm going to make my review later today. sorry for the delay, just tons on my plate right now :(

Fully understood. You are great at reviews so not worried. I’ll submit the kn plugins IT as separate PR that depends on this one.

Thanks!

pkg/kn/commands/plugin/plugin_handler.go
rhuss: AFAIS you don't have a test which uses an option (== argument starting with dash)

pkg/kn/commands/plugin/plugin_handler.go
```diff
@@ -0,0 +1,131 @@
...
+		PluginsDir:          pluginsDir,
+		LookupPluginsInPath: lookupInPath,
+	}
+}
+
+// Lookup implements PluginHandler
+func (h *DefaultPluginHandler) Lookup(name string) (string, bool) {
+	var pluginPath string
+	var err error
+	for _, prefix := range h.ValidPrefixes {
```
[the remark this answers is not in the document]
maximilien: No way for user to change this now.

pkg/kn/commands/plugin/plugin_handler.go
```diff
@@ -0,0 +1,131 @@
...
+		pluginPath = fmt.Sprintf("%s-%s", prefix, name)
+		if h.LookupPluginsInPath {
```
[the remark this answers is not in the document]
maximilien: Fair, I can do that.

pkg/kn/commands/plugin/path_verifier.go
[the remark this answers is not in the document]
maximilien: Of course that works using `string(os.PathSeparator)`. Just missed addressing it. 

Done.

pkg/kn/commands/plugin/plugin_handler.go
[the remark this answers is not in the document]
maximilien: Done

maximilien:
OK @rhuss see latest update. Passes CI and my local ITs---I will submit different issue/PR for that soon today or Monday worst case. Ping me if any other issues. I'll monitor slack. Thx.

pkg/kn/commands/plugin/plugin_handler.go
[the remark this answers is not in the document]
maximilien: Yup. Fixed.

maximilien:
@sixolet I think you need to also add your lgtm

rhuss:
> @sixolet I think you need to also add your lgtm

I think we are good for now, but with some homework left. 'happy to take that over. Thanks a ton to @maximilien for doing the hard work !

PR #245, opened 2019-07-08
Rename go files under command group dir

zhanggbj:
- Files under service dir
- Files under revision dir
- Files under plugin dir

[issue 229](https://github.com/knative/client/issues/229)

Issue #235, opened 2019-07-05
Rename `kn service describe` to `kn service show`


rhuss:
Why ? 

* `show` is shorter and easier to write than `describe`
* We already moved away from kubectl semantics when we switched from `get` to `list` and `list` won't return ever a singe element (but potentially a list with one element)
* We can add a `-o json` etc. to `kn service show` to allow retrieving the definition of a single service (not possible right now) without adding more confusion between `kubectl describe service` (no -o option) and a `kn service describe` (with then maybe an `-o options`).

zhangtbj:
How about change `describe` to `get`? I think it is shorter and more standard.

like `kn route get helloworld`

I see most of cli use `get` to show the detail info, but I think less use `show`, hope it is helpful :)

rhuss:
`get` is really kind of "burned" as it has a special meaning for "kubectl" (where it corresponds to our "list"). For kubectl it doesn't show you the details.

So I think we should either follow the kubectl semantics (with get & describe) or, in order to avoid any confusion complete deviate from it (hence the suggestion with list & show).

zhangtbj:
oh i see, thanks for clarify. 

I think if the result is same as kubectl, we can use the same `get` or `describe`

If the result is different, we should use the different subcommands like `show` or `list` to avoid the confusing.

rhuss:
Yes, it's different:

* `kubectl get` returns either a list (without or argument) or a single object (with name) argument. This is very confusing (i.e. that `get` returns a list and that in different formats), so that we decided to always return a list with `list`. Otherwise you have the same options for machine readable output (e.g. -o yaml).

* `kubectl describe` always show a human readable output, but does not support creating machine readable output `-o yaml`. Since `kn list` removed the option to get a single object, the idea was to allow `-o yaml` to `kn show` (as it is now different than `describe`).

That whole mapping is a mess, so I really believe the more we can differentiate us from kubectl the better. There already other differences like `kn <noun> <verb>` vs. `kubectl <verb> <noun>`, so let's make the difference clear by non-overlapping verbs for where `kn` and `kubectl` differs.

rhuss:
Closing this issue in favour of `kn service describe` as it has been decided. RIP // @duglin (that the issue I referred to)

PR #251, opened 2019-07-09
Fix issue #205 to add route describe feature and test

zhangtbj:
1, Implement kn route describe command with required parameters:
```
✔ $ go run cmd/kn/main.go route describe --help
Describe available route.

Usage:
  kn route describe NAME [flags]

Flags:
  -o, --output string                 Output format. One of: json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-file. (default "yaml")

Global Flags:
```

2, Add `gotest.tools` tests for kn route describe commands:
 - test without route name
 - test route describe with the default output
 - test route describe with the YAML output
 - test route describe with the JSON output
 - test route describe with the name output

Also rebased from master and change to use new knClient.

For issue: #205

### DIFF
--- a/pkg/kn/commands/route/route_describe_test.go
+++ b/pkg/kn/commands/route/route_describe_test.go
@@ -1,0 +1,130 @@
+// Copyright © 2019 The Knative Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package route
+
+import (
+	"encoding/json"
+	"strings"
+	"testing"
+
+	"github.com/knative/client/pkg/kn/commands"
+	"github.com/knative/serving/pkg/apis/serving/v1alpha1"
+	"gotest.tools/assert"
+	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	client_testing "k8s.io/client-go/testing"
+	"sigs.k8s.io/yaml"
+)
+
+func fakeRouteDescribe(args []string, response *v1alpha1.Route) (action client_testing.Action, output string, err error) {
+	knParams := &commands.KnParams{}
+	cmd, fakeRoute, buf := commands.CreateTestKnCommand(NewRouteCommand(knParams), knParams)
+	fakeRoute.AddReactor("*", "*",
+		func(a client_testing.Action) (bool, runtime.Object, error) {
+			action = a
+			return true, response, nil
+		})
+	cmd.SetArgs(args)
+	err = cmd.Execute()
+	if err != nil {
+		return
+	}
+	output = buf.String()
+	return
+}
+
+func TestCompletion(t *testing.T) {
+	var expectedRoute v1alpha1.Route
+
+	setup := func(t *testing.T) {
+		expectedRoute = v1alpha1.Route{
+			TypeMeta: metav1.TypeMeta{
+				Kind:       "Route",
+				APIVersion: "knative.dev/v1alpha1",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "foo",
+				Namespace: "default",
+			},
+		}
+	}
+
+	t.Run("requires the route name", func(t *testing.T) {
+		_, _, err := fakeRouteDescribe([]string{"route", "describe"}, &v1alpha1.Route{})
+		assert.Assert(t, err != nil)
+		assert.Assert(t, strings.Contains(err.Error(), "requires the route name."))
+	})
+
+	t.Run("describe a valid route with default output", func(t *testing.T) {
+		setup(t)
+
+		action, output, err := fakeRouteDescribe([]string{"route", "describe", "foo"}, &expectedRoute)
+		assert.Assert(t, err == nil)
+		assert.Assert(t, action != nil)
+		assert.Assert(t, action.Matches("get", "routes"))
+
+		jsonData, err := yaml.YAMLToJSON([]byte(output))
+		assert.Assert(t, err == nil)
+
+		var returnedRoute v1alpha1.Route
+		err = json.Unmarshal(jsonData, &returnedRoute)
+		assert.Assert(t, err == nil)
+		assert.Assert(t, equality.Semantic.DeepEqual(expectedRoute, returnedRoute))
+	})
+
+	t.Run("describe a valid route with special output", func(t *testing.T) {
+		t.Run("yaml", func(t *testing.T) {
+			setup(t)
+
+			action, output, err := fakeRouteDescribe([]string{"route", "describe", "foo", "-oyaml"}, &expectedRoute)
+			assert.Assert(t, err == nil)
+			assert.Assert(t, action != nil)
+			assert.Assert(t, action.Matches("get", "routes"))
+
+			jsonData, err := yaml.YAMLToJSON([]byte(output))
+			assert.Assert(t, err == nil)
+
+			var returnedRoute v1alpha1.Route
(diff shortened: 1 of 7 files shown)


zhangtbj:
Hi @maximilien , @rhuss and @navidshaikh ,

Sorry I closed the previous PR #228 by mistake, this is the new one which is rebased from master and change to use new knClient.

I think for the discussion for #206 and #235, it should be related to command Printer, should not change in each sub describe command.

For example:
1, without parameter, should show common properties, like resource name, ns, status....
2, with `--all`, show all info like now
3, with `--color`, no change for showing

So I think these should be the same for all subcommands, only change the implementation in Printer code.

I am not sure if we could accept this PR first then change or refine the `describe/show` format for all subcommands.

Thanks all!:)

zhangtbj:
Thanks for the review @rhuss 

About `I'm not sure why the renaming of the files was needed wrt/ this PR. Could you please revert ?`

Do you mean the change like this in PR?:
`pkg/kn/commands/route/list.go → pkg/kn/commands/route/route_list.go`

I took a look the other command file styles, they are all format like this `<noun>_<verb>.go`, like `revision_describe_test.go`, in route folder, there is only list subcommand and named as `list.go` directly.

I think it makes people confusing (what list?) and not sync with other commands.

If this is not a good place to fix it in this PR, I can revert it in here and just for `route describe` function.

rhuss:
> Do you mean the change like this in PR?:
> `pkg/kn/commands/route/list.go → pkg/kn/commands/route/route_list.go`

Yes. Tbh, we already have a PR pending which does it the other way round for the other files #245 . So please revert that change and we can discuss that in another context.

zhangtbj:
Sure thing @rhuss . I have reverted the renaming change. Pls review again. Thanks!

maximilien:
> As an intermediate step, I'm happy to include but we should go to a more human readable output as it is under the way for `kn service describe`.

Agree. Default output should be us humans :)

pkg/kn/commands/route/route_describe_test.go
```diff
@@ -0,0 +1,130 @@
...
+		assert.Assert(t, err == nil)
+
+		var returnedRoute v1alpha1.Route
+		err = json.Unmarshal(jsonData, &returnedRoute)
+		assert.Assert(t, err == nil)
+		assert.Assert(t, equality.Semantic.DeepEqual(expectedRoute, returnedRoute))
+	})
+
+	t.Run("describe a valid route with special output", func(t *testing.T) {
+		t.Run("describe a valid route with YAML output", func(t *testing.T) {
```
maximilien: to simplify and make the test run output easier, I'd avoid duplicating the descriptions when nesting these `t.Run("...")` so for instance, I'd name these as follows:
```go
[...]
t.Run("describe a valid route with special output", func(t *testing.T) {
		t.Run("YAML", func(t *testing.T) {
[...] 
  })
})
[...]
```
See output of test runs to see what I mean. Try to do the same in subsequent nested `t.Run(...)`

pkg/kn/commands/route/route_describe_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+			var returnedRoute v1alpha1.Route
+			err = json.Unmarshal([]byte(output), &returnedRoute)
+			assert.Assert(t, err == nil)
+			assert.Assert(t, equality.Semantic.DeepEqual(expectedRoute, returnedRoute))
+		})
+
+		t.Run("describe a valid route with name output", func(t *testing.T) {
+			setup(t)
+
```
maximilien: Are there a fix list for what `name` can be? If yes then IMO code should guard on these, unless APIs gives errors and you can return error with a nice message to user that the `-oname` they used is invalid...

pkg/serving/v1alpha1/client_test.go
```diff
@@ -285,6 +285,39 @@ func TestListRevisionForService(t *testing.T) {
...
+		})
+
+	t.Run("get known route by name returns route", func(t *testing.T) {
+		route, err := client.GetRoute(routeName)
+		assert.NilError(t, err)
+		assert.Equal(t, routeName, route.Name, "route name should be equal")
+		validateGroupVersionKind(t, route)
+	})
+
+	t.Run("get unknown route name returns error", func(t *testing.T) {
```
maximilien: Is there a need for a check for invalid route names?

pkg/kn/commands/route/route_describe_test.go
zhangtbj: Thanks Max, I have fixed in my PR

pkg/serving/v1alpha1/client_test.go
zhangtbj: Hi Max, I think it is better for us to cover the invalid failure case and I found we also have this invalid cases for service and revision.

pkg/kn/commands/route/route_describe_test.go
zhangtbj: the `-oname` returns the `resource tpye`/`resource name`, so I checked if its result contains the resource name directly.

pkg/serving/v1alpha1/client_test.go
maximilien: No I was thinking about user’s passing a route name like `@#$Q!$#` or anything with invalid HTTP char will be invalid. The API might have a specification as to what route names are valid.

The key for me is not so much that we guard against the valid or invalid name but that we tell user a message that’s easy to understand.

zhangtbj: Hi @maximilien , got it. I have tried, no matter it is a non-exist or invalid name, kn client can report `routes.serving.knative.dev "xxx" not found` correctly without any http error. such as:

```
jordan at jordandembp.cn.ibm.com in ~/Work/workspace/go/src/github.com/knative/client [master|✔]
✘-1 $ go run cmd/kn/main.go route describe he$low@rld
routes.serving.knative.dev "he@rld" not found
exit status 1

jordan at jordandembp.cn.ibm.com in ~/Work/workspace/go/src/github.com/knative/client [master|✔]
✘-1 $ go run cmd/kn/main.go route describe hello
routes.serving.knative.dev "hello" not found
exit status 1
```

So I also change the test route name to `r@ute-that-d$es-n#t-exist` for spceial case.

pkg/kn/commands/route/route_describe_test.go
```diff
@@ -0,0 +1,130 @@
...
+	t.Run("requires the route name", func(t *testing.T) {
+		_, _, err := fakeRouteDescribe([]string{"route", "describe"}, &v1alpha1.Route{})
+		assert.Assert(t, err != nil)
+		assert.Assert(t, strings.Contains(err.Error(), "requires the route name."))
+	})
+
+	t.Run("describe a valid route with default output", func(t *testing.T) {
+		setup(t)
+
+		action, output, err := fakeRouteDescribe([]string{"route", "describe", "test-foo"}, &expectedRoute)
```
rhuss: So you are requesting a route with name "test-foo" but return a route with `Route.Name == "foo"`. IMO the name you are are requesting and the returned name should be the same.

pkg/kn/commands/route/route_describe_test.go
rhuss: What is the benefit of `-oname` when you already have to provide the name as mandatory argument ?

pkg/serving/v1alpha1/client_test.go
rhuss: I think we can rely on the error message from the API client to point in the right direction. I don't think we need an extra validation of the name on our side.

pkg/kn/commands/route/route_describe_test.go
zhangtbj: the `-oname` is another format type to only show the resource name. I just add as anoter format type test. Like this:
```
jordanzt@zhangtaodembp:~/Works/workspaces/gopath/src/github.com/knative/client$ go run cmd/kn/main.go route describe helloworld -oname
route.serving.knative.dev/helloworld
```

pkg/kn/commands/route/route_describe_test.go
zhangtbj: Hi @rhuss , I have refined it. Thanks!

pkg/kn/commands/route/route_describe_test.go
rhuss: Still I believe that -oname is its a no-op here. Why would you call this command if you already know the name + group. Agreed, you could check for whether a revision exists or not. So I'm not totally against it and we might adapt this when route describe changes anyway. So I'm fine with it for now.

PR #252, opened 2019-07-09
feature(service describe): Output of service details

rhuss:
`kn service describe` mimics `kubectl describe <sth>` as that it output detail information in human-readable output. It also adds the possibility to print the service in machine-readable output e.g. via `-o yaml`. This is a difference to `kubectl describe` but necessary as `kn list -o yaml` will always print a list, not a single object. You can get the single object now with `kn describe -o yaml`, which also makes for a clear separation between list operations (multiple entities always) and single entity operation (one entity only). Hopefully, this is less confusing than `kubectl get` operation which has either list or single entity output depending on arguments.

This command shows information about the service itself, but also a summary of the associated revisions.

`kn service describe` has the following features:

* Show basic info like
  - name
  - namespace
  - external and internal URLs
  - age
  - annotations and labels of the service itself (shortened)
* Shows the list of attached revisions which gets traffic
  - name
  - traffic split percentage
  - image name
  - configuration generation when created
  - environment variables (shortened)
  - scale (min ... max)
  - concurrency options
  - resource requests and limits
  - age
* With `--details` more information can be printed:
  - annotation, labels, and envs are not shortened but all are printed
  - image digest is shown for revisions
  - all revisions are shown, not only those getting traffic
* Machine-readable output with `-o yaml|json|...` or `--template` of this single service

This PR is a subset of and replacement for #75. There is no color option included nor any hipster stuff 🤓

### Examle output

`kn service describe hello`

```
Name:         hello
Namespace:    default
URL:          http://hello.default.example.com
Address:      http://hello.default.svc.cluster.local
Annotations:  serving.knative.dev/creator=minikube-user, serving.knative.dev/lastModifier=minikube-user
Age:          1h

Revisions:
  100%  Name:   hello-tw5ds [3] (1h)
        Image:  gcr.io/knative-samples/helloworld-go
        Env:    hello=world, foo=bar
        Scale:  0 ... 10
        Memory: 50 Mb ... 100 Mb

Conditions:
  ++ ConfigurationsReady     1h
  ++ Ready                   1h
  ++ RoutesReady             1h
```

`kn service describe hello --details`

```
Name:         hello
Namespace:    default
URL:          http://hello.default.example.com
Address:      http://hello.default.svc.cluster.local
Annotations:  serving.knative.dev/creator=minikube-user
              serving.knative.dev/lastModifier=minikube-user
Age:          4m

Revisions:
  100%  Name:   hello-tw5ds [3] (16s)
        Image:  gcr.io/knative-samples/helloworld-go (5ea96ba4b872)
        Env:    hello=world
                foo=bar
        Scale:  0 ... 10
     -  Name:   hello-nc8xz [2] (3m)
        Image:  gcr.io/knative-samples/helloworld-go (5ea96ba4b872)
        Env:    hello=world
                foo=bar
     -  Name:   hello-fkl6l [1] (4m)
        Image:  gcr.io/knative-samples/helloworld-go (5ea96ba4b872)

Conditions:
  ++ ConfigurationsReady    13s
  ++ Ready                  13s
  ++ RoutesReady            13s
```


pkg/printers/prefixwriter.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (pw *prefixWriter) WriteLine(a ...interface{}) {
+	fmt.Fprintln(pw.out, a...)
+}
+
+func (pw *prefixWriter) Flush() error {
+	if f, ok := pw.out.(flusher); ok {
+		return f.Flush()
+	}
+	return fmt.Errorf("outpustream %v doesn't support Flush", pw.out)
```
navidshaikh: *output stream

pkg/kn/commands/service/service_show.go
```diff
@@ -0,0 +1,485 @@
...
+	env         []string
+	port        *int32
+}
+
+// [REMOVE COMMENT WHEN MOVING TO 0.7.0]
+// For transition to v1beta1 this command uses the migration approach as described
+// in https://docs.google.com/presentation/d/1mOhnhy8kA4-K9Necct-NeIwysxze_FUule-8u5ZHmwA/edit#slide=id.p
+// With serving 0.6.0 we are at step #1
+// I.e we first look at new fields of the v1alpha1 API before falling back to the original ones.
+// As this command does not any updates, it's just a matter of fallbacks.
```
navidshaikh: probably

*does not perform ?

pkg/kn/commands/service/service_show.go
```diff
@@ -0,0 +1,485 @@
...
+func writeService(dw printers.PrefixWriter, service *v1alpha1.Service) {
+	dw.WriteColsLn(printers.LEVEL_0, l("Name"), service.Name)
+	dw.WriteColsLn(printers.LEVEL_0, l("Namespace"), service.Namespace)
+	dw.WriteColsLn(printers.LEVEL_0, l("URL"), extractURL(service))
+	if service.Status.Address != nil {
+		url := service.Status.Address.GetURL()
+		dw.WriteColsLn(printers.LEVEL_0, l("Address"), url.String())
+	}
+	writeMapDesc(dw, printers.LEVEL_0, service.Labels, l("Labels"), "")
+	writeMapDesc(dw, printers.LEVEL_0, service.Annotations, l("Annotations"), "")
```
navidshaikh: One of thing I observed while testing this PR is, the annotations are configured at two places
`.metadata.annotations` (current)
and
`.spec.metadata.annotations`

the later one is where we set the autoscaling values  (`autoscaling.knative.dev/target: "10"`)
I think we should print these values as well part of `show`

rhuss: Yes, good idea. We should however show it in a form as they are provided during service creation/update, i.e. 'typed'

navidshaikh: or altogether a section in show's output which shows up only if autoscaling is configured for the service? (same for min-scale/max-scale)

test/e2e/service_options_test.go
```diff
@@ -33,16 +33,16 @@ func TestServiceOptions(t *testing.T) {
 	})
 
 	t.Run("returns valid concurrency options for hello service", func(t *testing.T) {
-		testServiceDescribeConcurrencyLimit(t, k, "hello", "250")
-		testServiceDescribeConcurrencyTarget(t, k, "hello", "300")
+		//testServiceDescribeConcurrencyLimit(t, k, "hello", "250")
```
cppforlife: whats going on with these tests being commented out?

go.sum
```diff
@@ -49,6 +49,8 @@ github.com/evanphx/json-patch v4.2.0+incompatible h1:fUDGZCv/7iAN7u0puUVhvKCcsR6
 github.com/evanphx/json-patch v4.2.0+incompatible/go.mod h1:50XU6AFN0ol/bzJsmQLiYLvXMP4fmwYFNcr97nuDLSk=
 github.com/evanphx/json-patch v4.5.0+incompatible h1:ouOWdg56aJriqS0huScTkVXPC5IcNrDCXZ6OoTAWu7M=
 github.com/evanphx/json-patch v4.5.0+incompatible/go.mod h1:50XU6AFN0ol/bzJsmQLiYLvXMP4fmwYFNcr97nuDLSk=
+github.com/fatih/color v1.7.0 h1:DkWD4oS2D8LGGgTQ6IvwJJXSL5Vp2ffcQg58nFV38Ys=
```
cppforlife: how come these deps are being bumped in this pr?

rhuss: Good point will revert

rhuss:
@sixolet @cppforlife @maximilien All features are now implemented, please have a look at the [updated issue description](https://github.com/knative/client/pull/252#issue-295779125) for all data that is shown. The focus was to show everything which also can be provided with `kn service create` as options.

What is missing is the unit test. However, because it really depends highly on the output to be shown I wanted to wait for first feedback if the current format is ok.

I would then also do the unit tests on then KnClient level, i.e. using a mock implementation of the KnClient interface (which needs to be introduced). Or is this something which we should leave for another PR? (this one is already huge). In that case, I would make a simplified test here (to keep the code coverage test happy), and work on another PR for the full coverage including mock support for KnClient.

pkg/kn/commands/service/service_describe.go
```diff
@@ -41,23 +122,515 @@ func NewServiceDescribeCommand(p *commands.KnParams) *cobra.Command {
...
+		return ""
+	}
+}
+
+// Return either image name with tag or together with its resolved digest
+func getImageDesc(desc *revisionDesc) string {
+	image := desc.image
+	if printDetails && desc.imageDigest != "" {
+		digest := "(" + shortenDigest(desc.imageDigest) + ")"
+		return fmt.Sprintf("%s %s", image, digest)
```
toVersus: I prefer to put things into one place here:

```suggestion
		return fmt.Sprintf("%s (%s)", image, shortenDigest(desc.imageDigest))
```

pkg/kn/commands/service/service_describe.go
```diff
@@ -41,23 +122,515 @@ func NewServiceDescribeCommand(p *commands.KnParams) *cobra.Command {
...
+// Write out main service information. Use colors for major items.
+func writeService(dw printers.PrefixWriter, service *v1alpha1.Service) {
+	dw.WriteColsLn(printers.LEVEL_0, l("Name"), service.Name)
+	dw.WriteColsLn(printers.LEVEL_0, l("Namespace"), service.Namespace)
+	dw.WriteColsLn(printers.LEVEL_0, l("URL"), extractURL(service))
+	if service.Status.Address != nil {
+		url := service.Status.Address.GetURL()
+		dw.WriteColsLn(printers.LEVEL_0, l("Address"), url.String())
+	}
+	writeMapDesc(dw, printers.LEVEL_0, service.Labels, l("Labels"), "")
```
toVersus: Should labels and annotations be printed in columns rather than a comma separated string? IMO, it would improve readability in case of using a large amount of labels and annotations.

rhuss: You get full line-by-line list when usin `--details`. Otherwise its a single line (comma separated), and truncated if too long.

That way I think its a good compromise between a concise initial overview and the full list with all environment variables.

One could argue to even leave out envs, annos and labes when not using `--details` if the comma separated format doesn't make sense.

toVersus: > truncated if too long.

Ah, sorry! I missed the logic of masking too long line.

pkg/kn/commands/service/service_show.go
rhuss: implemented as navid suggests

pkg/kn/commands/service/service_describe.go
```diff
@@ -41,23 +125,513 @@ func NewServiceDescribeCommand(p *commands.KnParams) *cobra.Command {
...
+// Write a map either compact in a single line (possibly truncated) or, if printDetails is set,
+// over multiple line, one line per key-value pair
+func writeMapDesc(dw printers.PrefixWriter, indent int, m map[string]string, label string, labelPrefix string) {
+	if len(m) == 0 {
+		return
+	}
+
+	if printDetails {
+		l := labelPrefix + label
+		for key, value := range m {
```
toVersus: I'm a little concerned that the kn prints a list of annotations and labels in random order. This is caused by looping a map. Can we put all the map keys in string slice and sort using [sort.Strings](https://golang.org/pkg/sort/#Strings) before writing them in the buffer ?

rhuss: Good catch. Let's print them in alphabetical order, will fix that.

rhuss:
@sixolet @cppforlife @maximilien @navidshaikh Uff, finally I think the PR is ready for prime time. Some tests have been added (coverage 84%), please have another look at the PR. 'would be super awesome to get it merged until Friday before I head off to PTO and the PR would lie around for another three weeks.

PR #256, opened 2019-07-11
Add e2e tests for service with concurrency options

### LINKED ISSUE #253: Add integration tests for concurrency options of service create and update
For `service create` and `service update`, we are missing coverage for following options
```
      --concurrency-limit int    Hard Limit of concurrent requests to be processed by a single replica.
      --concurrency-target int   Recommendation for when to scale up based on the concurrent number of incoming request. Defaults to --concurrency-limit when given.
```

toVersus:
Check the behaviour of creating/updating service with concurrency options

Fixes #253

### DIFF
--- a/test/e2e/basic_workflow_test.go
+++ b/test/e2e/basic_workflow_test.go
@@ -58,6 +58,12 @@ func TestBasicWorkflow(t *testing.T) {
 	testRouteListWithArgument(t, k, "hello")
 	testServiceDelete(t, k, "hello")
 	testServiceDelete(t, k, "svc2")
+	testServiceCreateWithOptions(t, k, "svc3", []string{"--concurrency-limit", "250", "--concurrency-target", "300"})
+	testServiceDescribeConcurrencyLimit(t, k, "svc3", "250")
+	testServiceDescribeConcurrencyTarget(t, k, "svc3", "300")
+	testServiceUpdate(t, k, "svc3", []string{"--concurrency-limit", "300"})
+	testServiceDescribeConcurrencyLimit(t, k, "svc3", "300")
+	testServiceDelete(t, k, "svc3")
 	testServiceListEmpty(t, k)
 }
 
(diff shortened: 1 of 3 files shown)


test/e2e/basic_workflow_test.go
```diff
@@ -87,6 +93,19 @@ func testServiceCreate(t *testing.T, k kn, serviceName string) {
 	}
 }
 
+func testServiceCreateWithOptions(t *testing.T, k kn, serviceName string, args []string) {
```
navidshaikh: probably rename args with options ?

test/e2e/basic_workflow_test.go
```diff
@@ -87,6 +93,19 @@ func testServiceCreate(t *testing.T, k kn, serviceName string) {
...
+	command := []string{"service", "create", fmt.Sprintf("%s", serviceName), "--image", KnDefaultTestImage}
```
navidshaikh: just use the serviceName directly here?

test/e2e/basic_workflow_test.go
```diff
@@ -87,6 +93,19 @@ func testServiceCreate(t *testing.T, k kn, serviceName string) {
...
+	command = append(command, args...)
+	out, err := k.RunWithOpts(command, runOpts{NoNamespace: false})
+	if err != nil {
+		t.Fatalf(fmt.Sprintf("Error executing 'kn service create' command. Error: %s", err.Error()))
+	}
+
+	if !strings.Contains(out, fmt.Sprintf("Service '%s' successfully created in namespace '%s'.", serviceName, k.namespace)) {
```
navidshaikh: In other tests, we are testing the output string to contain specific words, for eg:
https://github.com/knative/client/blob/9fb6a436e81d69935a67cc5b245cc634b2ede55b/test/e2e/revision_workflow_test.go#L48

toVersus: Thanks! I changed to use assert here, but it seems that raw `strings.Contains` is used for asserting output text in other e2e tests. Can I follow up to fix them in separate PR?

navidshaikh: sure!

test/e2e/service_advance_test.go
```diff
@@ -0,0 +1,68 @@
...
+		t.Fatalf(fmt.Sprintf("Error executing 'kn service describe' command. Error: %s", err.Error()))
+	}
+
+	expectedOutput := fmt.Sprintf("containerConcurrency: %s", concurrencyLimit)
+	assert.Check(t, util.ContainsAll(out, expectedOutput))
+}
+
+func testServiceDescribeConcurrencyTarget(t *testing.T, k kn, serviceName, concurrencyTarget string) {
+	out, err := k.RunWithOpts([]string{"service", "describe", serviceName}, runOpts{NoNamespace: false})
+	if err != nil {
```
rhuss: You can use `assert.NilError()` to avoid that explicit check.

test/e2e/service_advance_test.go
```diff
@@ -0,0 +1,68 @@
...
+		t.Fatalf(fmt.Sprintf("Error executing 'kn service create' command. Error: %s", err.Error()))
+	}
+
+	assert.Check(t, util.ContainsAll(out, "Service", serviceName, "successfully created in namespace", k.namespace, "OK"))
+}
+
+func testServiceDescribeConcurrencyLimit(t *testing.T, k kn, serviceName, concurrencyLimit string) {
+	out, err := k.RunWithOpts([]string{"service", "describe", serviceName}, runOpts{NoNamespace: false})
+	if err != nil {
```
rhuss: `assert.NilError()`

test/e2e/service_advance_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/knative/client/pkg/util"
+	"gotest.tools/assert"
+)
+
+func TestServiceAdvance(t *testing.T) {
```
rhuss: Please call it more specific like `TestServiceOptions` and also rename the file to `service_options_test.go`

I would then add also here the test for minScale and maxScale.

toVersus:
@rhuss Thanks for your review! I addressed the comments.

PR #263, opened 2019-07-12
Hide completion command from help.

evankanderson:
## Proposed Changes

* Hide the presence of the `kn completion` command so that users mostly see the active command surface.

```release-note
`kn completion` is no longer suggested by `kn help`. The command still exists
to enable command-line completion, but it should not be advertised in normal
usage.
```

### DIFF
--- a/pkg/kn/commands/completion.go
+++ b/pkg/kn/commands/completion.go
@@ -30,6 +30,7 @@ func NewCompletionCommand(p *KnParams) *cobra.Command {
 	completionCmd := &cobra.Command{
 		Use:   "completion",
 		Short: "Output shell completion code (default Bash)",
+		Hidden: true,  // Don't show this in help listing.
 		Run: func(cmd *cobra.Command, args []string) {
 			if completionFlags.Zsh {
 				cmd.Root().GenZshCompletion(os.Stdout)


duglin:
oh the mental conflict! :-)
On the one hand, +1 since I'm not fan of that command/noun to being with!
On the other hand, I'm not a fan of hiding comands from the user either. How would anyone know about this command at all? Would our docs still talk about it?

If we keep it, I'm leaning more towards @rhuss's suggestion of having the help show groupings and putting this under `misc` instead of lying to the user about what commands are actually available.

evankanderson:
Sample output:

```shell
$ .\kn help
Manage your Knative building blocks:

* Serving: Manage your services and release new software to them.
* Eventing: Manage event subscriptions and channels. Connect up event sources.

Usage:
  kn [command]

Available Commands:

Flags:
  -h, --help                help for kn

Use "kn [command] --help" for more information about a command.

$ kn help completion
Output shell completion code (default Bash)

Usage:
  kn completion [flags]

Flags:
  -h, --help   help for completion
      --zsh    Generates completion code for Zsh shell.

Global Flags:
```

The use of `kn completion` is documented here:

https://github.com/knative/client#docs

> Run the following command to enable BASH auto-completion:
> 
> ```shell
> $ source <(kn completion)
> ```

My $0.02 is that hiding `kn completion` in the help (but documenting it in the docs) is appropriate, as running the command on its own doesn't help, you need to do a special invocation (e.g. `source <(kn completion)`) for the command to work. As such, it makes more sense to document it a different way.

rhuss:
Hmm, not sure how this PR could pass the tests, as the documentation has not be regenerated, nor the autoformatting has been applied (e.g. has not been build with `hack/build.sh`.

I'm going to extend our tests that all generated files have been generated.

rhuss:
Added check & fix to #265

PR #264, opened 2019-07-13
Clean up e2e test assertion

toVersus:
This is a follow-up PR of #256 . We should use assert package (and utils) for testing the output string to keep consistency between unit and e2e tests.

### DIFF
--- a/test/e2e/basic_workflow_test.go
+++ b/test/e2e/basic_workflow_test.go
@@ -18,8 +18,10 @@ package e2e
 
 import (
 	"fmt"
-	"strings"
 	"testing"
+
+	"github.com/knative/client/pkg/util"
+	"gotest.tools/assert"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


test/e2e/basic_workflow_test.go
```diff
@@ -65,121 +67,76 @@ func TestBasicWorkflow(t *testing.T) {
...
-	}
-	outputLines := strings.Split(out, "\n")
-	for _, line := range outputLines[1:] {
-		if len(line) > 1 && !strings.HasPrefix(line, serviceName) {
-			t.Fatalf(fmt.Sprintf("Expected output incorrect, expecting line to start with service name: %s\nFound: %s", serviceName, line))
-		}
-	}
+	assert.NilError(t, err)
+
+	assert.Check(t, util.ContainsAll(out, serviceName, "True"))
```
rhuss: Not sure if this is equivalent to the previous behaviour.
Tbh, I'm not sure whether the test before really worked as intended, as the revsion list now starts with the revsion names, not the service name (this is in the second column). 
Also you should test that *every* line (starting from line 2) contains the service name.

Not sure what the "True" comes from.

toVersus: I think the test before is not working as expected but passed because service name is included in the part of revision name. Also, I intentionally changed the logic to check if both service name and status of the revision ("True") are contained in the full output text. But, I will follow your suggestion to avoid missing the verification in case of multiple revisions associated with a single service.

toVersus:
> In the single main test method, can we wrap the calls to the subtests into t.Run("description of subtest", func(t *testing.T) { .... }) kind of subtest calls.

Can I make sure your intention here? Just wrap the subtests in the single main test method (e.g. `TestBasicWorkflow`) at this level of granularity?

```go
func TestBasicWorkflow(t *testing.T) {
	teardown := Setup(t)
	defer teardown(t)

	t.Run("returns no service before running tests", func(t *testing.T) {
		testServiceListEmpty(t, k)
	})

	t.Run("create hello service and returns no error", func(t *testing.T) {
		testServiceCreate(t, k, "hello")
	}

	t.Run("returns rough and detailed information associated with hello service")
		testServiceList(t, k, "hello")
		testServiceDescribe(t, k, "hello")
	})
...
}
```

test/e2e/basic_workflow_test.go
rhuss: Good catch.

Maybe you could check for space before/and after the serviceName ? Like in `" " + serviceName + " "` ? That way you can be ensure that the serviceName is included.

But I would still iterate over all line and ensure that the condition is true for every line.

toVersus:
@rhuss Thanks for reviewing! I tried to add subtests in main test methods. Let me know if you have any  opinions about the level of granularity. BTW, Prow doesn't trigger the test due to a typo :smirk_cat:

PR #265, opened 2019-07-13
fix(build): Minor build optimizations

rhuss:
* Check that every generated file has been regenerated for CI
* No color when not on a tty which is useful for build logs
* Always do updates when called without args
* Removed -u option and added -c for codegen only (dep update, docs gen, formatting, license check)


rhuss:
@maximilien could you have a quick look please and potentially merge ? This PR is about making the build more error proof so (e.g. update will be automatically called so that it can't be forgotten + the presubmit CI check also checks for license + docs now)

PR #267, opened 2019-07-16
chore(revision): Remove unwanted log message in test

rhuss:
This showed up in the test run as single log message so removing it

### DIFF
--- a/pkg/kn/commands/revision/revision_list_test.go
+++ b/pkg/kn/commands/revision/revision_list_test.go
@@ -163,7 +163,7 @@ func createMockRevisionWithParams(name, svcName string) *v1alpha1.Revision {
 	revision := &v1alpha1.Revision{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Revision",
-			APIVersion: "serving.knative.dev/v1alpha1",
+			APIVersion: "knative.dev/v1alpha1",
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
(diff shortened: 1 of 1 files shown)


pkg/kn/commands/revision/revision_list_test.go
```diff
@@ -164,7 +163,7 @@ func createMockRevisionWithParams(name, svcName string) *v1alpha1.Revision {
 	revision := &v1alpha1.Revision{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Revision",
-			APIVersion: "knative.dev/v1alpha1",
+			APIVersion: "serving.knative.dev/v1alpha1",
```
maximilien: The change seems to be at odd with the title...

rhuss: Well, what's odd and the test is broken, as it does not broken of because of the wrong coordinates (in fact it the api group *is* [serving.knative.dev/v1alpha1](https://github.com/knative/serving/blob/master/docs/spec/spec.md#revision), it's just not checked in this test, so it's meaningless anyway.

Happy to revert back to the wrong coordinates though.

rhuss:
Reverted the unrelated fix, and will send in a new PR with fixes for all test with a broken API group

rhuss:
/retest 

The flaky test should be fixed with #271 .

rhuss:
Guess we have to wait until #271 is merged to get the test green again.

PR #270, opened 2019-07-18
Command conventions doc

sixolet:
Mostly copied from  https://docs.google.com/document/d/1Or3f51EIrdyuwlJ9IfhQdoe7F7WrjsUPEPa0Ijsv2go/edit#

... updated a little, too.

Codifies CLI naming and structure conventions.

### DIFF
--- a/conventions/cli.md
+++ b/conventions/cli.md
@@ -1,0 +1,56 @@
+# Guidelines for kn Commands
+
+Commands are generally of the form `kn <resource> <verb>`.
+
+## Resource
+
+The <group>.knative.dev Kind, singluar and lowercase. For example, `service` for
+`serving.knative.dev/service` or `trigger` for `eventing.knative.dev/trigger`.
+
+## Verb
+
+If the thing the user's doing can be described by the following commands, these
+should be the name of the verb:
+
+* `describe` prints detailed information about a single resource. It can include
+  status information of related or child resources, too.
+
+* `list` prints summary information about all resources of a type, possibly
+  filtered by parent or label selector.
+
+* `create` creates a resource. Accepts a `--force` flag to create-or-replace.
+
+* `update` updates a resource based on the changes the user would like to make.
+
+* `delete` deletes a resource
+
+For a given resource there should be parallelism between arguments to `--create`
+and `--update` as much as possible.
+
+Other domain-specific verbs are possible, like `set-traffic` for a Knative
+Serving Service.
+
+## Arguments
+
+### Positionals
+
+Where there's a single target resource, it should be a positional argument.
+
+```bash
+kn service create foo --image gcr.io/things/stuff:tag
+```
+In this case `foo` is positional, and refers to the service to create.
+
+### Flags
+
+* `--force` is a flag on all create commands, and will replace the resource if
+  it already exists (otherwise this is an error). The resource will be *mutated*
+  to have a spec exactly like the resource that would otherwise be created. It
+  is not deleted and recreated.
+
+* When a flag sets a particular field on create or update, it should be a short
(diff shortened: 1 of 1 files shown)


conventions/cli.md
```diff
@@ -0,0 +1,56 @@
...
+
+The <group>.knative.dev Kind, singluar and lowercase. For example, `service` for
+`serving.knative.dev/service` or `trigger` for `eventing.knative.dev/trigger`.
+
+## Verb
+
+If the thing the user's doing can be described by the following commands, these
+should be the name of the verb:
+
+* `describe` prints detailed information about a single resource. It can include
```
rhuss: At the moment we don't have really a `describe` implemented, so it would be nice to reflect on this for a moment (i.e. it's still easy to change)

We had some brief discussion whether we want to rename that to `show` to distinguish it from the `kubectl` case as with `kn service show myservice -o yaml` we will be able to get the resource description of a Knative resource.

We already moved from `get` to `list` so it would be consequent also to move away from `describe` to avoid any confusion. I don't insist on `show` (but its much shorter to type)

| Action | kubectl | kn |     
| ------ | ------- | --- |
| List resources - human readable | kubectl get <entity> |  kn <entity> list |
| List resources - machine readable with -o | list or single entity depending on arguments | always list |
| Resource details - human readable | kubectl describe <entity> | kn <entity> show |
| Resource details - machine readable with -o |  ----    | always single entity |

conventions/cli.md
```diff
@@ -0,0 +1,56 @@
...
+* `list` prints summary information about all resources of a type, possibly
+  filtered by parent or label selector.
+
+* `create` creates a resource. Accepts a `--force` flag to create-or-replace.
+
+* `update` updates a resource based on the changes the user would like to make.
+
+* `delete` deletes a resource
+
+For a given resource there should be parallelism between arguments to `--create`
```
rhuss: `create` instead of `--create` ?

conventions/cli.md
```diff
@@ -0,0 +1,56 @@
...
+and `--update` as much as possible.
```
rhuss: `update` instead of `--update` ?

conventions/cli.md
```diff
@@ -0,0 +1,56 @@
...
+
+Other domain-specific verbs are possible, like `set-traffic` for a Knative
+Serving Service.
+
+## Arguments
+
+### Positionals
+
+Where there's a single target resource, it should be a positional argument.
```
rhuss: Maybe add that this positional always reference an entity of <noun>,
e.g. for a `kn revision list <something>` , `<something>` is always the name of a revision (and not e.g. a service)

duglin: What is "it" in this sentence?

rhuss: the name of the target resource of the type specified with `<noun>`

duglin: Do you mean "the name of the resource"?  If so, let's spell that out.

conventions/cli.md
sixolet: I like "describe", personally. And I like not changing things — describe exists today, it's just got annoying format.

conventions/cli.md
sixolet: Thanks for the catch

conventions/cli.md
sixolet: Thanks

sixolet:
I'll be making competing PRs after this is merged for the different styles for boolean flags. People can vote with reaction emoji.

conventions/cli.md
rhuss: I'm fine with keeping "describe" then (as it looks like I'm the only one with this concern)

PR #271, opened 2019-07-18
feature(service): Wait on update for service to become ready.

rhuss:
The behaviour is now the same as for `kn service create`. This should also fix some flakiness in the tests again.


docs/cmd/kn_service_update.md
```diff
@@ -27,6 +27,7 @@ kn service update NAME [flags]
 ### Options
 
 ```
+      --async                    Create service and don't wait for it to become ready.
```
navidshaikh: Now this should be, Create or update.
https://github.com/knative/client/blob/0b348d7441fcb48f1e94d36b223674ad8251f9c6/pkg/kn/commands/wait_flags.go#L36

rhuss: Ah, right. Let me fix that (maybe using a switch to have different messags for create and update)

pkg/kn/commands/wait_flags.go
```diff
@@ -32,8 +32,8 @@ type WaitFlags struct {
 // Add flags which influence the sync/async behaviour when creating or updating
 // resources. Set `waitDefault` argument if the default behaviour is synchronous.
 // Use `what` for describing what is waited for.
-func (p *WaitFlags) AddConditionWaitFlags(command *cobra.Command, waitTimeoutDefault int, what string) {
-	waitUsage := fmt.Sprintf("Create %s and don't wait for it to become ready.", what)
+func (p *WaitFlags) AddConditionWaitFlags(command *cobra.Command, waitTimeoutDefault int, action string, what string) {
```
navidshaikh: Nice :+1:

rhuss:
@maximilien @sixolet @cppforlife this is an important feature/fix which fixes a test flake introduced lately (where a `kn route list` comes directly after `kn service update` and the test expects the route to be already up).

would be cool if this could be reviewed asap as e.g. #267 integration test is constantly failing because of this.

pkg/kn/commands/service/service.go
```diff
@@ -36,3 +43,16 @@ func NewServiceCommand(p *commands.KnParams) *cobra.Command {
 	serviceCmd.AddCommand(NewServiceUpdateCommand(p))
 	return serviceCmd
 }
+
+func waitForService(client serving_kn_v1alpha1.KnClient, serviceName string, out io.Writer, timeout int) error {
+	fmt.Fprintf(out, "Waiting for service '%s' to become ready ... ", serviceName)
```
maximilien: And what happens if user `CTRL-C` out. I assume the API request continues but `kn` fails with user cancel? Might want to be clear. Since alternatively you could trap `CTL-C` and try to cancel cleanly or even delete service. Not advocating for clean up at this point but users will want to cancel and let's anticipate that and aid them.

maximilien:
Nice 👍 work BTW, definitely a needed feature. Cheers 🍻

pkg/kn/commands/service/service.go
rhuss: Good point. But I would interpret CTL-C as "cancelling the waiting" (that is where it currently states). We could try to catch that and print out something like "Cancelled waiting for service to become ready". It then just turns into an async behaviour where you have to check the server state. 

But a complete rollback is imo out of scope (that would mean to implement a fully transactional behaviour, recording the previous state and as you are on the client you don't have full control what's happening on the server-side. At least some events have been produced, which might trigger some other, non-rollbackable actions etc. A client-side managed transaction with possible compensation actions is not something easy.

maximilien: OK on canceling the waiting. Best.

maximilien:
@rhuss build is failing due to:

```bash
# github.com/knative/client/pkg/kn/commands/service [github.com/knative/client/pkg/kn/commands/service.test]
pkg/kn/commands/service/service_update_test.go:94:19: undefined: serving.GetRevisionTemplate
pkg/kn/commands/service/service_update_test.go:107:18: undefined: serving.GetRevisionTemplate
?   	github.com/knative/client/cmd/kn	[no test files]
?   	github.com/knative/client/hack	[no test files]
ok  	github.com/knative/client/pkg	0.098s [no tests to run]
ok  	github.com/knative/client/pkg/kn/commands	0.020s [no tests to run]
ok  	github.com/knative/client/pkg/kn/commands/revision	0.015s [no tests to run]
ok  	github.com/knative/client/pkg/kn/commands/route	0.015s [no tests to run]
FAIL	github.com/knative/client/pkg/kn/commands/service [build failed]
...
```

rhuss:
@maximilien that's because of the last merge of #275 which renamed that method. Going to fix that over the weekend.

rhuss:
@maximilien rename issue fixed.

rhuss:
@sixolet @maximilien @cppforlife PTAL, it already has been /lgtmed but had to be rebased because of an upstream merge conflict (renaming of some methods). Would be cool to lgtm it again, some other PRs are waiting for this so that the integration tests are back green again.

pkg/kn/commands/service/service.go
```diff
@@ -15,7 +15,14 @@
 package service
 
 import (
+	"io"
+	"time"
+
 	"github.com/knative/client/pkg/kn/commands"
+	serving_kn_v1alpha1 "github.com/knative/client/pkg/serving/v1alpha1"
+
+	"fmt"
```
cppforlife: i thought we were grouping std imports together?

docs/cmd/kn_service_update.md
```diff
@@ -40,6 +41,7 @@ kn service update NAME [flags]
   -p, --port int32               The port where application listens on.
       --requests-cpu string      The requested CPU (e.g., 250m).
       --requests-memory string   The requested memory (e.g., 64Mi).
+      --wait-timeout int         Seconds to wait before giving up on waiting for service to be ready. (default 60)
```
cppforlife: shouldnt we set wait-timeout to be time.Duration. otherwise what's the unit of this flag?

docs/cmd/kn_service_update.md
```diff
@@ -27,6 +27,7 @@ kn service update NAME [flags]
 ### Options
 
 ```
+      --async                    Update service and don't wait for it to become ready.
```
cppforlife: we should rename this flag to --wait=bool. async is too confusing imho with serveless "async" handler configuration (no expected response).

pkg/kn/commands/service/service.go
cppforlife: > Since alternatively you could trap CTL-C and try to cancel cleanly or even undo change
this would be a very surprising behaviour to most users.

pkg/kn/commands/service/service_update_test.go
```diff
@@ -75,7 +122,7 @@ func TestServiceUpdateImage(t *testing.T) {
 	servinglib.UpdateImage(template, "gcr.io/foo/bar:baz")
 
 	action, updated, output, err := fakeServiceUpdate(orig, []string{
-		"service", "update", "foo", "--image", "gcr.io/foo/quux:xyzzy", "--namespace", "bar"})
+		"service", "update", "foo", "--image", "gcr.io/foo/quux:xyzzy", "--namespace", "bar", "--async"}, false)
```
cppforlife: it seems strange that we placed non-default behaviour in bunch of tests. given that these are unit tests i dont see a good reason why we would add --async to bunch of them instead of just having a dedicated test that tests --async.

sixolet:
Unit tests for the simpler behavior (asynchronous) make perfect sense to me. We should have tests for both; should we add more specific waiting tests you think?

pkg/kn/commands/service/service.go
rhuss: I still have to add that kind of trapping for CTRL-C and printing out something like

"Interrupt: Stop waiting on service completion. Please check the service's status yourself."

or so.

pkg/kn/commands/service/service.go
rhuss: We are running now "goimports" unconditionally on every build. So we take exactly what is spit out there, that there is no ambiguity anymore. Is there anything we could optimize ?

docs/cmd/kn_service_update.md
rhuss: As the message says, its "Seconds to wait". I think (hope) that's good enough, but we can improve on adding a unit, too (like moving to a string and parsing "1min"). personally, I think that's overkill.

docs/cmd/kn_service_update.md
rhuss: Happy to rename when we decided to on the general direction. See #283 and #284 for voting how to continue with boolean flags (like `--help=true`)

pkg/kn/commands/service/service_update_test.go
rhuss: I agree, that we should update most of them. The more involved part is to get the reactor right for simulating the sync behaviour, but we can change that.

There is at least one test for the sync behaviour.

That not all tests have been changed is mostly because of lack of time. As we should refactor the create/update tests anyway (think there is even an issue open for that), I thought its ok to postpone that, too.

PR #274, opened 2019-07-18
Improves(e2e): improves integration tests and to better run locally

### LINKED ISSUE #273: Flaky e2e local tests
## In what area(s)?

e2e integration tests

Classifications:
kind cleanup

```bash
➜  client git:(improve-it) ✗ ./kn version
Version:      v20190718-local-0b348d7-dirty
Build Date:   2019-07-18 23:07:07
Git Revision: 0b348d7
Dependencies:
- serving:    v0.6.0
```

0.7.x

## Expected Behavior

Should be able to run `./test/e2e-tests-local.sh` that is connected to my IKS or minikube cluster.
Should be able to run `./test/e2e-tests-local.sh` multiple times in sequence.

## Actual Behavior

I get errors across tests since the test namespace used across tests is the same and gets deleted after a test but cannot be used (immediately) in subsequent test since it's being deleted.

Multiple runs produce similar issue with creating the namespace.

```bash
./test/e2e-tests-local.sh
...
./test/e2e-tests-local.sh
```

maximilien:
Improves integration tests and to better run locally

Fixes #273 

## Proposed Changes

* refactors basic workflow (move some code to common)
* use different namespace for each test to isolate
* include wait logic in `CreateNamespace` to allow multiple runs
* include wait logic in `DeleteNamespace` to allow multiple runs

**Release Note**

```release-note
Important since we need more integration tests overall for kn and some of
these changes should allow `./test/e2e-tests-local.sh` to run on
local clusters (minikube or something other than test-infra) and
correctly behave during multiple runs.
```


test/e2e/basic_workflow_test.go
```diff
@@ -20,156 +20,146 @@ import (
...
-
-	t.Run("delete hello and svc2 services and returns no error", func(t *testing.T) {
-		testServiceDelete(t, k, "hello")
-		testServiceDelete(t, k, "svc2")
-	})
-
-	t.Run("returns no service after completing tests", func(t *testing.T) {
-		testServiceListEmpty(t, k)
-	})
+	testServiceListEmpty(t, k)
```
rhuss: What is the reason to revert this change we introduced lately ?

test/e2e/basic_workflow_test.go
```diff
@@ -20,156 +20,146 @@ import (
...
+		t.Fatalf("Expected output 'No resources found.' Instead found:\n%s\n", out)
+	}
 }
 
 func testServiceCreate(t *testing.T, k kn, serviceName string) {
 	out, err := k.RunWithOpts([]string{"service", "create",
 		fmt.Sprintf("%s", serviceName),
 		"--image", KnDefaultTestImage}, runOpts{NoNamespace: false})
-	assert.NilError(t, err)
+	if err != nil {
```
rhuss: Why do you revert all these asserts ?

test/e2e/common.go
```diff
@@ -35,15 +36,64 @@ type runOpts struct {
...
+	var (
+		created bool
+		retries int
+		out     string
+		err     error
+	)
+
+	for !created && retries < MaxRetries {
+		out, err = kubectlCreateNamespace()
+		if err != nil && retries >= MaxRetries {
```
rhuss: How can retries be larger than MaxRetries here (considering the loop condition) ?

rhuss: Can we use an assert.NotNil() here ?

test/e2e/common.go
```diff
@@ -35,15 +36,64 @@ type runOpts struct {
...
+			t.Fatalf("Failed creating test namespace:%s\n after %d retries", namespace, retries)
+		} else {
```
rhuss: `else` is unneeded

test/e2e/common.go
rhuss: I think the logic of this loop is flawed.

I would put this in an extra function returning error or nil. So you could immediately return nil when success and an error after the loop, like in 

```golang
func createNamespace() error {

   retries := 0
   var err Error
   for retries < MaxRetries {
     _, err := kubectlCreateNamespace()
     if err == nil {
	    return nil
     } 
     time.Sleep(30 * time.Second)
     retries++
   }
   return err
}

```

Not tested, but hopefully good enough to clarify the point.

test/e2e/common.go
```diff
@@ -63,6 +113,41 @@ func DeleteTestNamespace(t *testing.T, namespace string) {
...
+		if err != nil {
+			t.Fatalf(fmt.Sprintf("Error executing 'kubectl get namespace' command. Error: %s", err.Error()))
+			return ""
+		}
+
+		return out
+	}
+
+	logger.Debugf("Waiting for namespace: %s to be deleted", namespace)
+	deleted := false
```
rhuss: as above, if you put the following code in a function you could return with a nil error on success and avoid the usage of such flags (which I consider bad style as its hard to follow and are often unnecessary).

This would make the code much more readable.

test/e2e/common.go
maximilien: Either one but I prefer it in the logic since I am not testing here.

test/e2e/basic_workflow_test.go
maximilien: I may have done that unintentionally... Let me double check.

test/e2e/basic_workflow_test.go
maximilien: Reverted. Interesting that GH or git did not complain when I pushed. Unless I somehow overwrote the files. Anyhow reverted and improved some more as I add another IT for plugin. I'll include that in the #249 which means this PR is a pre-req.

test/e2e/basic_workflow_test.go
maximilien: See previous. Reverted.

test/e2e/common.go
maximilien: It's more complicated than that but adding a helper private for the retry logic is fine. See update.

test/e2e/common.go
maximilien: Again slightly more complicated. But see update that separates retry logic into private func.

maximilien:
OK see update @rhuss. I need this merged before #249 if you can. Thx.

rhuss:
Fails because of async update, fixed with #271 . The test should be fixed as soon as this PR is merged.

Nevertheless, trying 
/retest

maximilien:
OK thx @rhuss I see #271 is merged. I will rebase and push this again. Hopefully along with update for #249

maximilien:
OK updated @rhuss. Let's see if all tests go green

PR #279, opened 2019-07-19
Validate scale and container concurrency options when updating configuration resource

toVersus:
I would like to do more contributions on knative/client but currently some large PRs are under reviewing. Therefore, I decided to read and refactor current code bases without bothering core contributors. If it's too much troubles, just let me know!

## Proposed Changes

* Validate the scale and container concurrency options and returns error when invalid value is found.
* Add unit/e2e tests for handling invalid user inputs
* Use GetContainer method for extracting container spec from revision template
* Use constant values for annotation keys
* Add missing unit test for TestUpdateContainerImage method

**Release Note**

```release-note
None
```

### DIFF
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -51,10 +51,10 @@ func (p *ConfigurationEditFlags) AddUpdateFlags(command *cobra.Command) {
 	command.Flags().StringVar(&p.RequestsFlags.Memory, "requests-memory", "", "The requested memory (e.g., 64Mi).")
 	command.Flags().StringVar(&p.LimitsFlags.CPU, "limits-cpu", "", "The limits on the requested CPU (e.g., 1000m).")
 	command.Flags().StringVar(&p.LimitsFlags.Memory, "limits-memory", "", "The limits on the requested memory (e.g., 1024Mi).")
-	command.Flags().IntVar(&p.MinScale, "min-scale", -1, "Minimal number of replicas.")
-	command.Flags().IntVar(&p.MaxScale, "max-scale", -1, "Maximal number of replicas.")
+	command.Flags().IntVar(&p.MinScale, "min-scale", 0, "Minimal number of replicas.")
+	command.Flags().IntVar(&p.MaxScale, "max-scale", 0, "Maximal number of replicas.")
 	command.Flags().IntVar(&p.ConcurrencyTarget, "concurrency-target", 0, "Recommendation for when to scale up based on the concurrent number of incoming request. Defaults to --concurrency-limit when given.")
-	command.Flags().IntVar(&p.ConcurrencyLimit, "concurrency-limit", -1, "Hard Limit of concurrent requests to be processed by a single replica.")
+	command.Flags().IntVar(&p.ConcurrencyLimit, "concurrency-limit", 0, "Hard Limit of concurrent requests to be processed by a single replica.")
 	command.Flags().Int32VarP(&p.Port, "port", "p", 0, "The port where application listens on.")
 }
 
@@ -111,7 +111,21 @@ func (p *ConfigurationEditFlags) Apply(service *servingv1alpha1.Service, cmd *co
 		}
 	}
 
-	servinglib.UpdateConcurrencyConfiguration(template, p.MinScale, p.MaxScale, p.ConcurrencyTarget, p.ConcurrencyLimit)
+	if cmd.Flags().Changed("min-scale") {
+		servinglib.UpdateMinScale(template, p.MinScale)
+	}
+
+	if cmd.Flags().Changed("max-scale") {
+		servinglib.UpdateMaxScale(template, p.MaxScale)
+	}
+
+	if cmd.Flags().Changed("concurrency-target") {
+		servinglib.UpdateConcurrencyTarget(template, p.ConcurrencyTarget)
+	}
+
+	if cmd.Flags().Changed("concurrency-limit") {
+		servinglib.UpdateConcurrencyLimit(template, p.ConcurrencyLimit)
+	}
 
 	return nil
 }
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -51,10 +51,10 @@ func (p *ConfigurationEditFlags) AddUpdateFlags(command *cobra.Command) {
 	command.Flags().StringVar(&p.RequestsFlags.Memory, "requests-memory", "", "The requested memory (e.g., 64Mi).")
 	command.Flags().StringVar(&p.LimitsFlags.CPU, "limits-cpu", "", "The limits on the requested CPU (e.g., 1000m).")
 	command.Flags().StringVar(&p.LimitsFlags.Memory, "limits-memory", "", "The limits on the requested memory (e.g., 1024Mi).")
-	command.Flags().IntVar(&p.MinScale, "min-scale", 0, "Minimal number of replicas.")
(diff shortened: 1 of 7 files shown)


maximilien:
@toVersus Thanks for contribution. Yes, we have some pending PRs on core and things should clear up next week. Also, PR #274 will require some changes for your e2e but should not be too hard and clean things a bit even more. Leaving some initial review. Cheers 🍻

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -51,10 +51,10 @@ func (p *ConfigurationEditFlags) AddUpdateFlags(command *cobra.Command) {
 	command.Flags().StringVar(&p.RequestsFlags.Memory, "requests-memory", "", "The requested memory (e.g., 64Mi).")
 	command.Flags().StringVar(&p.LimitsFlags.CPU, "limits-cpu", "", "The limits on the requested CPU (e.g., 1000m).")
 	command.Flags().StringVar(&p.LimitsFlags.Memory, "limits-memory", "", "The limits on the requested memory (e.g., 1024Mi).")
-	command.Flags().IntVar(&p.MinScale, "min-scale", 0, "Minimal number of replicas.")
-	command.Flags().IntVar(&p.MaxScale, "max-scale", 0, "Maximal number of replicas.")
+	command.Flags().IntVar(&p.MinScale, "min-scale", -1, "Minimal number of replicas.")
```
maximilien: `-1` reads odd since it's kind of an impossible number. What does -1 minimum replica means... same for the others.

toVersus: I intentionally put the invalid number on these update options as default. Let's say, someone wants to change the minScale value for hello service from 1 to 0 and runs `kn service update hello --min-scale 0`. In the previous implementation, kn didn't allow to set 0 value to minScale option according to the validation below. Therefore, the users cannot update minScale value as they want. I think this behavior is not desirable.

https://github.com/knative/client/blob/144ccb875e1de81067623281e5628e67d2c982a4/pkg/serving/config_changes.go#L40-L42

That's why I set the invalid number to those update options as default. If we need the default values for each option, I guess we should put the invalid numbers on them and change the value only when the users pass the valid number for each option. Otherwise, leave it as it is.

sixolet: I think it's better to use https://godoc.org/github.com/spf13/pflag#FlagSet.Changed, which will tell you straight-up whether the user set it. That way things that read the defaults don't have to worry about weird invalid numbers.

toVersus: @sixolet Thanks for your suggestion! I updated to use FlagSet.Changed so we don't care about default value for those update options.

rhuss:
@toVersus have you rebased to the latest master commits ? If not, please do as this would fix the integration test error.

toVersus:
@rhuss Hmm, I have already rebased to the latest master commit, but it seems something wrong with the tests against the latest release job. I rebased to the previous commit and triggered the CI job again.

toVersus:
@rhuss It looks like an error comes from the flakiness of the test. Now, it's green!
I think this PR is ready for the review because #274 is already merged.

pkg/serving/config_changes.go
```diff
@@ -35,34 +37,59 @@ func UpdateEnvVars(template *servingv1alpha1.RevisionTemplateSpec, vars map[stri
...
+func UpdateMaxScale(template *servingv1alpha1.RevisionTemplateSpec, max int) error {
+	return UpdateAnnotation(template, autoscaling.MaxScaleAnnotationKey, strconv.Itoa(max))
+}
+
+// UpdateConcurrencyTarget updates container concurrency annotation
+func UpdateConcurrencyTarget(template *servingv1alpha1.RevisionTemplateSpec, target int) error {
+	// TODO(toVersus): Remove the following validation once serving library is updated to v0.8.0
+	// and just rely on ValidateAnnotations method.
+	if target < autoscaling.TargetMin {
+		return fmt.Errorf("Invalid %s annotation value: must be an intger greater than 0", autoscaling.TargetAnnotationKey)
```
navidshaikh: *integer

navidshaikh: How about printing error message as
```
("Invalid `concurrency-target` value: %d , must be an integer > 0", target)
```
easy for user to know which flag value to be adjusted than printing the annotation key
WDYT ?

pkg/serving/config_changes.go
```diff
@@ -35,34 +37,59 @@ func UpdateEnvVars(template *servingv1alpha1.RevisionTemplateSpec, vars map[stri
...
+	return UpdateAnnotation(template, autoscaling.TargetAnnotationKey, strconv.Itoa(target))
+}
+
+// UpdateConcurrencyLimit updates container concurrency limit
+func UpdateConcurrencyLimit(template *servingv1alpha1.RevisionTemplateSpec, limit int) error {
+	cc := servingv1beta1.RevisionContainerConcurrencyType(limit)
+	// Validate input limit
+	ctx := context.Background()
+	if err := cc.Validate(ctx).ViaField("spec.containerConcurrency"); err != nil {
+		return fmt.Errorf("Invalid containerConcurrency revision spec: %s", err)
```
navidshaikh: again, the field referenced in error should match the flag name ?

pkg/serving/config_changes.go
```diff
@@ -35,34 +37,59 @@ func UpdateEnvVars(template *servingv1alpha1.RevisionTemplateSpec, vars map[stri
...
-func UpdateAnnotation(template *servingv1alpha1.RevisionTemplateSpec, annotation string, value string) {
+// UpdateAnnotation updates (or adds) an annotation to the given service
+func UpdateAnnotation(template *servingv1alpha1.RevisionTemplateSpec, annotation string, value string) error {
 	annoMap := template.Annotations
 	if annoMap == nil {
 		annoMap = make(map[string]string)
 		template.Annotations = annoMap
 	}
+
+	// Validate autoscaling annotations and returns the same value as before if input value is invalid
```
navidshaikh: *retuns error if invalid input provided without changing the existing spec

pkg/serving/v1alpha1/client_test.go
```diff
@@ -423,8 +423,8 @@ func newRoute(name string) *v1alpha1.Route {
 
 func getServiceEvents(name string) []watch.Event {
 	return []watch.Event{
-		{watch.Added, wait.CreateTestServiceWithConditions(name, corev1.ConditionUnknown, corev1.ConditionUnknown, "")},
-		{watch.Modified, wait.CreateTestServiceWithConditions(name, corev1.ConditionUnknown, corev1.ConditionTrue, "")},
-		{watch.Modified, wait.CreateTestServiceWithConditions(name, corev1.ConditionTrue, corev1.ConditionTrue, "")},
+		{Type: watch.Added, Object: wait.CreateTestServiceWithConditions(name, corev1.ConditionUnknown, corev1.ConditionUnknown, "")},
+		{Type: watch.Modified, Object: wait.CreateTestServiceWithConditions(name, corev1.ConditionUnknown, corev1.ConditionTrue, "")},
+		{Type: watch.Modified, Object: wait.CreateTestServiceWithConditions(name, corev1.ConditionTrue, corev1.ConditionTrue, "")},
```
navidshaikh: doesnt seem like a related change ?

pkg/serving/config_changes.go
toVersus: Yeah, that’s more like it! Thanks.

toVersus:
@navidshaikh Thanks for reviewing! I addressed your comments.

Issue #276, opened 2019-07-19
Allow bring-your-own revision name


sixolet:
## In what area(s)?
Supported parameters

## Describe the feature:

v1beta1 style (even in v1alpha1) allows "bring your own" revision names. Let's allow users to bring 'em in service creation .

`--revision-name` or `--revision-name-prefix` (or even `--name-prefix`) can be the name of the flag.

duglin:
why the "prefix" variants? It's not a prefix w.r.t. naming the revision.

Of those I like `--revision-name` even thought it's kind of long.  `--name` would be nice but that's probably not clear enough on which thing we're naming. `--revname` might be too short, but it is an option.

`kn service create myapp --image ns/repo --revname myapp-v1`

re: prefix... what's interesting is that if we don't ask for the full name, if anything it would be `postfix` since I think all revNames must start with a prefix that matches the service name.

`kn service create myapp --image ns/repo --revpostfix=v1` ?
`kn service create myapp --image ns/repo --posfix=v1` ?
`kn service create myapp --image ns/repo --name-postfix=v1` ?
`kn service create myapp --image ns/repo --revision=v1` ?
`kn service create myapp --image ns/repo --version=v1` ? 

We just need to make sure that whatever we use it'll be an obvious mapping to the revision name in the traffic section.

`kn service create myapp --image ns/repo --version v1 --traffic v1:50 --traffic v2:50`
`kn service create myapp --image ns/repo --revision v1 --traffic v1:50 --traffic v2:50`

I can't decide if it's really helpful to not force them to duplicate the KnService name over and over, or hiding too much.  But it that does feel nicer than something like:
`kn service create myapp --image ns/repo --revision-name myapp-v1 --traffic myapp-v1:50 --traffic myapp-v2:50`

So far, I like this one the best:
`kn service create myapp --image ns/repo --revision v1 --traffic v1:50 --traffic v2:50`

rhuss:
> I can't decide if it's really helpful to not force them to duplicate the KnService name over and over, or hiding too much.

Think about `kn revision list` (or `kubectl get revision`) with multiple services active. Of course, you have a service column but the column is sorted after revision name. Having a common prefix like the service name groups them nicely in such name sorted list.

duglin:
yup - wasn't questioning it in lists like "revision list" - was just wondering about it on the cmd line for specifying traffic or revision name. E.g.
[...]

Issue #110, opened 2019-05-15
Split ConfigurationEditFlags and create ConfigurationCreateFlags


navidshaikh:
We have `--force` option while creating a service, the `ConfigurationEditFlags` structure can be restructured as suggested here https://github.com/knative/client/pull/79#discussion_r283004004

navidshaikh:
/reopen
/remove-lifecycle stale

Added refactor label to it and we can consider this while separating the flags into their own package for re-usability by plugins.

rhuss:
/remove-lifecycle stale

I think this issue is still valid

PR #282, opened 2019-07-19
Specify names on service update and generate names client-side.

### LINKED ISSUE #277: Allow automatic client-side revision name generation
## In what area(s)?
Workflow and client conventions

## Describe the feature:

Previously, in the client-conventions.md (still in serving, still meaning to port it over here), we suggested adding a nonce to force creation of a new revision and to make that new revision findable once you'd created it.

Now that v1beta1 style (even in v1alpha1) allows for bring-your-own-revision-name, it's much more natural to autogenerate a name for the revision on the client side. We should support random name generation on the client side, and have it on by default, so that the revisions are findable, at least, and so that when you deploy an image by a tag that's changed it forces creation of a new revision.

sixolet:
Fixes #277 #276 

## Proposed Changes

* `--revision-name` flag allows setting revision names in service update. If you don't provide the service name as a prefix, it'll get auto-added for you.
* `--generate-revision-name` flag (default TRUE) makes revision names client side by default.

Whether to include the generation is up for discussion. Otherwise this is substantially similar to the server-side generateName(), up to and including our lack of vowels.

**Release Note**

```release-note

```

### DIFF
--- a/docs/cmd/kn_service_update.md
+++ b/docs/cmd/kn_service_update.md
@@ -39,11 +39,11 @@ kn service update NAME [flags]
       --limits-memory string     The limits on the requested memory (e.g., 1024Mi).
       --max-scale int            Maximal number of replicas.
       --min-scale int            Minimal number of replicas.
-      --name string              The revision name prefix to set. Implies --generate-revision-name=false
   -n, --namespace string         List the requested object(s) in given namespace.
   -p, --port int32               The port where application listens on.
       --requests-cpu string      The requested CPU (e.g., 250m).
       --requests-memory string   The requested memory (e.g., 64Mi).
+      --revision-name string     The revision name to set. If you don't add the service name as a prefix, it'll be added for you.
       --wait-timeout int         Seconds to wait before giving up on waiting for service to be ready. (default 60)
 ```
 
--- a/docs/cmd/kn_service_update.md
+++ b/docs/cmd/kn_service_update.md
@@ -42,7 +42,7 @@ kn service update NAME [flags]
   -p, --port int32               The port where application listens on.
       --requests-cpu string      The requested CPU (e.g., 250m).
       --requests-memory string   The requested memory (e.g., 64Mi).
-      --revision-name string     The revision name to set. Must start with the service name and a dash as a prefix. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
+      --revision-name string     The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
       --wait-timeout int         Seconds to wait before giving up on waiting for service to be ready. (default 60)
 ```
 
--- a/docs/cmd/kn_service_update.md
+++ b/docs/cmd/kn_service_update.md
@@ -31,6 +31,7 @@ kn service update NAME [flags]
       --concurrency-limit int    Hard Limit of concurrent requests to be processed by a single replica.
       --concurrency-target int   Recommendation for when to scale up based on the concurrent number of incoming request. Defaults to --concurrency-limit when given.
   -e, --env stringArray          Environment variable to set. NAME=value; you may provide this flag any number of times to set multiple environment variables. To unset, specify the environment variable name followed by a "-" (e.g., NAME-).
+      --generate-revision-name   Automatically generate a revision name client-side. If false, the revision name is cleared. (default true)
   -h, --help                     help for update
       --image string             Image to run.
   -l, --label stringArray        Service label to set. name=value; you may provide this flag any number of times to set multiple labels. To unset, specify the label name followed by a "-" (e.g., name-).
@@ -38,6 +39,7 @@ kn service update NAME [flags]
       --limits-memory string     The limits on the requested memory (e.g., 1024Mi).
       --max-scale int            Maximal number of replicas.
       --min-scale int            Minimal number of replicas.
+      --name string              The revision name prefix to set. Implies --generate-revision-name=false
   -n, --namespace string         List the requested object(s) in given namespace.
   -p, --port int32               The port where application listens on.
       --requests-cpu string      The requested CPU (e.g., 250m).
--- a/docs/cmd/kn_service_update.md
+++ b/docs/cmd/kn_service_update.md
@@ -31,7 +31,6 @@ kn service update NAME [flags]
       --concurrency-limit int    Hard Limit of concurrent requests to be processed by a single replica.
       --concurrency-target int   Recommendation for when to scale up based on the concurrent number of incoming request. Defaults to --concurrency-limit when given.
   -e, --env stringArray          Environment variable to set. NAME=value; you may provide this flag any number of times to set multiple environment variables. To unset, specify the environment variable name followed by a "-" (e.g., NAME-).
-      --generate-revision-name   Automatically generate a revision name client-side. If false, the revision name is cleared. (default true)
   -h, --help                     help for update
       --image string             Image to run.
   -l, --label stringArray        Service label to set. name=value; you may provide this flag any number of times to set multiple labels. To unset, specify the label name followed by a "-" (e.g., name-).
@@ -43,7 +42,7 @@ kn service update NAME [flags]
   -p, --port int32               The port where application listens on.
       --requests-cpu string      The requested CPU (e.g., 250m).
       --requests-memory string   The requested memory (e.g., 64Mi).
-      --revision-name string     The revision name to set. If you don't add the service name as a prefix, it'll be added for you.
+      --revision-name string     The revision name to set. Must start with the service name and a dash as a prefix. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
       --wait-timeout int         Seconds to wait before giving up on waiting for service to be ready. (default 60)
 ```
 
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -41,8 +41,7 @@ type ConfigurationEditFlags struct {
 	RevisionName               string
 
 	// Preferences about how to do the action.
-	GenerateRevisionName bool
-	ForceCreate          bool
+	ForceCreate bool
 
 	// Bookkeeping
 	flags []string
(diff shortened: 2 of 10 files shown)


sixolet:
Cool, now that I've dealt with the old-api cases I feel happier about this.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -85,6 +127,23 @@ func (p *ConfigurationEditFlags) Apply(service *servingv1alpha1.Service, cmd *co
 		return err
 	}
 
+	name := ""
```
rhuss: This `apply()` method is already quite a mouthful could this revision name handling block be extracted in an own method e.g. `setRevisionName(cmd)` ?

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -26,51 +26,93 @@ import (
...
+	flags []string
 }
 
 type ResourceFlags struct {
 	CPU    string
 	Memory string
 }
 
-func (p *ConfigurationEditFlags) AddUpdateFlags(command *cobra.Command) {
+func (p *ConfigurationEditFlags) addSharedFlags(command *cobra.Command) {
```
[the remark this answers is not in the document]
duglin: Can you add a comment to explain that "Shared" means that these flags are common between both create() and update() ?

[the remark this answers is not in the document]
duglin: Let's get this PR in before #345 so that #345 can add its flags in here too!

pkg/serving/service.go
```diff
@@ -38,6 +41,23 @@ func RevisionTemplateOfService(service *servingv1alpha1.Service) (*servingv1alph
 	return config.DeprecatedRevisionTemplate, nil
 }
 
+var charChoices = []string{
+	"b", "c", "d", "f", "g", "h", "j", "k", "l", "m", "n", "p", "q", "r", "s", "t", "v", "w", "x",
```
duglin: I gotta ask... did you purposely remove some chars to reduce the chances of inappropriate words?  :-)

rhuss: @duglin see initial PR comment:

> Otherwise this is substantially similar to the server-side generateName(), up to and including our lack of vowels.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -85,6 +127,23 @@ func (p *ConfigurationEditFlags) Apply(service *servingv1alpha1.Service, cmd *co
...
+		name = p.RevisionName
+		if !strings.HasPrefix(name, service.Name+"-") {
+			name = service.Name + "-" + name
+		}
+	} else if p.GenerateRevisionName {
+		name = servinglib.GenerateRevisionName(service)
+	}
+	if anyMutation {
+		err = servinglib.UpdateName(template, name)
+		if err == servinglib.ApiTooOldError && !cmd.Flags().Changed("revision-name") {
```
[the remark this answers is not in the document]
rhuss: I'm fine with both. It's not hard to understand, just unconventional for my eyes (not a Python guy, and in Perlland we never used such an idiom, we had better ways to obfuscate ;-P

pkg/kn/commands/service/service_update_test.go
```diff
@@ -398,6 +491,23 @@ func newEmptyService() *v1alpha1.Service {
 	}
 }
 
+func newEmptyServiceBeta() *v1alpha1.Service {
```
rhuss: doesnt look very 'beta'

pkg/serving/service.go
```diff
@@ -38,6 +42,40 @@ func RevisionTemplateOfService(service *servingv1alpha1.Service) (*servingv1alph
...
+	chars := make([]string, 0, l)
+	for i := 0; i < l; i++ {
+		chars = append(chars, charChoices[rand.Int()%len(charChoices)])
+	}
+	return strings.Join(chars, "")
+}
+
+// GenerateRevisionName returns an automatically-generated name suitable for the
+// next revision of the given service.
+func GenerateRevisionName(nameTempl string, service *servingv1alpha1.Service) (string, error) {
```
rhuss: Can we add some tests for this method ? (and maybe also when given as argument from outside in service_update_test ?)

rhuss:
If I understand the motivation of client-side generation as described in https://github.com/knative/client/blob/master/conventions/client.md#change-non-code-attributes then it's about the potential risk of a race condition so that `latestCreatedRevisionName` could not be used for obtaining the image digest.

I doubt a bit that this could be the case as you always have an optimistic lock on the service resource, so when two clients simultaneously try to update the service the first one wins, and the second one has to re-fetch the resource (with the updated `latestCreatedRevisionName`). I'm not sure whether this update of the latestCreatedRevisionName is done in an admission controller or async. 

In the latter case, there is still a slight chance for a race, but is this really that severe?

* Two clients have to operate at the same time
* The digest behind the image tag must have been changed. This is likely the case for `latest`. It could be the case for a version like tag (`1.2`) but although possible and sometimes done, I consider this a very unlikely event.

So if there should be a race for `latest` which this might result in an hick-up of a development workflow (as `latest` is not used in production).

I don't question the algorithm in the document at all, it has been the result of a lengthy (and probably painful) discussion. I just want to point out that I personally would take that risk and just reuse `lastesCreatedRevisionName` always for finding the active revision, regardless of who has created the name (since we are in an update or change situation we will create a new revision anyway, no need for a trigger in this case)

sixolet:
`latestCreatedRevisionName` is updated async, which is the race opportunity, yeah.

The chance of the race is low, but the consequences can be high (running code you specifically did not intend to run)

pkg/kn/commands/service/service_update_test.go
sixolet: Renamed.

pkg/serving/service.go
sixolet: Done.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -85,6 +127,23 @@ func (p *ConfigurationEditFlags) Apply(service *servingv1alpha1.Service, cmd *co
...
 	}
 
+	name := ""
+	if cmd.Flags().Changed("revision-name") {
+		name = p.RevisionName
+		if !strings.HasPrefix(name, service.Name+"-") {
+			name = service.Name + "-" + name
+		}
+	} else if p.GenerateRevisionName {
+		name = servinglib.GenerateRevisionName(service)
```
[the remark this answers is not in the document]
sixolet: The server will reject. I think this'll work as designed with people able to template their own.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -85,6 +127,23 @@ func (p *ConfigurationEditFlags) Apply(service *servingv1alpha1.Service, cmd *co
...
+	}
+	if anyMutation {
```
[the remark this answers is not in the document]
sixolet: done

pkg/kn/commands/service/configuration_edit_flags.go
sixolet: Done

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -26,45 +26,90 @@ import (
...
+	command.Flags().IntVar(&p.ConcurrencyTarget, "concurrency-target", 0,
+		"Recommendation for when to scale up based on the concurrent number of incoming request. "+
+			"Defaults to --concurrency-limit when given.")
+	p.markFlagMakesRevision("concurrency-target")
+	command.Flags().IntVar(&p.ConcurrencyLimit, "concurrency-limit", 0,
+		"Hard Limit of concurrent requests to be processed by a single replica.")
+	p.markFlagMakesRevision("concurrency-limit")
 	command.Flags().Int32VarP(&p.Port, "port", "p", 0, "The port where application listens on.")
+	p.markFlagMakesRevision("port")
+	command.Flags().StringVar(&p.RevisionName, "revision-name", "{{.Service}}-{{.Random 5}}-{{.Generation}}",
```
duglin: this seems excessively long :-)

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -165,3 +224,14 @@ func (p *ConfigurationEditFlags) computeResources(resourceFlags ResourceFlags) (
 
 	return resourceList, nil
 }
+
+// AnyMutation returns tue if there are any revision template mutations in the
```
duglin: s/tue/true/

docs/cmd/kn_service_update.md
```diff
@@ -41,6 +41,7 @@ kn service update NAME [flags]
   -p, --port int32               The port where application listens on.
       --requests-cpu string      The requested CPU (e.g., 250m).
       --requests-memory string   The requested memory (e.g., 64Mi).
+      --revision-name string     The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
```
duglin: Since I suspect this flag will be used a lot, what do you think about having it available via `-r` too ?

pkg/kn/commands/service/configuration_edit_flags.go
sixolet: It inspired #351 . I think it's better to describe clearly.

duglin: only the first time, after that when people do --help it'll just annoy them

docs/cmd/kn_service_update.md
sixolet: Easy enough to do later.  Let's do the simpler thing for now — I'm nervous-er about doing shortopts w/o consensus much more than longopts.

pkg/kn/commands/service/configuration_edit_flags.go
sixolet: Hum. Roland suggested I _add_ the part about the empty string. I think the part about the template is valuable, as it is not discoverable otherwise. 

We could make it so, but we don't have a good place for that yet. I'd like to defer shortening it until we have a good place for other doc on options.

docs/cmd/kn_service_update.md
rhuss: Now that I know the service name prefix is mandatory for a revision, I wonder whether we can help the user to do this for her automatically if not already existant? E.g. allowing for name `{{.Random 5}}-{{.Generation}}` (which would be a good benefit because of the verboseness of Go templates).

If not, we should at least add client-side validation for this constraint.

sixolet:
If you pass the empty string, as the revision name, it's equivalent

rhuss:
> If you pass the empty string, as the revision name, it's equivalent

Cool. So the last thing which needs clarification whether we enforce a `{{.Service}}-` prefix or allow everything and wait for the server-side error. I liked that this was added implicitly when not given.

sixolet:
Oh I think I misread you as saying the opposite earlier. Easy enough!

rhuss:
Yeah, sorry, totally missed that the service name as the prefix is a mandatory server-side requirement. Now that this is clear to me, I think we already should check and enforce that client-side. Apologies for my ignorance.

sixolet:
Did you mean check and error, or fix it up for the user? I'm of two minds myself that's why I keep changing it back and forth

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -26,45 +26,90 @@ import (
...
+		"The revision name to set. Must start with the service name and a dash as a prefix. "+
+			"Empty revision name will result in the server generating a name for the revision. "+
+			"Accepts golang templates, allowing {{.Service}} for the service name, "+
+			"{{.Generation}} for the generation, and {{.Random [n]}} for n random consonants.")
+	p.markFlagMakesRevision("revision-name")
 }
 
+// AddUpdateFlags adds the flags specific to update.
+func (p *ConfigurationEditFlags) AddUpdateFlags(command *cobra.Command) {
+	p.addSharedFlags(command)
```
navidshaikh: The better way to handle this separation would be to implement the suggestion from @cppforlife https://github.com/knative/client/pull/79#discussion_r283004004 logged as issue at https://github.com/knative/client/issues/110, that should go as separate PR though and shouldn't block this one.

pkg/kn/commands/service/configuration_edit_flags.go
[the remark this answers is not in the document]
navidshaikh: @duglin We add traffic flags separately on `service update` command. IMO they don't need to show up in `ConfigurationEditFlags` as updating traffic block is not configuration-edit operation and only traffic flags change don't result into a new revision.

rhuss:
@sixolet I clearly would opt for fixing it for the user, as this is what the user would have to do anyway manually. Tbh I found it very annoying if a command knows how to fix sth, but doesn't do it automatically when there is a 99% certainty about the intention of the user.

My favourite (counter) example is:
```
$ find -name bla .
find: paths must precede expression: .
Usage: find [-H] [-L] [-P] [-Olevel] [-D help|tree|search|stat|rates|opt|exec] [path...] [expression]
```

(of course there could be ambiguities in certain cases, and then one can still throw an error. But for this very common 'user error' there should be an automatic "fix").

We could 'educate' the user by sending a warning if not `service-` prefixed, but its maybe also just fine to add this to the success message like

```
kn service create --image gcr.io/bla --revision-name="{{Rand 5}}-{{.Generation}}"
Service "myservice" with revision "myservice-dgnde-1" has been created.
```

and maybe something similar for `kn service update`.

pkg/kn/commands/service/configuration_edit_flags.go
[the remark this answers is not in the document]
duglin: From a UX perspective, I'd like to be able to setup the traffic during "service create" - that's all I'm asking for.

duglin:
Seeing something odd:
```
$ ./kn service create echo --image duglin/echo --revision-name echo-v1
the service is using too old of an API format for the operation
```
But this works:
```
$ ./kn service create echo --image duglin/echo
Service 'echo' successfully created in namespace 'default'.
Waiting for service 'echo' to become ready ... OK

Service URL:
http://echo-default.kndev.us-south.containers.appdomain.cloud

$ ./kn service update echo --revision-name echo-v1
Waiting for service 'echo' to become ready ... OK
Service 'echo' updated in namespace 'default'.
```
I'm pretty sure I have both alpha1 and beta1 installed in my cluster.

sixolet:
@duglin that's due to a related thing that I think should be fixed in a different PR: we default to creating services in the old api format, so a newly created service "is using the old api format" until it's created and then the server converts it.

duglin:
ok, so when #361 is merged then all should be good, right?
btw - rebase needed on this one.

docs/cmd/kn_service_update.md
duglin: Do we describe what "Generation" is in our docs some place?

duglin:
/lgtm

Couple of comments/question though:
- I'm assuming once #361 goes in you'll create a PR to handle the `kn service create` variant, right?
- any reason we're no updating the docs in https://github.com/knative/client/blob/master/docs/cmd/kn_service_update.md with info about this?

Rebase needed.

sixolet:
There we go. Rebased.

PR #283, opened 2019-07-21
Document boolean flags by presence

sixolet:
Add documentation that boolean flags should be specified by flag presence. 

This is mutally exclusive with https://github.com/knative/client/pull/284 . 

Vote by thumbs up reaction emoji. Winner gets merged.

### DIFF
--- a/conventions/cli.md
+++ b/conventions/cli.md
@@ -64,6 +64,13 @@ In this case `foo` is positional, and refers to the service to create.
   example, `--image=img.repo/asdf` in Knative Serving sets
   `spec.template.containers[0].image`
 
+* Flags that control a boolean behavior (eg. generate a name or not) are
+  specified by their presence. The default happens when the flag is not present,
+  and adding the flag marks the user's desire for the non-default thing. When
+  the flag *disables* a default behavior which is to do something, it should
+  start with `no` (eg. `--no-generate-name` when the default is to generate a
+  name).
+
 #### Output
 
 Commands that output information should support `--output` with a shorthand of


conventions/cli.md
```diff
@@ -64,6 +64,13 @@ In this case `foo` is positional, and refers to the service to create.
   example, `--image=img.repo/asdf` in Knative Serving sets
   `spec.template.containers[0].image`
 
+* Flags that control a boolean behavior (eg. generate a name or not) are
+  specified by their presence. The default happens when the flag is not present,
+  and adding the flag marks the user's desire for the non-default thing. When
+  the flag *disables* a default behavior which is to do something, it should
+  start with `no` (eg. `--no-generate-name` when the default is to generate a
+  name).
+
```
rhuss: Actually, both variations (with and without `--no`) should be present in this case, regardless of the default value  (using the variant that corresponds to the default value should be possible, too). In this example, e.g. `--generate-name` and `--no-generate-name` should be both available all the time (corresponding `--generate-name=true` and `--generate-name=false`).

The reason why I prefer this variant is:

* There is no ambiguity what should be used as boolean value ("true", "yes" and "on" are commonly used as such values)
* It's shorter
* It stands in the tradition of GNU getopts
* Many boolean options are modelled like this in Unix (they are called 'flags' in this case)
* We already use that style e.g. for `kn --help` (or would we like to switch to `kn --help=true` ?). Other examples: `kn service create --force` or `kn service create --async` would have change to `kn service create --force=true` and `kn service create --async=true` (or maybe in the latter case `--sync=false`)
* For consistencies reason we would always have to require options with a value.

toVersus: > Actually, both variations (with and without --no) should be present in this case, regardless of the default value (using the variant that corresponds to the default value should be possible, too).

I have the same feeling.

duglin:
I think the most compelling argument made was:
> It stands in the tradition of GNU getopts (sometimes you see the variation --with- and --no- for true and false)
> Many boolean options are modelled like this in Unix (they are called 'flags' in this case)

The use of `..=true` and `..=false` always felt a bit odd to me.

Either PR can work though... my only really BIG sore point is if we allow `--no-xxx=false`. The double negative is just painful for people. So if we support `..=BOOL` then we should never allow the name to start with `no`.

conventions/cli.md
cppforlife: just noticed your statement here:

>We already use that style e.g. for kn --help (or would we like to switch to kn --help=true ?). Other examples: kn service create --force or kn service create --async would have change to kn service create --force=true and kn service create --async=true (or maybe in the latter case --sync=false)

`--help` continues to work as cobra defaults it to mean `true`. so full form `--flag=false` is only necessary when turning things off.

rhuss: @cppforlife Have you tried `kn --help=false` ?

Issue #210, opened 2019-06-27
Support setting up traffic splitting of service


maximilien:
With implementations of `kn route` in the works (issues #205, #201, #204), it's perhaps time to think about how to allow users to set traffic splitting on their services. 

Not only as potentially a `kn route update ...` with percentage values for different revisions, but also whether other types of traffic routing splitting is warranted. 

I know @duglin has ideas on "feature flag traffic splitting" so let's use this issue to discuss the client UX. Perhaps the MVP and the v2 and nice to have.

duglin:
I'd prefer if users didn't think about traffic splitting as a "route" manipulation, because it's not. It's a modification to a KnService resource. As an example, let's say I have this KnService:
```
apiVersion: serving.knative.dev/v1alpha1
kind: Service
metadata:
  name: echo
spec:
  template:
    spec:
      containers:
        - image: duglin/echo
  traffic:
  - tag: old
    revisionName: echo-v1
    percent: 50
  - tag: new
    revisionName: echo-v2
    percent: 50
  - tag: latest
    latestRevision: true
    percent: 0
```
I'd like to discuss how this would appear on a `kn service` command - probably under `update` but only if the syntax fits, if not then a new `verb` might be needed.

Just some brainstorming/ramblings....
- there are 3 pieces of information that are associated with each possible revisions/route: tag, revisionName and percent
- revisionName is the only one that's (sort of) static because it acts as the unique identifier for the revision we're talking about for that traffic split/route
- that leads me to think that each traffic split would need to be identified/configured by at least a single word value (revisionName), and at most a triple word value
- something like:  `echo-v1` vs `echo-v1:50` vs `echo-v1:50:old`
- the reason I think things need to be merged into a single string is because the user will need to specify all possible routes on a single command line - and trying to do some sort of correlation of values across args feels like it'll be a challenge and messy for people.

So, with that, the above KnService update command might look like:
`kn service update echo --traffic echo-v1:50:old --traffic echo-v2:50:new --traffic latest:0:latest`

Note that I think the general syntax for the arg would be:
`revisionName[:[percent][:tag]]`
- if `percent` is absent then it defaults to zero, just like serving does
- if `tag` is absent then there is not default value and there is no custom URL for that revision
[...]

Issue #232, opened 2019-07-05
Add command for tagging revisions


navidshaikh:
This is followup issue of #231, the idea is to introduce leaf command for labeling revisions.
Couple of verb choices: `tag`, `label` 
I personally like `kn revision tag`.

rhuss:
I definitley all for `tag` to avoid any confustion with `metadata.labels`

maximilien:
Yup. We had that in `knctl`. Should be also useful for how we end up doing #210.

sixolet:
Fixed by #345

PR #285, opened 2019-07-22
WIP: Adds Traffic Splitting doc

navidshaikh:
Enlisting the operations, commands and examples.

This is followup to https://github.com/knative/client/issues/210#issuecomment-513862170

### DIFF
--- a/docs/TrafficSplitting.md
+++ b/docs/TrafficSplitting.md
@@ -25,25 +25,27 @@ User can tag traffic targets (revisions) to
 
 #### Tagging commands
 
-#### 1. `kn service tag-target [RevisionName1:Tag1] [RevisionName2:Tag2]`
+#### 1. `kn service tag-target serviceName [RevisionName1:Tag1] [RevisionName2:Tag2]`
 
+ - Takes the first argument as service name to operate on
  - Tags given revision with specified tag thereby creating a custom URL
  - Does not change the traffic percent of the target
  - Multiple `RevisionName:Tag` combination can be specified
  - Multiple tags for same revision can be specified
-   - `kn service tag-target echo-v1:stable echo-v1:current`
+   - `kn service tag-target svc echo-v1:stable echo-v1:current`
  - If the target doesn't exist in deployed service, it creates a target entry with 0% traffic portion
  - For updating tags of existing traffic targets, use `kn service untag-target` followed by `kn service tag-target`, because
     - The deployed service might have multiple tags referring to a `RevisionName`
     - The specified `RevisionName:Tag` combination on CLI might request to have different tags for same revision
 
-#### 2. `kn service untag-target [RevisionName:Tag] [RevisionName [--all]] [Tag]`
+#### 2. `kn service untag-target serviceName [Tag] [RevisionName:Tag] [RevisionName [--all]]`
 
- - Takes a single reference of a traffic target
+ - Takes the first argument as service name to operate on
+ - Takes single reference of a traffic target
  - Does not change the traffic percent of the target
  - Can be specified either via
-   -  `RevisionName`, removes the tag for target where `RevisionName` matches, errors if multiple tags for a revision exists
    -  `Tag`, removes the specified tag from traffic targets
+   -  `RevisionName`, removes the tag for target where `RevisionName` matches, errors if multiple tags for a revision exists
    -  `RevisionName:Tag` removes `Tag` for `RevisionName`, useful for case where a single revision is tagged with multiple tags
    - `--all` flag can be optionally specified along with `RevisionName` to remove all tags for given revision
  - If a target has 0% traffic portion and it is untagged, target is removed from traffic block altogether
@@ -75,7 +77,7 @@ A user can
  - Referencing `$` to tag a target or set traffic percent will update target in traffic block with Revision reference to `latestRevision: true`
 
 ```
-kn service tag-target $:current
+kn service tag-target svc $:current
 ```
 will form the traffic block as below before updating service
 ```
@@ -87,7 +89,7 @@ will form the traffic block as below before updating service
 
 and,
 ```
-kn service set-traffic svc1 $:100   #OR 'kn service set-traffic svc1 current:100'
+kn service set-traffic svc $:100   #OR 'kn service set-traffic svc current:100'
 ```
 will form the traffic block
 ```
@@ -102,54 +104,54 @@ will form the traffic block
 
 1. Tag revisions `echo-v1` and `echo-v2` as `stable` and `staging`
 ```
-kn service tag-target echo-v1:stable echo-v2:staging
+kn service tag-target svc echo-v1:stable echo-v2:staging
 ```
 
 2. Ramp up/down revision echo-v3 to 20%, adjusting other traffic to accommodate:
 ```
-kn service set-traffic svc1 echo-v3=20 echo-v2=*
+kn service set-traffic svc echo-v3=20 echo-v2=*
 ```
 
 3. Give revision echo-v3 the tag candidate, without otherwise changing any traffic split:
 ```
-kn service tag-target echo-v3:candidate
+kn service tag-target svc echo-v3:candidate
 ```
 
 4. Give echo-v3 the tag candidate, and 2% of traffic adjusting other traffic to go to revision echo-v2:
 ```
-kn service tag-target echo-v3:candidate
-kn service set-traffic svc1 candidate=2 echo-v2=*
+kn service tag-target svc echo-v3:candidate
+kn service set-traffic svc candidate=2 echo-v2=*
 ```
 
 5. Give whatever revision has the tag candidate 10% of the traffic, adjusting the traffic on whatever revision has the tag current to accommodate:
 ```
-kn service set-traffic svc1 candidate=10 current=*
+kn service set-traffic svc candidate=10 current=*
 ```
 
 6. Update the tag for echo-v3 from candidate to current:
 ```
-kn service untag-target echo-v3:candidate
-kn service tag-target echo-v3 current
+kn service untag-target svc echo-v3:candidate
+kn service tag-target svc echo-v3 current
 ```
 
 7. Remove the tag current from echo-v3:
 ```
-kn service untag-target echo-v3:current
+kn service untag-target svc echo-v3:current
 ```
 
 8. Remove echo-v3 from the traffic assignments entirely, adjusting echo-v2 to fill up:
 ```
-kn service set-traffic svc1 echo-v3=0 echo-v2=*  #echo-v2 gets 100% traffic portion here
-kn service untag-target echo-v3:current
+kn service set-traffic svc echo-v3=0 echo-v2=*  #echo-v2 gets 100% traffic portion here
+kn service untag-target svc echo-v3:current
 ```
 
 9. Tag revision echo-v1 as stable and current with 50-50% traffic split
 ```
-kn service tag-target echo-v1:stable echo-v1:current
-kn service set-traffic svc1 stable=50 current=50
+kn service tag-target svc echo-v1:stable echo-v1:current
+kn service set-traffic svc stable=50 current=50
 ```
 
 10. Revert all the traffic to latest revision of service
 ```
-kn service set-traffic svc1 $=100
+kn service set-traffic svc $=100
 ```
--- a/docs/TrafficSplitting.md
+++ b/docs/TrafficSplitting.md
@@ -67,6 +67,36 @@ A user can
  - Wildcard character `*` can be specified to allocate portion remaining after, summing specified portions and substracting it from 100
  - Specified traffic allocations replace existing traffic block of deployed service
 
+#### Referencing `latestRevision:true` (`$`) target
+ - Serving allows having `latestRevision:true` field in traffic block as traffic target
+ - This field indicates that the latest ready Revision of the Configuration should be used for this traffic target
+ - We can refer this as placeholder to point to the latest Revision which will be generated for the Service
+ - To reference this special target in traffic block, we can use character `$`, which can not be a Revision name to avoid conflicts
+ - Referencing `$` to tag a target or set traffic percent will update target in traffic block with Revision reference to `latestRevision: true`
+
+```
+kn service tag-target $:current
+```
+will form the traffic block as below before updating service
+```
+  targets:
+  - latestRevision: true
+    tag: current
+    percent: xx
+```
+
+and,
+```
+kn service set-traffic svc1 $:100   #OR 'kn service set-traffic svc1 current:100'
+```
+will form the traffic block
+```
+  targets:
+  - latestRevision: true
+    tag: current
+    percent: 100
+```
+
 
 ## Examples:
 
@@ -117,4 +147,9 @@ kn service untag-target echo-v3:current
 ```
 kn service tag-target echo-v1:stable echo-v1:current
 kn service set-traffic svc1 stable=50 current=50
+```
+
+10. Revert all the traffic to latest revision of service
+```
+kn service set-traffic svc1 $=100
 ```
--- a/docs/TrafficSplitting.md
+++ b/docs/TrafficSplitting.md
@@ -1,0 +1,120 @@
+# Traffic Splitting
+
+The purpose of this section of the kn documentation is to describe the command verbs for traffic splitting operations and present some examples.
+
+`kn` client can work with following properties of a traffic target defined in traffic block of a service spec.
+ - **RevisionName**: RevisionName of a specific revision to which to send a portion of traffic.
+ - **Tag**: Tag is optionally used to expose a dedicated url for referencing this target exclusively.
+ - **Percent**: Percent specifies percent of the traffic to this Revision.
+
+## Operations
+
+Operations required to perform traffic splitting use cases, for example: blue/green deployment, etc can be summarized as below:
+ - Tag targets
+ - Split traffic
+
+### Tag Targets
+
+User can tag traffic targets (revisions) to
+ - Create custom URLs
+ - Create a human-friendly reference to refer to a traffic target
+ - Create static routes and change the underlying revisions as needed
+ - Remove the custom URLs OR remove the tags of targets
+ - Remove a target from traffic block altogether
+
+
+#### Tagging commands
+
+#### 1. `kn service tag-target [RevisionName1:Tag1] [RevisionName2:Tag2]`
+
+ - Tags given revision with specified tag thereby creating a custom URL
+ - Does not change the traffic percent of the target
+ - Multiple `RevisionName:Tag` combination can be specified
+ - Multiple tags for same revision can be specified
+   - `kn service tag-target echo-v1:stable echo-v1:current`
+ - If the target doesn't exist in deployed service, it creates a target entry with 0% traffic portion
+ - For updating tags of existing traffic targets, use `kn service untag-target` followed by `kn service tag-target`, because
+    - The deployed service might have multiple tags referring to a `RevisionName`
+    - The specified `RevisionName:Tag` combination on CLI might request to have different tags for same revision
+
+#### 2. `kn service untag-target [RevisionName:Tag] [RevisionName [--all]] [Tag]`
+
+ - Takes a single reference of a traffic target
+ - Does not change the traffic percent of the target
+ - Can be specified either via
+   -  `RevisionName`, removes the tag for target where `RevisionName` matches, errors if multiple tags for a revision exists
+   -  `Tag`, removes the specified tag from traffic targets
+   -  `RevisionName:Tag` removes `Tag` for `RevisionName`, useful for case where a single revision is tagged with multiple tags
+   - `--all` flag can be optionally specified along with `RevisionName` to remove all tags for given revision
+ - If a target has 0% traffic portion and it is untagged, target is removed from traffic block altogether
+
+
+## Split Traffic
+
+A user can
+ - Specify the traffic percent portion to route to a particular target
+ - Adjust the traffic percent of existing traffic targets
+ - Reference a traffic target either by `RevisionName` or `Tag`
+
+### Traffic splitting command
+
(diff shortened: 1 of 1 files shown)


docs/TrafficSplitting.md
```diff
@@ -0,0 +1,120 @@
...
+ - Create custom URLs
+ - Create a human-friendly reference to refer to a traffic target
+ - Create static routes and change the underlying revisions as needed
+ - Remove the custom URLs OR remove the tags of targets
+ - Remove a target from traffic block altogether
+
+
+#### Tagging commands
+
+#### 1. `kn service tag-target [RevisionName1:Tag1] [RevisionName2:Tag2]`
```
duglin: It's not clear to me how this works when there could be more than one traffic block with the specified revisionName. How does this work when there is no revisionName on the block - just `latestRevision:true`? Asking the user to look at the status section to find the current revisionName in that block will be cumbersome.

docs/TrafficSplitting.md
```diff
@@ -0,0 +1,120 @@
+# Traffic Splitting
+
+The purpose of this section of the kn documentation is to describe the command verbs for traffic splitting operations and present some examples.
+
+`kn` client can work with following properties of a traffic target defined in traffic block of a service spec.
+ - **RevisionName**: RevisionName of a specific revision to which to send a portion of traffic.
+ - **Tag**: Tag is optionally used to expose a dedicated url for referencing this target exclusively.
+ - **Percent**: Percent specifies percent of the traffic to this Revision.
```
duglin: Just wanted to point out some key things in case people are unaware:
- The `tag` property on a traffic block is optional.  This means that we CAN NOT identify a traffic block by this value, since not all blocks will have one.
- The `revisionName` property on a traffic block may not be unique. More than one traffic block can refer to the same revisionName. This means that we CAN NOT identify a traffic block by this value.
- Additionally, since people can use `latestRevision: true`, there is no consistent revisionName to search on for this block (it'll change each time a new revision is created) - which requires the user to query the service each time they want to modify this traffic block to see if the value changed.  Otherwise, per this PR, a new block will be create w/o them realizing it (I think).

IMO this means any operation that tries to modify an existing traffic block probably can't because there's no

docs/TrafficSplitting.md
```diff
@@ -0,0 +1,120 @@
...
+ - Does not change the traffic percent of the target
+ - Multiple `RevisionName:Tag` combination can be specified
+ - Multiple tags for same revision can be specified
+   - `kn service tag-target echo-v1:stable echo-v1:current`
+ - If the target doesn't exist in deployed service, it creates a target entry with 0% traffic portion
+ - For updating tags of existing traffic targets, use `kn service untag-target` followed by `kn service tag-target`, because
+    - The deployed service might have multiple tags referring to a `RevisionName`
+    - The specified `RevisionName:Tag` combination on CLI might request to have different tags for same revision
+
+#### 2. `kn service untag-target [RevisionName:Tag] [RevisionName [--all]] [Tag]`
```
duglin: Since 'tag' must be unique across all traffic blocks, can't you just make it:
`kn service untag-target TagName` ?

docs/TrafficSplitting.md
duglin: And checking the status section is racey

docs/TrafficSplitting.md
```diff
@@ -0,0 +1,120 @@
...
+ - Adjust the traffic percent of existing traffic targets
+ - Reference a traffic target either by `RevisionName` or `Tag`
+
+### Traffic splitting command
+
+#### `kn service set-traffic serviceName [revisionName1=Percent] [Tag2=Percent] [Tag3=*]`
+
+ - Takes the first argument as service name to operate on
+ - Subsequent arguments specifies the traffic targets
+ - Traffic target can be referenced either via `RevisionName` or `Tag`
```
duglin: This does not guarantee a unique look-up of the block in question

duglin:
Repeating here:

Some key things in case people are unaware:

-    The tag property on a traffic block is optional. This means that we CAN NOT identify a traffic block by this value, since not all blocks will have one.
-    The revisionName property on a traffic block may not be unique. More than one traffic block can refer to the same revisionName. This means that we CAN NOT identify a traffic block by this value.
-    Additionally, since people can use latestRevision: true, there is no consistent revisionName to search on for this block (it'll change each time a new revision is created) - which requires the user to query the service each time they want to modify this traffic block to see if the value changed. Otherwise, per this PR, a new block will be create w/o them realizing it (I think).

IMO this means any operation that tries to modify an existing traffic block probably can't because there's no guaranteed way to find what you're looking for (in some cases) and in others it requires extra work, which is also racey.

- - -

I'm saying all of this given the current state of what's defined in Serving. If we want to suggest changes to serving to make things easier then that will change things. But I'm trying to see what can work given what we have today, and I don't see a good way to do any kind of reliable (and non-racey) searching to find a specific traffic target to make any modifications.

That's why I suggested: https://github.com/knative/client/issues/210#issuecomment-506141432

docs/TrafficSplitting.md
rhuss: > It's not clear to me how this works when there could be more than one traffic block with the specified revisionName.

You can repeat the same revision name multiple times

> How does this work when there is no revisionName on the block - just `latestRevision:true`

True, that might be an issue. So I would recommend a special revision name for specifying the target referenced with latestResvision. Something which is otherwise an illegal revision name, like "$" or "+".

BTW, a user can always find out all revision names attached to a service via `kn service describe`

docs/TrafficSplitting.md
rhuss: wdym ? Because a tag can be named like a revision ? I consider this case super seldom and we should verify this and throw an error, if then the tag and revision differs (as said (a) its very unlikely that a revision and tag will ever have the same name and (b) if this should be the case its very likely that it specifies the same block).

If you mean that when the user specifies a tag or revision which is not present, then of course again an error should be thrown.

So IMO it does guarantee a unique lookup (to 99.99999%) and simplifies the UI considerably.

docs/TrafficSplitting.md
rhuss: > * The `tag` property on a traffic block is optional.  This means that we CAN NOT identify a traffic block by this value, since not all blocks will have one.

True, therefore a user can use revision or tag  (or both in case of multiple revisions being present). As mentioned above we might need to add a special name to target the "latestRevision" traffic block.

> * The `revisionName` property on a traffic block may not be unique. More than one traffic block can refer to the same revisionName. This means that we CAN NOT identify a traffic block by this value.

But the combination tag:revision is unique. That's how you would need to address such a block.

docs/TrafficSplitting.md
duglin: > >    It's not clear to me how this works when there could be more than one traffic block with the specified revisionName.

> You can repeat the same revision name multiple times

I wasn't clear... the same revisionName might appear on 5 traffic blocks but I want to set a tag on just one of them....how do I specify that one of the 5?

docs/TrafficSplitting.md
duglin: only if tag is present - which it may not be because it's optional.  I think having a different traffic-block-picking syntax based on the data in the traffic sections will be a bit confusing.
We need a cmd that works consistently.

docs/TrafficSplitting.md
duglin: I mean, since the revisionName is not unique, `[revisionName1=Percent]` is indeterminate.

rhuss: * If the revisionName appears once, the operation is performed
* If the revisionName appears 0 times, an error is thrown
* If the revisionName appears more than once, an error is thrown and the user is hinted that she has to use a tag (multiple revisions without tags doesn't make any sense, or ?)

docs/TrafficSplitting.md
navidshaikh: > It's not clear to me how this works when there could be more than one traffic block with the specified revisionName.

As Roland mentioned, just repeat the same revision multiple times.

> How does this work when there is no revisionName on the block - just latestRevision:true

I think referencing the entry in traffic block having `latestRevsion:true` is technically confusing, as with every update to service (with `traffic-tag` or `set-traffic` commands) there will be a new revision generated, so which revision should be referenced as `latestRevision` ? The one which was there before issuing the command OR the one that will be created after issuing the command?
I think here, we want the traffic block entry referencing the most latest revision for service (after issuing the current command), and for that we need to reference that via a word or character, which should translate to `latestRevision:true` field in traffic block, but not use the `latestRevision` as is in the CLI (because that could be confusing).

navidshaikh: > I wasn't clear... the same revisionName might appear on 5 traffic blocks but I want to set a tag on just one of them....how do I specify that one of the 5?

@duglin 
I think you are saying we need a service reference in that tag command ?

duglin: hmmm I wasn't but I think you might be right about that. Should the service name appear in the cmd ? Or are you assuming that the prefix of the revision name can be used to find the service?  Not sure it can if there are dashes in the service name or revision postfix string. Even if it could I just wonder if that's a confusing UX since most people might expect the cmd to be: `kn service NAME tag-target ...`. This might be where we could use just the unique part of the revision name instead of the full revision name, like we talked about in another thread.

e.g. `kn service echo tag-target v1:sometag` instead of
`kn tag-target echo-v1:sometag` or
`kn service echo tag-target echo-v1:sometag`

docs/TrafficSplitting.md
navidshaikh: > only if tag is present - which it may not be because it's optional.

Yes, and which gives us opportunity to design UX to remove the ambiguities. 

> I think having a different traffic-block-picking syntax based on the data in the traffic sections will be a bit confusing.

yes, and this is why we need to give users a way to not confuse them. We should ask users to create unique identity for ambiguous targets, which is `Tag`. We ask users to create `Tag` for a revision if it has to present more than once in traffic block. So that sub-subsequent operations can be performed without ambiguity. `Tag` is optional, I agree, but we can leverage it for identifying a target uniquely.

There are a bunch of combinations, given that serving traffic block allows fields as optional and aliases and I think we should work towards making the client UX easy to understand and execute for users, and defining the path to reach to a desired state.

docs/TrafficSplitting.md
navidshaikh: > are you assuming that the prefix of the revision name can be used to find the service? 

No. We can find the service name of the revision by lookup in revision's labels as
https://github.com/knative/client/blob/9513dedfbae45d38df80ff903da67f569b926c8d/pkg/kn/commands/revision/human_readable_flags.go#L58

this is how we list the service for revision in `kn revision list` command and thats why `kn service tag-target` command was proposed with requiring to refer a service name as we can always figure it out.

duglin: ah I see. I do wonder though if from a user's perspective that's confusing because the main resource we're trying to get them to work on is a "service" and yet we don't ask them to provide that. Yes we may not need it, but will a user need it in their mental model they're dealing with. I don't ever expect to expose my users to "revisions".  Just something to think about.

navidshaikh: Sure, if we see the `kn service set-traffic` command, it requires a service name. We could just bring the same experience for `kn service tag-target` command, i.e. requiring the service name as first positional argument (this also saves the lookup for service name and we can then verify that requested revisions to be tagged belongs to the service requested to operate on). 

@rhuss : WDYT about requiring a service name for `kn service tag-target` command as first positional argument ?

docs/TrafficSplitting.md
navidshaikh: Yes, we can implement that as first iteration to require only `TagName`. The subsequent iteration though should support the other options too.

docs/TrafficSplitting.md
rhuss: (just updated my comment above which overlooked a case)

docs/TrafficSplitting.md
rhuss: According to our CLI conventions spec we should *always* use the name of the _noun_ entity as the first argument, so +1 for using the service name here.

I'm wondering whether we could shorten `tag-target` to `tag` as I think `target` does not add much info in addition (and `tag` then just means tag within the route)

navidshaikh: > although there is `AGE` column but we can ease the lookup by printing the numbers.

:+1: I will update the PR for requiring service name for `tag-target`.

> I'm wondering whether we could shorten tag-target to tag as I think target does not add much info in addition (and tag then just means tag within the route)

IMO, `tag-target` is more explicit specifying its the tag for the traffic target and not the tag for service itself.  Also `tag` word is used to label the resources at other places for example ` knctl`, where `knctl revision tag ` will **label** the revision object as specified. We also have an open issue https://github.com/knative/client/issues/232 on the same line.

navidshaikh:
> The tag property on a traffic block is optional. This means that we CAN NOT identify a traffic block by this value, since not all blocks will have one.
The revisionName property on a traffic block may not be unique. More than one traffic block can refer to the same revisionName. This means that we CAN NOT identify a traffic block by this value.
Additionally, since people can use latestRevision: true, there is no consistent revisionName to search on for this block (it'll change each time a new revision is created) - which requires the user to query the service each time they want to modify this traffic block to see if the value changed. Otherwise, per this PR, a new block will be create w/o them realizing it (I think).
IMO this means any operation that tries to modify an existing traffic block probably can't because there's no guaranteed way to find what you're looking for (in some cases) and in others it requires extra work, which is also racey.

Replied at https://github.com/knative/client/pull/285#discussion_r306699172

>I'm saying all of this given the current state of what's defined in Serving. If we want to suggest changes to serving to make things easier then that will change things. But I'm trying to see what can work given what we have today, and I don't see a good way to do any kind of reliable (and non-racey) searching to find a specific traffic target to make any modifications.

The other cases where revisionName could be present multiple times is covered (check above comments).

The only case we need to cover is `latestRevision:true`, which I think, we should refer as **placeholder** field for referring to a particular (always latest) revision at current state of service, than actual revision reference. (because before and after issuing the command `set-traffic` or `tag-target` or even `service update`, that field's reference will be updated by serving any way)

So we use that field to say, for e.g. whatever the latest revision of this service will be, give that a tag `current` and 10% traffic.
 **We can identify that field by a character which can not be revision name**, for e.g. `$` (as it denotes to last character in regexp, so some similarity as last generated revision for this service).
You could then say
```
kn service tag-target $:current echo-v2:stable
kn service set-traffic current=10 echo-v2=90
```
which should form the final spec of the traffic block of the service as
```
[..]
    traffic:
    - tag: stable
      RevisionName: echo-v2
      percent: 90
    - tag: latest
      latestRevision: true
      percent: 10
```

@duglin : Does this make sense ? 

I think we covered the cases you mentioned.

duglin:
@navidshaikh what do you think about supporting this type of flow:
```
$ kn service create myapp --image echo:v1 --revision-name=v1
$ kn service update myapp --image echo:v2 --revision-name=v2 --traffic=v1:90 --traffic=v2:10
```
I think think type of flow might be popular so forcing them to do it in more commands might be cumbersome.

docs/TrafficSplitting.md
navidshaikh: I have updated the PR to require `serviceName` as first argument for tag and untag commands.

navidshaikh:
> I think think type of flow might be popular so forcing them to do it in more commands might be cumbersome.

@duglin:
Technically, the operations (flags) on the `service update` command updates the `Configuration` of the service, thereby creating a new `Revision`,  while updating the traffic block of the service doesn't update the `Configuration` and thus doesn't create a new `Revision`. We should have a separate path for what we're updating.

duglin:
don't think of from an inner-Kn perspective, think of it from a user perspective. They just created a new revision and want the routing to take effect immediately.

duglin:
But, if you really want to ... :-)  I can do exactly that `kn` command via a single `kubectl` command - so it seems natural to me that `kn` could match that functionality - just in a non-yaml way

rhuss:
> don't think of from an inner-Kn perspective, think of it from a user perspective. They just created a new revision and want the routing to take effect immediately.

I would love to avoid the multiplied `--traffic` args, which makes it much easier from a _user perspective_

```
kn service set-traffic myservice myservice-v1=10% myservice-v2=30% myservice-backup=*
```

vs

```
kn service update myservice --traffic myservice-v1=10% --traffic myservice-v2=30% --traffic myservice-backup=*
```

duglin:
the syntax is less important to me than being able to do it all in one cmd :-)  so if dropping the flags works, that's fine.

But keep in mind, we may need to allow people to specify it on a `kn service create` so we'd need consistency there.  Not sure yet about the % splitting yet since we just have one revision, but definitely for tagging.

Another thing to remember in all of this... today we have % splitting, we need to make sure whatever design we come up with allows for splitting on something else at some point. See https://github.com/knative/serving/issues/4736

docs/TrafficSplitting.md
```diff
@@ -0,0 +1,157 @@
...
+   - `--all` flag can be optionally specified along with `RevisionName` to remove all tags for given revision
+ - If a target has 0% traffic portion and it is untagged, target is removed from traffic block altogether
+
+
+## Split Traffic
+
+A user can
+ - Specify the traffic percent portion to route to a particular target
+ - Adjust the traffic percent of existing traffic targets
+ - Reference a traffic target either by `RevisionName` or `Tag`
```
duglin: We should mention that in the future people may be able to specify something other than "percent" as the criteria, so we need to design a syntax that doesn't preclude us from supporting that later on.

Since right now I believe all of our proposals have done something of the form `...:%` - then as long as the `%` specification is at the end of the syntax then I think we might be ok. We can then assume that whatever string is after the `:` is the criteria - even if it's a complex expression. We may need to allow for it to be of the form `type:expression` (e.g. `path:/foo` or `header:user=john`) but I think that's workable.  So, nothing needs to be changed w.r.t. any proposal (except the comment above to mention it) - I just wanted to bring this up so people are thinking about it.

sixolet:
Ping; Navid do you want to re-do this with the decisions we've made, or shall we start a different PR?

navidshaikh:
@sixolet : I think we'll need a different PR, the doc should be more of doc-style guide than a proposal-style one.

docs/TrafficSplitting.md
```diff
@@ -0,0 +1,157 @@
+# Traffic Splitting
+
+The purpose of this section of the kn documentation is to describe the command verbs for traffic splitting operations and present some examples.
+
+`kn` client can work with following properties of a traffic target defined in traffic block of a service spec.
+ - **RevisionName**: RevisionName of a specific revision to which to send a portion of traffic.
```
maximilien: Nit.. Do we need to repeat the word when it's already bolded and delimits the paragraph :) ?

docs/TrafficSplitting.md
```diff
@@ -0,0 +1,157 @@
...
+ - **Tag**: Tag is optionally used to expose a dedicated url for referencing this target exclusively.
+ - **Percent**: Percent specifies percent of the traffic to this Revision.
+
+## Operations
+
+Operations required to perform traffic splitting use cases, for example: blue/green deployment, etc can be summarized as below:
+ - Tag targets
```
maximilien: I would have proposed names for these and small explanation if possible. Since one goal here is to see what other operations people reading this spec can come up with...

docs/TrafficSplitting.md
```diff
@@ -0,0 +1,157 @@
...
+#### `kn service set-traffic serviceName [revisionName1=Percent] [Tag2=Percent] [Tag3=*]`
+
+ - Takes the first argument as service name to operate on
+ - Subsequent arguments specifies the traffic targets
+ - Traffic target can be referenced either via `RevisionName` or `Tag`
+ - Traffic portion to allocate is specified after `=`
+ - Wildcard character `*` can be specified to allocate portion remaining after, summing specified portions and substracting it from 100
+ - Specified traffic allocations replace existing traffic block of deployed service
+
+#### Referencing `latestRevision:true` (`$`) target
```
maximilien: Not a huge fan of the `$` here... Maybe there is a precedence in other CLIs or reasoning. Can you elucidate this choice @navidshaikh? Thanks.

Not a huge concern if there is precedence. If not, then `*latestRevision` or `&latestRevision` could be perhaps more descriptive?

rhuss: `$` is a take on regexps where it means 'the end of a line' I guess (like in 'the end of a line of revisions') ;-). A bit made up, we could definitely use something different. However we should choose best something which does not has a meaning for a shell. E.g '*latestRevision' would expand to a filename if in the current directory is a file like 'mylatestRevision' before it's handed over to `kn`. `&` will put the process in the background if unescaped. `$` alone is safe but when combined with another character it specifies a variable. 

I'm afraid regardless of what we chose we won't find something so intuitive that it's obvious without reading the documentation.

I think `$` is ok-ish but maybe we find something better ?

docs/TrafficSplitting.md
rhuss: Would also rephrase to '**RevisionName**: Name of a specific revision ...'

zhanggbj:
@maximilien Sure, no problem for me. BTW, @navidshaikh @toVersus and I planned to raise a new doc PR with the voting result and discussions here, just started to break down to pieces so we can collaborate, thx!

docs/TrafficSplitting.md
navidshaikh: sure, we'll have this incorporated in a different PR.

docs/TrafficSplitting.md
navidshaikh: one of the suggestion Naomi mentioned during WG meeting is to use flag `--traffic-latest` which should refer the `latestRevision` field in traffic block.

We'll need an identifier for (a) traffic (b) tag operations, we can use
`--traffic-latest` and `--tag-latest` flags, for eg:
1. `--traffic-latest 2`
2. `--tag-latest current`

navidshaikh:
Thank you all for review and feedback, since the feedback required re-write of the doc in this PR, here is the new PR https://github.com/knative/client/pull/331 with additional description of the flags, operation precedence and more examples, PTAL.

Closing this now.

PR #291, opened 2019-07-24
Add E2E tests for Service, Revision, Route

mgencur:
* service, revision, route describe and describe with print flags
* route list with print flags
* service create for a duplicate service

Fixes #

## Proposed Changes

* increase test coverage via new E2E tests

**Release Note**

```release-note

```

### DIFF
--- a/test/e2e/basic_workflow_test.go
+++ b/test/e2e/basic_workflow_test.go
@@ -77,6 +77,10 @@ func TestBasicWorkflow(t *testing.T) {
 		test.serviceDelete(t, "svc2")
 	})
 
+	t.Run("delete hello service again and get an error", func(t *testing.T) {
+		test.serviceDeleteNonexistent(t, "hello")
+	})
+
 	t.Run("returns no service after completing tests", func(t *testing.T) {
 		test.serviceListEmpty(t)
 	})
@@ -92,8 +96,7 @@ func (test *e2eTest) serviceListEmpty(t *testing.T) {
 }
 
 func (test *e2eTest) serviceCreate(t *testing.T, serviceName string) {
-	out, err := test.kn.RunWithOpts([]string{"service", "create",
-		fmt.Sprintf("%s", serviceName),
+	out, err := test.kn.RunWithOpts([]string{"service", "create", serviceName,
 		"--image", KnDefaultTestImage}, runOpts{NoNamespace: false})
 	assert.NilError(t, err)
 
@@ -111,8 +114,7 @@ func (test *e2eTest) serviceCreateDuplicate(t *testing.T, serviceName string) {
 	out, err := test.kn.RunWithOpts([]string{"service", "list", serviceName}, runOpts{NoNamespace: false})
 	assert.Check(t, strings.Contains(out, serviceName), "The service does not exist yet")
 
-	_, err = test.kn.RunWithOpts([]string{"service", "create",
-		fmt.Sprintf("%s", serviceName),
+	_, err = test.kn.RunWithOpts([]string{"service", "create", serviceName,
 		"--image", KnDefaultTestImage}, runOpts{NoNamespace: false, AllowError: true})
 
 	assert.ErrorContains(t, err, "the service already exists")
@@ -158,7 +160,7 @@ func (test *e2eTest) serviceDescribeWithPrintFlags(t *testing.T, serviceName str
 	assert.NilError(t, err)
 
 	expectedName := fmt.Sprintf("service.serving.knative.dev/%s", serviceName)
-	assert.Check(t, strings.Contains(out, expectedName))
+	assert.Equal(t, strings.TrimSpace(out), expectedName)
 }
 
 func (test *e2eTest) routeList(t *testing.T) {
@@ -184,6 +186,16 @@ func (test *e2eTest) serviceDelete(t *testing.T, serviceName string) {
 	assert.Check(t, util.ContainsAll(out, "Service", serviceName, "successfully deleted in namespace", test.kn.namespace))
 }
 
+func (test *e2eTest) serviceDeleteNonexistent(t *testing.T, serviceName string) {
+	out, err := test.kn.RunWithOpts([]string{"service", "list", serviceName}, runOpts{NoNamespace: false})
+	assert.Check(t, !strings.Contains(out, serviceName), "The service exists")
+
+	_, err = test.kn.RunWithOpts([]string{"service", "delete", serviceName}, runOpts{NoNamespace: false, AllowError: true})
+
+	expectedErr := fmt.Sprintf(`services.serving.knative.dev "%s" not found`, serviceName)
+	assert.ErrorContains(t, err, expectedErr)
+}
+
 func (test *e2eTest) routeDescribe(t *testing.T, routeName string) {
 	out, err := test.kn.RunWithOpts([]string{"route", "describe", routeName}, runOpts{})
 	assert.NilError(t, err)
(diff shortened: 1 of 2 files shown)


test/e2e/basic_workflow_test.go
```diff
@@ -132,6 +153,14 @@ func (test *e2eTest) serviceUpdate(t *testing.T, serviceName string, args []stri
 	assert.Check(t, util.ContainsAll(out, expectedOutput))
 }
 
+func (test *e2eTest) serviceDescribeWithPrintFlags(t *testing.T, serviceName string) {
+	out, err := test.kn.RunWithOpts([]string{"service", "describe", serviceName, "-o=name"}, runOpts{})
+	assert.NilError(t, err)
+
+	expectedName := fmt.Sprintf("service.serving.knative.dev/%s", serviceName)
+	assert.Check(t, strings.Contains(out, expectedName))
```
navidshaikh: this should be an exact string match IMO, i..e `assert.Equal(t, out, expectedName)`

test/e2e/basic_workflow_test.go
```diff
@@ -154,3 +183,29 @@ func (test *e2eTest) serviceDelete(t *testing.T, serviceName string) {
...
+	expectedServiceLabel := fmt.Sprintf("serving.knative.dev/service: %s", routeName)
+	assert.Check(t, util.ContainsAll(out, expectedGVK, expectedNamespace, expectedServiceLabel))
+}
+
+func (test *e2eTest) routeDescribeWithPrintFlags(t *testing.T, routeName string) {
+	out, err := test.kn.RunWithOpts([]string{"route", "describe", routeName, "-o=name"}, runOpts{})
+	assert.NilError(t, err)
+
+	expectedName := fmt.Sprintf("route.serving.knative.dev/%s", routeName)
+	assert.Check(t, strings.Contains(out, expectedName))
```
navidshaikh: same as above about exact name matching

test/e2e/basic_workflow_test.go
navidshaikh: as this is not string manipulation and printing on the client side, but machine readable output

test/e2e/basic_workflow_test.go
```diff
@@ -93,8 +104,18 @@ func (test *e2eTest) serviceList(t *testing.T, serviceName string) {
 	out, err := test.kn.RunWithOpts([]string{"service", "list", serviceName}, runOpts{NoNamespace: false})
 	assert.NilError(t, err)
 
-	expectedOutput := fmt.Sprintf("%s", serviceName)
-	assert.Check(t, util.ContainsAll(out, expectedOutput))
+	assert.Check(t, util.ContainsAll(out, serviceName))
+}
+
+func (test *e2eTest) serviceCreateDuplicate(t *testing.T, serviceName string) {
```
navidshaikh: IMO, this particular test case doesn't belong to `basic_workflow_test` as the filename suggest, but kind of testing for error handling mechanism, we should probably create another workflow which specifically aims at error handling test cases.

test/e2e/revision_workflow_test.go
```diff
@@ -33,26 +34,57 @@ func TestRevisionWorkflow(t *testing.T) {
...
+}
+
+func (test *e2eTest) revisionDescribeWithPrintFlags(t *testing.T, serviceName string) {
+	revName := test.findRevision(t, serviceName)
+
+	out, err := test.kn.RunWithOpts([]string{"revision", "describe", revName, "-o=name"}, runOpts{})
+	assert.NilError(t, err)
+
+	expectedName := fmt.Sprintf("revision.serving.knative.dev/%s", revName)
+	assert.Check(t, strings.Contains(out, expectedName))
```
navidshaikh: same comment as above, just check for exact match

test/e2e/basic_workflow_test.go
```diff
@@ -38,9 +38,14 @@ func TestBasicWorkflow(t *testing.T) {
 		test.serviceCreate(t, "hello")
 	})
 
+	t.Run("create hello service again and get service already exists error", func(t *testing.T) {
+		test.serviceCreateDuplicate(t, "hello")
```
maximilien: Nice nice, I wonder if we should also try to delete a service that’s already deleted? I would expect an error in case where service is already deleted (non-existent) and if the service is being deleted. Of course in the end the API should do the right thing and the client should return a “sensible” error.

test/e2e/basic_workflow_test.go
maximilien: If you all decide to move this to a more thorough IT (I am ambivalent at this point), please do the same for suggestion above on delete service.

test/e2e/basic_workflow_test.go
mgencur: Yeah. Good point. I can add a test for deleting a non-existent service to this PR.

mgencur:
hey! Thanks for the reviews. I have fixed all things except moving the test. I also added one more test that checks deleting a non-existent service. 
Regarding moving the test from the basic workflow, I spoke with @navidshaikh and I will shortly create an issue with a proposal for restructuring the test suite a little bit. When we agree on what to do we can move the test(s) to appropriate locations.

mgencur:
@navidshaikh the failure doesn't seem to be related to my changes. Can you advise what to do, pls?

rhuss:
I'm not sure about that flake, because I thought it was fixed with #271 . It might be that we have to increase the timeout for the wait (as if there's a timeout, then the previous update command returns prematurely).

I just triggered a retest. Let's see how it goes ...

mgencur:
Passed this time:)

maximilien:
@mgencur not quite...

```bash
Running 'kubectl delete namespace kne2etests0'...
--- FAIL: TestBasicWorkflow (126.82s)
    --- PASS: TestBasicWorkflow/returns_no_service_before_running_tests (0.05s)
    --- PASS: TestBasicWorkflow/create_hello_service_and_returns_no_error (24.58s)
    --- PASS: TestBasicWorkflow/create_hello_service_again_and_get_service_already_exists_error (0.12s)
    --- PASS: TestBasicWorkflow/returns_valid_info_about_hello_service (0.14s)
    --- PASS: TestBasicWorkflow/update_hello_service's_configuration_and_returns_no_error (0.11s)
    --- PASS: TestBasicWorkflow/create_another_service_and_returns_no_error (2.99s)
    --- FAIL: TestBasicWorkflow/returns_a_list_of_revisions_associated_with_hello_and_svc2_services (0.13s)
        basic_workflow_test.go:132: assertion failed: 
            Actual output: hello-kxgvk   hello     3s    1 OK / 5     Unknown   Deploying
            Missing strings: True
...
```

Might be flaky and timing to allow status to update.

test/e2e/basic_workflow_test.go
navidshaikh: should be addressed as part of resolution to https://github.com/knative/client/issues/301

test/e2e/basic_workflow_test.go
```diff
@@ -132,6 +155,14 @@ func (test *e2eTest) serviceUpdate(t *testing.T, serviceName string, args []stri
 	assert.Check(t, util.ContainsAll(out, expectedOutput))
 }
 
+func (test *e2eTest) serviceDescribeWithPrintFlags(t *testing.T, serviceName string) {
+	out, err := test.kn.RunWithOpts([]string{"service", "describe", serviceName, "-o=name"}, runOpts{})
+	assert.NilError(t, err)
+
+	expectedName := fmt.Sprintf("service.serving.knative.dev/%s", serviceName)
```
navidshaikh: this error message needs to be fixed on kn side, I will have a PR for this fix.

PR #299, opened 2019-07-24
Move client conventions doc from Serving

sixolet:
This moves client_conventions.md from knative/serving/doc to here. After this is merged I expect to make a PR over in Serving to change that doc to a link back over here.

I also did a pass to account for bring-your-own revision names, as they're more convenient and to-the-point than nonces.

### DIFF
--- a/conventions/client.md
+++ b/conventions/client.md
@@ -1,0 +1,195 @@
+# Client Conventions
+
+This document describes conventions that Knative domain-specific clients can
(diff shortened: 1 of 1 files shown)


rhuss:
Do we need to put this into a new top-level dir ? Or couldn't `/conventions` be a subdir of `/docs` ? Maybe call it even `specs` or `proposals` so that it might not only host a single file.

PR #300, opened 2019-07-25
test(service): Add e2e tests for min, max scale options

### LINKED ISSUE #254: Add integration tests for min/max scale options of service create and update
For service create and service update, we are missing coverage for following options
```
```

navidshaikh:
/lint

Fixes #254 

## Proposed Changes
* add e2e tests for min, max scale options for `service create` and `service update` operations.


test/e2e/service_options_test.go
```diff
@@ -49,6 +49,30 @@ func TestServiceOptions(t *testing.T) {
 	t.Run("delete hello service and returns no error", func(t *testing.T) {
 		test.serviceDelete(t, "hello")
 	})
+
+	t.Run("create hello service with min/max scale options and returns no error", func(t *testing.T) {
+		test.serviceCreateWithOptions(t, "hello", []string{"--min-scale", "1", "--max-scale", "3"})
+	})
+
+	t.Run("validate min scale value for service", func(t *testing.T) {
+		test.validateServiceMinScale(t, "hello", "1")
```
rhuss: I would add the validation step after the service call itself but in the same subtest, to make it not that fine granular.

rhuss:
@navidshaikh I highly suspect that the inegration test error is because the `kn delete` right before that step is async and the route as dependenct object is not yet deleted when you create a new service. 

For now, I'd suggest to just rename the service from "hello" to something like "hello-2"

navidshaikh:
@rhuss : Thanks Roland, I've renamed the service.

test/e2e/service_options_test.go
navidshaikh: fixed it, thanks!

navidshaikh:
Failure of `pull-knative-client-integration-tests-latest-release` job was because it was returning the deprecated fields while we were after new fields https://github.com/knative/client/pull/302#issuecomment-515297563.

Updated PR to fallback to work with old and new fields for jsonpath lookup.

rhuss:
@googlebot rescan

PR #302, opened 2019-07-25
improves(e2e): Update concurrency tests

navidshaikh:
/lint

## Proposed Changes

* updates the validation of concurrency limits/targets with upcoming change to `service describe` output, uses `service list` with `jsonpath` output format
* validation commands now return specific field in output than returning complete output

### DIFF
--- a/test/e2e/service_options_test.go
+++ b/test/e2e/service_options_test.go
@@ -17,7 +17,6 @@
 package e2e
 
 import (
-	"fmt"
 	"testing"
 
 	"github.com/knative/client/pkg/util"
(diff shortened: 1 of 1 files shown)


navidshaikh:
interesting! `NotOwned: There is an existing Route "hello" that we do not own.`
excerpts from https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_client/300/pull-knative-client-integration-tests-latest-release/1154332359860948996
```
Running 'kn service delete hello -n kne2etests2'...
=== RUN   TestServiceOptions/create_and_validate_service_'hello'_with_min/max_scale_options_
Running 'kn service create hello --image gcr.io/knative-samples/helloworld-go --min-scale 1 --max-scale 3 -n kne2etests2'...
Failed to successfully execute 'kn service create hello --image gcr.io/knative-samples/helloworld-go --min-scale 1 --max-scale 3 -n kne2etests2': Execution error: stderr: 'NotOwned: There is an existing Route "hello" that we do not own.
```

navidshaikh:
quote mismatch
```
        service_options_test.go:75: assertion failed: '300' (out string) != 300 (concurrencyTarget string)
```
updating..

rhuss:
> interesting! `NotOwned: There is an existing Route "hello" that we do not own.`
> excerpts from https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_client/300/pull-knative-client-integration-tests-latest-release/1154332359860948996

that's because the `kn delete` before is async when it comes to delete dependent objects and the route just had not been deleted.

navidshaikh:
jsonpath doesnt seem to yield result against `latest-release` e2e job while it works fine for v0.6.0
```
--- FAIL: TestServiceOptions (137.14s)
    --- FAIL: TestServiceOptions/create_and_validate_service_with_concurrency_options (4.02s)
        service_options_test.go:75: assertion failed:  (out string) != 300 (concurrencyTarget string)
[..]
```

test/e2e/service_options_test.go
```diff
@@ -29,62 +28,55 @@ func TestServiceOptions(t *testing.T) {
...
+	//jsonpath := "jsonpath={.items[0].spec.template.spec.containerConcurrency}"
+	//debug
+	out, err := test.kn.RunWithOpts([]string{"service", "describe", serviceName}, runOpts{})
+	//debug
+	assert.Equal(t, "foo", out)
 	assert.NilError(t, err)
-
-	expectedOutput := fmt.Sprintf("containerConcurrency: %s", concurrencyLimit)
-	assert.Check(t, util.ContainsAll(out, expectedOutput))
+	//out, err := test.kn.RunWithOpts([]string{"service", "list", serviceName, "-o", jsonpath}, runOpts{})
```
maximilien: Can we remove commented code?

test/e2e/service_options_test.go
```diff
@@ -29,62 +28,55 @@ func TestServiceOptions(t *testing.T) {
...
-		test.serviceDescribeConcurrencyTarget(t, "hello", "300")
+	t.Run("update and validate service with concurrency limit", func(t *testing.T) {
+		test.serviceUpdate(t, "svc1", []string{"--concurrency-limit", "300"})
+		test.validateServiceConcurrencyLimit(t, "svc1", "300")
 	})
 
-	t.Run("update concurrency limit for hello service and returns no error", func(t *testing.T) {
-		test.serviceUpdate(t, "hello", []string{"--concurrency-limit", "300"})
+	t.Run("update concurrency options with invalid values for service", func(t *testing.T) {
+		test.serviceUpdateWithInvalidValue(t, "svc1", []string{"--concurrency-limit", "-1", "--concurrency-target", "0"})
```
maximilien: The one thing I don’t like here is that the method will test an invalid case in two steps. Perhaps better is to have helper method that does service update and returns error that in this “test” is validated to make sure failure is correct. Or implement all here. The reason for the helpers is that they are reusable but one that tests an invalid condition is likely not reusable IMO. What do you think?

maximilien:
Odd that @navidshaikh has cla/google failing? He's one of our top committers! Something wrong here.

navidshaikh:
The tests failure is due to different fields returned by different e2e jobs we have namely
1. pull-knative-integration-tests  [logs](https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_client/302/pull-knative-client-integration-tests/1154430760137003009)
2. pull-knative-client-integration-tests-latest-release [logs](https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_client/302/pull-knative-client-integration-tests-latest-release/1154430760137003011)

where

1. returns the `service` with fields like `spec.template.spec.containerConcurrency` while
2. returns the `service` with fields `spec.runLatest.configuration.revisionTemplate.spec.containerConcurrency`

Thus, the `jsonpath` we've used to extract the specific fields can't return the expected value for 2.

If checked logs, we can see that distinction, I'd added a debug commit to reflect that.

This has been the reason for failure of job 2 in this  PR and #300. 
I will have this accommodated in both PRs.

cc @rhuss

test/e2e/service_options_test.go
navidshaikh: Sure, this was added as debug commit to check the reason for continuous failure of e2e job `pull-knative-client-integration-tests-latest-release` we've, please check comment https://github.com/knative/client/pull/302#issuecomment-515297563 ,  I'll replace the debug commit with a fix. Thanks!

test/e2e/service_options_test.go
navidshaikh: Ok I see. Although the other tests here uses the helper validation methods to validate particular field of service, but they are used more than once. This particular test and its helper method is checking for a particular error.
As we move along, we'll need to bring more of such negative test cases, but I am fine with implementing that two liner helper method here, we can always have it separated into helper if more tests require it.

navidshaikh:
> Odd that @navidshaikh has cla/google failing?

@maximilien : Yeah, all of my PRs are complaining the cla failure, this seems like flake and noticed on slack that other folks complained about it too.

cc: @adrcunha

rhuss:
@navidshaikh I wonder whether we should bump up to 0.7.0 for KNATIVE_VERSION in `test/presubmit-integration-tests-latest-release.sh` ? I mean we are testing against head of knative serving master by default and then we should test against the latest release.

I suspect that this error is already gone in 0.7.0 as according to lemonade only the new fields should be populated since 0.7.0

adrcunha:
@googlebot rescan

adrcunha:
@googlebot rescan

adrcunha:
@googlebot rescan

adrcunha:
The bot states that nshaikh@redhat.com didn't sign a CLA. Can you double check that using the links provided in the bot's response above? Also, replying "I signed it" should force it to recheck the status (rescanning causes the bot to set the "cla: no" label again).

navidshaikh:
@googlebot I signed it

navidshaikh:
@googlebot I consent

rhuss:
@googlebot rescan

PR #306, opened 2019-07-26
feature(testing): Introduce a mock implementation for KnClient

rhuss:
Commands must only use the `KnClient` API and their unit tests should also
only use a mock implementation for this interface.

This commit introduces such a Mock implementation and works like in
the example below for creating a simple service in a synchronous
way

```golang
// New mock client
client := knclient.NewMockKnClient(t)

// Recording:
r := client.Recorder()
// Check for existing service --> no
r.GetService("foo", nil, errors.NewNotFound(v1alpha1.Resource("service"), "foo"))
// Create service (don't validate given service --> "Any()" arg is allowed)
r.CreateService(knclient.Any(), nil)
// Wait for service to become ready
r.WaitForService("foo", knclient.Any(), nil)
// Get for showing the URL
r.GetService("foo", getServiceWithUrl("foo", "http://foo.example.com"), nil)

// Testing:
output, err := executeCommand(client, "create", "foo", "--image", "gcr.io/foo/bar:baz")
assert.NilError(t, err)
assert.Assert(t, util.ContainsAll(output, "created", "foo", "http://foo.example.com", "Waiting"))

// Validate that all recorded API methods have been called
r.Validate()
```

Such tests have three phases:

* A recording phase where the mock client is prepared. In this phase a
  recorder is called with the expected arguments and the return values
  it can return. The arguments can be also functions with
  signature `func (t *testing.T, actual interface{}, expected interface{})`
  and will be called to verify a given argument. Such a function should
  `t.Fail()` on its own if the validation fails.
  A convenient `Any()` method is added to allow no validation on an argument
  (see example below).
  Method can be called multiple times, but the order needs to reflect
  the actual calling order
* A playback phase where the test executed which in turn calls out to the
  Mocks
* A validation phase to check the expected output. The recorder can be
  also validated to verify that all recorded mock calls has been
  used during the test.

  See `service_create_mock_test.go` for a full example.

### DIFF
--- a/pkg/kn/commands/service/service_create_mock_test.go
+++ b/pkg/kn/commands/service/service_create_mock_test.go
@@ -26,7 +26,6 @@ import (
 
 	"github.com/knative/client/pkg/kn/commands"
 	knclient "github.com/knative/client/pkg/serving/v1alpha1"
-	mock "github.com/knative/client/pkg/serving/v1alpha1_test"
 
 	"github.com/knative/client/pkg/util"
 )
--- a/pkg/kn/commands/service/service_create_mock_test.go
+++ b/pkg/kn/commands/service/service_create_mock_test.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/knative/client/pkg/kn/commands"
 	knclient "github.com/knative/client/pkg/serving/v1alpha1"
+	mock "github.com/knative/client/pkg/serving/v1alpha1_test"
 
 	"github.com/knative/client/pkg/util"
 )
--- a/pkg/kn/commands/service/service_create_mock_test.go
+++ b/pkg/kn/commands/service/service_create_mock_test.go
@@ -1,0 +1,79 @@
+// Copyright © 2019 The Knative Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package service
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/knative/pkg/apis"
+	"gotest.tools/assert"
+	"k8s.io/apimachinery/pkg/api/errors"
+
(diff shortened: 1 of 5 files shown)


rhuss:
@maximilien you might be interested in this PR as it would simplify command testing considerably.

pkg/kn/commands/namespaced.go
```diff
@@ -57,7 +58,10 @@ func (params *KnParams) GetNamespace(cmd *cobra.Command) (string, error) {
 		var err error
 		namespace, err = params.CurrentNamespace()
 		if err != nil {
-			return "", err
+			if !clientcmd.IsEmptyConfig(err) {
```
rhuss: This is PR #305 included here.

rhuss:
The reasons why I decided against a Mock framework (GoMock, counterfeiter, pegomock, ...) are:

* I wanted to avoid code generation
* The simplest implementation possible which suites our needs

wrt/ to maintenance, yes its a bit more then calling a code generator but not so much as that's all you have to add for a new `KnClient` method:

```
// Get a revision by name
func (r *Recorder) GetRevision(name interface{}, revision *v1alpha1.Revision, err error) {
	r.add("GetRevision", apiMethodCall{[]interface{}{name}, []interface{}{revision, err}})
}

func (c *MockKnClient) GetRevision(name string) (*v1alpha1.Revision, error) {
	call := c.getCall("GetRevision")
	c.verifyArgs(call, name)
	return call.result[0].(*v1alpha1.Revision), errorOrNil(call.result[1])
}
```

As mentioned, simplicity was the main driving factor (which makes up also for a simplified call to the recording method which combines expected arguments and return values in a single argument list.

Could be that's too simplistic. If so, we, of course, switch over to a full Mock framework.

maximilien:
I am not going to oppose this but my vote would be using a mock framework. Specifically because it's less custom code to maintain. 

However, since we have not used any mock framework so far in this project, the decision of using a framework is heavier than a custom mock...hence why I'd compromise with your solution. I want to be easy 🏖 and trust your judgement here. 

OK this message and I'll lgtm. Cheers :beers:

rhuss:
Some other benefits of using that kind of mocks for testing commands:

* If we want to offer a KnClient API as stated as a goal, offering a typed test API is a plus for users of this API. 
* Using that Mock interface exclusively in the Kn commands unit tests, we have a clear separation and also an indication that no command uses the plain KnApis directly but only via the KnClient abstraction. 

The Mock API is kept deliberately very simplistic to make it easy (even trivial) to extend or to adapt it.
However, I totally agree with @sixolet  that an excellent test coverage of the KnClient implementation is essential if we not only want to rely on the integration tests.

Issue #301, opened 2019-07-25
Restructure E2E test suite


mgencur:
## In what area(s)?

Structure of the E2E test suite

## Ask your question here:

Should we use a different director/file/test structure for the E2E tests?
Let me propose some changes which we briefly discussed with Navid:
* create the following folders in test/e2e folder which would hold specific tests for these different kinds of objects/commands: service, revision, route.
* place service_options_test.go in the service folder
* move revision_workflow_test.go to the revision folder and rename it to something that doesn't include "workflow" in the name; IMO, there's nothing like a "revision" workflow. Workflows show be defined at a higher level and related to goals that the end user wants to achieve: basic, advanced, trafic_split, etc.
* keep basic_workflow_test.go in test/e2e folder or move it to a new subfolder called "workflow"? 
* pull out sub-tests which are not really basic ones from the basic workflow and place them in a more specific test file

Example:
```
test/e2e/basic_workflow_test.go
test/e2e/service/service_options_test.go
test/e2e/revision/revision_test.go
test/e2e/route/..
test/e2e/env.go
test/e2e/e2e_flags.go
test/e2e/...
```

Note: I think it is good to keep some workflows at the top level because they're nicely showing the basic usage of kn which includes invocations of different kn commands, not just specific to service or revision.

navidshaikh:
@mgencur 

> move revision_workflow_test.go to the revision folder 

As you mentioned, just re-group the mentioned file's tests into specific tests under `revision` folder, and remove `revision_workflow_test.go` file altogether.

> keep basic_workflow_test.go in test/e2e folder or move it to a new subfolder called "workflow" ? 

+1 for moving the workflows under `workflow` directory, we could then add more there as we develop.

PR #307, opened 2019-07-26
Restructure E2E test suite

mgencur:
* move out more tests from the basic workflow and put them in separate
files, including tests for routes
* rename revision_workflow_test.go to revision_test.go

Fixes #https://github.com/knative/client/issues/301

## Proposed Changes

I've also excluded things related to routes from the basic workflow. IMO, it is not required to touch routes in the basic workflow. We get information about the route from "service describe". OTOH, I included `revision describe` in the basic workflow because I think that will be a common operation to get information about the given revision.

Moving test files into sub-folders such as revision, service, etc. is currently not possible with the current way we run tests. There are two problems:
1) When running `go test command` for all tests we'd have to list multiple/many directories
2) We can't really define methods for the e2eTest type outside the e2e package, this is not allowed in Go. So we can't define our methods in e.g. service package for the e2eType

**Release Note**

```release-note

```

### DIFF
--- a/test/e2e/revision_test.go
+++ b/test/e2e/revision_test.go
@@ -26,6 +26,7 @@ import (
 )
 
 func TestRevision(t *testing.T) {
+	t.Parallel()
 	test := NewE2eTest(t)
 	test.Setup(t)
 	defer test.Teardown(t)
--- a/test/e2e/revision_test.go
+++ b/test/e2e/revision_test.go
@@ -25,25 +25,24 @@ import (
 	"gotest.tools/assert"
 )
 
-func TestRevisionWorkflow(t *testing.T) {
+func TestRevision(t *testing.T) {
 	test := NewE2eTest(t)
 	test.Setup(t)
 	defer test.Teardown(t)
 
-	t.Run("create hello service and returns no error", func(t *testing.T) {
+	t.Run("create hello service and return no error", func(t *testing.T) {
 		test.serviceCreate(t, "hello")
 	})
 
-	t.Run("describe revision from hello service", func(t *testing.T) {
-		test.revisionDescribe(t, "hello")
+	t.Run("describe revision from hello service with print flags", func(t *testing.T) {
 		test.revisionDescribeWithPrintFlags(t, "hello")
 	})
 
-	t.Run("delete latest revision from hello service and returns no error", func(t *testing.T) {
+	t.Run("delete latest revision from hello service and return no error", func(t *testing.T) {
 		test.revisionDelete(t, "hello")
 	})
 
-	t.Run("delete hello service and returns no error", func(t *testing.T) {
+	t.Run("delete hello service and return no error", func(t *testing.T) {
 		test.serviceDelete(t, "hello")
 	})
 }
@@ -55,19 +54,6 @@ func (test *e2eTest) revisionDelete(t *testing.T, serviceName string) {
 	assert.NilError(t, err)
 
 	assert.Check(t, util.ContainsAll(out, "Revision", revName, "deleted", "namespace", test.kn.namespace))
-}
-
-func (test *e2eTest) revisionDescribe(t *testing.T, serviceName string) {
-	revName := test.findRevision(t, serviceName)
-
-	out, err := test.kn.RunWithOpts([]string{"revision", "describe", revName}, runOpts{})
-	assert.NilError(t, err)
-
-	expectedGVK := `apiVersion: serving.knative.dev/v1alpha1
-kind: Revision`
-	expectedNamespace := fmt.Sprintf("namespace: %s", test.kn.namespace)
-	expectedServiceLabel := fmt.Sprintf("serving.knative.dev/service: %s", serviceName)
-	assert.Check(t, util.ContainsAll(out, expectedGVK, expectedNamespace, expectedServiceLabel))
 }
 
 func (test *e2eTest) revisionDescribeWithPrintFlags(t *testing.T, serviceName string) {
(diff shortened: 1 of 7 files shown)


mgencur:
I increase the timeout because the tests ran more than 10mins. What takes most time is deleting the namespace multiple times. I'm not sure what's the purpose of creating/deleting the namespace in each test and I guess we could save a lot of test execution time by only creating/deleting the namespace once - and having one common namespace for all tests.

rhuss:
> Why are you changing "returns" to "return" in the comment?

Because its mostly used as an imperative like in 

> delete hello and svc2 services and return no error

If you want to have a declarative sentence (without subject though), then you should also use "deletes" instead of "delete". 

BTW, we are now entering the realm of heavy nit-picking imo. I think we have bigger fish to fry.

mgencur:
@maximilien thanks for the review. Some comments:
* return vs. returns - I think @rhuss explained it nicely. That was my point. My intention was not to dive into any big discussions around it, though. I just did the change while doing some other changes.
* I didn't create sub-packages in this PR even though I wanted to. Perhaps I agree that right now it doesn't have a big value. It just complicates stuff. We can consider adding sub-packages later, in my opinion.

toVersus:
> I'm not sure what's the purpose of creating/deleting the namespace in each test

@mgencur I think this is for the use case of running single e2e test. Let's say, I would like to run `TestServiceOptions` only because this change is related to the service creation with options. But, I'm also sick of waiting for each test to delete its namespace every time through the e2e test suites.

> and having one common namespace for all tests.

I guess we can safely run e2e tests [in parallel](https://golang.org/pkg/testing/#T.Parallel) thanks to the namespace isolation, which saves a lot of our time. (In my local setup, it takes 311.766s.)

mgencur:
@toVersus does it make sense to put t.Parallel() in all the tests to make them run parallel? When I do that the whole test suite runs only for 133s on my localhost.

toVersus:
Sure, I think we should do that. @maximilien @rhuss, will it cause any problems?

mgencur:
Rebased.

toVersus:
> On the other hand, if I create the namespace only once at the beginning and delete once at the end, I can run the whole test suite serially under 80 seconds.

I think it may cause a race condition when we use the same resource identifier (service name, revision name, ...) and we have to keep in mind that it is unacceptable to bring any duplications throughout the test suites. IMO, instead of doing that, we should create a namespace per test scenario and guarantee the isolation. Also, it makes impossible for us to run single test for confirmation.

rhuss:
I think if we can settle on a naming convention for the resources created (like prefixing them uniquely according to testsuite) then we could come away without creating a new namespace. In general, the faster tests are running the better because they are running often.

I 👍  to stay in one single namespace per overall testrun.

mgencur:
My apologies. I provided wrong numbers for the "serial" execution because I was not deleting the namespace at the end by mistake. When I run the tests serially and delete the namespace at the end it takes about 230 seconds which is longer than the parallel execution. 
I suppose at this point it makes sense to enable parallel execution because it is faster, even with multiple namespaces.

mgencur:
@rhuss, thanks. We can try using a single namespace in the future via a separate PR.

test/e2e/revision_test.go
```diff
@@ -25,25 +25,25 @@ import (
...
 
-func TestRevisionWorkflow(t *testing.T) {
+func TestRevision(t *testing.T) {
+	t.Parallel()
 	test := NewE2eTest(t)
 	test.Setup(t)
 	defer test.Teardown(t)
 
-	t.Run("create hello service and returns no error", func(t *testing.T) {
+	t.Run("create hello service and return no error", func(t *testing.T) {
```
maximilien: this is a small issue, however, I disagree with this change as I mentioned before. The title of the test is for humans, “returns” is the correct Eglish. Changing to “return” here has the feeling that the PR author is just looking for things to change.

Cc: @rhuss I guess this is already merged but perhaps we should have resolutions on various reviewers before merging

rhuss: Have you seen my comment https://github.com/knative/client/pull/307#issuecomment-515656277 ?

"return" is perfectly valid as the sentence is an imperative sentence. If you change this to a declarative sentence, then you should use "returns" but also "creates". Is this hard to understand ? I think we should start a vote whether we want to use the imperative or declarative form for *test* output (that no users see).

rhuss: > Cc: @rhuss I guess this is already merged but perhaps we should have resolutions on various reviewers before merging

Can we agree that the imperative form is ok ? It the same form that is also recommended for git commit message and it says: "Hey, test, create hello service and return no error"

You could also say "this test create**s** hello service and return**s** no error"

Regardless what, the sentence "create hello service and returns no error" is plain wrong.

rhuss: But you probably just overlooked https://github.com/knative/client/pull/307#issuecomment-515656277 because I haven't seen any answer or rejection of it. So I thought that you silently agreed (like for my other review comments elsewhere. maybe.)

rhuss: And as also stated before, I firmly believe we should not waste our energy here. there are much bigger challenges ahead.

mgencur: My intention was really just to fix the test name so as to use the imperative style of the sentence. I was not looking for random things to change, really.

maximilien: Did not see the comment. I am cool. I also really like the Parallel change which justifies this change but clearly came after.

We are cool @mgencur and certainly appreciate the contributions. Cheers 🍻

maximilien: In general, as you know from previous interactions, I try not to be picky about things. My goal is to ship things: the best UX for Knative. All of this is a means to an end.

That said, the more consistent the code and our work products the better. So just wanted to be clear. We are now. Enough words on this issue. Cheers all. Next :)

PR #313, opened 2019-07-27
refactor(plugins): Improved and simplified verifier and plugin list

rhuss:
First sweep:

* Added proper executable check for the current user (+ extensive tests for all combinations)
* Refactor plugin_list to make it more straightforward
* Separate errors and warnings
* Don't return an error when no plugin is installed and `plugin list` is called
* Simplified tests
* Check assumption that a prefix is given in verify()
* Fixed broken homedir expansion ( `~/` (not `~`) must be expanded only when at the beginning of a path)

After this:

* Remove plugin options from the command line
* Remove multiple prefix support

This PR is a result of the discussion on #249 where we agreed to merge now (as the PR has been aged and grown considerably) but work on improvements right after.


rhuss:
Note to reviewers: As this a refactoring PR the diff is not easy to read. All changes are confined to the `plungin` package.

I suggest to start at those files with these entry points:

* For the refactored verifier look here --> https://github.com/knative/client/pull/313/files#diff-cf339cceb78d7b2e50576e72699ffce3R60
* For the simplified `plugin list` command here --> https://github.com/knative/client/blob/c2bd8d25377e5117c43b49a2a36967375ca1b7aa/pkg/kn/commands/plugin/plugin_list.go#L61
* For the changed tests
  - plugin_list_test: https://github.com/knative/client/blob/c2bd8d25377e5117c43b49a2a36967375ca1b7aa/pkg/kn/commands/plugin/plugin_list_test.go#L93
  - plugin_verifier: https://github.com/knative/client/blob/c2bd8d25377e5117c43b49a2a36967375ca1b7aa/pkg/kn/commands/plugin/plugin_verifier_test.go#L46

maximilien:
@rhuss I will get to this next week. Lots of things to tie up this week and also taking a few days off for a quick trip to London.

I will move this up my TODO list once I am back or if I have WiFi in plane on Tuesday. Promise :)

duglin:
Can you also address this:
```
$ kn --lookup-plugins-in-path=true plugin list
Unable read directory '/root/.kn/plugins' from your plugins path: open /root/.kn/plugins: no such file or directory. Skipping...The following compatible plugins are available, using options:
  - plugins dir: '~/.kn/plugins'
  - lookup plugins in path: 'true'

/root/bin/kn-abc
Unable read directory '/root/src/k8s.io/kubernetes/_output/dockerized/bin/linux/amd64' from your plugins path: open /root/src/k8s.io/kubernetes/_output/dockerized/bin/linux/amd64: no such file or directory. Skipping...
```

That's just weird output  ;-)

and
```
$ kn --lookup-plugins-in-path=true plugin list
Unable read directory '/root/.kn/plugins' from your plugins path: open /root/.kn/plugins: no such file or directory. Skipping...Unable read directory '/root/src/k8s.io/kubernetes/_output/dockerized/bin/linux/amd64' from your plugins path: open /root/src/k8s.io/kubernetes/_output/dockerized/bin/linux/amd64: no such file or directory. Skipping...
warning: unable to find any kn plugins in your plugin path: '[/root/.kn/plugins /root/bin /usr/local/sbin /usr/local/bin /usr/sbin /usr/bin /sbin /bin /usr/games /usr/local/games /usr/local/go/bin /root/src/bin /root/src/k8s.io/kubernetes/_output/local/go/bin /root/src/k8s.io/kubernetes/_output/dockerized/bin/linux/amd64]'
```
Should be empty.

```
$ ./kn plugin list
No plugins found in path /root/.kn/plugins
```

It should show nothing IMO

rhuss:
@duglin 

* Non-existing path directories are ignored (like for real path handling), so no output here anymore
* Configuration of where to lookup configuration is only shown with `--verbose`
* For no plugin installed `kn plugin list` shows

```
No plugins found in ~/.kn/plugins
```

The later error message could be improved to say "No plugins are found in ~/.kn/plugins nor in your path" if path lookup is enable

rhuss:
> It should show nothing IMO

Yes, that could be done, too. I suggest to print nothing, but with `--verbose` we print the line above.

duglin:
>  For no plugin installed kn plugin list shows
> > No plugins found in ~/.kn/plugins

I don't see the value in showing this - let it be controlled via `--verbose` as you said. But if we MUST show something w/o --verbose then I would suggest you remove the `in ~/.kn/plugins` since for consistency you should then show the $PATH value when we search $PATH - and that's just annoying.  Also, we don't need to remind people where the dir is.

pkg/kn/commands/plugin/plugin_verifier.go
```diff
@@ -0,0 +1,252 @@
...
+
+	return eaw
+}
+
+func (v *pluginVerifier) addWarningIfAlreadySeen(eaw errorsAndWarnings, path string) errorsAndWarnings {
+	fileName := filepath.Base(path)
+	if existingPath, ok := v.seenPlugins[fileName]; ok {
+		return eaw.addWarning("%s is ignored because it is shadowed by a equally named plugin: %s.", path, existingPath)
+	}
+	v.seenPlugins[fileName] = path
```
sixolet: Can we move the part where we check off we've seen it to the end of `verify`, so the `addWarningIf...` methods don't mutate anything but the `eaw` and not the acceptor?

pkg/kn/commands/plugin/plugin_list_test.go
```diff
@@ -15,226 +15,208 @@
...
+	if inPath {
+		path = ctx.pathDir
+	}
+	return ctx.createTestPluginWithPath(pluginName, fileMode, path)
+}
 
-		pluginsDir = filepath.Join(tmpPathDir, "plugins")
-		pluginsDirFlag = fmt.Sprintf("--plugins-dir=%s", pluginsDir)
+func (ctx *testContext) createTestPluginWithPath(pluginName string, fileMode os.FileMode, path string) error {
+	if runtime.GOOS == "windows" {
```
maximilien: Nice. I completely forgot and ignored Windows users. But good on this.

maximilien: Though I have no way to test actual Windows plugin

pkg/kn/commands/plugin/plugin_verifier_test.go
```diff
@@ -0,0 +1,228 @@
...
+				eaw = verifier.verify(eaw, overwritingPluginPath)
+				assert.Assert(t, len(eaw.errors) == 1)
+				assert.Assert(t, len(eaw.warnings) == 0)
+				assert.Assert(t, util.ContainsAll(eaw.errors[0], "overwrite", "kn-plugin"))
+			})
+		})
+	})
+}
+
+func isExecutable(plugin string) bool {
```
maximilien: Why are we ignoring output here? Wonder if we should save this for debug purposes?

pkg/kn/commands/plugin/plugin_verifier.go
```diff
@@ -0,0 +1,259 @@
...
+func printSection(out io.Writer, label string, values []string) {
+	if len(values) > 0 {
+		printLabelWithConditionalPluralS(out, label, len(values))
+		for _, value := range values {
+			fmt.Fprintf(out, "  - %s\n", value)
+		}
+	}
+}
+
+func printLabelWithConditionalPluralS(out io.Writer, label string, nr int) {
```
maximilien: `...PluralS`... why not `...Plurals`?

pkg/kn/commands/plugin/plugin_verifier.go
```diff
@@ -0,0 +1,259 @@
...
+	// Only pick the first prefix as it is very like that it will be reduced to
+	// a single prefix anyway (PR pending)
+	prefix := ValidPluginFilenamePrefixes[0]
+	if !strings.HasPrefix(fileName, prefix+"-") {
+		eaw.addWarning("%s plugin doesn't start with plugin prefix %s", fileName, prefix)
+	}
+	return eaw
+}
+
+func (v *pluginVerifier) addWarningIfNotExecutable(eaw errorsAndWarnings, path string, fileInfo os.FileInfo) errorsAndWarnings {
```
maximilien: Would it make sense to divide this into smaller helper func for each of the different cases?  This func is close to 70 lines long...

pkg/kn/commands/plugin/plugin_list_test.go
rhuss: same for me ;-( I think we need at least a CI test suite somehow. I wonder how other CLI's like kubectl does it ?

pkg/kn/commands/plugin/plugin_verifier.go
rhuss: Because I refer to the plural "S" to be added at the end, and I don't refer to the plural of plural ;-) (which would be "plurals", supposed that something like this exists)

rhuss:
> So overall for me errors and warnings are important to give as much info as possible to the user for that command.

warnings and errors are printed even without `--verbose`. It's only that the static configuration information where plugins are looked up is hidden without `--verbose`.

pkg/kn/commands/plugin/plugin_verifier.go
maximilien: OK cool... weird, but cool :)

PR #316, opened 2019-07-29
Update dev docs for some missing bits

duglin:
Just noticed some missing things while installing on a new system

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -25,6 +25,9 @@ You must install these tools:
 
 1. [`go`](https://golang.org/doc/install): The language `Knative Client` is
    built in (1.11.4 or later)
+1. [`goimports`](https://godoc.org/golang.org/x/tools/cmd/goimports)
+1. `gcc` compiler: Used during testing. Not needed if golang is installed
+   via the installer
 1. [`git`](https://help.github.com/articles/set-up-git/): For source control
 1. [`kubectl`](https://kubernetes.io/docs/tasks/tools/install-kubectl/): For
    managing development environments


DEVELOPMENT.md
```diff
@@ -25,6 +25,8 @@ You must install these tools:
 
 1. [`go`](https://golang.org/doc/install): The language `Knative Client` is
    built in (1.11.4 or later)
+1. [`goimports`](https://godoc.org/golang.org/x/tools/cmd/goimports)
```
rhuss: fyi, you don't have to. If not there is a fall back to gofmt and do the import properly on your own. But happy to keep that (as it doesnt harm)

DEVELOPMENT.md
```diff
@@ -25,6 +25,8 @@ You must install these tools:
...
+1. `gcc` compiler: Used during testing
```
rhuss: Why that ?

DEVELOPMENT.md
duglin: I kept getting a warning about goimports not being there, and this seemed to fix it

DEVELOPMENT.md
duglin: No idea, surprised me too but the testing failed for me w/o installing gcc.  I was on Ubuntu - perhaps gcc is there by default for some other distros.

rhuss: Good to know, because I'm not aware of any low-level system integrations which requires gcc.
I also think that we should still support windows as a build platform, as we have some platform specific handling (i.e. for plugins) which is currently untested for Windows. We would need a Windows CI system for that. 

@duglin do you remember the error message ? Really would love to track that done, but can't right now.

DEVELOPMENT.md
rhuss: Yep. It's a warning "only". But of course nicer to have that installed, so let's keep that.

DEVELOPMENT.md
duglin: I just uninstalled gcc and the error message doesn't appear, I'm wondering if some stuff is left around. My assumption was that golang needed some gcc stuff during the testing phase.

I managed to recreate it in a clean docker container:
```
⚖️ License
📖 Docs
🚧 Compile
🧪 Test
🔥 Failure
# runtime/cgo
exec: "gcc": executable file not found in $PATH
=== RUN   TestDummy
--- PASS: TestDummy (0.00s)
PASS
ok  	github.com/knative/client/pkg	0.001s
FAIL	github.com/knative/client/pkg/kn/commands [build failed]
FAIL	github.com/knative/client/pkg/kn/commands/plugin [build failed]
FAIL	github.com/knative/client/pkg/kn/commands/revision [build failed]
FAIL	github.com/knative/client/pkg/kn/commands/route [build failed]
FAIL	github.com/knative/client/pkg/kn/commands/service [build failed]
FAIL	github.com/knative/client/pkg/kn/core [build failed]
FAIL	github.com/knative/client/pkg/serving [build failed]
FAIL	github.com/knative/client/pkg/serving/v1alpha1 [build failed]
=== RUN   TestContainsAll
--- PASS: TestContainsAll (0.00s)
PASS
ok  	github.com/knative/client/pkg/util	0.002s
FAIL	github.com/knative/client/pkg/wait [build failed]
```

What's interesting is that I only get this when I installed golang via a `tar` command. Installing it via `apt-get install golang` didn't show the issue. I'm guessing that installs some variant of gcc under the covers.

duglin: I updated the text to so you only need to install gcc if you don't use the golang installer.

Issue #286, opened 2019-07-23
`kn service update` should print an error message when invalid or no option is passed


toVersus:
## In what area(s)?

/kind good-first-issue
/kind feature

## Describe the feature:

When users run `kn service update` command with invalid options, nothing is changed. But in current implementation, kn prints the following output:

```
$ ./kn service update hello --max-scale -100
Waiting for service 'hello' to become ready ... OK
Service 'hello' updated in namespace 'default'.
```

IMO, this is a little confusing behavior. Kn should print `unchanged` when it changed nothing as kubectl does. I think we can implement this feature by comparing the generation before and after [updating the service resource](https://github.com/knative/client/blob/75a41def4e2753fac76d113c331f036499228a95/pkg/serving/v1alpha1/client.go#L177) and returning changed/unchanged state (boolean).

```
$ ./kn service update hello --max-scale -100
Service 'hello' unchanged in namespace 'default'.
```

Can I take this, if you can accept this proposal?

rhuss:
Shouldn't we just print an error in this case ? like "-100 is an invalid value for max-scale".

Anything else would be surprising.

rhuss:
And then the "unchanged" message is unnecessary (you perfectly can assume that the service is not changed if an error occurs)

toVersus:
> Shouldn't we just print an error in this case ? 

Ah, maybe I mixed up things, so aren't there any strong opinions in current behavior, which silently keeps "unchanged" state if users input invalid values. Can we require to pass at least one option at anytime for avoiding the following case ?

```
$ ./kn service update hello
Waiting for service 'hello' to become ready ... OK
Service 'hello' updated in namespace 'default'.
```

rhuss:
Good point. We should catch that and return an error for such a no-op.

Of course, when we update with a new value == old value, we should not worry about (and should also print the success message)

zhanggbj:
Hi @toVersus @rhuss ,

If nobody's working on this, I would like to take it. Thanks!
From the above comments, I think two items need to fix.

- Add input validation for `kn service update` flags, for e.g., if `--max-scale -100` is specified, we should return an error.

- If no input for `kn service update`, we should return an error.

rhuss:
@zhanggbj agreed on the action items.

/assign zhanggbj

toVersus:
@zhanggbj Thanks for picking it up! I worked on the first item in #279. I would appreciate if you look into the PR before implementing the second action item.

zhanggbj:
[...]

PR #318, opened 2019-07-30
Report an error if no flag(s) set in service update

zhanggbj:
For now if no flag(s) set, service update will still try to do an update as below, it should return an error instead.

```
$ ./kn service update hello
Waiting for service 'hello' to become ready ... OK
Service 'hello' updated in namespace 'default'.
```

Fixes [issue 286](https://github.com/knative/client/issues/286)

## Proposed Changes

* Return an error if no flag(s) set in service update

**Release Note**

```release-note

```

### DIFF
--- a/pkg/kn/commands/service/service_update.go
+++ b/pkg/kn/commands/service/service_update.go
@@ -29,7 +29,7 @@ func NewServiceUpdateCommand(p *commands.KnParams) *cobra.Command {
 	var waitFlags commands.WaitFlags
 
 	serviceUpdateCommand := &cobra.Command{
-		Use:   "update NAME",
+		Use:   "update NAME [flags]",
 		Short: "Update a service.",
 		Example: `
   # Updates a service 'mysvc' with new environment variables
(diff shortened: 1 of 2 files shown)


pkg/kn/commands/service/service_update_test.go
```diff
@@ -88,6 +88,26 @@ func fakeServiceUpdate(original *v1alpha1.Service, args []string, sync bool) (
...
+	orig := newEmptyService()
+
+	action, _, _, err := fakeServiceUpdate(orig, []string{
+		"service", "update", "foo"}, false)
+
+	if action != nil {
+		t.Errorf("Unexpected action if no flag(s) set")
+	}
+
+	if err != nil {
```
toVersus: I think you can reduce a if-else depth by swapping the if-else block and using t.Fatal.

pkg/kn/commands/service/service_update.go
```diff
@@ -50,6 +50,10 @@ func NewServiceUpdateCommand(p *commands.KnParams) *cobra.Command {
 				return err
 			}
 
+			if cmd.Flags().NFlag() == 0 {
```
maximilien: The other option would be to do this in a `PreRun` function which should allow separation of the checks to the `Run`. See this [comment](https://github.com/spf13/cobra/issues/206) from Cobra community.

Since this check is simple, it's likely OK here but if we start adding more such checks or other commands then `PreRun` could be a nice way to write these pre-condition code. Cheers :)

zhanggbj: @maximilien As you mentioned, there may be more checks needed later, and people even mentioned about the odd error message when wrong args here # 317, so I moved this part to `PreRun`. Thanks!

pkg/kn/commands/service/service_update_test.go
zhanggbj: @toVersus I addressed it, thanks!

maximilien:
Thanks @zhanggbj and @toVersus for the review

/lgtm

PR #326, opened 2019-08-01
Add --log-http option

sixolet:
It prints all the HTTP requests and responses to stderr. 

Quick and dirty, I just found it useful when debugging what requests `kn` was sending, maybe other people will too.

Prints the http requests before they get their tokens added, so it's not that much of a security risk.


cppforlife:
@sixolet is that similar to how kubectl implements -v=8 option?

sixolet:
Looks like yes-ish, but they use the `github.com/golang/glog` module to read the global log level in the factory for deciding whether to generate the verbose transport or not, and can generate the transport at creation time, which didn't make its way into client-go. However, reading more code made me realize a slightly better way to do it that nixes the downcast, I think.

pkg/util/logging_http_transport.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func (t *LoggingHttpTransport) RoundTrip(r *http.Request) (*http.Response, error) {
+	stream := t.stream
+	if stream == nil {
+		stream = os.Stderr
+	}
+	reqBytes, _ := httputil.DumpRequestOut(r, true)
+	fmt.Fprintln(stream, "===== REQUEST =====")
+	fmt.Fprintln(stream, string(reqBytes))
```
cppforlife: a little concerning that this will include auth headers etc.

sixolet:
Doesn't, actually, experientially, since the logger is outside the layer that adds them. I can add that to the test.

sixolet:
... actually never mind, I was looking at the response. I'll fix.

pkg/util/logging_http_transport.go
sixolet: Should be fixed now.

pkg/kn/commands/types_test.go
```diff
@@ -70,6 +109,11 @@ func TestGetClientConfig(t *testing.T) {
 
 	multiConfigs = multiConfigs
 	for _, tc := range []typeTestCase{
+		{
+			"",
+			clientcmd.NewDefaultClientConfigLoadingRules().ExplicitPath,
+			"",
+		},
```
rhuss: What's the difference to the test case below ?

pkg/util/logging_http_transport.go
```diff
@@ -0,0 +1,74 @@
...
+	stream := t.stream
+	if stream == nil {
+		stream = os.Stderr
+	}
+	reqCopy := *r
+	reqCopy.Header = make(http.Header, len(r.Header))
+	for k, v := range r.Header {
+		sensitive := SENSITIVE_HEADERS[k]
+		if sensitive {
+			reqCopy.Header.Set(k, "ELIDED")
```
rhuss: Side note: As a non-native speaker, I had to look up "ELIDED", (now I know it ;-). Maybe replace with "******" as this is a common idiom for representing sensitive data in an obfuscated way?

(nothing mission critical though)

pkg/util/logging_http_transport.go
```diff
@@ -0,0 +1,74 @@
+// Copyright © 2018 The Knative Authors
```
toVersus:
```suggestion
// Copyright © 2019 The Knative Authors
```

pkg/util/logging_http_transport.go
```diff
@@ -0,0 +1,74 @@
...
+
+func NewLoggingTransport(transport http.RoundTripper) http.RoundTripper {
+	return &LoggingHttpTransport{transport, nil}
+}
+
+func NewLoggingTransportWithStream(transport http.RoundTripper, s io.Writer) http.RoundTripper {
+	return &LoggingHttpTransport{transport, s}
+}
+
+var SENSITIVE_HEADERS = map[string]bool{
```
toVersus: To use an empty struct instead of a boolean for saving cost (bytes) is my taste.

```golang
var SENSITIVE_HEADERS = map[string]struct{}{
	"Authorization":       {},
	"WWW-Authenticate":    {},
	"Cookie":              {},
	"Proxy-Authorization": {},
}

...

	for k, v := range r.Header {
		_, sensitive := SENSITIVE_HEADERS[k]
		if sensitive {
...


```

sixolet: I think the boolean is clearer, personally.

pkg/kn/commands/types_test.go
sixolet: Oops. Accidentally left a thing in from messing around.

pkg/util/logging_http_transport.go
sixolet: ... and I checked, it's the idiom https://blog.golang.org/go-maps-in-action

sixolet:
Let's merge this into `--verbose` as part of or shortly after that change?

rhuss:
> Let's merge this into `--verbose` as part of or shortly after that change?

Let's do it after this PR is merged. I opened #333  for this.

/approve
/lgtm

Issue #329, opened 2019-08-05
Integration test flake "service-account default not found"


rhuss:
There is currently an integration test flaking popping up from time to time with an error:

```
Running 'kn service create hello --image gcr.io/knative-samples/helloworld-go -n kne2etests2'...
Failed to successfully execute 'kn service create hello --image gcr.io/knative-samples/helloworld-go -n kne2etests3': Execution error: stderr: 'RevisionFailed: Revision "hello-qj4nm" failed with message: Unable to fetch image "gcr.io/knative-samples/helloworld-go": serviceaccounts "default" not found.
' error: 'exit status 1'FAIL	github.com/knative/client/test/e2e	1.024s
```

See e.g. https://storage.googleapis.com/knative-prow/pr-logs/pull/knative_client/312/pull-knative-client-integration-tests/1157313391212105729/build-log.txt

It smells like an issue in the cluster setup but not so much like a race as it happens after some successful image pulls in those tests.

@adrcunha do you have any idea what the cause could be ?

adrcunha:
This doesn't sound like a cluster setup to me. A few tests passed before the one failing, and the service account was (correctly) set before the tests. Since this is a flake, my guess is that one of the tests is doing some sort of teardown and thus causing  `TestRevision/create_hello_service_and_return_no_error` to fail.

```
=== RUN   TestRoute/create_hello_service_and_return_no_error
Running 'kn service create hello --image gcr.io/knative-samples/helloworld-go -n kne2etests4'...
=== RUN   TestServiceOptions/create_and_validate_service_with_concurrency_options
Running 'kn service create svc1 --image gcr.io/knative-samples/helloworld-go --concurrency-limit 250 --concurrency-target 300 -n kne2etests1'...
=== RUN   TestBasicWorkflow/returns_no_service_before_running_tests
Running 'kn service list -n kne2etests0'...
=== RUN   TestService/create_hello_service_duplicate_and_get_service_already_exists_error
Running 'kn service create hello --image gcr.io/knative-samples/helloworld-go -n kne2etests3'...
=== RUN   TestBasicWorkflow/create_hello_service_and_return_no_error
Running 'kn service create hello --image gcr.io/knative-samples/helloworld-go -n kne2etests0'...
=== RUN   TestRevision/create_hello_service_and_return_no_error
Running 'kn service create hello --image gcr.io/knative-samples/helloworld-go -n kne2etests2'...
[...]

PR #330, opened 2019-08-05
List services sorted by alphabetical order

### LINKED ISSUE #324: `kn service list` is not sorted
```
$ kn version
Version:      v20190731-71c39e8
Build Date:   2019-07-31 10:04:14
Git Revision: 71c39e8
Dependencies:
- serving:    v0.6.0
```

## Expected Behavior

`$ kn service list` lists services sorted by alphabetical order.

## Actual Behavior

`$ kn service list` does not seem to have any particular sorting.

daisy-ycguo:
Fixes #324 

## Proposed Changes

* `kn service list` lists services sorted by alphabetical order.

### DIFF
--- a/pkg/kn/commands/service/service_list.go
+++ b/pkg/kn/commands/service/service_list.go
@@ -16,6 +16,7 @@ package service
 
 import (
 	"fmt"
+	"sort"
 
 	"github.com/knative/client/pkg/kn/commands"
 	v1alpha12 "github.com/knative/client/pkg/serving/v1alpha1"
@@ -60,6 +61,11 @@ func NewServiceListCommand(p *commands.KnParams) *cobra.Command {
 			if err != nil {
 				return err
 			}
+
+			// Sort serviceList by name
+			sort.SliceStable(serviceList.Items, func(i, j int) bool {
+				return serviceList.Items[i].ObjectMeta.Name < serviceList.Items[j].ObjectMeta.Name
+			})
 
 			err = printer.PrintObj(serviceList, cmd.OutOrStdout())
 			if err != nil {
--- a/pkg/kn/commands/service/service_list_test.go
+++ b/pkg/kn/commands/service/service_list_test.go
@@ -76,9 +76,10 @@ func TestGetEmpty(t *testing.T) {
 }
 
 func TestServiceListDefaultOutput(t *testing.T) {
-	service1 := createMockServiceWithParams("foo", "http://foo.default.example.com", 1)
-	service2 := createMockServiceWithParams("bar", "http://bar.default.example.com", 2)
-	serviceList := &v1alpha1.ServiceList{Items: []v1alpha1.Service{*service1, *service2}}
+	service1 := createMockServiceWithParams("foo", "http://foo.default.example.com", 2)
+	service3 := createMockServiceWithParams("sss", "http://sss.default.example.com", 3)
+	service2 := createMockServiceWithParams("bar", "http://bar.default.example.com", 1)
+	serviceList := &v1alpha1.ServiceList{Items: []v1alpha1.Service{*service1, *service2, *service3}}
 	action, output, err := fakeServiceList([]string{"service", "list"}, serviceList)
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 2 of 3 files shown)


pkg/kn/commands/service/service_list.go
```diff
@@ -52,6 +53,10 @@ func NewServiceListCommand(p *commands.KnParams) *cobra.Command {
 			if err != nil {
 				return err
 			}
+			//sort serviceList by name
+			sorter := NewServicesSorter(serviceList)
+			sort.Sort(sorter)
```
navidshaikh: How about sorting the list just before calling `PrintObj`?

daisy-ycguo: Good suggestion. Moved just before PrintObj. Thanks.

pkg/kn/commands/service/service_list_test.go
```diff
@@ -88,9 +89,11 @@ func TestServiceListDefaultOutput(t *testing.T) {
 	} else if !action.Matches("list", "services") {
 		t.Errorf("Bad action %v", action)
 	}
+	//sorted output
```
edwardstudy: same comment format as above

pkg/kn/commands/service/service_list.go
```diff
@@ -61,6 +62,10 @@ func NewServiceListCommand(p *commands.KnParams) *cobra.Command {
 				return err
 			}
 
+			//sort serviceList by name
```
edwardstudy: nit: Comments should be capitalized and have one space at the beginning

navidshaikh:
@daisy-ycguo : I think the service list we receive from server comes sorted by name already, check https://github.com/knative/client/issues/324#issuecomment-518606771

navidshaikh:
@daisy-ycguo : I think we could instead update this PR to ensure (by adding unit tests) that the list printed is sorted.

pkg/kn/commands/service/service_list.go
```diff
@@ -61,6 +62,10 @@ func NewServiceListCommand(p *commands.KnParams) *cobra.Command {
...
+			sorter := NewServicesSorter(serviceList)
+			sort.Sort(sorter)
```
navidshaikh: If we are to implement this sorting via `sort.Sort`, I'd recommend renaming to re-present the sorting key, for example:
```
sort.Sort(ByServiceName[serviceList])
```

makes the operation readable.

or, we could simply use this one liner

```suggestion
			sort.SliceStable(serviceList.Items, func(i, j int) bool { return serviceList.Items[i].Name < serviceList.Items[j].Name })
```

navidshaikh:
> I think we could instead update this PR to ensure (by adding unit tests) that the list printed is sorted.

@daisy-ycguo : Ignore this comment as mentioned on the issue that other implementation might not return the sorted service list.

pkg/kn/commands/service/service_list.go
sixolet: I like having the named sorter, and I agree with your naming, Navid.

One thing to make it simpler and more idiomatic is to use the pattern the sort library built into golang does: instead of making a struct with a ServiceList in it, make a typedef for ServiceList with the appropriate methods:

Example from golang source https://golang.org/src/sort/sort.go?s=7078:7105#L270

```
type ByServiceName *v1alpha1.ServiceList

func (s ByServiceName) Len() int { ... }
```
etc

pkg/kn/commands/service/service_list_test.go
daisy-ycguo: Fixed

pkg/kn/commands/service/service_list.go
daisy-ycguo: fixed

pkg/kn/commands/service/service_list.go
daisy-ycguo: Thanks for the suggestion. I used Navid's suggested change, because it looks much simpler.

daisy-ycguo: Please review again.

rhuss:
The  integration test flake

```
Failed to successfully execute 'kn service create hello --image gcr.io/knative-samples/helloworld-go -n kne2etests1': Execution error: stderr: 'RevisionFailed: Revision "hello-l58wg" failed with message: Unable to fetch image "gcr.io/knative-samples/helloworld-go": serviceaccounts "default" not found.
```

is really annoying. 'hope we get that fixed but no idea how (see #329 for tracking this). Seems that it happens more often recently.

daisy-ycguo:
> @daisy-ycguo : Thanks! Please update the CHANGELOG.

I did. Thanks for the instruction.

rhuss:
thanks !

/lgtm
/approve

Issue #333, opened 2019-08-06
Introduce --debug and --verbose as global options


rhuss:
## Describe the feature:

Debugging and verbose output should be treated in a consistent way over all commands. Currently, we have 

* `kn plugin list --verbose` as in #313 for printing out plugin configuration
* `kn service describe --details` as in #252 for showing additional service information
* `--log-http` as a global option to track HTTP API communication coming in with #326

As there are supposed to be more such output modifying flags coming in over time, I suggest to use `--debug` and `--verbose` with categories to fine tune this:

* Adding `--debug` for very low-level information. Only useful for developer of `kn` itself
* Adding a `--verbose` option for increasing the detail level of the information presented. Examples are
  * `kn service describe --verbose` to add more details to a service (like the image digests, or all env vars/annotations/labels)
  * `kn plugin list --verbose` to also print the plugin directories where plugins are looked up
  * `kn service create --debug` to print also the HTTP headers

In order to tune debug and verbose output, I'd suggest to introduce categories:

* `--debug` --> all debug info
* `--debug=all` --> all debug info
* `--debug=http` --> only http related debugging
* `--verbose` --> all  verbose output
* `--verbose=config` --> config related information, too
* `--verbose=config,details` --> config information and details (made up here to show how to combine categories)

This suggestion would introduce exactly two options, `--debug` and `--verbose` (with the possible values shown in the help page). The alternative would be to have an extra option for every category like in `--log-http`, but this would blow up the help pages and UX surface considerably over time.

rhuss:
Should be tackled together with #182 IMO

rhuss:
Related to this issue 

* #182 -- `--quiet` is kind of the opposite of `--verbose`
* #390 -- `--log-http` should be included as part of `--debug=http`

PR #339, opened 2019-08-06
Don't show warnings that are not bad conditions

duglin:
Don't show warnings while running `kn plugin list` when:
- a dir in your plug-in path doesn't exist - e.g. right now it shows it for a missing ~/.kn/plugins - which is a valid state for someone who just downloaded `kn` but doesn't have, or need, a .kn dir
- there are no plugins at all - that's a valid state and doesn't deserve a warning

To me warning are things people need to be aware of (or things they need to
take action to fix).  Neither of those case apply here. Instead they user
will just be annoyed by this:

```
$ kn plugin list
Unable read directory '/root/.kn/plugins' from your plugins path: open /root/.kn/plugins: no such file or directory. Skipping...
warning: unable to find any kn plugins in your plugin path: '[/root/.kn/plugins]'
```

When they should see no output at all. Or, if someone wants to add it we could output `no plugins found` but not call it a "warning" since it's not.

/lint

### DIFF
--- a/pkg/kn/commands/plugin/plugin_list.go
+++ b/pkg/kn/commands/plugin/plugin_list.go
@@ -110,7 +110,6 @@ func (o *PluginFlags) complete(cmd *cobra.Command) error {
 }
 
 func (o *PluginFlags) run() error {
-	pluginsFound := false
 	isFirstFile := true
 	pluginErrors := []error{}
 	pluginWarnings := 0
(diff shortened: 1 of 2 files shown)


rhuss:
Fixed with #313

rhuss:
@duglin would it be ok to close this and focus on #313 ? Happy to adapt this over there, but as #313 has a wider scope and also include a fix for this issue, I would love to continue over there to avoid fixing conflicts.

duglin:
@rhuss as long as the conditions I listed don't result in warning then yes!  :-)

duglin:
Does #313 cover the case of a dir in the plugin path not existing?  Can't tell for sure from the description

maximilien:
PR #313 changes how you would do this. I also think using the verbose feature #333 is also how more or less output should be controlled.

rhuss:
ok, let's close this and continue on #333

PR #342, opened 2019-08-06
Service and Revision labels

### LINKED ISSUE #328: Adding label
## In what area(s)?
`kn` command line option

Classifications:
> /kind feature

## Describe the feature:

Ability to add user defined labels to the service and its artifacts. Certain implementation specific features depend on labels. Being able to add a label with `kn` will avoid additional overhead of unnecessarily updating the service (and avoiding a new revision) just for adding a label.

Example
`kn service create myservice --label=app.kubernetes.io/part-of=mygroup`

aaron-lerner:
Fixes #328

## Proposed Changes

* Adds a --label (alias -l) flag to set service and revision template labels
* Allow unsetting env vars and labels

**Release Note**

```release-note
Added --label (alias -l) on `kn service create` and `kn service update` commands to set service and revision template labels.
```


pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -165,3 +175,15 @@ func (p *ConfigurationEditFlags) computeResources(resourceFlags ResourceFlags) (
 
 	return resourceList, nil
 }
+
```
sixolet: Move into `/pkg/kn/commands/`, probably in a new file.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -35,6 +35,7 @@ type ConfigurationEditFlags struct {
 	ConcurrencyTarget          int
 	ConcurrencyLimit           int
 	Port                       int32
+	Label                      []string
```
duglin: s/Label/Labels/ since its an array

pkg/serving/config_changes.go
```diff
@@ -152,6 +152,17 @@ func UpdateResources(template *servingv1alpha1.RevisionTemplateSpec, requestsRes
...
 }
 
+// UpdateServiceLabels updates the labels on a service
+func UpdateServiceLabels(service *servingv1alpha1.Service, vars map[string]string) error {
+	if service.ObjectMeta.Labels == nil {
+		service.ObjectMeta.Labels = make(map[string]string)
+	}
+	for key, value := range vars {
+		service.ObjectMeta.Labels[key] = value
+	}
```
sixolet:
```
if value == "" {
  delete(service.ObjectMeta.Labels, key)
}

File: pkg/kn/commands/service/configuration_edit_flags.go (hunk 1 above)
Comment (sixolet) [reply pkg/kn/commands/service/configuration_edit_flags.go]: ... and write a test for it separately :)

File: pkg/kn/commands/service/configuration_edit_flags.go [hunk 4]
```diff
@@ -165,3 +175,15 @@ func (p *ConfigurationEditFlags) computeResources(resourceFlags ResourceFlags) (
 
 	return resourceList, nil
 }
+
+func (p *ConfigurationEditFlags) mapFromArray(arr []string, delimiter string, flag string) (map[string]string, error) {
+	returnMap := map[string]string{}
+	for _, pairStr := range arr {
+		pairSlice := strings.SplitN(pairStr, delimiter, 2)
+		if len(pairSlice) <= 1 {
+			return nil, fmt.Errorf("%s argument requires a value that contains the '=' character; got %s", flag, pairStr)
```
duglin: use %q instead so it quotes them, otherwise sometimes that work plus the rest of the sentence could make sense as an english sentence and confuse people.

pkg/kn/commands/service/service_create_test.go
```diff
@@ -467,3 +467,30 @@ func TestServiceCreateEnvForce(t *testing.T) {
 		t.Fatalf("wrong output: %s", output)
 	}
 }
+
+func TestServiceCreateLabel(t *testing.T) {
+	action, created, _, err := fakeServiceCreate([]string{
+		"service", "create", "foo", "--image", "gcr.io/foo/bar:baz", "-l", "a=mouse", "--label", "b=cookie", "--async"}, false, false)
```
duglin: just wondering, does the option parsing support: `--label=foo=bar` ? If so, we should probably test for that variant too.

duglin:
overall looks good. Did you want to support unsetting a label too?

pkg/kn/commands/service/service_create_test.go
aaron-lerner: I'm not sure what you mean, there's already a `--label b=cookie`. Unless you mean the difference between `--label=foo=bar` and `-label foo=bar`?

duglin: Yes I mean —label=foo=bar . Some options parsers support the merging of the option name with the value all in one cmd line arg separated but an =  So if ours does we should probably test it.  Especially in the case f removing a label.

pkg/serving/config_changes.go
aaron-lerner: Thanks, meant to ask about that.

pkg/kn/commands/service/service_create_test.go
aaron-lerner: It looks like cobra does support that. I added a test case to `service_update_test.go` specifically for removing a label using that format. However, it doesn't seem that any other flag test uses that formatting.

pkg/kn/commands/service/configuration_edit_flags.go
aaron-lerner: I didn't love having to pass the flag but wasn't aware wrapping was an option. Thanks for the suggestion!

aaron-lerner:
Added support for unsetting by passing an empty value, ie `--label foo=` will remove the label named `foo`.

rhuss:
> Added support for unsetting by passing an empty value, ie `--label foo=` will remove the label named `foo`.

Does the `env` var support this, too ? It should be treated the same imo. AFAIK, empty label values, are allowed, too, so not sure how this could be achieved if an empty value means to remove a label.

rhuss:
I'm also wondering whether how we could allow setting labels on services, but also on the pod templates. I think both might be useful, and I wonder how the UX for this would look like.

* `--labels` for service labels
* `--pod-labels` (?) for labels on pods

Or should `--labels` set the label on both ? Would be the simplest and maybe that's already good enough ?

rhuss:
I see service labels are already applied to the generated pods (if I understand that correctly :). So then please forget my last comment :)

rhuss:
BTW, you see how a revision get the labels from the template, not the sevice itself here: https://github.com/knative/serving/blob/c6860bf7c0663ddbf71d8fa3b48cbfbd9b54c205/pkg/reconciler/configuration/resources/revision.go#L30-L33

pkg/serving/config_changes.go
```diff
@@ -152,6 +152,22 @@ func UpdateResources(template *servingv1alpha1.RevisionTemplateSpec, requestsRes
 	return nil
 }
 
+// UpdateServiceLabels updates the labels on a service
+func UpdateServiceLabels(service *servingv1alpha1.Service, vars map[string]string) error {
+	if service.ObjectMeta.Labels == nil {
+		service.ObjectMeta.Labels = make(map[string]string)
```
rhuss: The question is whether the labels of the service or the labels of the revision template which should be set.

aaron-lerner:
> Does the `env` var support this, too ? It should be treated the same imo. AFAIK, empty label values, are allowed, too, so not sure how this could be achieved if an empty value means to remove a label.

It doesn't look like `kn` has a way to remove an env var once set. If you pass an empty string, the env var remains in the spec but its `Value` field is unset. I'm not too sure on the utility of a set, but empty env var or label. Are they a feature that needs to be supported? It seems adding the same functionality to unset an env var with an empty string would be convenient.

Do you have any suggestions on an alternative way to unset env vars and labels? I suppose a simple option is to add a `--remove-[env|label]` flag. It looks like [kubectl has you to pass `VAR-`](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#-em-env-em-) where the minus sign indicates removal.

pkg/serving/config_changes.go
aaron-lerner: I think this should probably be decided in #328. I like the simplicity of `--label` setting both, but not sure if preventing users from setting separate labels on services and revision templates would be an issue.

rhuss: > Do you have any suggestions on an alternative way to unset env vars and labels? I suppose a simple option is to add a `--remove-[env|label]` flag. It looks like [kubectl has you to pass `VAR-`](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#-em-env-em-) where the minus sign indicates removal.

yes, I think we should settle on a similar convention. It's just people sometimes use just empty env vars as kind of a boolean flag (not recommended though). But we need a way to differentiate between an empty value (key with empty value) and not-existing entry (no key)

rhuss:
> Do you have any suggestions on an alternative way to unset env vars and labels? I suppose a simple option is to add a `--remove-[env|label]` flag. It looks like [kubectl has you to pass `VAR-`](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#-em-env-em-) where the minus sign indicates removal.

yes, I think we should settle on a similar convention. It's just people sometimes use just empty env vars as kind of a boolean flag (not recommended though). But we need a way to differentiate between an empty value (key with empty value) and not-existing entry (no key)

pkg/serving/config_changes.go
rhuss: My gut feeling is that people most often want to set labels for their pods, not the service. So we could settle on something like that:

* `--label mylabel=myvalue` would set it on both, the service and the pod
* `--label ^mylabel=myvalue` would set it only on the `Service` (the "outer" entity)
* `--label  :mylabel=myvalue` would set it only on the revision template (the "inner" entity)

That would cover 90% of all use case by setting both, but allows for tuning, too. Not sure about te concrete characters (one should also check whether there are special shell meanings so that those chars can be used unescaped), but using any character which can not be used within label keys should be cool.

wdyt ?

aaron-lerner: It seems at that point there might as well be separate `--revision-label`, `--service-label` , and `--label` flags. It would be much more explicit both in documentation and I think code that way. Plus, special characters change (slightly) across different shell environments and if [this list](https://docstore.mik.ua/orelly/unix/upt/ch08_19.htm) is any indication, it may not be worth the time and effort to find and confirm which characters are valid in a reasonable number of shell environments.

aaron-lerner:
I like the idea of using the minus, I suppose it can't be in front or it may be confused as another flag. To use it, I'd suggest something like:

* `--env KEY=value` sets `EnvVar{Name: "KEY", Value: "value"}` (current behavior)
* `--env KEY=` sets `EnvVar{Name: "KEY"}` (current behavior)
* `--env KEY` invalid
* `--env KEY-` removes the env var matching `EnvVar{Name: "KEY", ...}`
* `--env KEY-=` sets `EnvVar{Name: "KEY-"}` (current behavior)

And a similar behavior for labels. Thoughts?

pkg/kn/commands/parsing_helper_test.go
```diff
@@ -0,0 +1,49 @@
...
+package commands
+
+import (
+	"reflect"
+	"testing"
+
+	"gotest.tools/assert"
+)
+
+func TestMapFromArray(t *testing.T) {
```
maximilien: What about being defensive for user errors, e.g., leaving extra “,” or “=“ at the end as well as other similar cases. At a minimum should error out or warn.

pkg/serving/config_changes.go
maximilien: I too like the simplicity of `—label` for all labels as the object is already decided in the command group. Adding more complexity and notations to remember is unnecessary as it complicates the “language” of the CLI IMO.

pkg/kn/commands/parsing_helper_test.go
aaron-lerner: A delimiter at the end (e.g. `--label empty=`) is valid for setting an empty string as the value. An extra delimiter is considered a regular character and included in the value (e.g. `--label key=value=` leads to `{"key": "value="}`. I added a few more test cases to showcase this behavior. Is that what you meant or are there other cases you were thinking of that should error/warn?

aaron-lerner:
Changed the unset behavior to match kubectl. So pass the name/key of the environment variable followed by `-` to unset, e.g. `--env NAME-` unsets the environment variable with name `NAME`, same behavior for --label. Let me know if you'd like the changes to --env to be moved out into another PR, I don't think all the changes necessarily need to be entangled in this one, although they are heavily related.

With this, you can now set a label with an empty string and the already existing behavior of setting an environment variable with no value still works.

rhuss:
> * `--env KEY=value` sets `EnvVar{Name: "KEY", Value: "value"}` (current behavior)

:+1:

> * `--env KEY=` sets `EnvVar{Name: "KEY"}` (current behavior)
> * `--env KEY` invalid

Not sure whether both cases can be easily distinguished with Cobra. If not, I would think treat both variants as the same

> * `--env KEY-` removes the env var matching `EnvVar{Name: "KEY", ...}`

👍  as this is similar to the the kubectl behaviour.

> * `--env KEY-=` sets `EnvVar{Name: "KEY-"}` (current behavior)

This is an invalid env var name, an env var can't end with a `-`. Again not sure whether it can be easily distinguished from the case above, so I would just treat it the same as above, i.e. deleting the env variable.

pkg/serving/config_changes.go
rhuss: > It seems at that point there might as well be separate `--revision-label`, `--service-label` , and `--label` flags. It would be much more explicit both in documentation and I think code that way. Plus, special characters change (slightly) across different shell environments and if [this list](https://docstore.mik.ua/orelly/unix/upt/ch08_19.htm) is any indication, it may not be worth the time and effort to find and confirm which characters are valid in a reasonable number of shell environments.

This is always a tradeoff between options bloat and being to obfuscated. Currently there is a tendency to add 1-5 new options per new features, and I'm doing my best to fight against the bloat. At least asking, whether there could be a compromise and whether all those options are needed (e.g. see https://github.com/knative/client/pull/345#issuecomment-519385669 or https://github.com/knative/client/pull/326#pullrequestreview-270639537 or https://github.com/knative/client/pull/282#pullrequestreview-271866769 how to reduce the UX surface and so the cognitive load). 

Of course, we should no go over the top, and using special chars like that might be the case here (although `^` and `:` are safe shell chars so no need to escape).

What's about that compromise: Let's implement `--label` to label both, service and template and wait on feedback if the other use cases (individual labelling of serving/template) are actually requested ?

rhuss: From a single feature's PR POV adding 3 new options might not be a big deal, but its the sum which might lead to multi page help outputs which are really overwhelming (especially when alphabetically ordered, and `--revision-label`, `--sevice-label` and `--label` will apear on totally different positions in a list of 50+ options)

pkg/serving/config_changes.go
rhuss: ? is this comment still valid as we have now a dedicated mechanism to remove labels (in contrast to having empty labels, which is also allowed)

pkg/serving/config_changes_test.go
```diff
@@ -270,11 +304,89 @@ func testUpdateEnvVarsBoth(t *testing.T, template *servingv1alpha1.RevisionTempl
...
+	assert.NilError(t, err)
+	expected := map[string]string{
+		"a": "notfoo",
+		"b": "bar",
+		"c": "bat",
+		"d": "",
+	}
+
+	actual := service.ObjectMeta.Labels
+	if !reflect.DeepEqual(expected, actual) {
```
rhuss: again, my recommendation 

```
assert.Assert(t, cmp.DeepEqual(expected, actual))
```

applies to the other tests, too.

navidshaikh:
> --env KEY=value sets EnvVar{Name: "KEY", Value: "value"} (current behavior)
> --env KEY= sets EnvVar{Name: "KEY"} (current behavior)
> --env KEY- removes the env var matching EnvVar{Name: "KEY", ...}

+1 for support these operations. 

> --env KEY invalid

We should treat this same as `--env KEY=`. I agree with @rhuss .

pkg/serving/config_changes.go
aaron-lerner: No, this is no longer valid.

aaron-lerner:
> This is an invalid env var name, an env var can't end with a `-`. I would just treat it the same as above, i.e. deleting the env variable.

I assumed the lack of validation client-side on names and values meant this was being done server-side. However, I am allowed to set an environment variable with a name ending in `-`. Are you suggesting adding client-side validation to names in this PR? I'm afraid something like that starts to bloat it a bit too much.

rhuss:
> I assumed the lack of validation client-side on names and values meant this was being done server-side. However, I am allowed to set an environment variable with a name ending in `-`. Are you suggesting adding client-side validation to names in this PR? I'm afraid something like that starts to bloat it a bit too much.

I just meant that `--env VAR-` and `--env VAR-=` should be treated the same, i.e. deleting the variable. No need to do further validation: If the var ends with `-` then delete it (if present, otherwise ignore)

pkg/serving/config_changes.go
maximilien: OK on compromise!

pkg/kn/commands/service/service_create_test.go
```diff
@@ -467,3 +467,30 @@ func TestServiceCreateEnvForce(t *testing.T) {
 		t.Fatalf("wrong output: %s", output)
 	}
 }
+
+func TestServiceCreateLabel(t *testing.T) {
+	action, created, _, err := fakeServiceCreate([]string{
```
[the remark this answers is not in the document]
aaron-lerner: Haven't gotten to this yet

aaron-lerner:
Updated to have env and label keys passed without values map to empty values rather than throw an error. Additionally, keys that indicate deletion (`-` suffix) will always attempt to remove the associated env/label, regardless of whether or not a value is included with them.

pkg/serving/config_changes.go
sixolet: Aside: I'm starting to think we should implement our own flags groupings, which might help with this.

aaron-lerner: spf13/cobra#836 has some discussion on flag groupings in help text. Unfortunately, it doesn’t look like there’s currently first class support for it.

pkg/kn/commands/service/service_create_test.go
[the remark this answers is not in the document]
aaron-lerner: Done

PR #343, opened 2019-08-06
Only test in watch mode if passed test flag

aaron-lerner:
I believe this should be the behavior, based on https://github.com/knative/client/blob/master/DEVELOPMENT.md

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -49,7 +49,10 @@ run() {
   if $(has_flag --watch -w); then
     # Build and test first
     go_build
-    go_test
+
+    if $(has_flag --test -t); then
+       go_test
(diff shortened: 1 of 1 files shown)


sixolet:
From the doc ref: 
> build.sh - Compile, test, generate docs and format source code

... should test by default.

aaron-lerner:
This is for watch mode, as opposed to watch *and* test mode:

> `build.sh -w` - Enter watch mode for automatic recompilation
> `build.sh -w -t` - Enter watch mode for automatic recompilation & running tests

After entering watch mode it only tests if the `-t` flag is passed. It seems odd to initially test without the flag.

Issue #303, opened 2019-07-26
kn and Cloud Run fully managed


steren:
Classifications:
/kind feature
/kind proposal

Cloud Run (https://cloud.run) is an implementation of the knative/serving API specification and runtime container contract on top of Google Cloud's serverless stack.

I could not find in the `kn` docs how to configure `kn` to manipulate Cloud Run resources. I propose that we ensure `kn` works with Cloud Run and with other potential future implementations of the knative/serving spec.

The Cloud Run control plane:

* currently implements portions of the Knative/serving  API spec (accepting both `v1alpha1` and `v1beta1` shapes, but returning `v1beta1` shape under a `v1alpha1` endpoint at the moment),
* is using the same [authentication mechanism as other Google APIs](https://developers.google.com/identity/protocols/OAuth2),
* does not expose any underlying Kubernetes resources,
* does not support write operations on KRoute and KConfigurations.
* supports `GET`, `POST` and `PUT`, but does not support `PATCH` as far as I know.
* is documented [here](https://cloud.google.com/run/docs/reference/rest/)

For example, to manipulate services in the region `us-central1` in GCP project `my-project`, the following URI should be used: `https://us-central1-run.googleapis.com/apis/serving.knative.dev/v1alpha1/namespaces/my-project/services`

I am the Product Manager for Cloud Run. I hope to find time to contribute, but my experience is limited to a single digit number of Go programming hours (maybe one day I'll work on a JS client :)

duglin:
What kind of stuff are you looking for? Once you have your CLI environment (ala `kubectl`/KUBECONFIG) setup to point to your cluster, `kn` should just work.

rhuss:
>     * only support write operations on KServices (KRevisions, KRoute and KConfigurations are read-only)

I understand that Configuration and Routes are fully managed within a Service, but how can you clean up Revisions in Google Cloud Run ? (i.e. delete a revision). This is something that kn supports via `kn revision delete`.

rhuss:
[...]

PR #344, opened 2019-08-06
Connect using an OAuth2 JWT flow.

sixolet:
Allows you to connect using an oauth2 jwt flow if you have an appropriate json jwt key file.

An example of a "json jwt key file" is a Google Cloud Service account, so this fully solves #303 if you also have a kubeconfig about this minimal included:

```
apiVersion: v1
Kind: Config
clusters:
- cluster:
    server: https://us-central1-run.googleapis.com
  name: us-central1
users:
- name: u
contexts:
- context:
    user: u
    cluster: us-central1
    namespace: lol-its-my-project
  name: c
current-context: c
```

Invocation: 
```
kn service list  --jwt-key ~/my-key.json  --jwt-scope "https://www.googleapis.com/auth/cloud-platform"
```

The goal is to have this be an option for easily configuring kn from inside CI (and be easier than having to have a second oauth token printing command installed) if what you have is a jwt private key file.


sixolet:
(experimental so far; doesn't have tests yet)

rhuss:
Just to understand the scope of this PR:

* The functionality provided here could be equally achieved by preparing a proper kubeconfig file and using that with `--kubeconfig`. Is this correct ?
* Which other platforms (except of GCP) uses the same JWT based authentication ?

Without judging the usefulness of this PR, I see that it adds three new global options for a very specific use case. My general attitude is that we keep options at the possible minimum to not overwhelm users and so I always looking for how to keep the UX surface small or even reduce it (see also the discussion around logging/verbose/debug).  Adding three global options for a valid, but small use-case doesn't feel right to me.

I understand that it's easier to provide options to `kn` but what's about of having another (non-kn) program which just does the same but writes out that kubeconfig file which then could be used by kn.  This could be written in a single go file, compiled on the fly and run before `kn` is called. Not much more work for a CI and would help to keep kn focussed.

sixolet:
> The functionality provided here could be equally achieved by preparing a proper kubeconfig file and using that with --kubeconfig. Is this correct ?

Unfortunately, you can't get a kubeconfig that does this on its own, usefully. The closest you can get is one that shells out to another command to go do an oauth flow and print out an appropriate bearer token, which is the walkthrough I listed in #303 — but that requires you have other software installed and configured. Alternately you can provide the bearer token in the kubectl, but then it'll expire in an hour. This document explains it: https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens

> Which other platforms (except of GCP) uses the same JWT based authentication ?

According to that article, "notably Azure Active Directory, Salesforce, and Google"

> I understand that it's easier to provide options to kn but what's about of having another (non-kn) program which just does the same but writes out that kubeconfig file which then could be used by kn. This could be written in a single go file, compiled on the fly and run before kn is called. Not much more work for a CI and would help to keep kn focussed.

This might be workable, but I want to figure out something where a CI flow makes sense using a standard kn deployer image, even against providers that use jwt auth.  Maybe the answer is a previous step in CI needs to write the kubeconfig for them?

sixolet:
... putting on hold for discussion

rhuss:
@sixolet I wonder how we should proceed with this PR ? 

It would need some major refactoring because of the change in global config handling.
Do you want to continue to work on this PR or should be closed and the feature revisited at a later point in time ?

rhuss:
@sixolet this PR is now nearly 1 year old. How should we progress ? I suggest we close it for now and then reopen it later ?

If you don't mind and you don't have any objections until next Tuesday, I would close the PR (but feel free to reopen it any time).

rhuss:
closing it for now. feel free to rebase and reopen if the PR is still needed.

PR #345, opened 2019-08-07
feature(service): Implements traffic splitting and tagging targets

### LINKED ISSUE #210: Support setting up traffic splitting of service

navidshaikh:
Fix #210 
  
 Introduced 3 flags for working with traffic: `--traffic`, `--tag`,  `--untag`.
 Use `@latest` identifier to reference latest revision for `--traffic` and `--tag` flags.

 Added first cut, please give it a try by building the client from this PR,
 see if there are use cases you need and not covered here.

 PR is WIP and requires following to be complete:
- ~~Add check to verify if given revision exists~~ (open https://github.com/knative/serving/issues/5094)
 - [x] Better error handling
 - [x] Better error reporting
 - [x] Code comments
 - [x] Unit tests
 - [x] e2e tests
 - [x] Fix marked TODOs
 - [x] client side verification for traffic percent sum to 100

### DIFF
--- a/docs/cmd/kn_service_update.md
+++ b/docs/cmd/kn_service_update.md
@@ -14,14 +14,25 @@ kn service update NAME [flags]
 
 ```
 
-  # Updates a service 'mysvc' with new environment variables
-  kn service update mysvc --env KEY1=VALUE1 --env KEY2=VALUE2
+  # Updates a service 'svc' with new environment variables
+  kn service update svc --env KEY1=VALUE1 --env KEY2=VALUE2
 
-  # Update a service 'mysvc' with new port
-  kn service update mysvc --port 80
+  # Update a service 'svc' with new port
+  kn service update svc --port 80
 
-  # Updates a service 'mysvc' with new requests and limits parameters
-  kn service update mysvc --requests-cpu 500m --limits-memory 1024Mi
+  # Updates a service 'svc' with new requests and limits parameters
+  kn service update svc --requests-cpu 500m --limits-memory 1024Mi
+
+  # Assign tag 'latest' and 'stable' to revisions 'echo-v2' and 'echo-v1' respectively
+  kn service update svc --tag echo-v2=latest --tag echo-v1=stable
+  OR
+  kn service update svc --tag echo-v2=latest,echo-v1=stable
+
+  # Update tag from 'testing' to 'staging' for latest ready revision of service
+  kn service update svc --untag testing --tag @latest=staging
+
+  # Add tag 'test' to echo-v3 revision with 10% traffic and rest to latest ready revision of service
+  kn service update svc --tag echo-v3=test --traffic test=10,@latest=90
 ```
 
 ### Options
@@ -43,6 +54,9 @@ kn service update NAME [flags]
       --requests-cpu string      The requested CPU (e.g., 250m).
       --requests-memory string   The requested memory (e.g., 64Mi).
       --revision-name string     The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
+      --tag strings              Set tag (format: --tag revisionRef=tagName) where revisionRef can be a revision or '@latest' string representing latest ready revision. This flag can be specified multiple times.
+      --traffic strings          Set traffic distribution (format: --traffic revisionRef=percent) where revisionRef can be a revision or a tag or '@latest' string representing latest ready revision. This flag can be given multiple times with percent summing up to 100%.
+      --untag strings            Untag revision (format: --untag tagName). This flag can be spcified multiple times.
       --wait-timeout int         Seconds to wait before giving up on waiting for service to be ready. (default 60)
 ```
 
--- a/pkg/kn/commands/flags/traffic.go
+++ b/pkg/kn/commands/flags/traffic.go
@@ -1,0 +1,56 @@
+// Copyright © 2019 The Knative Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package flags
+
+import (
+	"github.com/spf13/cobra"
+)
+
+type Traffic struct {
+	RevisionsPercentages []string
+	RevisionsTags        []string
+	UntagRevisions       []string
+}
+
+func (t *Traffic) Add(cmd *cobra.Command) {
+	cmd.Flags().StringSliceVar(&t.RevisionsPercentages,
+		"traffic",
+		nil,
+		"Set traffic distribution (format: --traffic revisionRef=percent) where revisionRef can be a revision or a tag or '@latest' string "+
+			"representing latest ready revision. This flag can be given multiple times with percent summing up to 100%.")
+
+	cmd.Flags().StringSliceVar(&t.RevisionsTags,
+		"tag",
+		nil,
+		"Set tag (format: --tag revisionRef=tagName) where revisionRef can be a revision or '@latest' string representing latest ready revision. "+
+			"This flag can be specified multiple times.")
+
+	cmd.Flags().StringSliceVar(&t.UntagRevisions,
+		"untag",
+		nil,
+		"Untag revision (format: --untag tagName). This flag can be spcified multiple times.")
+}
+
+func (t *Traffic) PercentagesChanged(cmd *cobra.Command) bool {
+	return cmd.Flags().Changed("traffic")
+}
+
+func (t *Traffic) TagsChanged(cmd *cobra.Command) bool {
+	return cmd.Flags().Changed("tag") || cmd.Flags().Changed("untag")
+}
+
+func (t *Traffic) Changed(cmd *cobra.Command) bool {
+	return t.PercentagesChanged(cmd) || t.TagsChanged(cmd)
+}
--- a/pkg/kn/traffic/compute.go
+++ b/pkg/kn/traffic/compute.go
@@ -1,0 +1,369 @@
+// Copyright © 2019 The Knative Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package traffic
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/knative/client/pkg/kn/commands/flags"
+	"github.com/knative/pkg/ptr"
+	"github.com/knative/serving/pkg/apis/serving/v1alpha1"
+	"github.com/spf13/cobra"
+)
+
+var latestRevisionRef = "@latest"
+
+// ServiceTraffic type for operating on service traffic targets
+type ServiceTraffic []v1alpha1.TrafficTarget
+
+func newServiceTraffic(traffic []v1alpha1.TrafficTarget) ServiceTraffic {
+	return ServiceTraffic(traffic)
+}
+
+func splitByEqualSign(pair string) (string, string, error) {
+	parts := strings.Split(pair, "=")
+	if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+		return "", "", fmt.Errorf("expecting the value format in value1=value2, given %s", pair)
+	}
+	return parts[0], strings.TrimSuffix(parts[1], "%"), nil
+}
+
+func newTarget(tag, revision string, percent int, latestRevision bool) (target v1alpha1.TrafficTarget) {
+	target.Percent = percent
+	target.Tag = tag
+	if latestRevision {
+		target.LatestRevision = ptr.Bool(true)
+	} else {
+		// as LatestRevision and RevisionName can't be specfied together for a target
+		target.LatestRevision = ptr.Bool(false)
+		target.RevisionName = revision
+	}
+	return
+}
+
+func (e ServiceTraffic) isTagPresentOnRevision(tag, revision string) bool {
+	for _, target := range e {
+		if target.Tag == tag && target.RevisionName == revision {
+			return true
+		}
+	}
+	return false
+}
+
+func (e ServiceTraffic) tagOfLatestReadyRevision() string {
+	for _, target := range e {
+		if *target.LatestRevision {
+			return target.Tag
+		}
+	}
+	return ""
+}
+
+func (e ServiceTraffic) isTagPresent(tag string) bool {
+	for _, target := range e {
+		if target.Tag == tag {
+			return true
+		}
+	}
+	return false
+}
+
+func (e ServiceTraffic) untagRevision(tag string) {
+	for i, target := range e {
+		if target.Tag == tag {
+			e[i].Tag = ""
+			break
+		}
+	}
+}
+
+func (e ServiceTraffic) isRevisionPresent(revision string) bool {
+	for _, target := range e {
+		if target.RevisionName == revision {
+			return true
+		}
+	}
+	return false
+}
+
+func (e ServiceTraffic) isLatestRevisionTrue() bool {
+	for _, target := range e {
+		if *target.LatestRevision == true {
+			return true
+		}
+	}
+	return false
+}
+
+// TagRevision assigns given tag to a revision
+func (e ServiceTraffic) TagRevision(tag, revision string) ServiceTraffic {
+	for i, target := range e {
+		if target.RevisionName == revision {
+			if target.Tag != "" { // referenced revision is requested to have multiple tags
+				break
+			} else {
+				e[i].Tag = tag // referenced revision doesn't have tag, tag it
+				return e
+			}
+		}
+	}
+	// append a new target if revision doesn't exist in traffic block
+	// or if referenced revision is requested to have multiple tags
+	e = append(e, newTarget(tag, revision, 0, false))
+	return e
+}
+
+// TagLatestRevision assigns given tag to latest ready revision
+func (e ServiceTraffic) TagLatestRevision(tag string) ServiceTraffic {
+	for i, target := range e {
+		if *target.LatestRevision {
+			e[i].Tag = tag
+			return e
+		}
+	}
+	e = append(e, newTarget(tag, "", 0, true))
+	return e
+}
+
+// SetTrafficByRevision checks given revision in existing traffic block and sets given percent if found
+func (e ServiceTraffic) SetTrafficByRevision(revision string, percent int) {
+	for i, target := range e {
+		if target.RevisionName == revision {
+			e[i].Percent = percent
+			break
+		}
+	}
+}
+
+// SetTrafficByTag checks given tag in existing traffic block and sets given percent if found
+func (e ServiceTraffic) SetTrafficByTag(tag string, percent int) {
+	for i, target := range e {
+		if target.Tag == tag {
+			e[i].Percent = percent
+			break
+		}
+	}
+}
+
+// SetTrafficByLatestRevision sets given percent to latest ready revision of service
+func (e ServiceTraffic) SetTrafficByLatestRevision(percent int) {
+	for i, target := range e {
+		if *target.LatestRevision {
+			e[i].Percent = percent
+			break
+		}
+	}
+}
+
+// ResetAllTargetPercent resets (0) 'Percent' field for all the traffic targets
+func (e ServiceTraffic) ResetAllTargetPercent() {
+	for i := range e {
+		e[i].Percent = 0
+	}
+}
+
+// RemoveNullTargets removes targets from traffic block if they don't have and 0 percent traffic
+func (e ServiceTraffic) RemoveNullTargets() (newTraffic ServiceTraffic) {
+	for _, target := range e {
+		if target.Tag == "" && target.Percent == 0 {
+		} else {
+			newTraffic = append(newTraffic, target)
+		}
+	}
+	return newTraffic
+}
+
+func errorOverWritingtagOfLatestReadyRevision(existingTag, requestedTag string) error {
+	return fmt.Errorf("tag '%s' exists on latest ready revision of service, "+
+		"refusing to overwrite existing tag with '%s', "+
+		"add flag '--untag %s' in command to untag it", existingTag, requestedTag, existingTag)
+}
+func errorOverWritingTag(tag string) error {
+	return fmt.Errorf("refusing to overwrite existing tag in service, "+
+		"add flag '--untag %s' in command to untag it", tag)
+}
+
+func errorRepeatingRevision(forFlag string, name string) error {
+	if name == latestRevisionRef {
+		name = "identifier " + latestRevisionRef
+	} else {
+		name = "revision reference " + name
+	}
+	return fmt.Errorf("repetition of %s "+
+		"is not allowed, use only once with %s flag", name, forFlag)
+}
+
+// verifies if user has repeated @latest field in --tag or --traffic flags
+// verifyInputSanity checks:
+// - if user has repeated @latest field in --tag or --traffic flags
+// - if provided traffic portion are integers
+func verifyInputSanity(trafficFlags *flags.Traffic) error {
+	var latestRevisionTag = false
+	var sum = 0
+
+	for _, each := range trafficFlags.RevisionsTags {
+		revision, _, err := splitByEqualSign(each)
+		if err != nil {
+			return err
+		}
+
+		if latestRevisionTag && revision == latestRevisionRef {
+			return errorRepeatingRevision("--tag", latestRevisionRef)
+		}
+
+		if revision == latestRevisionRef {
+			latestRevisionTag = true
+		}
+	}
+
+	revisionRefMap := make(map[string]int)
+	for i, each := range trafficFlags.RevisionsPercentages {
+		revisionRef, percent, err := splitByEqualSign(each)
+		if err != nil {
+			return err
+		}
+
+		// To check if there are duplicate revision names in traffic flags
+		if _, exist := revisionRefMap[revisionRef]; exist {
+			return errorRepeatingRevision("--traffic", revisionRef)
+		} else {
+			revisionRefMap[revisionRef] = i
+		}
+
+		percentInt, err := strconv.Atoi(percent)
+		if err != nil {
+			return fmt.Errorf("error converting given %s to integer value for traffic distribution", percent)
+		}
+
+		if percentInt < 0 || percentInt > 100 {
+			return fmt.Errorf("invalid value for traffic percent %d, expected 0 <= percent <= 100", percentInt)
+		}
+
+		sum += percentInt
+	}
+
+	// equivalent check for `cmd.Flags().Changed("traffic")` as we don't have `cmd` in this function
+	if len(trafficFlags.RevisionsPercentages) > 0 && sum != 100 {
+		return fmt.Errorf("given traffic percents sum to %d, want 100", sum)
+	}
+
+	return nil
+}
+
+// Compute takes service traffic targets and updates per given traffic flags
+func Compute(cmd *cobra.Command, targets []v1alpha1.TrafficTarget, trafficFlags *flags.Traffic) ([]v1alpha1.TrafficTarget, error) {
+	err := verifyInputSanity(trafficFlags)
+	if err != nil {
+		return nil, err
+	}
+
+	traffic := newServiceTraffic(targets)
+
+	// First precedence: Untag revisions
+	for _, tag := range trafficFlags.UntagRevisions {
+		traffic.untagRevision(tag)
+	}
+
+	for _, each := range trafficFlags.RevisionsTags {
+		revision, tag, _ := splitByEqualSign(each) // err is checked in verifyInputSanity
+
+		// Second precedence: Tag latestRevision
+		if revision == latestRevisionRef {
+			existingTagOnLatestRevision := traffic.tagOfLatestReadyRevision()
+
+			// just pass if existing == requested
+			if existingTagOnLatestRevision == tag {
+				continue
+			}
+
+			// apply requested tag only if it doesnt exist in traffic block
+			if traffic.isTagPresent(tag) {
+				return nil, errorOverWritingTag(tag)
+			}
+
+			if existingTagOnLatestRevision == "" {
+				traffic = traffic.TagLatestRevision(tag)
+				continue
+			} else {
+				return nil, errorOverWritingtagOfLatestReadyRevision(existingTagOnLatestRevision, tag)
+			}
+
+		}
+
+		// Third precedence: Tag other revisions
+		// dont throw error if the tag present == requested tag
+		if traffic.isTagPresentOnRevision(tag, revision) {
+			continue
+		}
+
+		// error if the tag is assigned to some other revision
+		if traffic.isTagPresent(tag) {
+			return nil, errorOverWritingTag(tag)
+		}
+
+		traffic = traffic.TagRevision(tag, revision)
+	}
+
+	if cmd.Flags().Changed("traffic") {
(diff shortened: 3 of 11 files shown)


duglin:
started to play with it... one initial comment:
the use of `--traffic-latest` is kind of awkward since all other traffic settings are under `--traffic`. Since all revision names must start with `service-`, how about we just use `--traffic latest:50` since we can assume that `latest` can never be a valid revision name... no `-` in it.

duglin:
Unless we drop the requirement of all revision names starting with `service-` for the CLI - which we did discuss but I don't remember the results of that chat.

navidshaikh:
@duglin : Thanks for testing it.

> the use of --traffic-latest is kind of awkward since all other traffic settings are under --traffic.

The way I'd use that flag as easy way to re-present the desired state on CLI is, for eg:
I want to rollout 10% traffic to my new app while keeping rest on known version
```
kn service update myapp --image myapp:latest --traffic-latest 10 --traffic stable:90
```
helps make identifying latestRevision reference in traffic block by a flag than an assumed value.


>  Since all revision names must start with service-, how about we just use --traffic latest:50 since we can assume that latest can never be a valid revision name... no - in it.

Personally, -1 for any assumptions, I'd rather want to write exact revision names on CLIs. BTW, did you notice you can also do
```
kn service update myapp --traffic echo-v1:20,echo-v2:20 --traffic echo-v3:60`
```
If one doesn't want to repeat `--traffic` flag, could just append the values comma separated, also works in combination (was simple but useful switch to use right method from cobra).

duglin:
oh wait, using `--traffic latest;50` won't work due to us being able to use tag names, not just revision names in there.  

I still think:
```
kn service update myapp --traffic echo-v1:20,echo-v2:20,@latest:60`
```
is more friendly than:
```
kn service update myapp --traffic echo-v1:20,echo-v2:20 --traffic-latest 60`
```

But oh well.  Still need to play more with tag names and multiple "latest" blocks....

maximilien:
> ```
> kn service update myapp --traffic echo-v1:20,echo-v2:20,@latest:60`
> ```
> 
> is more friendly than:
> 
> ```
> kn service update myapp --traffic echo-v1:20,echo-v2:20 --traffic-latest 60`
> ```
 What’s the ``` at the end of your command here? Typo?

maximilien:
> If one doesn't want to repeat `--traffic` flag, could just append the values comma separated, also works in combination (was simple but useful switch to use right method from cobra).

👍🏽👍🏽

docs/cmd/kn_service_update.md
```diff
@@ -41,6 +41,11 @@ kn service update NAME [flags]
   -p, --port int32               The port where application listens on.
       --requests-cpu string      The requested CPU (e.g., 250m).
       --requests-memory string   The requested memory (e.g., 64Mi).
+      --tag-latest string        Tag latest ready revision, format: --tag-latest tag , example: --tag-latest current
+      --tag-revision strings     Tag revisions, format: --tag-revision revision:tag , example: --tag-revision echo-abcde:current (can be specified multiple times)
```
maximilien: I assume these `—tag-revision` and `—untag-revision` will match the `—label` work being done in #328? If yes, then we need to agree on one name IMO. Different names for same thing will only confuse user.

rhuss: #328 is mostly unrelated to this feature. #328 is more like `kn service create --env` which we already have. 

However, instead of `--untag-revision current`  we coluld also use `--tag-revision :current` (i.e. keep the revision part empty). Not sure if this looks much cleaner though (especiall if we might consider to switch to `=` as separator then this would mutate to `--tag-revision =current` or even `--tag-revision==current`.

Normally I'm always for introducing as less options as possible, but in this case, I think a dedicated `--untag-revision` is justified.

Another question is around whether we want to have `--tag-revision` or whether its not already clear from the context (tags can be only on revisions), that we could reduce that to `--tag` and `--untag`

rhuss:
wrt. to `--tag-latest` and a special tag name like `@latest`, e..g

```
kn service update myapp --traffic echo-v1:20,echo-v2:20,@latest:60
```

vs 

```
kn service update myapp --traffic echo-v1:20,echo-v2:20 --traffic-latest:60
```

I'm clearly for the `@latest` variant as this would save us to options (`--traffic-latest` and `--tag-latest`). `kn service update` already is 15 options, more to come (like for annotations and labels), so adding an additional 5 options is a heavy addition to the cognitive load. Saving 2 options here is a clear benefit (we could even save 3 if we would remove `--untag-revision` and use a value convention on `--tag-revision`.

So my suggestion would be:

* `--tag`
* `--untag`
* `--traffic`

with _latest_ handling done by a marker value. This looks like a nice and concise UX.

navidshaikh:
> What’s the ` at the end of your command here? Typo?

@maximilien : Typo, fixed it.

navidshaikh:
For discussion on having dedicated flag to reference latest revision, I'd prefer to use separate flags (`--traffic-latest`, `--tag-latest`), but  it seems there is more consensus on using an assumed special word for referencing same. 

We'll need bring docs for this, IMO we had that 'special word' represented by flag `--traffic-latest` with cobra native docs for this, but then there is point raise about number of flags.

My take on re-presenting the YAMLs via CLI UX is to give users convenient flags that are easy to remember and use, I don't think number of flags should be a concern if they are used for specific operations, like these flags will be used for specific operations and we'll have 2 for traffic and 3 for tagging, that number sounds fine to me.

navidshaikh:
If we're to use special word and omit the `--traffic-latest` and `--tag-latest`, k8s has restrictions for resource names as as described here [ref](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names), excerpts
```
By convention, the names of Kubernetes resources should be up to maximum length of 253
characters and consist of lower case alphanumeric characters, -, and ., but certain resources
have more specific restrictions.
```
Something to note here is that it allows only `lower` case characters, so using a special assumed word in upper case for eg: `LATEST`  without any special characters like `@` is safe IMO, helps avoid interference with scripting languages `kn` could be used with.

Would `LATEST` word be a convenient choice? WDYT ?

duglin:
it's actually because `@` isn't allowed as a name that we SHOULD use it  :-)  Then there's no chance over overlap with a user's name. The presence of the `@latest` in there should be detected by the CLI and converted to the appropriate yaml sent to the server, meaning the server should never see it.

rhuss:
@duglin LATEST is equally forbidden and there is no chance of interference with shell special characters nor that it can be used as a regular revision name. The same here: LATEST will never be sent but has the special meaning for the client to specify the 'latest'. 

I have no super strong opinion, but think LATEST (all upper case) might make more sense (except when the user gets confused with the casing, in which case a special prefix character like `@`, `.`, or `:` might make sense.

rhuss:
Let's make a vote for these three options

* --tag-latest
* LATEST
* @latest

rhuss:
Vote for `--tag-latest`

rhuss:
Vote for `LATEST` as special revision name

rhuss:
Vote vor `@latest` or `@LATEST` (i.e. in any case) as special name.

navidshaikh:
> Another question is around whether we want to have --tag-revision or whether its not already clear from the context (tags can be only on revisions), that we could reduce that to --tag and --untag

@rhuss : I like your suggestion.

I am +1 to change the current `--tag-revision` and `--untag-revision` to `--tag` and `--untag`, since `tag` should be referred in terms of traffic block of service, while for general tagging we have `label`. Update coming..

duglin:
+1 to removing the unnecessary `-revision` suffix on the `--tag` option

navidshaikh:
@duglin @rhuss : Updated the PR per suggestions, please git it a spin. If the UX surface looks good, I;ll go ahead by adding rest of the items in the checklist.

duglin:
yep - I know LATEST is disallowed too, it's just less clear to a newbie that it's special. I think a special char (like `@`) makes it stand-out more that this is special and it won't look like all of the others when they look at the status section. "@" almost looks like a pointer reference to me - which is kind of what it is.   Plus it's easier to type :-)   But, yeah, either *can* work.

duglin:
we should have a follow-on PR to support this on `kn service create` so that people can define their traffic block immediately. The cases I'm most interested in are:
- defining a tag, so I get a dedicated URL
- defining 2 sections - one for "latest" and one for "v1" (current), and both with dedicated URLs

I don't see any reason to force people to do it in two steps - especially when kubectl can do it all in one.

docs/cmd/kn_service_update.md
```diff
@@ -41,6 +41,9 @@ kn service update NAME [flags]
   -p, --port int32               The port where application listens on.
       --requests-cpu string      The requested CPU (e.g., 250m).
       --requests-memory string   The requested memory (e.g., 64Mi).
+      --tag strings              Tag revisions, format: --tag revision=tag , example: --tag echo-abcde=current (can be specified multiple times). Use identifier @latest to refer to latest ready revision, for e.g.: --tag LATEST=new (LATEST can be used only once with --tag flag).
```
duglin: I think the LATEST in the e.g. needs to be udpated to @latest

duglin:
Mixing this PR with #282:

```
$ ./kn service create echo --image duglin/echo
$ ./kn service update echo --revision-name=v1 --tag echo-v1=echo-v1,test=@latest --traffic echo-v1=100
```
results in:
```
  traffic:
  - latestRevision: false
    percent: 100
    revisionName: echo-v1
    tag: echo-v1
  - latestRevision: false
    percent: 0
    revisionName: test
    tag: '@latest'
```
So close!  just the last `tag` looks funky.  But aside from that, so far so good...

Tried to do it all in on the `create` but got this:
```
./kn service create echo --image duglin/echo --revision-name=v1 ...
The service is using too old of an API format for the operation
```
but I think the issue is with #282

navidshaikh:
> $ ./kn service update echo --revision-name=v1 --tag echo-v1=echo-v1,test=@latest --traffic echo-v1=100


@duglin : There is an item in check list of this PR to verify if the given revision exists before updating it in traffic block, so there would be an error that revision `test` doesn't exist.

About using character '@' for tag, I think there needs to be some validation done at serving level to detect that, as I just tried creating a service using YAML with
```
    traffic:
    - latestRevision: false
      percent: 50
      revisionName: svc1-bg9cs
      tag: '@latest'
    - latestRevision: false
      percent: 50
      revisionName: svc1-mr9lz
      tag: current

```
and now service status reports
```
Unknown   OutOfDate : The Route is still working to reflect the latest desired specification.
```
if it were never to reach desired state there should've been an error I think.

docs/cmd/kn_service_update.md
navidshaikh: thanks, fixed the typo.

navidshaikh:
> There is an item in check list of this PR to verify if the given revision exists before updating it 

I wonder though if client should start validating the resources/values as, if we verify it for revisions' existence and pass for invalid tag names, that'd still keep the service in unknown state.

rhuss:
There is an issue open for the missing revision name validation at https://github.com/knative/serving/issues/5094

I guess the same validation is missing for tag-names, too.

duglin:
ah, my bad, I had the --tag flag backwards, when I switch it to be `@latest=test` then I get:
```
    traffic:
    - latestRevision: true
      percent: 0
      tag: test
    - latestRevision: false
      percent: 100
      revisionName: echo-v1
      tag: echo-v1
```
which is correct!

duglin:
Not sure what to make of this:
```
#!/bin/bash

set -x

./kn service delete echo || true
./kn service create echo --image duglin/echo
./kn service update echo --revision-name=v1 --tag echo-v1=echo-v1,@latest=test --traffic echo-v1=100
./kn service update echo --revision-name=v1 --tag echo-v1=v1,@latest=test --traffic echo-v1=100

./kn service update echo --revision-name=v1 --untag test --tag echo-v1=v1,@latest=test --traffic echo-v1=100
```

The middle update fails with:
```
refusing to overwrite existing tag in service, add flag '--untag test' in command to untag it
```
I'm not sure the error should happen since I'm technically not changing anything in the `test` traffic block.


btw - as I'm playing with `kn` I'm finding typing `kn service describe...` to be annoying and I really miss the shorter `kn service get ...`   :-)  just saying

rhuss:
> btw - as I'm playing with `kn` I'm finding typing `kn service describe...` to be annoying and I really miss the shorter `kn service get ...` :-) just saying

5 chars more, come on ;-) There was a chance for `show`, but that window closed ;-)

At least we have now a proper `kn service describe`, finally. Even now with a `-o yaml`.

navidshaikh:
> I'm not sure the error should happen since I'm technically not changing anything in the test traffic block.

Yes, its only checking for presence of same tag in block, you are right it should also check if the revision=tag pair is same and ignore as delta is none. I've updated the PR to check for this case.

duglin:
> 5 chars more, come on ;-) There was a chance for show, but that window closed ;-)

LOL I don't think it's the number of chars, it's the muscle memory aspect. My fingers just want to type `get` and I find myself doing `get^h^h^hdescribe` a lot  :-)

sixolet:
Let's not make up a new syntax. The API has different fields, the client should have different flags.

sixolet:
I don't think the number of options is particularly worrisome, since the alternative is a sigil that is a new thing to learn about an not an option at all, which is more cognitive load.

pkg/kn/traffic/compute.go
```diff
@@ -0,0 +1,336 @@
...
+func newServiceTraffic(traffic []v1alpha1.TrafficTarget) ServiceTraffic {
+	return ServiceTraffic(traffic)
+}
+
+func splitByEqualSign(pair string) (string, string, error) {
+	parts := strings.SplitN(pair, "=", 2)
+	if len(parts) != 2 {
+		return "", "", errors.New(fmt.Sprintf("expecting the value format in value1=value2, given %s", pair))
+	}
+	return parts[0], strings.TrimSuffix(parts[1], "%"), nil
```
zhanggbj: Do we still need `strings.TrimSuffix(parts[1], "%")` here, it is harmless but as we don't support `--traffic revision=2%`, `--traffic revision=2` is enough.

docs/cmd/kn_service_update.md
```diff
@@ -41,6 +41,9 @@ kn service update NAME [flags]
   -p, --port int32               The port where application listens on.
       --requests-cpu string      The requested CPU (e.g., 250m).
       --requests-memory string   The requested memory (e.g., 64Mi).
+      --tag strings              Tag revisions, format: --tag revision=tag , example: --tag echo-abcde=current (can be specified multiple times). Use identifier @latest to refer to latest ready revision, for e.g.: --tag @latest=new (@latest can be used only once with --tag flag).
```
rhuss: examples should go in an extra section, not part of the flag description.

docs/cmd/kn_service_update.md
```diff
@@ -41,6 +41,9 @@ kn service update NAME [flags]
...
+      --traffic strings          Set traffic percentage, format: --traffic revision=percent , example: --traffic echo-abcde=50) (can be specified multiple times). Use identifier @latest to refer to latest ready revision, for e.g.: --traffic @latest=100 (@latest can be used only once with --traffic flag).
+      --untag strings            Untag revision, format: --untag tag , example: --untag current
```
rhuss: please no examples here, otherwise the description gets too long. It's meant as a reference.

pkg/kn/traffic/compute.go
navidshaikh: Right, but we can support both with TrimSuffix, the example section should cover that.

docs/cmd/kn_service_update.md
navidshaikh: Right, I missed that, updating..

docs/cmd/kn_service_update.md
navidshaikh: Ack!

pkg/kn/commands/flags/traffic.go
```diff
@@ -0,0 +1,70 @@
...
+	UntagRevisions       []string
+}
+
+func (t *Traffic) Add(cmd *cobra.Command) {
+	cmd.Flags().StringSliceVar(&t.RevisionsPercentages,
+		"traffic",
+		nil,
+		"Set traffic percentage, format: --traffic revision=percent , example: --traffic echo-abcde=50) (can be specified multiple times). "+
+			"Use identifier @latest to refer to latest ready revision, for e.g.: --traffic @latest=100 (@latest can be used only once with --traffic flag).")
+
```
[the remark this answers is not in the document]
navidshaikh: Sure.

rhuss:
> I don't think the number of options is particularly worrisome, since the alternative is a sigil that is a new thing to learn about an not an option at all, which is more cognitive load.

Slightly disagree if the alternative is easy to understand and fits in the help description of the option like in 

> Set traffic distribution in the format "revOrTag1=percent,revOrTag2=percent,..." where name can be a revision, a tag or '@latest' for the latest ready revision. This option can be given multiple times, summing up to 100%.

Especially for splitting up the traffic, why should be one portion be treated special with an extra option, when they are all fractions of the same thing (traffic) ? Also, it wouldn't be possible to add all parts into a single `--traffic rev1=10,@latest=90` then

Said that, I would be fine with both but favor the @latest syntax.

docs/cmd/kn_service_update.md
rhuss: Or as shown below which would be more complete:

> Set traffic distribution in the format "revOrTag1=percent,revOrTag2=percent,..." where name can be a revision, a tag or '@latest' for the latest ready revision. This option can be given multiple times, summing up to 100%.

toVersus:
> The API has different fields, the client should have different flags.

I think this is the most important thing we should consider when developing an reference implementation of knative client. I know magic variables are useful but should we do that kind of works as extensions?

Even though kn provides a powerful tab completion feature and detailed help message, why are we worried about the number of options or the option name length?

pkg/kn/commands/flags/traffic.go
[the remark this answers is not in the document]
navidshaikh: We'll move the examples off the flag description, as discussed above.

pkg/kn/traffic/compute.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+	e = append(e, newTarget(tag, "", 0, true))
+	return e
+}
+
+func (e ServiceTraffic) SetTrafficByRevision(revision string, percent int) {
+	for i, target := range e {
+		if target.RevisionName == revision {
+			target.Percent = percent
+			e[i] = target
```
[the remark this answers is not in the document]
navidshaikh: Fixed, thanks!

docs/cmd/kn_service_update.md
maximilien: I'd vote for `--tag` `--untag` since context is in the group name.

maximilien:
> btw - as I'm playing with `kn` I'm finding typing `kn service describe...` to be annoying and I really miss the shorter `kn service get ...` :-) just saying

We need `kn s d ...` for power users :)

rhuss:
@toVersus not sure how that goes, but why could you commit and push changed to @navidshaikh 's PR ? I mean, technically, why is this possible ?

sixolet:
That's the "suggest changes" feature of GitHub. It correctly attributes the suggested changes, and then the CLA check gets all messy.

duglin:
are you sure? The commit were added, not just suggested. Unless @navidshaikh accepted them...

navidshaikh:
I'd given collaborator access on my branch to @toVersus and @zhanggbj as they wanted to collaborate on this implementation.

>  but one or more commits were authored or co-authored by someone other than the pull request submitter.

@toVersus : As we'd discussed on slack to collaborate, let's propose your changes via PRs against my pr/traffic branch, we'll merge the changes and see if CLA issue resolves.
Raising a PR helps avoid merge conflicts and duplication of efforts, I already worked on few new commits and was working on another one and didn't see any collaborator PR so I could plan accordingly

```
Author: Navid Shaikh <nshaikh@redhat.com>
Date:   Sat Aug 10 20:25:32 2019 +0530

    Moves verification of traffic portion in input sanity check
    
     renames the verification function to verifyInputSanity

 pkg/kn/traffic/compute.go | 35 +++++++++++++++++++----------------
 1 file changed, 19 insertions(+), 16 deletions(-)

commit 81b93a11be3c5719c9b7d0879368ee0abceb7881
Author: Navid Shaikh <nshaikh@redhat.com>
Date:   Sat Aug 10 20:15:11 2019 +0530

    Removes unnecessary check whether flag has changed
    
     as if the flag changed the for loop will execute

 pkg/kn/traffic/compute.go | 74 ++++++++++++++++++++++++++++++++++++--------------------------------------
 1 file changed, 36 insertions(+), 38 deletions(-)

commit 4103bf5a32aac8126731bf8eaa68b56aac96cbff (origin/pr/traffic)
Author: Navid Shaikh <nshaikh@redhat.com>
Date:   Sat Aug 10 19:36:55 2019 +0530

```

I backed up the current state of this branch to [pr/traffic-messed-up](https://github.com/navidshaikh/client/pull/new/pr/traffic-messed-up), you can pull copy of your commits from there, I'll reset and force to push to this PR to restore the CLA issue.

sixolet:
I guess "that's just how it always happens to me"

toVersus:
Guys, sorry for bothering you, but just pretend you didn't see that! There is no magic in there. As @navidshaikh said, I got write permission to Navid's `pr/traffic` branch for collaboration and pushed my commits to its branch directly. (My fault...) I raised a PR against Navid's `pr/traffic`, so I hope everything will go well.

duglin:
No worries, just hadn’t seen it before from someone without write access to the repo 🙂

navidshaikh:
@toVersus : Try telling googleabot about consent in comment as suggested above?

toVersus:
@googlebot I consent.

rhuss:
The integration test error looks like a real one:

```
=== RUN   TestTrafficSplit/tag_a_revision_as_candidate,_without_otherwise_changing_any_traffic_split
Running 'kn service create echo --image gcr.io/knative-samples/helloworld-go -n kne2etests0'...
Failed to successfully execute 'kn service create echo --image gcr.io/knative-samples/helloworld-go -n kne2etests0': Execution error: stderr: 'RevisionMissing: Revision "echo-d27lh" referenced in traffic not found.
' error: 'exit status 1'FAIL	github.com/knative/client/test/e2e	74.074s
Finished run, return code is 1
XML report written to /logs/artifacts/junit_M7PBUUmm.xml
```

navidshaikh:
@rhuss : The one you pasted popping up as it *creates* the service, I also noticed in other e2e-test runs that Revisions are missing in different tests randomly. These all tests went well on my e2e local setup.

pkg/kn/commands/flags/traffic.go
```diff
@@ -0,0 +1,56 @@
...
+	cmd.Flags().StringSliceVar(&t.RevisionsTags,
+		"tag",
+		nil,
+		"Set tag (format: --tag revisionRef=tag) where revisionRef can be a revision or '@latest' string representing latest ready revision. "+
+			"This flag can be specified multiple times.")
+
+	cmd.Flags().StringSliceVar(&t.UntagRevisions,
+		"untag",
+		nil,
+		"Untag revision (format: --untag tag). This flag can be spcified multiple times.")
```
maximilien: Don't use `tag` as the tag for the `--untag` example. That confuses things IMO since tag is a flag. Would make more sense to use a more concrete example, e.g., `v0.1` or similar?

navidshaikh: how about changing the all `tagName`? 
(the `format` in flag description is field reference in traffic block)

navidshaikh:
/test pull-knative-client-integration-tests says
```
*** Test killed with quit: ran too long (10m0s).
```

while the other e2e-test pull-knative-client-integration-tests-latest-release  passed.

navidshaikh:
This PR adds multiple test cases each updating a service in sync multiple times to have multiple revisions of service for scenarios to run traffic splitting test cases. This makes total e2e tests duration go beyond the default timeout duration (10minutes) for  `go test`.  Logged the issue at test-infra to support passing timeout flag and value as parameter to `go_test_e2e` test-infra library function we're using to run e2e tests.

//cc: @adrcunha

navidshaikh:
To ensure the e2e tests for this PR passes, I am updating our test script to run the same command as go_test_e2e invokes ([ref](https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_client/345/pull-knative-client-integration-tests/1161176727611772928#0:build-log.txt%3A477)) with `timeout` flag

pkg/kn/traffic/compute.go
```diff
@@ -0,0 +1,362 @@
...
+
+func errorRepeatingLatestRevision(forFlag string) error {
+	return fmt.Errorf("repetition of identifier %s "+
+		"is not allowed, use only once with %s flag", latestRevisionRef, forFlag)
+}
+
+// verifies if user has repeated @latest field in --tag or --traffic flags
+// verifyInputSanity checks:
+// - if user has repeated @latest field in --tag or --traffic flags
+// - if provided traffic portion are integers
```
zhanggbj: Hi @navidshaikh 

I think except `@latest`, we should also report error if there're duplicated revision name in the traffic flags, for e.g. so far we will fail util the server side, we can check this from the client side. I submitted a PR to your branch and please take a review. Thanks!

```
./kn service update canary --traffic gracecanary-blue=40 --traffic gracecanary-blue=60
Internal error occurred: admission webhook "webhook.serving.knative.dev" denied the request: mutation failed: Traffic targets sum to 50, want 100: spec.traffic
```

navidshaikh: Hi @zhanggbj , sounds good. 
If we are to add this check, I'd say, repetition of revision reference in `--traffic` block is not allowed (than saying revision name), because we allow `revisionName`, `tag` and `@latest` with `--traffic` flag.

duglin:
Scenario:
- I have 2 different users for a service
- Each user will be given their own URL to the specific version that they should use - ie. each will point to the traffic "tag" that was created for them
- The specific version that each user points to will change over time - as they migrate from version to version.

When they initially point to different revisions, the yaml can look like:
```
traffic:
- tag: coke
  revisionName: svc-v1
- tag: pepsi
  revisionName: svc-v2
- latestRevision: true
  percent: 100
```
and the `kn` command would be `kn service update svc --tag svc-v1=coke --tag svc-v2=pepsi --traffic @latest=100`. I think.

However, there will be times when both users end up pointing to the same revision. In that case the yaml might look like:
```
traffic:
- tag: coke
  revisionName: svc-v2
- tag: pepsi
  revisionName: svc-v2
- latestRevision: true
  percent: 100
```
It's not clear to me how I would define this state using `kn`. Is this possible?

navidshaikh:
New service with 100% to '@ latest'
```
✗  kn service describe echo -o "jsonpath={range .status.traffic[*]}{.tag}{' '}{.revisionName}{' '}{.percent}{' '}{.latestRevision}{'|'}{end}"
 echo-x8gkv 100 true|
```

3 revisions available:
```
✗  kn revision list                                                                                                                          
NAME         SERVICE   GENERATION   AGE     CONDITIONS   READY   REASON
echo-x8gkv   echo      3            3m39s   3 OK / 4     True    
echo-5brl6   echo      2            9m46s   3 OK / 4     True    
echo-d62xr   echo      1            64m     3 OK / 4     True    
```

Add 2 more targets with tags coke and pepsi with different revisions
```
✗  kn service update echo --tag echo-5brl6=coke --tag echo-x8gkv=pepsi
Waiting for service 'echo' to become ready ... OK
Service 'echo' updated in namespace 'default'.
```

current state :
```
✗  kn service describe echo -o "jsonpath={range .status.traffic[*]}{.tag}{' '}{.revisionName}{' '}{.percent}{' '}{.latestRevision}{'|'}{end}"
 echo-x8gkv 100 true|coke echo-5brl6 0 false|pepsi echo-x8gkv 0 false|%                                                                                                                       
```

Now we want coke and pepsi to point to same revision
```
✗  kn service update echo --untag coke,pepsi --tag echo-x8gkv=coke,echo-x8gkv=pepsi                                                          
Waiting for service 'echo' to become ready ... OK
Service 'echo' updated in namespace 'default'.
```

and its state
```
✗  kn service describe echo -o "jsonpath={range .status.traffic[*]}{.tag}{' '}{.revisionName}{' '}{.percent}{' '}{.latestRevision}{'|'}{end}"
 echo-x8gkv 100 true|coke echo-x8gkv 0 false|pepsi echo-x8gkv 0 false|%       
```

duglin:
ah, so I would have to know that both coke and pepsi are going to point to the same revision and then include both in the same cmd. I can't set just move one of them. Do I have that right?

navidshaikh:
@duglin :

>  so I would have to know that both coke and pepsi are going to point to the same revision

Yes, I believe thats the desired state you want to apply.

> and then include both in the same cmd. 

Not really. You can also set them separately as well.

The desired state and commands needed depend on what's there in existing traffic block. You can combine operations as needed to reach desired state.
Operations:
1. Remove a target entirely from traffic block having traffic 0%   `--untag pepsi` 
2. Update tag of a target                                                            `--untag oldtag --tag revision=newTag`  
3. Append new target in traffic block with 
    a.newRevisionName=newTagName                                    `--tag echo-v3=coke`   
    b. existingRevisionName=newTagName         `--tag echo-v2=pepsi`

duglin:
```
$ ./kn service update echo --traffic echo-v1=50,echo-v1=50 --tag echo-v1=coke
admission webhook "webhook.serving.knative.dev" denied the request: mutation failed: Traffic targets sum to 50, want 100: spec.traffic
```

I'm not sure how i can build up the scenario I want on different commands. Isn't the thing on the left side of the "=" the search string? And if there more than one traffic block with that revision name, how does it know which one to use?

test/e2e-tests.sh
```diff
@@ -59,6 +59,5 @@ initialize $@
 
 header "Running tests for Knative serving $KNATIVE_VERSION"
 
-go_test_e2e ./test/e2e || fail_test
-
+go_test_e2e -timeout=30m ./test/e2e || fail_test
```
maximilien: Interesting. You did this because you've seen the test take upward of 30 minutes? Or are you just being cautious? A 30 minutes test is not a good thing obviously...

test/e2e/common.go
```diff
@@ -72,6 +73,14 @@ func getNamespaceCountAndIncrement() int {
 	return current
 }
 
+func getServiceNameAndIncrement(base string) string {
+	m.Lock()
+	defer m.Unlock()
```
maximilien: Nice nice, thanks for handling the new parallel fashion we have now for tests.

test/e2e-tests.sh
navidshaikh: There was a timeout for going beyond default value, 10m. I kept it this 3 times to be sure.

navidshaikh:
> Isn't the thing on the left side of the "=" the search string? 

It is.

> And if there more than one traffic block with that revision name, how does it know which one to use?

It can't and its not safe to guess right? --traffic supports referencing tag or revision.
If user wants to have multiple targets referring to same revision, user must tag them. 
```
✗ kn service update svc --tag svc-5r2dp=pepsi,svc-5r2dp=coke --traffic pepsi=50,coke=50
```

zhanggbj:
@googlebot I consent.

duglin:
> It can't and its not safe to guess right? --traffic supports referencing tag or revision.
> If user wants to have multiple targets referring to same revision, user must tag them.
> 
> ```
> ✗ kn service update svc --tag svc-5r2dp=pepsi,svc-5r2dp=coke --traffic pepsi=50,coke=50
> Waiting for service 'svc' to become ready ... OK
> Service 'svc' updated in namespace 'default'.
> ```

ok, so this is what I'm seeing:
```
$ ./kn service update echo --tag echo-v1=coke,echo-v1=pepsi --traffic coke=50,pepsi=50
```
does appear to work, so that's good. However, my first attempt at it I tried to just setup the traffic block w/o any tags and just have 2 50% sections:
```
$ ./kn service update echo --traffic echo-v1=50,echo-v1=50
admission webhook "webhook.serving.knative.dev" denied the request: mutation failed: Traffic targets sum to 50, want 100: spec.traffic
```
Should I be able to do this?

zhanggbj:
@duglin About the second item as below, I think we disallow duplicated revision reference. I have added a check and it will report an error from the client instead of the server, the code should have been included in this PR request.
```
$ ./kn service update echo --traffic echo-v1=50,echo-v1=50
admission webhook "webhook.serving.knative.dev" denied the request: mutation failed: Traffic targets sum to 50, want 100: spec.traffic
```


Now you will get something as below,
```
$ ./kn service update echo --traffic echo-v1=50,echo-v1=50
repetition of revision reference echo-v1 is not allowed, use only once with --traffic flag
```

navidshaikh:
> Should I be able to do this?

If the desired state expect to have same revisions in two targets, user must tag them for `kn` to identify target uniquely for subsequent operations.

navidshaikh:
@sixolet @maximilien @cppforlife: I've rebased and squashed, PTAL.

maximilien:
@navidshaikh and @zhanggbj I think it’s not merging because you need to rebase, conflict with: `docs/cmd/kn_service_update.md`

PR #346, opened 2019-08-08
Bool flags in the paired `--foo` and `--no-foo` format

sixolet:
Adds a library to add and understand paired boolean flags in the format we agreed upon by voting. Uses it for `--log-http`, for now.

### DIFF
--- a/pkg/kn/flags/bool.go
+++ b/pkg/kn/flags/bool.go
@@ -48,10 +48,11 @@ func AddBothBoolFlags(f *pflag.FlagSet, p *bool, name, short string, value bool,
 }
 
 // ReconcileBoolFlags sets the value of the all the "--foo" flags based on
-// "--no-foo" if provided, and returns an error if both were provided.
+// "--no-foo" if provided, and returns an error if both were provided or an
+// explicit value of false was provided to either (as that's confusing).
 func ReconcileBoolFlags(f *pflag.FlagSet) error {
 	var err error
-	f.Visit(func(flag *pflag.Flag) {
+	f.VisitAll(func(flag *pflag.Flag) {
 		// Return early from our comprehension
 		if err != nil {
 			return
@@ -62,17 +63,41 @@ func ReconcileBoolFlags(f *pflag.FlagSet) error {
 		if strings.HasPrefix(flag.Name, "no-") {
 			positiveName := flag.Name[len(negPrefix):len(flag.Name)]
 			positive := f.Lookup(positiveName)
-			if positive.Changed {
-				err = fmt.Errorf("only one of --%s and --%s may be specified",
-					flag.Name, positiveName)
-				return
+			if flag.Changed {
+				if positive.Changed {
+					err = fmt.Errorf("only one of --%s and --%s may be specified",
+						flag.Name, positiveName)
+					return
+				}
+				var noValue bool
+				noValue, err = strconv.ParseBool(flag.Value.String())
+				if err != nil {
+					return
+				}
+				if !noValue {
+					err = fmt.Errorf("use --%s instead of providing false to --%s",
+						positiveName, flag.Name)
+					if err != nil {
+						return
+					}
+				}
+				err = positive.Value.Set(strconv.FormatBool(!noValue))
+			} else if positive.Changed {
+				// For the positive version, just check it wasn't set to the
+				// confusing "false" value.
+				var yesValue bool
+				yesValue, err = strconv.ParseBool(positive.Value.String())
+				if err != nil {
+					return
+				}
+				if !yesValue {
+					err = fmt.Errorf("use --%s instead of providing false to --%s",
+						flag.Name, positiveName)
+					if err != nil {
+						return
+					}
+				}
 			}
-			var noValue bool
-			noValue, err = strconv.ParseBool(flag.Value.String())
-			if err != nil {
-				return
-			}
-			err = positive.Value.Set(strconv.FormatBool(!noValue))
 		}
 	})
 	return err
--- a/pkg/kn/flags/bool.go
+++ b/pkg/kn/flags/bool.go
@@ -1,4 +1,4 @@
-// Copyright © 2018 The Knative Authors
+// Copyright © 2019 The Knative Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/pkg/kn/flags/bool.go
+++ b/pkg/kn/flags/bool.go
@@ -1,0 +1,74 @@
+// Copyright © 2018 The Knative Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package flags
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/spf13/pflag"
+)
+
+var negPrefix = "no-"
+
+func AddBothBoolFlags(f *pflag.FlagSet, p *bool, name, short string, value bool, usage string) {
+
+	negativeName := negPrefix + name
+
+	f.BoolVarP(p, name, short, value, usage)
+	f.Bool(negativeName, !value, "do not "+usage)
+
+	if value {
+		err := f.MarkHidden(name)
+		if err != nil {
+			panic(err)
+		}
+	} else {
+		err := f.MarkHidden(negativeName)
+		if err != nil {
+			panic(err)
+		}
+	}
+}
+
+func ReconcileBoolFlags(f *pflag.FlagSet) error {
+	var err error
+	f.Visit(func(flag *pflag.Flag) {
+		// Return early from our comprehension
+		if err != nil {
+			return
+		}
+		// Walk the "no-" versions of the flags. Make sure we didn't set
+		// both, and set the positive value to the opposite of the "no-"
+		// value if it exists.
+		if strings.HasPrefix(flag.Name, "no-") {
+			positiveName := flag.Name[len(negPrefix):len(flag.Name)]
+			positive := f.Lookup(positiveName)
+			if positive.Changed {
+				err = fmt.Errorf("only one of --%s and --%s may be specified",
+					flag.Name, positiveName)
+				return
+			}
+			var noValue bool
+			noValue, err = strconv.ParseBool(flag.Value.String())
+			if err != nil {
+				return
+			}
+			err = positive.Value.Set(strconv.FormatBool(!noValue))
+		}
+	})
+	return err
+}
--- a/pkg/kn/flags/bool.go
+++ b/pkg/kn/flags/bool.go
@@ -24,6 +24,9 @@ import (
 
 var negPrefix = "no-"
 
+// AddBothBoolFlags adds the given flag in both `--foo` and `--no-foo` variants.
+// If you do this, make sure you call ReconcileBoolFlags later to catch errors and
+// set the relationship between the flag values.
 func AddBothBoolFlags(f *pflag.FlagSet, p *bool, name, short string, value bool, usage string) {
 
 	negativeName := negPrefix + name
@@ -44,6 +47,8 @@ func AddBothBoolFlags(f *pflag.FlagSet, p *bool, name, short string, value bool,
 	}
 }
 
+// ReconcileBoolFlags sets the value of the all the "--foo" flags based on
+// "--no-foo" if provided, and returns an error if both were provided.
 func ReconcileBoolFlags(f *pflag.FlagSet) error {
 	var err error
 	f.Visit(func(flag *pflag.Flag) {
--- a/pkg/kn/flags/bool.go
+++ b/pkg/kn/flags/bool.go
@@ -61,7 +61,7 @@ func ReconcileBoolFlags(f *pflag.FlagSet) error {
 		// both, and set the positive value to the opposite of the "no-"
 		// value if it exists.
 		if strings.HasPrefix(flag.Name, "no-") {
-			positiveName := flag.Name[len(negPrefix):len(flag.Name)]
+			positiveName := flag.Name[len(negPrefix):]
 			positive := f.Lookup(positiveName)
 			if flag.Changed {
 				if positive.Changed {
--- a/pkg/kn/flags/bool_test.go
+++ b/pkg/kn/flags/bool_test.go
@@ -66,7 +66,7 @@ func TestBooleanPair(t *testing.T) {
 		if c.expectedErrText != "" {
 			assert.ErrorContains(t, err, c.expectedErrText)
 		} else {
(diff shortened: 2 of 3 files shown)


pkg/kn/flags/bool.go
```diff
@@ -0,0 +1,79 @@
...
+var negPrefix = "no-"
+
+// AddBothBoolFlags adds the given flag in both `--foo` and `--no-foo` variants.
+// If you do this, make sure you call ReconcileBoolFlags later to catch errors and
+// set the relationship between the flag values.
+func AddBothBoolFlags(f *pflag.FlagSet, p *bool, name, short string, value bool, usage string) {
+
+	negativeName := negPrefix + name
+
+	f.BoolVarP(p, name, short, value, usage)
```
navidshaikh: I think we can add switch here based on the value of `short` whether to use to `f.BoolVarP` or `f.BoolVar`,  if a flag doesn't specify the short hand we don't need to use `BoolVarP` for same, case for `log-http`

pkg/kn/flags/bool.go
```diff
@@ -0,0 +1,79 @@
+// Copyright © 2018 The Knative Authors
```
navidshaikh: *2019

pkg/kn/flags/bool.go
```diff
@@ -0,0 +1,79 @@
...
+// set the relationship between the flag values.
+func AddBothBoolFlags(f *pflag.FlagSet, p *bool, name, short string, value bool, usage string) {
+
+	negativeName := negPrefix + name
+
+	f.BoolVarP(p, name, short, value, usage)
+	f.Bool(negativeName, !value, "do not "+usage)
+
+	if value {
+		err := f.MarkHidden(name)
```
rhuss: maybe just not add it if the option already corresponds to the default value? Not sure, but that might help to catch errors as calling the hidden command will result in a no-op which might be hiding errors.

On the other hand, it's good to have those in case we want to change the default value later to the opposites. That would be for sure a backwards-incompatible change, but a user could already safeguard against such a change by always using the appropriate option.

So I'm for to keep it like it is (so ignore my comment ;-)

rhuss: btw, and somewhat unrelated, that's my general take on default values in general (included in the book I wrote). Opinionated, so I don't expect that everyone agrees on this ;-)

> Default values make life easier, as they take away the burden of selecting a value for a configuration parameter you might not even know exists. They also play a significant role in the _convention over configuration_ paradigm. However, defaults are not always a good idea. Sometimes they might even be an antipattern for an evolving application.
>
> This is because _changing_ default values retrospectively is a difficult task.
First, changing default values means replacing them within the code, which requires a rebuild. Second, people relying on defaults (either by convention or consciously) will always be surprised when a default value changes. We have to communicate the change, and the user of such an application probably has to modify the calling code as well.
> 
> Changes in default values, however, often make sense, because it is hard to get default values right from the very beginning. It's essential that we consider a change in a default values as a _major change_, and if semantic versioning is in use, such a modification justifies a bump in the major version number. If unsatisfied with a given default value, it is often better to remove the default altogether and throw an error if the user does not provide a configuration value. This will at least break the application early and prominently instead of it doing something different and unexpected silently.
> 
> Considering all these issues, it is often the best solution to _avoid default values_ from the very beginning if you cannot be 90% sure that a reasonable default will last for a long time. Passwords or database connection parameters are good candidates for not providing default values, as they depend highly on the environment and often cannot be reliably predicted. Also, if we do not use default values, the configuration information has to be provided explicitly, which serves as documentation too.

pkg/kn/flags/bool.go
sixolet: All `BoolVar` does is call `BoolVarP` with an empty string.

pkg/kn/flags/bool.go
```diff
@@ -0,0 +1,79 @@
...
+		if strings.HasPrefix(flag.Name, "no-") {
+			positiveName := flag.Name[len(negPrefix):len(flag.Name)]
+			positive := f.Lookup(positiveName)
+			if positive.Changed {
+				err = fmt.Errorf("only one of --%s and --%s may be specified",
+					flag.Name, positiveName)
+				return
+			}
+			var noValue bool
+			noValue, err = strconv.ParseBool(flag.Value.String())
```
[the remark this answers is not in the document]
sixolet: I opened https://github.com/spf13/pflag/issues/214 upstream; I read the code and determined that this case is difficult-to-impossible to disable without deep surgery into the pflags code.

pkg/kn/flags/bool_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+func TestBooleanPair(t *testing.T) {
+	cases := []*boolPairTestCase{
+		{"foo", true, []string{}, true, ""},
+		{"foo", true, []string{"--foo"}, true, ""},
+		{"foo", true, []string{"--no-foo"}, false, ""},
+		{"foo", false, []string{"--foo"}, true, ""},
+		{"foo", false, []string{}, false, ""},
+		{"foo", false, []string{"--no-foo"}, false, ""},
+		{"foo", true, []string{"--foo", "--no-foo"}, false, "only one of"},
```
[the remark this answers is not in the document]
sixolet: Agreed, but what we have now allows the syntax we want, just doesn't *disallow* the syntax we don't want. 

If the `pflags` folks agree that https://github.com/spf13/pflag/issues/214 is worth looking at, I can pretty easily make a PR up there to do what we want, but for now I'm not sure there's recourse other than getting to the arguments before they're parsed, which is, IMHO, more mess than it's worth.

pkg/kn/flags/bool.go
sixolet: Yeah, I went back and forth on the whole "do you have a flag that basically is a noop to do the default thing" issue, but was settled by the case where it's also a config file-settable variable, which effectively makes a context-dependent default, and you definitely want both flags. So leaving them in.

pkg/kn/flags/bool.go
sixolet: done

sixolet:
@maximilien I think I did address @rhuss's comments, though unfortunately by pointing to an issue I opened upstream. I don't think I can do better than that RN.

pkg/kn/flags/bool.go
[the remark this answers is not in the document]
rhuss: So if I understand correctly, then we can't prevent the user from providing a value, so we need to check it ?

What if we just ignore the value ? (just asking, I see that `--no-foo=false` would be still possible, but if we respect this right now, it's hard to remove later when this (hopefully) becomes an error)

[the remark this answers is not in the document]
rhuss: I mean, couldn't we just check for `flag.Changed()` to check whether the option was provided instead of evaluating the value (which could be not want we want it to be) ?

rhuss:
I'm fine, but maybe we could just check for the existence of a flag via `flag.Lookup()` instead of evaluating its value? (see https://github.com/knative/client/pull/346#discussion_r312385501)

pkg/kn/flags/bool.go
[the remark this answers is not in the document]
sixolet: I see what you're saying, though I think you mean `.Changed`, and I think I disagree. While you _can_ provide `true` or `false`, having it not matter what you provide is even more confusing, I think.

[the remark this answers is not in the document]
sixolet: Or I guess I could guard against unexpected values...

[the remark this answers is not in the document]
rhuss: Yeah, guarding would be the best solution IMO (and btw, 'Changed()' probably doesn't fire when you use the default value. not sure, but had the impression when looking at the implementation). 

And then we should add tests for this behaviour, too.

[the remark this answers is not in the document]
sixolet: Tested it, and `Changed` works.

pkg/kn/flags/bool_test.go
```diff
@@ -0,0 +1,72 @@
...
+	for _, c := range cases {
+		f := &pflag.FlagSet{}
+		var result bool
+		AddBothBoolFlags(f, &result, c.name, "", c.defaultVal, "set "+c.name)
+		f.Parse(c.flags)
+		err := ReconcileBoolFlags(f)
+		if c.expectedErrText != "" {
+			assert.ErrorContains(t, err, c.expectedErrText)
+		} else {
+			assert.Assert(t, cmp.Equal(result, c.expectedResult))
```
rhuss:
```suggestion
			assert.Equal(t, result, c.expectedResult)
```

pkg/kn/flags/bool.go
```diff
@@ -48,10 +48,11 @@ func AddBothBoolFlags(f *pflag.FlagSet, p *bool, name, short string, value bool,
...
-// "--no-foo" if provided, and returns an error if both were provided.
+// "--no-foo" if provided, and returns an error if both were provided or an
+// explicit value of false was provided to either (as that's confusing).
 func ReconcileBoolFlags(f *pflag.FlagSet) error {
 	var err error
-	f.Visit(func(flag *pflag.Flag) {
+	f.VisitAll(func(flag *pflag.Flag) {
 		// Return early from our comprehension
 		if err != nil {
 			return
```
maximilien: Style nit... It's interesting that you are simply using `return` here without explicitly stating the `error` being returned and have not defined the function with a named return --- something I have never been a fan of.

So to be clear, I'd expect one of two things: 
a) `return err`

or 

b) `return` if the function was defined as `func ReconcileBoolFlags(f *pflag.FlagSet) (err error)`

Maybe I am missing something... obviously am since this compiles. Anyhow.

maximilien: Oh I see this return is for the `VisitAll` anonymous function. Oh life makes sense again :) Ignore above.

pkg/kn/flags/bool.go
```diff
@@ -0,0 +1,104 @@
...
+// If you do this, make sure you call ReconcileBoolFlags later to catch errors and
+// set the relationship between the flag values.
+func AddBothBoolFlags(f *pflag.FlagSet, p *bool, name, short string, value bool, usage string) {
+
+	negativeName := negPrefix + name
+
+	f.BoolVarP(p, name, short, value, usage)
+	f.Bool(negativeName, !value, "do not "+usage)
+
+	if value {
```
duglin: Just want make sure I grok this... if the default value is `true` then we'll show the non-no flag. Isn't that backwards? Shouldn't we show the `--no-xxx` flag instead since the more obvious option is to turn it off?  Or an I reading this backwards?

pkg/kn/flags/bool.go
```diff
@@ -0,0 +1,104 @@
...
+		err := f.MarkHidden(negativeName)
+		if err != nil {
+			panic(err)
+		}
+	}
+}
+
+// ReconcileBoolFlags sets the value of the all the "--foo" flags based on
+// "--no-foo" if provided, and returns an error if both were provided or an
+// explicit value of false was provided to either (as that's confusing).
```
duglin: Bummer that there's no way to tell cobra to not allow `=xxx`.

pkg/kn/flags/bool.go
```diff
@@ -0,0 +1,104 @@
...
+	var err error
+	f.VisitAll(func(flag *pflag.Flag) {
+		// Return early from our comprehension
+		if err != nil {
+			return
+		}
+		// Walk the "no-" versions of the flags. Make sure we didn't set
+		// both, and set the positive value to the opposite of the "no-"
+		// value if it exists.
+		if strings.HasPrefix(flag.Name, "no-") {
```
duglin: While not too likely, I think it would be better to also check to make sure it's a bool flag and there's the non-no flag present. I'm concerned about someone defining just a `--no-foo` flag by itself and this if-statement picking it up as one of these special ones.  The non-no check might be annoying, if so, can we at least check that it's a bool?

duglin: Also, on the next line you use `negPrefix` but above you use `no-` - for consistency let's use `negPrefix` above too.

pkg/kn/flags/bool.go
```diff
@@ -0,0 +1,104 @@
...
+			positiveName := flag.Name[len(negPrefix):]
+			positive := f.Lookup(positiveName)
```
duglin: ok so you do look for the positive here... let's move that outside of this if-block and then you can use that in the if-statement.  Then you can avoid the NPE if for some reason `positive` is nil.

pkg/kn/flags/bool.go
```diff
@@ -0,0 +1,104 @@
...
+						flag.Name, positiveName)
+					return
+				}
+				var noValue bool
+				noValue, err = strconv.ParseBool(flag.Value.String())
+				if err != nil {
+					return
+				}
+				if !noValue {
+					err = fmt.Errorf("use --%s instead of providing false to --%s",
```
duglin: can you quote `false` so people know it's a reserved word and not an English one?

duglin: actually, why not print an error no matter the value?  I would expect an error for `--no-foo=bar`. Then you don't even need to check for a valid bool.

pkg/kn/flags/bool.go
```diff
@@ -0,0 +1,104 @@
...
+					if err != nil {
+						return
+					}
+				}
+				err = positive.Value.Set(strconv.FormatBool(!noValue))
+			} else if positive.Changed {
+				// For the positive version, just check it wasn't set to the
+				// confusing "false" value.
+				var yesValue bool
+				yesValue, err = strconv.ParseBool(positive.Value.String())
```
duglin: ditto here - let's just flag it if there's any value at all.  It would be more consistent than to allow some bool value but not the other.

duglin:
overall it looks good - just a few minor comments

pkg/kn/flags/bool.go
navidshaikh: Take example of `--force` flag:
default value is `false`, we don't show `--no-force` because thats implied, we instead show `--force` so user choose if they want to change the behavior by flagging `--force`.


Another example of `--headers` flag for listing:
default value true, we don't show what is obvious and default behavior, 
but show `--no-headers` to change the default behavior.

pkg/kn/flags/bool.go
navidshaikh: > Also, on the next line you use negPrefix but above you use no- - for consistency let's use negPrefix above too.

I think above is actual flag value while negPrefix is variable.

pkg/kn/flags/bool.go
navidshaikh: I revisited this, thought its better to represent intent than writing actual value, false should be fine.

pkg/kn/flags/bool.go
duglin: It's the use of `positive.` on the following lines that will cause the NPE

pkg/kn/flags/bool.go
duglin: ok never mind - I was reading it backwards :-)

pkg/kn/flags/bool.go
sixolet: Fixed.

pkg/kn/flags/bool.go
```diff
@@ -62,17 +63,41 @@ func ReconcileBoolFlags(f *pflag.FlagSet) error {
...
+				}
+				var noValue bool
+				noValue, err = strconv.ParseBool(flag.Value.String())
+				if err != nil {
+					return
+				}
+				if !noValue {
+					err = fmt.Errorf("use --%s instead of providing false to --%s",
+						positiveName, flag.Name)
+					if err != nil {
```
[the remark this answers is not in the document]
sixolet: ok

PR #347, opened 2019-08-08
Supports setting comma separated env vars

navidshaikh:
/lint

## Proposed Changes
 `--env ENV1=VAL1 --env ENV2=VAL2,ENV3=VAL3`

 should work specifying by separate `--env` or specifying comma separated
 values to `--env` or combination of both.

**Release Note**

```release-note
Supports specifying comma separated environment variables as `--env KEY1=VAL1,KEY2=VAL2`
```

### DIFF
--- a/pkg/kn/commands/service/service_update_test.go
+++ b/pkg/kn/commands/service/service_update_test.go
@@ -258,6 +258,65 @@ func TestServiceUpdateEnv(t *testing.T) {
 	}
 }
 
+func TestServiceUpdateMultipleEnvs(t *testing.T) {
+	orig := &v1alpha1.Service{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Service",
+			APIVersion: "knative.dev/v1alpha1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "foo",
+			Namespace: "default",
+		},
+		Spec: v1alpha1.ServiceSpec{
+			DeprecatedRunLatest: &v1alpha1.RunLatestType{
+				Configuration: v1alpha1.ConfigurationSpec{
+					DeprecatedRevisionTemplate: &v1alpha1.RevisionTemplateSpec{
+						Spec: v1alpha1.RevisionSpec{
+							DeprecatedContainer: &corev1.Container{},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	template, err := servinglib.RevisionTemplateOfService(orig)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	servinglib.UpdateImage(template, "gcr.io/foo/bar:baz")
+
+	action, updated, _, err := fakeServiceUpdate(orig, []string{
+		"service", "update", "foo", "-e", "TARGET=Awesome,FOR=EnvTest", "--async"}, false)
+
+	if err != nil {
+		t.Fatal(err)
+	} else if !action.Matches("update", "services") {
+		t.Fatalf("Bad action %v", action)
+	}
+	expectedEnvVar1 := corev1.EnvVar{
+		Name:  "TARGET",
+		Value: "Awesome",
+	}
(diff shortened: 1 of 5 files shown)


rhuss:
Fine with me, but please note that this might conflict with #342 which also has some minor tweaks on env vars. Maybe we can hold off with this one until #342 is merged ? (its soonish)

rhuss:
@navidshaikh don't forget to add to CHANGELOG.adoc, too

navidshaikh:
>  this might conflict with #342

@rhuss : I thought about that, separating the values by comma for `--env` should be handled by cobra and we receive the values in list before we interpret the values as mentioned #342. 

Without comma
```
--env KEY=value sets EnvVar{Name: "KEY", Value: "value"} (current behavior)
--env KEY= sets EnvVar{Name: "KEY"} (current behavior)
--env KEY
--env KEY- removes the env var matching EnvVar{Name: "KEY", ...}
```

With comma
```
--env KEY=value,KEY2=value2
--env KEY=,KEY2=
--env KEY,KEY2
--env KEY-,KEY2-
```
I think this should just work fine without any additional processing.

duglin:
Not strong opinion either way, but do people have a guess as to how often someone may use a comma in an env var? Is there a way to escape it via cobra?

navidshaikh:
> but do people have a guess as to how often someone may use a comma in an env var? Is there a way to escape it via cobra?

Supporting this would disallow users from setting commas in env var values, need to check about escaping but single/double quotes on only values or even whole pair, back/forward slash didnt work.

```
➜  client git:(pr/multiple-envs)  ./kn service update svc --env TARGET="hello2,"
--env argument requires a value that contains the '=' character; got 

➜  client git:(pr/multiple-envs)  ./kn service update svc --env TARGET='hello2,'
--env argument requires a value that contains the '=' character; got 

➜  client git:(pr/multiple-envs)  ./kn service update svc --env TARGET=hello2\,  
--env argument requires a value that contains the '=' character; got 

➜  client git:(pr/multiple-envs)  ./kn service update svc --env TARGET=hello2/,
--env argument requires a value that contains the '=' character; got 

➜  client git:(pr/multiple-envs)  ./kn service update svc --env "TARGET=hello2,"
--env argument requires a value that contains the '=' character; got 

# the current behavior allows commas
➜  client git:(pr/multiple-envs)  git checkout master                           
➜  client git:(master)  go run ./cmd/kn/main.go service update svc --env TARGET=hello, 

➜  client git:(master)  curl http://svc.default.apps-crc.testing                      
Hello hello,!
```
The current forma though support it (check last service update command above).

duglin:
I did some digging and found this:
https://github.com/spf13/pflag/pull/59/files

And this does seem to work:
```
./kn service update echo --env=\"abc=def,def=qwe\"
```

So if we document this then I think we're good

duglin:
We should add a test for it too :-)

navidshaikh:
@duglin : Nice! I can do this as well
```
kn service update svc --env \"TARGET=hello,=\",TARGET=overwrite
```

pkg/kn/commands/service/service_update_test.go
```diff
@@ -258,6 +258,65 @@ func TestServiceUpdateEnv(t *testing.T) {
...
+	}
+
+	template, err := servinglib.RevisionTemplateOfService(orig)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	servinglib.UpdateImage(template, "gcr.io/foo/bar:baz")
+
+	action, updated, _, err := fakeServiceUpdate(orig, []string{
```
sixolet: Add test for quoting (to have a comma in an env var), and for repeating the flag, to make sure we cover all the bases.

jimcurtis:
LGTM

navidshaikh:
repeat --env flag to specify multiple env vars, escaping might create more confusion.

PR #351, opened 2019-08-08
Wrap flag usage to terminal size

sixolet:
It's weird that the most idiomatic library to go get a term size is in a crypto lib, but it's not actually that bad of a dep and greatly improves the experience on a smaller-width terminal.

We don't bother to wrap if there's no gettable terminal size.


go.mod
```diff
@@ -6,16 +6,17 @@ require (
 	github.com/ghodss/yaml v1.0.0 // indirect
 	github.com/gogo/protobuf v1.2.1 // indirect
 	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b // indirect
-	github.com/golang/protobuf v1.3.2 // indirect
+	github.com/golang/protobuf v1.3.1 // indirect
```
rhuss: why a downgrade ?

go.mod
```diff
@@ -6,16 +6,17 @@ require (
...
-	github.com/google/go-containerregistry v0.0.0-20190723205234-74f764b9c575 // indirect
+	github.com/google/go-containerregistry v0.0.0-20190623150931-ca8b66cb1b79 // indirect
+	github.com/google/gofuzz v1.0.0 // indirect
 	github.com/googleapis/gnostic v0.3.0 // indirect
 	github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79 // indirect
 	github.com/hashicorp/golang-lru v0.5.1 // indirect
 	github.com/imdario/mergo v0.3.7 // indirect
 	github.com/inconshreveable/mousetrap v1.0.0 // indirect
-	github.com/json-iterator/go v1.1.7 // indirect
+	github.com/json-iterator/go v1.1.6 // indirect
```
rhuss: downgrade ?

go.mod
```diff
@@ -28,7 +29,8 @@ require (
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/pflag v1.0.3
 	github.com/spf13/viper v1.3.1
-	golang.org/x/net v0.0.0-20190724013045-ca1201d0de80 // indirect
+	golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2
```
rhuss: Is there a released version we could use ?

go.mod
sixolet: I have no idea, I just let `./hack/build.sh` do its thing.

go.mod
sixolet: I have no idea, I just let `./hack/build.sh` do its thing.

go.mod
sixolet: I'll look.

go.mod
rhuss: Maybe a rebase is needed ?

go.mod
sixolet: No. Or, kind of, but using it ends up with even more downgrades.

go.mod
sixolet: Should be sorted now.

PR #352, opened 2019-08-09
Fix index out of range error

duglin:
Fixed:
```
$ ./kn plugin list --lookup-plugins-in-path
panic: runtime error: index out of range

goroutine 1 [running]:
github.com/knative/client/pkg/kn/core.extractKnPluginFlags(0xc00003c0c0, 0x4, 0x4, 0xc0002be280, 0xc000000180, 0xc0002ba180, 0xc0000de670, 0xc000117f68)
        /root/go/src/knative.dev/client/pkg/kn/core/root.go:222 +0x666
github.com/knative/client/pkg/kn/core.NewDefaultKnCommand(0xc000000000)
        /root/go/src/knative.dev/client/pkg/kn/core/root.go:46 +0x79
main.main()
        /root/go/src/knative.dev/client/cmd/kn/main.go:33 +0x46
```

While in there I fixed it so that we only look for --flag at the start
of the arg, and handle the --flag, --flag=, and --flag=xxx cases.
If there's --flagX we skip it because it's some other flag and not one
of interest to us.

Also, fixed some error processing that seemed a bit off.

/lint

### DIFF
--- a/pkg/kn/core/root.go
+++ b/pkg/kn/core/root.go
@@ -45,7 +45,8 @@ func NewDefaultKnCommand() *cobra.Command {
 	// and will not be parsed
 	pluginsDir, lookupPluginsInPath, err := extractKnPluginFlags(os.Args)
 	if err != nil {
-		panic("Invalid plugin flag value")
+		fmt.Fprintf(os.Stderr, "%s\n", err)
+		os.Exit(1)
 	}
 
 	pluginHandler := plugin.NewDefaultPluginHandler(plugin.ValidPluginFilenamePrefixes,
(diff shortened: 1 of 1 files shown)


pkg/kn/core/root.go
```diff
@@ -45,7 +45,8 @@ func NewDefaultKnCommand() *cobra.Command {
 	// and will not be parsed
 	pluginsDir, lookupPluginsInPath, err := extractKnPluginFlags(os.Args)
 	if err != nil {
-		panic("Invalid plugin flag value")
+		fmt.Fprintf(os.Stderr, "%s\n", err)
+		os.Exit(1)
```
duglin: Changed this because there's no reason to show a stack trace to the end user for a bad cmd line arg. Just show the error message.

PR #355, opened 2019-08-09
docs: References golang 1.12.7 or later in docs

### LINKED ISSUE #353: Minimum golang version for kn
I am running Fedora 29 and golang versions as below
```
13:37 ➜  client git:(fix-docs)  cat /etc/redhat-release 
Fedora release 29 (Twenty Nine)
13:38 ➜  client git:(fix-docs)  rpm -q golang
golang-1.11.11-1.fc29.x86_64
```

latest master fails to build for me as it requires golang 1.12 for strings.ReplaceAll,
```
13:39 ➜  client git:(master)  ./hack/build.sh 
🕸️ Update
🧽 Format
⚖️ License
📖 Docs
# github.com/knative/client/pkg/kn/commands/plugin
pkg/kn/commands/plugin/plugin_verifier.go:280:12: undefined: strings.ReplaceAll
```

Our docs point 1.11.4 or later, that should be replaced with 1.12.7 or latest (the latest one available https://golang.org/dl/)

navidshaikh:
Fixes #353

As we require minimum 1.12.7 for building  kn

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -24,7 +24,7 @@ Start by creating [a GitHub account](https://github.com/join), then set up
 You must install these tools:
 
 1. [`go`](https://golang.org/doc/install): The language `Knative Client` is
-   built in (1.11.4 or later)
+   built in (1.12 or later)
 1. [`goimports`](https://godoc.org/golang.org/x/tools/cmd/goimports)
 1. `gcc` compiler: Used during testing. Not needed if golang is installed
    via the installer


rhuss:
Not sure if 1.12.7 specifically is required or if 1.12 in general is enough. I would omit the patchlevel from the version

navidshaikh:
@rhuss:

The page <https://golang.org/dl/> we're referencing for downloading golang has listing of (only) 1.12.7 in `Stable Version`  section (for 1.12.x variants).

rhuss:
We then would have to upgrade our docs for every patch level version then. Still, think that 1.12 is good enough to mention (and if there is only 1.12.7 to download, then fine, but people with 1.12.3 still can compile it, I think).

Nit-picking on high-level though, I'm fine with either.

PR #356, opened 2019-08-09
Add support for -A variant of --all-namespaces

duglin:
-A is my newest favorite kubectl flag, it's so much easier/faster to type
than --all-namespaces. kn users should benefit from it too!

/lint

**Release Note**

```release-note
The kubectl `-A` variant of the `--all-namespaces` flag is now supported.
```

### DIFF
--- a/pkg/kn/commands/namespaced_test.go
+++ b/pkg/kn/commands/namespaced_test.go
@@ -78,6 +78,18 @@ func TestGetNamespaceAllNamespacesSet(t *testing.T) {
 	if actualNamespace != expectedNamespace {
 		t.Fatalf("Incorrect namespace retrieved: %v, expected: %v", actualNamespace, expectedNamespace)
 	}
+
+	// Now do it again using -A instead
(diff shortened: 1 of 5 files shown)


pkg/kn/commands/namespaced_test.go
```diff
@@ -95,6 +107,18 @@ func TestGetNamespaceDefaultAllNamespacesUnset(t *testing.T) {
 	if actualNamespace != expectedNamespace {
 		t.Fatalf("Incorrect namespace retrieved: %v, expected: %v", actualNamespace, expectedNamespace)
 	}
+
+	// Now do it again using -A instead
+	testCmd.SetArgs([]string{"-A"})
```
rhuss: Could we make those tests in a loop like in 

```golang
for _, arg := range []string { "--all-namespaces", "-A" } {
   ....

}
```

This would avoid copy & paste code. Same for the other test.

duglin: sure

duglin: done

PR #359, opened 2019-08-09
chore(plugins, services, routes): Aligned "Not found" messages

rhuss:
For the sake of consistency with out own service commands and for kubectl,

`kn plugin list` should print:

```
No plugins found.
```

`kn service list` should print:

```
No services found.
```

and not `No resources found.` like it does now.

Same for `kn routes list`.

### DIFF
--- a/pkg/kn/commands/plugin/plugin_list.go
+++ b/pkg/kn/commands/plugin/plugin_list.go
@@ -80,7 +80,9 @@ func listPlugins(cmd *cobra.Command, flags pluginListFlags) error {
 
 	if len(pluginsFound) == 0 {
 		if flags.verbose {
-			fmt.Fprintf(out, "No plugins found in path %s\n", pluginPath)
+			fmt.Fprintf(out, "No plugins found in path %s.\n", pluginPath)
+		} else {
+			fmt.Fprintln(out, "No plugins found.")
 		}
 		return nil
 	}
--- a/pkg/kn/commands/route/list.go
+++ b/pkg/kn/commands/route/list.go
@@ -15,6 +15,7 @@
 package route
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/knative/client/pkg/kn/commands"
@@ -57,13 +58,13 @@ func NewRouteListCommand(p *commands.KnParams) *cobra.Command {
 			case 1:
 				routeList, err = client.ListRoutes(v1alpha12.WithName(args[0]))
 			default:
-				return fmt.Errorf("'kn route list' accepts maximum 1 argument.")
+				return errors.New("'kn route list' accepts only one additional argument")
 			}
 			if err != nil {
 				return err
 			}
 			if len(routeList.Items) == 0 {
-				fmt.Fprintf(cmd.OutOrStdout(), "No resources found.\n")
+				fmt.Fprintf(cmd.OutOrStdout(), "No routes found.\n")
 				return nil
 			}
 			printer, err := routeListFlags.ToPrinter()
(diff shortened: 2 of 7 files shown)


pkg/kn/commands/plugin/plugin_list.go
```diff
@@ -80,7 +80,9 @@ func listPlugins(cmd *cobra.Command, flags pluginListFlags) error {
 
 	if len(pluginsFound) == 0 {
 		if flags.verbose {
-			fmt.Fprintf(out, "No plugins found in path %s\n", pluginPath)
+			fmt.Fprintf(out, "No plugins found in path %s.\n", pluginPath)
```
duglin: How about we remove the "in path..." part here since it makes for a really long and hard to read message, and if we still want to show the PATH (which I think is silly) then do it above at line 78 like all of the other settings we show.

duglin:
+1 to the idea

pkg/kn/commands/plugin/plugin_list.go
rhuss: Actually this would show the whole path which has been used(including $PATH if enabled). this is not exactly the same as in 78 as this doesn't show the $PATH. And as this is written only on `--verbose` I think it could be even redundant.

As I guess you not a typical `--verbose` user ;-P, this shouldn't be so critical ?

rhuss:
Will fix the integration test later. Now: Barbecue ;-)

pkg/kn/commands/route/list.go
```diff
@@ -57,13 +57,13 @@ func NewRouteListCommand(p *commands.KnParams) *cobra.Command {
 			case 1:
 				routeList, err = client.ListRoutes(v1alpha12.WithName(args[0]))
 			default:
-				return fmt.Errorf("'kn route list' accepts maximum 1 argument.")
+				return fmt.Errorf("'kn route list' accepts no or one name argument, but not %d arguments.", len(args))
```
maximilien: Is this just me... I find this bit: `accepts no or one name argument` hard to parse. Sorry.

rhuss: maybe "none or a single name argument"   is better ? Hope for some native speaker to kick. @abrennan89 any suggestion for a better wording here ?

pkg/kn/commands/plugin/plugin_list.go
duglin: yep - I tend to be a minimalist by default, and as you've said, that's what --verbose is for.
So w/o any kind of --verbose flag I want to see the bare minimum info, just the facts. And showing me the PATH just clutters up the screen. It's only useful when things go wrong (which most of the time they won't), but when they do, that's when they can add --verbose and see all of the gory details of where we looked.

pkg/kn/commands/route/list.go
duglin: Perhaps: `'kn route list' does not accept more than one name argument` ?  Since they passed in more than 1.

I don't think we need to show how many they passed in, they can count it if they want :-)

rhuss: > Perhaps: `'kn route list' does not accept more than one name argument` ? Since they passed in more than 1.

That does not explicit mention that no argument is allowed as well. I settled now to a simple message "allows only one additional argument"

maximilien: I like the other changes. Any agreement on this one? I like @duglin's suggestion but would consider something else (if proposed). For me I'd go with the affirmative version: `'kn route list' accepts 0 or 1 argument` or similar

pkg/kn/commands/route/list.go
```diff
@@ -57,13 +58,13 @@ func NewRouteListCommand(p *commands.KnParams) *cobra.Command {
 			case 1:
 				routeList, err = client.ListRoutes(v1alpha12.WithName(args[0]))
 			default:
-				return fmt.Errorf("'kn route list' accepts maximum 1 argument.")
+				return errors.New("'kn route list' accepts only one additional argument")
```
navidshaikh: @maximilien : You had a review comment on this line, does this look good to you now ?

PR #361, opened 2019-08-14
Always create in new format rather than old

sixolet:
This is the version of the pull request where we don't make a flag to determine whether we want to use the "old format" or the "new lemonadey v1beta1 format", we just create new services in the new format.

This will DROP SUPPORT for 0.5

We could also have a flag, but this PR is quicker and easier to make, so I'm making it and we can discuss.

### DIFF
--- a/pkg/kn/commands/service/service_create_mock_test.go
+++ b/pkg/kn/commands/service/service_create_mock_test.go
@@ -75,7 +75,7 @@ func TestServiceCreateLabel(t *testing.T) {
 	template.Spec.Containers[0].Image = "gcr.io/foo/bar:baz"
 	r.CreateService(service, nil)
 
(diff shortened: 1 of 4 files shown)


duglin:
SGTM

pkg/kn/commands/service/service_create_mock_test.go
```diff
@@ -75,7 +75,7 @@ func TestServiceCreateLabel(t *testing.T) {
 	template.Spec.Containers[0].Image = "gcr.io/foo/bar:baz"
 	r.CreateService(service, nil)
 
-	output, err := executeServiceCommand(client, "create", "foo", "--image", "gcr.io/foo/bar:baz", "-l", "a=mouse", "--label", "b=cookie", "--label=empty", "--async")
+	output, err := executeServiceCommand(client, "create", "foo", "--image", "gcr.io/foo/bar:baz", "-l", "a=mouse", "--label", "b=cookie", "--label=empty", "--async", "--revision-name=")
```
navidshaikh: Right.

PR #366, opened 2019-08-15
feature(service list): Print NAMESPACE column as the first column when --all-namespaces is specified

toVersus:
/lint

## Proposed Changes

`kn service list` should print NAMESPACE column when specifying `--all-namespaces` option, which follows the [kubectl convention](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-cli/kubectl-conventions.md#output-conventions).

> NAMESPACE should be displayed as the first column when --all-namespaces is specified

I know we can find a namespace we deployed our service by inspecting an URL column. It tells us a namespace by default because its FQDN is composed of `{{.Name}}.{{.Namespace}}.{{.Domain}}`. However, we can easily change this FQDN template by modifying `domainTemplate` field in [config-network.yaml](https://github.com/knative/serving/blob/d37bd91a9355ea602819de1bf2a25fcf2294c06c/config/config-network.yaml#L116). In this case, we have no idea to see a namespace from the output of `kn service list --all-namespace` at a glance. Also, I think it is useful to see a bunch of services in the same namespace by sorting in alphabetical order.

* Print NAMESPACE as the first column when `--all-namespaces` is specified in `kn service list` command
* Show services in default namespace at the top of the list
* Add unit tests using fake services as well as mock framework

Before:

```
$ ./kn service list -A
NAME    URL                                GENERATION   AGE     CONDITIONS   READY   REASON
bar     http://bar-bar.example.com         1            73m     3 OK / 3     True
echo    http://echo-bar.example.com        1            73m     3 OK / 3     True
echo    http://echo-default.example.com    1            6h10m   3 OK / 3     True
echo    http://echo-foo.example.com        1            74m     3 OK / 3     True
foo     http://foo-default.example.com     1            6h9m    3 OK / 3     True
foo     http://foo-foo.example.com         1            74m     3 OK / 3     True
hello   http://hello-bar.example.com       1            72m     3 OK / 3     True
hello   http://hello-default.example.com   1            42h     3 OK / 3     True
hello   http://hello-foo.example.com       1            73m     3 OK / 3     True
```

After:

```
$ ./kn service list -A
NAMESPACE   NAME    URL                                GENERATION   AGE     CONDITIONS   READY   REASON
default     echo    http://echo-default.example.com    1            5h51m   3 OK / 3     True
default     foo     http://foo-default.example.com     1            5h51m   3 OK / 3     True
default     hello   http://hello-default.example.com   1            42h     3 OK / 3     True
bar         bar     http://bar-bar.example.com         1            54m     3 OK / 3     True
bar         echo    http://echo-bar.example.com        1            54m     3 OK / 3     True
bar         hello   http://hello-bar.example.com       1            54m     3 OK / 3     True
foo         echo    http://echo-foo.example.com        1            55m     3 OK / 3     True
foo         foo     http://foo-foo.example.com         1            56m     3 OK / 3     True
foo         hello   http://hello-foo.example.com       1            55m     3 OK / 3     True
```

### DIFF
--- a/pkg/kn/commands/human_readable_flags.go
+++ b/pkg/kn/commands/human_readable_flags.go
@@ -62,7 +62,6 @@ func NewHumanPrintFlags() *HumanPrintFlags {
 // EnsureWithNamespace sets the "WithNamespace" humanreadable option to true.
 func (f *HumanPrintFlags) EnsureWithNamespace() {
 	f.WithNamespace = true
-	return
 }
 
 // Private functions
--- a/pkg/kn/commands/human_readable_flags.go
+++ b/pkg/kn/commands/human_readable_flags.go
@@ -30,7 +30,7 @@ import (
 // Given the following flag values, a printer can be requested that knows
 // how to handle printing based on these values.
 type HumanPrintFlags struct {
-	//TODO: Add more flags as required
+	AllNamespaces bool
 }
 
 // AllowedFormats returns more customized formating options
@@ -42,7 +42,7 @@ func (f *HumanPrintFlags) AllowedFormats() []string {
 // ToPrinter receives returns a printer capable of
 // handling human-readable output.
 func (f *HumanPrintFlags) ToPrinter(getHandlerFunc func(h hprinters.PrintHandler)) (hprinters.ResourcePrinter, error) {
-	p := hprinters.NewTablePrinter(hprinters.PrintOptions{})
+	p := hprinters.NewTablePrinter(hprinters.PrintOptions{f.AllNamespaces})
 	getHandlerFunc(p)
 	return p, nil
 }
@@ -50,7 +50,12 @@ func (f *HumanPrintFlags) ToPrinter(getHandlerFunc func(h hprinters.PrintHandler
 // AddFlags receives a *cobra.Command reference and binds
 // flags related to human-readable printing to it
 func (f *HumanPrintFlags) AddFlags(c *cobra.Command) {
-	//TODO: Add more flags as required
+	// Bind AllNamespaces flag
+	all, err := c.Flags().GetBool("all-namespaces")
+	if err != nil {
+		fmt.Fprintln(c.OutOrStdout(), err)
+	}
+	f.AllNamespaces = all
 }
 
 // NewHumanPrintFlags returns flags associated with
--- a/pkg/kn/commands/human_readable_flags.go
+++ b/pkg/kn/commands/human_readable_flags.go
@@ -30,7 +30,7 @@ import (
 // Given the following flag values, a printer can be requested that knows
 // how to handle printing based on these values.
 type HumanPrintFlags struct {
-	AllNamespaces bool
+	WithNamespace bool
 }
 
 // AllowedFormats returns more customized formating options
@@ -42,7 +42,7 @@ func (f *HumanPrintFlags) AllowedFormats() []string {
 // ToPrinter receives returns a printer capable of
 // handling human-readable output.
 func (f *HumanPrintFlags) ToPrinter(getHandlerFunc func(h hprinters.PrintHandler)) (hprinters.ResourcePrinter, error) {
-	p := hprinters.NewTablePrinter(hprinters.PrintOptions{f.AllNamespaces})
+	p := hprinters.NewTablePrinter(hprinters.PrintOptions{f.WithNamespace})
 	getHandlerFunc(p)
 	return p, nil
 }
@@ -50,18 +50,19 @@ func (f *HumanPrintFlags) ToPrinter(getHandlerFunc func(h hprinters.PrintHandler
 // AddFlags receives a *cobra.Command reference and binds
 // flags related to human-readable printing to it
 func (f *HumanPrintFlags) AddFlags(c *cobra.Command) {
-	// Bind AllNamespaces flag
-	all, err := c.Flags().GetBool("all-namespaces")
-	if err != nil {
-		fmt.Fprintln(c.OutOrStdout(), err)
-	}
-	f.AllNamespaces = all
+	// TODO: Add more flags as required
 }
 
 // NewHumanPrintFlags returns flags associated with
 // human-readable printing, with default values set.
 func NewHumanPrintFlags() *HumanPrintFlags {
 	return &HumanPrintFlags{}
+}
+
+// EnsureWithNamespace sets the "WithNamespace" humanreadable option to true.
+func (f *HumanPrintFlags) EnsureWithNamespace() {
+	f.WithNamespace = true
+	return
 }
 
 // Private functions
--- a/pkg/kn/commands/service/human_readable_flags.go
+++ b/pkg/kn/commands/service/human_readable_flags.go
@@ -33,8 +33,6 @@ func ServiceListHandlers(h hprinters.PrintHandler) {
 		{Name: "Namespace", Type: "string", Description: "Namespace of the Knative service", Priority: 0},
 		{Name: "Name", Type: "string", Description: "Name of the Knative service.", Priority: 1},
 		{Name: "Url", Type: "string", Description: "URL of the Knative service.", Priority: 1},
-		//{Name: "LastCreatedRevision", Type: "string", Description: "Name of last revision created.", Priority: 1},
-		//{Name: "LastReadyRevision", Type: "string", Description: "Name of last ready revision.", Priority: 1},
 		{Name: "Generation", Type: "integer", Description: "Sequence number of 'Generation' of the service that was last processed by the controller.", Priority: 1},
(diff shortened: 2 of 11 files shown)


duglin:
Clearly including the NS in there is +1  :-)

I'm just wondering about people using `kn` within scripts and possibly parsing the output, and whether having the number of columns change based on whether `-A` is used or not is a good or bad thing.  Meaning, should `NAMESPACE` be there even w/o `-A`? Honestly, I can't decide. It might depend on whether people consider the use of `-A` to make it a totally new command or just a variant of `service list` and the same "output processor" should work for both.

I do parse output from cmds like this a lot, but I don't know if I'm in a very small minority.

toVersus:
I'm not an enthusiast to use `kn` within my custom scripts, so I'm not sure about this :)

> Meaning, should NAMESPACE be there even w/o -A

I think it feels redundant because `kn service list` or `kn service list -n <namespace>` will only print the services in specific namespace. It is clear for the users without `NAMESPACE` column.

> It might depend on whether people consider the use of -A to make it a totally new command or just a variant of service list and the same "output processor" should work for both.

IMO, the motivation of `-A` option is to ensure the human readability, which is the same as `-o wide` option. It is not designed for improving machine readability for the use in scripts, so it is acceptable to change the number of columns.

pkg/kn/commands/human_readable_flags.go
```diff
@@ -42,15 +42,20 @@ func (f *HumanPrintFlags) AllowedFormats() []string {
...
 // AddFlags receives a *cobra.Command reference and binds
 // flags related to human-readable printing to it
 func (f *HumanPrintFlags) AddFlags(c *cobra.Command) {
-	//TODO: Add more flags as required
+	// Bind AllNamespaces flag
+	all, err := c.Flags().GetBool("all-namespaces")
+	if err != nil {
+		fmt.Fprintln(c.OutOrStdout(), err)
+	}
+	f.AllNamespaces = all
```
navidshaikh: Can we do make this other way round and handle it in more generic way ?

As in, flags defined in HumanPrintFlags, gets added to all resource listing commands.

ToPrinter handlers it for services, revisions and routes.

WDYT ?

pkg/kn/commands/service/human_readable_flags.go
```diff
@@ -15,26 +15,40 @@
...
+var baseKServiceColumnDefinitions = []metav1beta1.TableColumnDefinition{
+	{Name: "Name", Type: "string", Description: "Name of the Knative service."},
+	{Name: "Url", Type: "string", Description: "URL of the Knative service."},
+	//{Name: "LastCreatedRevision", Type: "string", Description: "Name of last revision created."},
+	//{Name: "LastReadyRevision", Type: "string", Description: "Name of last ready revision."},
+	{Name: "Generation", Type: "integer", Description: "Sequence number of 'Generation' of the service that was last processed by the controller."},
+	{Name: "Age", Type: "string", Description: "Age of the service."},
+	{Name: "Conditions", Type: "string", Description: "Conditions describing statuses of service components."},
+	{Name: "Ready", Type: "string", Description: "Ready condition status of the service."},
+	{Name: "Reason", Type: "string", Description: "Reason for non-ready condition of the service."},
```
navidshaikh: Take a look at priority field as done in kubectl, we could do something like
priority 0 = namespace
priority 1 = default
priority 2 = reserved for ` -o wide` option

if all-namespace flag is given, print priority 0,1
if all-namespace flag is not give, print priority 1

WDYT ?

pkg/kn/commands/service/service_list_flags.go
```diff
@@ -47,6 +47,13 @@ func (f *ServiceListFlags) ToPrinter() (hprinters.ResourcePrinter, error) {
 		return p, nil
 	}
 	// if no flags specified, use the table printing
+	if f.HumanReadableFlags.AllNamespaces {
+		p, err := f.HumanReadableFlags.ToPrinter(ServiceListAllNamespaceHandlers)
```
navidshaikh: We should be able to use single handler and adjust based on the printer flags.

pkg/kn/commands/human_readable_flags.go
toVersus: I'm not sure I can handle it in more generic way, but I changed to handle `--all-namespace` option in a similar way to [kubectl get](https://github.com/kubernetes/kubernetes/blob/2ca023f270779a7f017cf10b0e56d008262ed577/pkg/kubectl/cmd/get/humanreadable_flags.go#L30).

pkg/kn/commands/service/human_readable_flags.go
toVersus: SGTM. I updated to use priority field for switching output of column definitions 👍

pkg/kn/commands/service/human_readable_flags.go
```diff
@@ -25,16 +30,18 @@ import (
...
-		//{Name: "LastReadyRevision", Type: "string", Description: "Name of last ready revision."},
-		{Name: "Generation", Type: "integer", Description: "Sequence number of 'Generation' of the service that was last processed by the controller."},
-		{Name: "Age", Type: "string", Description: "Age of the service."},
-		{Name: "Conditions", Type: "string", Description: "Conditions describing statuses of service components."},
-		{Name: "Ready", Type: "string", Description: "Ready condition status of the service."},
-		{Name: "Reason", Type: "string", Description: "Reason for non-ready condition of the service."},
+		{Name: "Namespace", Type: "string", Description: "Namespace of the Knative service", Priority: 0},
+		{Name: "Name", Type: "string", Description: "Name of the Knative service.", Priority: 1},
+		{Name: "Url", Type: "string", Description: "URL of the Knative service.", Priority: 1},
+		//{Name: "LastCreatedRevision", Type: "string", Description: "Name of last revision created.", Priority: 1},
```
maximilien: Why comments. Can we remove

toVersus: Thanks! I'm not sure the reason why these commented codes are alive in there, but I'll delete it.

pkg/kn/commands/human_readable_flags.go
```diff
@@ -59,6 +59,12 @@ func NewHumanPrintFlags() *HumanPrintFlags {
 	return &HumanPrintFlags{}
 }
 
+// EnsureWithNamespace sets the "WithNamespace" humanreadable option to true.
+func (f *HumanPrintFlags) EnsureWithNamespace() {
+	f.WithNamespace = true
+	return
```
navidshaikh: remove return ?

pkg/kn/commands/human_readable_flags.go
```diff
@@ -42,15 +42,15 @@ func (f *HumanPrintFlags) AllowedFormats() []string {
...
+	p := hprinters.NewTablePrinter(hprinters.PrintOptions{f.WithNamespace})
 	getHandlerFunc(p)
 	return p, nil
 }
 
 // AddFlags receives a *cobra.Command reference and binds
 // flags related to human-readable printing to it
 func (f *HumanPrintFlags) AddFlags(c *cobra.Command) {
-	//TODO: Add more flags as required
+	// TODO: Add more flags as required
```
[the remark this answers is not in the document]
toVersus: That makes sense. We can easily introduce `--with-namespace` flag riding on this logic. After merging this PR, I bet someone will take this approach. If not, I will do it :grin:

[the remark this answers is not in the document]
navidshaikh: I asked this question in yesterday's WG call and we decided not to introduce new flag.

pkg/kn/commands/service/service_list_mock_test.go
```diff
@@ -0,0 +1,85 @@
...
+	assert.Assert(t, util.ContainsAll(output, "created", "svc3", "bar", "Waiting"))
+
+	output, err = executeServiceCommand(client, "list", "--all-namespaces")
+	assert.NilError(t, err)
+
+	outputLines := strings.Split(output, "\n")
+	assert.Assert(t, util.ContainsAll(outputLines[0], "NAMESPACE", "NAME", "URL", "GENERATION", "AGE", "CONDITIONS", "READY", "REASON"))
+	assert.Assert(t, util.ContainsAll(outputLines[1], "svc1", "default"))
+	assert.Assert(t, util.ContainsAll(outputLines[2], "svc3", "bar"))
+	assert.Assert(t, util.ContainsAll(outputLines[3], "svc2", "foo"))
```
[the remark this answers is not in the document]
toVersus: Good catch! I fixed to compare raw string output of service list.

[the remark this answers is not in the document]
toVersus: Thanks for providing your clear intention in Slack, I reverted a commit and swapped the input order of util.ContainsAll.

PR #368, opened 2019-08-15
Update serving dependency to 0.8; change all import paths

sixolet:
Serving moved from github.com/knative/serving to knative.dev/serving, and some duck type paths moved too. Update to using the new paths throughout.


sixolet:
It looks like coverage is failing due to a GOPATH issue. @adrcunha can you advise?

sixolet:
@maximilien coverage is breaking as a result of this change specifically, and I've been talking out of band with @adrcunha about how we'll commit this and then fix coverage in the prow config, and then turn the requirement back on.

PR #371, opened 2019-08-16
Adds option in hack/build.sh to build cross platform binaries

navidshaikh:
### Proposed Changes

* Use `./hack/build.sh -x` or `./hack/build.sh --build-cross`
* Doesn't touch any of the existing flow, providing these options will build binaries and exit


Harwayne:
> LGTM

I think this has to be in the form `/lgtm` with nothing else on the line for the prow bot to recognize it.

PR #376, opened 2019-08-19
Fix json path to latest revision in e2e tests to something reliable

sixolet:
This is just a small change that should make a helper function in the tests reliably return the latest revision name even when traffic doesn't have anything set to latest. No change to the actual running of the tests.


test/e2e/traffic_split_test.go
```diff
@@ -35,7 +35,7 @@ var targetFieldsLength = 4
 var targetsJsonPath = "jsonpath={range .status.traffic[*]}{.tag}{','}{.revisionName}{','}{.percent}{','}{.latestRevision}{'|'}{end}"
 
 // returns deployed service latest revision name
-var latestRevisionJsonPath = "jsonpath={.status.traffic[?(@.latestRevision==true)].revisionName}"
+var latestRevisionJsonPath = "jsonpath={.status.latestCreatedRevisionName}"
```
navidshaikh: Thanks!

PR #379, opened 2019-08-20
Service describe:  information about revision status, tags, latest [created, ready] revisions, and latest traffic

sixolet:
Before, the output was misleading because we:
* Didn't show the latest created revision if it wasn't getting traffic. The latest created revision is always relevant, and especially so if it's broken somehow.
* Didn't show what was traffic to a revision name, and what was traffic because it happened to be latest.
* Didn't show any information about revision statuses.
* Didn't show any information about traffic tags.
* Was liable to not add up to 100% if the same revision was in the traffic targets list twice, once because it was latest and once because it was tagged under its own right.

This change makes a number of small changes to the describe output to fix all that. It's best shown by example:

**Before**:
```
$ ./kn service describe bar
Name:         bar
Namespace:    default
URL:          http://bar.default.example.com
Address:      http://bar.default.svc.cluster.local
Labels:       cool=beans
Annotations:  client.knative.dev/user-image=gcr.io/knative-samples/helloworld-nodejs:latest, kubectl.kubernete ...                                                                                         
Age:          175d

Revisions:
     -  Name:   bar-vkqsk-18 [13] (3h)
        Image:  gcr.io/knative-samples/helloworld-nodejs@sha256:ae8c741379bbe004df62c1f29bbfd9bc92c84060a6d4af409d220e54f29959f6                                                                           
        Env:    ASDF=tame
        CPU:    400m
     -  Name:   bar-whee-11 [12] (3h)
        Image:  gcr.io/knative-samples/helloworld-nodejs:latest
        Env:    ASDF=wild
        CPU:    400m
   90%  Name:   bar-bmchl-10 [11] (3h)
        Image:  gcr.io/knative-samples/helloworld-nodejs:latest
        Env:    ASDF=yep
        CPU:    400m

Conditions:
  -- ConfigurationsReady     2h RevisionFailed
  -- Ready                   2h RevisionFailed
  ++ RoutesReady             2h
```

**After**:
```
$ ./kn service describe bar 
Name:       bar
Namespace:  default
URL:        http://bar.default.example.com
Address:    http://bar.default.svc.cluster.local
Labels:     cool=beans
Age:        217d

Revisions:
     !  bar-bywmr-32 (latest created) [21] (39d)
        Error:  ContainerMissing
        Image:  gcr.io/knative-samples/does-not-exist
   10%  @latest (bar-vkqsk-18) #latest [13] (42d)
        Image:  gcr.io/knative-samples/helloworld-nodejs:latest (pinned to ae8c74)
     +  bar-vkqsk-18 (current @latest) #weird [13] (42d)
        Image:  gcr.io/knative-samples/helloworld-nodejs:latest (pinned to ae8c74)
     +  bar-whee-11 #candidate [12] (42d)
        Image:  gcr.io/knative-samples/helloworld-nodejs:latest (at ae8c74)
   90%  bar-bmchl-10 #current [11] (42d)
        Image:  gcr.io/knative-samples/helloworld-nodejs:latest (at ae8c74)

Conditions:
  !! ConfigurationsReady    39d RevisionFailed
  !! Ready                  39d RevisionFailed
  ++ RoutesReady            42d
```

* Now shows latest created & latest ready revisions even if they are not routed.
* Shows `!` on revisions with errors
* Shows `?` on revisions with unknown status
* Condition readiness symbols changed to `++` (ok) `??` (unknown), `!!` (error)
* Limit per-revision output to only image and error unless the user requests it verbose.
* Distinguish between revisions routed *because* they're latest from revisions that are routed by revision name that *happen* to be latest.


sixolet:
Oh, also: non-verbose annotations were dominated by boilerplate, which this PR removes. Verbosely, all the annots are still present.

navidshaikh:
> Didn't show the latest created revision if it wasn't getting traffic. The latest created revision is always relevant, and especially so if it's broken somehow.

:+1: 

> Didn't show what was traffic to a revision name,

I think it did?
```
✗  kn service describe svc          
Name:         svc
Namespace:    default
URL:          http://svc.default.apps-crc.testing
Address:      http://svc.default.svc.cluster.local
Annotations:  serving.knative.dev/creator=kube:admin, serving.knative.dev/lastModifier=kube:admin
Age:          19h

Revisions:
     -  Name:    svc-qkqxx-3 [3] (3m)
        Image:   docker.io/swordphilic/go-helloworld
        Env:     TARGET=kn2
        Memory:  100M ... 200M
        CPU:     400m ... 1
   50%  Name:    svc-rlplg-2 [2] (4m)
        Image:   docker.io/swordphilic/go-helloworld
        Env:     TARGET=kn
        Memory:  100M ... 200M
        CPU:     400m ... 1
   50%  Name:    svc-pftkz-1 [1] (19h)
        Image:   docker.io/swordphilic/go-helloworld
        Memory:  100M ... 200M
        CPU:     400m ... 1

Conditions:
  ++ ConfigurationsReady     3m 
  ++ Ready                   1m 
  ++ RoutesReady             1m 
```

> Didn't show any information about traffic tags.

:+1: 
The tags are though isn't easy to locate, how about placing it in between traffic portion and revision name? (close to traffic portion)
Also, when given `--verbose` flag, we could add URL field per target (only if tag is present)
```
   10%  @latest at bar-vkqsk-18 #latest [13] (3h)
          Image:  gcr.io/knative-samples/helloworld-nodejs@sha256:ae8c741379bbe004df62c1f29bbfd9bc92c84060a6d4af409d220e54f29959f6
          Env:    ASDF=tame
          CPU:    400m
          URL:    latest-bar.domain.com
```

> Was liable to not add up to 100% if the same revision was in the traffic targets list twice, once because it was latest and once because it was tagged under its own right.

Are you suggesting that revision listing should be just once in this scenario, any example for this?

> Didn't show any information about revision statuses.

If  a revision is not ready and not receiving any traffic, would this show up in the default output ?

sixolet:
> > Didn't show what was traffic to a revision name,
> 
> I think it did?
To be clearer: It didn't differentiate traffic to a revision because it was latest from traffic to a revision because it was called out by name. The new form does this. 
> > Didn't show any information about traffic tags.
> 
> 👍
> The tags are though isn't easy to locate, how about placing it in between traffic portion and revision name? (close to traffic portion)

Yeah, I considered doing that. I liked the thing I implemented better, but only by a little. 

A couple options:
* foo-asdf with tag current
  - `10% foo-asdf #current`
  - `10% current: foo-asdf`
* latest at foo-qwer with tag new
  - `10% @latest at foo-qwer #new`
  - `10% new: @latest at foo-qwer`

> Also, when given `--verbose` flag, we could add URL field per target (only if tag is present)
Sure.

> > Was liable to not add up to 100% if the same revision was in the traffic targets list twice, once because it was latest and once because it was tagged under its own right.
> 
> Are you suggesting that revision listing should be just once in this scenario, any example for this?

Look at the example I provided. I'm suggesting listing the same revision more than once when necessary.
> 
> > Didn't show any information about revision statuses.
> 
> If a revision is not ready and not receiving any traffic, would this show up in the default output ?

Yes, if it's the latest created revision. That's important information to show.

navidshaikh:
> Yeah, I considered doing that. I liked the thing I implemented better, but only by a little.

>A couple options:

> - foo-asdf with tag current
>     - 10% foo-asdf #current
>     - 10% current: foo-asdf
> - latest at foo-qwer with tag new
>    - 10% @latest at foo-qwer #new
>    - 10% new: @latest at foo-qwer

I like # before tag names, helps it stand out as tag identifier among  other info present there (specially when user-specified revision-names could be human-readable )

 - foo-asdf with tag current
    - 10% #current foo-asd
 - latest at foo-qwer with tag new
   - 10% #new foo-qwer (@latest)

Do we really need `at` there when `@` is already there ? 
Also there will be only one @latest, so IMO it can be present after revision name.
Tags; with # (which will be case for almost all revisions receiving traffic) get their own column between traffic and revision-name.

toVersus:
The idea of adding hashtag before tag names looks fine! I think it's better to place tag name with hashtag after revision name in the same order when tagging revision name using `--tag` option. Also, it makes easy to compare the revision name with random characters in case of having revisions associated with more than one tag.

```
  Revisions:
     !  bar-jnbtg-19 [14] (2h)
          Image:  gcr.io/knative-samples/does-not-exist
     +  bar-vkqsk-18 #weird [13] (3h)
          Image:  gcr.io/knative-samples/helloworld-nodejs@sha256:ae8c741379bbe004df62c1f29bbfd9bc92c84060a6d4af409d220e54f29959f6
   10%  bar-vkqsk-18 (@latest) #latest [13] (3h)    <-- same revision as above
          Image:  gcr.io/knative-samples/helloworld-nodejs@sha256:ae8c741379bbe004df62c1f29bbfd9bc92c84060a6d4af409d220e54f29959f6
     +  bar-whee-11 #candidate [12] (3h)
          Image:  gcr.io/knative-samples/helloworld-nodejs:latest
   90%  bar-bmchl-10 #current [11] (3h)
          Image:  gcr.io/knative-samples/helloworld-nodejs:latest
```

sixolet:
@toVersus after experimenting with what you suggested it feels like `@latest` is better the first thing, because that is "why is the traffic getting routed here" — the exact revision it is is almost a sidenote, as soon as someone pushes another revision, the revision will change, but the traffic remains going to the latest.

@navidshaikh Yeah, I want tags to get their own column but that just looks bad when I try it. Boo.

pkg/serving/v1alpha1/client_mock.go
```diff
@@ -137,6 +137,7 @@ func (r *Recorder) GetRevision(name interface{}, revision *v1alpha1.Revision, er
 }
 
 func (c *MockKnClient) GetRevision(name string) (*v1alpha1.Revision, error) {
+	c.t.Log("GetRevision " + name)
```
toVersus: Debugging code remained?

pkg/kn/commands/service/service_describe.go
```diff
@@ -438,56 +479,75 @@ func age(t time.Time) string {
...
+func completeWithUntargetedRevisions(client serving_kn_v1alpha1.KnClient, service *v1alpha1.Service, revisionsSeen sets.String, descs []*revisionDesc) ([]*revisionDesc, error) {
 	revisions, err := client.ListRevisions(serving_kn_v1alpha1.WithService(service.Name))
 	if err != nil {
-		return err
+		return nil, err
 	}
 	for _, revision := range revisions.Items {
-		if _, ok := descs[revision.Name]; !ok {
-			descs[revision.Name], err = newRevisionDesc(&revision, nil)
+		if !revisionsSeen.Has(revision.Name) {
```
toVersus: I like this kind of early return style:

```golang
		if revisionsSeen.Has(revisionName) {
			continue
		}

		rev, err := client.GetRevision(revisionName)
		if err != nil {
			return nil, err
		}
		newDesc, err := newRevisionDesc(rev, nil)
		if err != nil {
			return nil, err
		}

		descs = append(descs, newDesc)
```

pkg/kn/commands/service/service_describe.go
```diff
@@ -438,56 +479,75 @@ func age(t time.Time) string {
...
-func orderByConfigurationGeneration(descs map[string]*revisionDesc) []*revisionDesc {
-	descsList := make([]*revisionDesc, len(descs))
-	idx := 0
-	for _, desc := range descs {
-		descsList[idx] = desc
-		idx++
-	}
-	sort.SliceStable(descsList, func(i, j int) bool {
-		return descsList[i].configurationGeneration > descsList[j].configurationGeneration
+func orderByConfigurationGeneration(descs []*revisionDesc) []*revisionDesc {
```
toVersus: 👍

pkg/kn/commands/service/service_describe.go
```diff
@@ -421,11 +451,22 @@ func joinAndTruncate(sortedKeys []string, m map[string]string) string {
...
+			symbol = "%"
+		}
+	case v1.ConditionFalse:
+		symbol = "!"
+	default:
+		symbol = "?"
+	}
 	if percentage == 0 {
-		return "   -"
+		return fmt.Sprintf("   %s", symbol)
```
toVersus:
```suggestion
		return "   " + symbol
```

pkg/kn/commands/service/service_describe_test.go
```diff
@@ -74,6 +74,121 @@ func TestServiceDescribeBasic(t *testing.T) {
...
+func TestServiceDescribeLatest(t *testing.T) {
+
+	// New mock client
+	client := knclient.NewMockKnClient(t)
+
+	// Recording:
+	r := client.Recorder()
+	// Prepare service
+	expectedService := createTestService("foo", []string{"rev1"}, goodConditions())
+	expectedService.Status.Traffic[0].LatestRevision = new(bool)
```
toVersus: Can we use "knative.dev/pkg/ptr" for doing this?

```golang
	expectedService.Status.Traffic[0].LatestRevision = ptr.Bool(true)
```

pkg/kn/commands/service/service_describe_test.go
```diff
@@ -76,6 +77,112 @@ func TestServiceDescribeBasic(t *testing.T) {
...
+	// New mock client
+	client := knclient.NewMockKnClient(t)
+
+	// Recording:
+	r := client.Recorder()
+	// Prepare service
+	expectedService := createTestService("foo", []string{"rev1", "rev1"}, goodConditions())
+	expectedService.Status.Traffic[0].LatestRevision = new(bool)
+	expectedService.Status.Traffic[0].Tag = "latest"
+	*expectedService.Status.Traffic[0].LatestRevision = true
```
toVersus:
```suggestion
	expectedService.Status.Traffic[0].LatestRevision = ptr.Bool(true)
```

pkg/kn/commands/service/service_describe_test.go
```diff
@@ -76,6 +77,112 @@ func TestServiceDescribeBasic(t *testing.T) {
...
+func TestServiceDescribeLatestCreatedIsBroken(t *testing.T) {
+	// New mock client
+	client := knclient.NewMockKnClient(t)
+
+	// Recording:
+	r := client.Recorder()
+	// Prepare service
+	expectedService := createTestService("foo", []string{"rev1"}, goodConditions())
+	expectedService.Status.Traffic[0].LatestRevision = new(bool)
+	*expectedService.Status.Traffic[0].LatestRevision = true
```
toVersus:
```suggestion
	expectedService.Status.Traffic[0].LatestRevision = ptr.Bool(true)
```

sixolet:
Should be fixed now.

navidshaikh:
> Didn't show the latest created revision if it wasn't getting traffic. The latest created revision is always relevant, and especially so if it's broken somehow.

I think we want to print `@latest` identifier for the latest ready revision in describe default output format in each of the following combinations (if its not broken as well)

  1. @latest has tag and traffic                                    :heavy_check_mark: 
  2. @latest has tag and no traffic                               :heavy_check_mark: 
  3. @latest has no tag but traffic                                :heavy_check_mark: 
  4. @latest has neither tag nor traffic                         :heavy_multiplication_x: 

Case 4: Latest ready revision neither has tag nor traffic, `@latest` identifier is **not** shown in output
```
12:42 ➜  client git:(improve-describe)  ./kn service describe hello                       
  Name:       hello
  Namespace:  default
  URL:        http://hello.default.apps-crc.testing
  Address:    http://hello.default.svc.cluster.local
  Age:        16h

  Revisions:
     +  hello-v4 [4] (14s)
          Image:   index.docker.io/swordphilic/go-helloworld@sha256:4069f1b68c02d0ad3930259811538ecbadbadfe102244bec803dcb4616f1218f (at 4069f1b68c02)
          Env:     TARGET=kn
          Memory:  100M ... 200M
          CPU:     400m ... 1
  100%  hello-fpfws-1 [1] (16h)
          Image:   docker.io/swordphilic/go-helloworld (at 4069f1b68c02)
          Memory:  100M ... 200M
          CPU:     400m ... 1

  Conditions:
  ++ ConfigurationsReady     5s 
  ++ Ready                   5s 
  ++ RoutesReady            16h 
```

sixolet:
Ok. Added something to mark the latest ready and latest created revisions, but not `@latest` as the headline unless that's why they have a traffic percentage associated with them — I want to keep the revision as the headline except in the case where there's traffic or a URL that's only going there because of the latest-ness of the revision, and it'll be a different revision soon.

The first bit of the revision description is *what is being described by the traffic to the left*, and so that's only `@latest` when we're corresponding to a traffic target that'll send to a different revision if a new revision is pushed.

rhuss:
The good thing of being away for three weeks is that I can have an initial and innocent view on the new output. So here's my initial comment without digging into the PR comments first:

I have a sample output like:

``` 
  Name:         random-generator
  Namespace:    default
  URL:          http://random-generator.default.apps.rhuss-dev.devcluster.openshift.com
  Address:      http://random-generator.default.svc.cluster.local
  Annotations:  serving.knative.dev/creator=kube:admin
                serving.knative.dev/lastModifier=kube:admin
  Age:          46m

  Revisions:
  100%  @latest (random-generator-zsksx-2) [2] (9m)
          Image:  k8spatterns/random-generator:2.0 (at 5ceb6f182f36)
          Scale:  2 ... 10
          Concurrency:
                     Limit:   10
                     Target:  5
          Memory:  100M ... 200M
          CPU:     400m ... 1
     +  random-generator-qgwlj-1 [1] (46m)
          Image:  k8spatterns/random-generator:1.0 (at ca2bb6b69bf3)
          Scale:  2 ... 10
          Concurrency:
                     Limit:   10
                     Target:  5
          Memory:  100M ... 200M
          CPU:     400m ... 1

  Conditions:
  ++ ConfigurationsReady     8m
  ++ Ready                   8m
  ++ RoutesReady             8m
```

My first questions where: 

* What does the `+` mean in front of the revision name mean (used to be a `-` to indicate that there is no traffic routed, so this column was reserved purely for the traffic split portions) ? What would a `!` mean ?
* Why is the revision name once in parens and once without ? And why it is once in the fist position and once in the second ? I think we should keep the position of the revision name stable.
* Why is the layout messed up ? (see the concurrency indention and I also don't get why the "Fields" are not aligned anymore with the first line (e.g. "Image" does not start below "@latest" but with 2 spaced indented.
* Why is everything indented by 2 spaces, wasting extra screen real estate ?

When I do a `kn service update --tag random-generator-zsksx-2=v2` I get 

```
  Name:         random-generator
  Namespace:    default
  URL:          http://random-generator.default.apps.rhuss-dev.devcluster.openshift.com
  Address:      http://random-generator.default.svc.cluster.local
  Annotations:  serving.knative.dev/creator=kube:admin
                serving.knative.dev/lastModifier=kube:admin
  Age:          57m

  Revisions:
  100%  @latest (random-generator-zsksx-2) [2] (20m)
          Image:  k8spatterns/random-generator:2.0 (at 5ceb6f182f36)
          Scale:  2 ... 10
          Concurrency:
                     Limit:   10
                     Target:  5
          Memory:  100M ... 200M
          CPU:     400m ... 1
     +  random-generator-zsksx-2 (current @latest) #v2 [2] (20m)
          Image:  k8spatterns/random-generator:2.0 (at 5ceb6f182f36)
          Scale:  2 ... 10
          Concurrency:
                     Limit:   10
                     Target:  5
          Memory:  100M ... 200M
          CPU:     400m ... 1
     +  random-generator-qgwlj-1 [1] (57m)
          Image:  k8spatterns/random-generator:1.0 (at ca2bb6b69bf3)
          Scale:  2 ... 10
          Concurrency:
                     Limit:   10
                     Target:  5
          Memory:  100M ... 200M
          CPU:     400m ... 1

  Conditions:
  ++ ConfigurationsReady    19m
  ++ Ready                  50s
  ++ RoutesReady            50s
```

* Why is "random-generator-zsksx-2" listed twice now ?
* What does this `(current @latest)` mean ? What is the difference of this line s. having @latest righter after the traffic split portion ?

When doing then a `kn service update random-generator --traffic v1=50 --traffic v2=50` (after tagging v1, too):

```
  Name:         random-generator
  Namespace:    default
  URL:          http://random-generator.default.apps.rhuss-dev.devcluster.openshift.com
  Address:      http://random-generator.default.svc.cluster.local
  Annotations:  serving.knative.dev/creator=kube:admin
                serving.knative.dev/lastModifier=kube:admin
  Age:          1h

  Revisions:
   50%  random-generator-zsksx-2 (current @latest) #v2 [2] (30m)
          Image:  k8spatterns/random-generator:2.0 (at 5ceb6f182f36)
          Scale:  2 ... 10
          Concurrency:
                     Limit:   10
                     Target:  5
          Memory:  100M ... 200M
          CPU:     400m ... 1
   50%  random-generator-qgwlj-1 #v1 [1] (1h)
          Image:  k8spatterns/random-generator:1.0 (at ca2bb6b69bf3)
          Scale:  2 ... 10
          Concurrency:
                     Limit:   10
                     Target:  5
          Memory:  100M ... 200M
          CPU:     400m ... 1

  Conditions:
  ++ ConfigurationsReady    29m
  ++ Ready                  12s
  ++ RoutesReady            12s
```

* Why is there now no `@latest` entry anymore and only 2 entries and not 3 as before ?

```
  Name:       random-generator
  Namespace:  default
  URL:        http://random-generator.default.apps.rhuss-dev.devcluster.openshift.com
  Address:    http://random-generator.default.svc.cluster.local
  Age:        1h

  Revisions:
     ?  random-generator-fdcys-7 (latest created) [4] (1m)
          Image:  k8spatterns/random-generator:2.0 (at 5ceb6f182f36)
          Scale:  2 ... 10
          Concurrency:
                     Limit:   10
                     Target:  5
          Memory:  100M ... 200M
          CPU:     400m ... 1
     +  random-generator-lnyhd-6 (current @latest) [3] (3m)
          Image:  k8spatterns/random-generator:1.0 (at ca2bb6b69bf3)
          Scale:  2 ... 10
          Concurrency:
                     Limit:   10
                     Target:  5
          Memory:  100M ... 200M
          CPU:     400m ... 1
   50%  random-generator-zsksx-2 #v2 [2] (36m)
          Image:  k8spatterns/random-generator:2.0 (at 5ceb6f182f36)
          Scale:  2 ... 10
          Concurrency:
                     Limit:   10
                     Target:  5
          Memory:  100M ... 200M
          CPU:     400m ... 1
   50%  random-generator-qgwlj-1 #v1 [1] (1h)
          Image:  k8spatterns/random-generator:1.0 (at ca2bb6b69bf3)
          Scale:  2 ... 10
          Concurrency:
                     Limit:   10
                     Target:  5
          Memory:  100M ... 200M
          CPU:     400m ... 1
```

* What does the `?` mean ? If it indicates that the latest revision is "Unknown" isn't that redundant with the Service's "ConfigurationReady" status ?
* Is "(latest created)" really needed as we already have a temporal order and also the time since the revision is created printed ? It is easily to detect that its always the revision with the highest generation, printed on the top with the lowest creation time is the "latest created" revision.

rhuss:
BTW, what happens when an revision which has targeted traffic goes into an error / unknown state ? How is this indicated here ? 

Its a very good idea to have the revision status in `service describe`, I just wouldn't show it in union with the traffic distribution.

rhuss:
Another suggestion (but we could tackle that in a followup PR):

Instead of 

```
k8spatterns/random-generator:1.0 (at ca2bb6b69bf3)
```

with a redundant `(at ...)` that dosn't add extra value, we could use

```
k8spatterns/random-generator:1.0@ca2bb6
```

for normal display and the full shasum with `-v`:

```
k8spatterns/random-generator:1.0@sha256:ca2bb6b69bf3ef123456799effaaae
```

navidshaikh:
```
15:17 ➜  client git:(a67e2f3)  ./kn service create svc --image gcr.io/knative-samples/helloworld-go
Service 'svc' successfully created in namespace 'default'.

Service URL:
http://svc.default.apps-crc.testing

15:17 ➜  client git:(a67e2f3)  ./kn service update svc --env TARGET=kn

15:18 ➜  client git:(a67e2f3)  ./kn revision list -s svc
NAME          SERVICE   GENERATION   AGE   CONDITIONS   READY   REASON
svc-zrcgp-2   svc       2            15s   4 OK / 4     True    
svc-xsnlf-1   svc       1            48s   4 OK / 4     True    

15:18 ➜  client git:(a67e2f3)  ./kn service update svc --traffic=svc-xsnlf-1=100

15:18 ➜  client git:(a67e2f3)  ./kn service describe svc
  Name:       svc
  Namespace:  default
  URL:        http://svc.default.apps-crc.testing
  Address:    http://svc.default.svc.cluster.local
  Age:        1m

  Revisions:
     +  svc-zrcgp-2 (current @latest) [2] (37s)
          Image:   gcr.io/knative-samples/helloworld-go (pinned to 5ea96ba4b872)
          Env:     TARGET=kn
          Memory:  100M ... 200M
          CPU:     400m ... 1
  100%  svc-xsnlf-1 [1] (1m)
          Image:   gcr.io/knative-samples/helloworld-go (at 5ea96ba4b872)
          Memory:  100M ... 200M
          CPU:     400m ... 1

  Conditions:
  ++ ConfigurationsReady    26s 
  ++ Ready                   5s 
  ++ RoutesReady             5s 

15:18 ➜  client git:(a67e2f3)  ./kn service update svc --image does.not/exist    
Waiting for service 'svc' to become ready ... 
RevisionFailed: Revision "svc-kbgbj-4" failed with message: Unable to fetch image "does.not/exist": Get https://does.not/v2/: dial tcp: lookup does.not on 172.30.0.10:53: no such host.

15:20 ➜  client git:(a67e2f3)  ./kn service describe svc                     
  Name:       svc
  Namespace:  default
  URL:        http://svc.default.apps-crc.testing
  Address:    http://svc.default.svc.cluster.local
  Age:        2m

  Revisions:
     !  svc-kbgbj-4 (latest created) [3] (6s)
          Error:   ContainerMissing
          Image:   does.not/exist
          Env:     TARGET=kn
          Memory:  100M ... 200M
          CPU:     400m ... 1
     +  svc-zrcgp-2 (current @latest) [2] (2m)
          Image:   gcr.io/knative-samples/helloworld-go (pinned to 5ea96ba4b872)
          Env:     TARGET=kn
          Memory:  100M ... 200M
          CPU:     400m ... 1
  100%  svc-xsnlf-1 [1] (2m)
          Image:   gcr.io/knative-samples/helloworld-go (at 5ea96ba4b872)
          Memory:  100M ... 200M
          CPU:     400m ... 1

  Conditions:
  !! ConfigurationsReady     1s RevisionFailed
  !! Ready                   1s RevisionFailed
  ++ RoutesReady             1m 
```

I see the new changes representing and distinguishing the latest ready/non-ready revision by
```
current @latest
```
and
```
latest created
```
We are re-presenting this distinction in other ways too, like `+`/ `!` and the revision generation number `[4]` or `[3]`. I wonder if this is making it more explicit or might create confusion with tags of revision, although they are prepended by `#`.


Some difference in how git-sha is represented for different revisions:
```
      Image:   gcr.io/knative-samples/helloworld-go (pinned to 5ea96ba4b872)
```
and
```
          Image:   gcr.io/knative-samples/helloworld-go (at 5ea96ba4b872)
```

i.e. `pinned to ` and `at`.  Is there some re-presentation meaning that I am missing here ?

If `pinned to` representing here the last time user requested update with `--image` flag then a subsequent update with `--image` flag showed `at` for me

```
 ./kn service describe svc                                           
  Name:       svc
  Namespace:  default
  URL:        http://svc.default.apps-crc.testing
  Address:    http://svc.default.svc.cluster.local
  Age:        4m

  Revisions:
     +  svc-cswmp-5 (current @latest) [4] (40s)
          Image:   gcr.io/knative-samples/helloworld-go (at 5ea96ba4b872)
          Env:     TARGET=kn
          Memory:  100M ... 200M
          CPU:     400m ... 1
  100%  svc-lmkbw-1 [1] (4m)
          Image:   gcr.io/knative-samples/helloworld-go (at 5ea96ba4b872)
          Memory:  100M ... 200M
          CPU:     400m ... 1

  Conditions:
  ++ ConfigurationsReady    27s 
  ++ Ready                  27s 
  ++ RoutesReady             2m 
```

I went ahead and updated the service with different image altogether, I don't see `pinned to` for the latest ready revision
```
17:14 ➜  client git:(a67e2f3)  ./kn service describe svc          
  Name:       svc
  Namespace:  default
  URL:        http://svc.default.apps-crc.testing
  Address:    http://svc.default.svc.cluster.local
  Age:        12m

  Revisions:
     +  svc-zlrlm-6 (current @latest) [5] (1m)
          Image:   docker.io/swordphilic/go-helloworld (at 4069f1b68c02)
          Env:     TARGET=kn
          Memory:  100M ... 200M
          CPU:     400m ... 1
  100%  svc-lmkbw-1 [1] (12m)
          Image:   gcr.io/knative-samples/helloworld-go (at 5ea96ba4b872)
          Memory:  100M ... 200M
          CPU:     400m ... 1

  Conditions:
  ++ ConfigurationsReady     1m 
  ++ Ready                   1m 
  ++ RoutesReady            10m 
```

Summarizing, two questions
1. Do we need `current @latest` or `latest created` strings for re-presentation and distinction while we are re-presenting this distinction via other means `!` and `[$generation]`
2. When `pinned to` is shown infront of image ?

rhuss:
@sixolet sorry for the nit-picky review, but I really think we should stay consistent (e.g not indent by 2 spaces every line) and with a clean layout. Happy to help to work on this PR, too, and make a suggestion via code (if this would be ok for you).

sixolet:
> * What does the `+` mean in front of the revision name mean (used to be a `-` to indicate that there is no traffic routed, so this column was reserved purely for the traffic split portions) ? What would a `!` mean ?

A `-` to indicate no traffic routed was confusing when a `-` below meant that the status wasn't ready. In my formulation, 
`+` — ready but no %age traffic routed
`%` — ready and %age traffic routed
`!` — ready-false
`?` — ready-unknown

> * Why is the revision name once in parens and once without ? And why it is once in the fist position and once in the second ? I think we should keep the position of the revision name stable.

There are two ways traffic can be directed to a revision: because it's latest, and because it's named. 100% of traffic is primarily to `@latest`, which happens to be a particular revision. The other revision is listed qua itself. I actually don't know why it's displayed — it looks like it's neither the latest created or latest ready revision. Maybe the server left a 0% line for it in the traffic targets in the status.

Position stability is a good point. They should be stable. I'll check the code.

> * Why is the layout messed up ? (see the concurrency indention and I also don't get why the "Fields" are not aligned anymore with the first line (e.g. "Image" does not start below "@latest" but with 2 spaced indented.

It looks better when fields are indented from the first line. You can see the grouping better that way.

> * Why is everything indented by 2 spaces, wasting extra screen real estate

See above: because I thought it looked better. I admit I use hella columns in my terminals.
> 
> When I do a `kn service update --tag random-generator-zsksx-2=v2` I get
...
> 
> * Why is "random-generator-zsksx-2" listed twice now ?

Because you've tagged it by revision, but traffic to it is because it's `@latest`. So it deserves two lines.

> * What does this `(current @latest)` mean ? What is the difference of this line s. having @latest righter after the traffic split portion ?

The parenthetical is extra information, telling you what revision-by-name happens to be latest. `@latest` as the title tells you that traffic is to that revision because it's latest, and puts the particular revision in the parenthetical. 
> 
> When doing then a `kn service update random-generator --traffic v1=50 --traffic v2=50` (after tagging v1, too):
...
> 
> * Why is there now no `@latest` entry anymore and only 2 entries and not 3 as before ?

Because there's no traffic to `@latest`. You still get parenthetically told what the latest revision is.

> ...
> * What does the `?` mean ? If it indicates that the latest revision is "Unknown" isn't that redundant with the Service's "ConfigurationReady" status ?

It's unknown. Mildly redundant. It's ok.

> * Is "(latest created)" really needed as we already have a temporal order and also the time since the revision is created printed ? It is easily to detect that its always the revision with the highest generation, printed on the top with the lowest creation time is the "latest created" revision.

Is it really needed? No. But personally I like being told.

Let's chat about this, maybe in a small breakout meeting.

sixolet:
> @sixolet sorry for the nit-picky review, but I really think we should stay consistent (e.g not indent by 2 spaces every line) and with a clean layout. Happy to help to work on this PR, too, and make a suggestion via code (if this would be ok for you).

We may end up just disagreeing about what "clean layout" looks like. I tried lots of different things trying to find the one that looked best to me.

sixolet:
> 1. Do we need `current @latest` or `latest created` strings for re-presentation and distinction while we are re-presenting this distinction via other means `!` and `[$generation]`
You're right about [generation] but I like the explicitness.

The `!` means the revision is in an error state, which is why you're likely to see it when latest ready != latest created.

> 2. When `pinned to` is shown infront of image ?

It should be "when the `user-image` annotation is present. I may have gotten it slightly wrong and done "when it's present and differs from the actual contents of image". I can fix that, either in this PR or a different one.

rhuss:
Ok, I still found the output a bit confusing (in addition that I do not really agree with the ragged layout, but that's not so important for the moment)

So maybe let's go a step back and nail down the use cases which we want to solve with the traffic list, which I believe can be subsumed to

As a user I want ...

* ... set the current traffic distribution along with their revisions
* ... see whether traffic targets (== one entry in the traffic list) have the flag `LatestRevision` set and also what this latest revision is currently.
* ... uniform entries for every TrafficTarget item so that I can compare entries easily and can easily navigate through the list by skimming over it.
* ... conditionally (--verbose) also see other Revisions which are not routed (but by default only routed revisions should be shown)
* ... quickly see which revision is the latest revision
* ... see the relevant information for each TrafficTarget like tag (if present), revision pointing to it and whether this is a dynamic entry for the latest revision.
* ... for each traffic target shown, see a summary of the revision like image, status, and other information (more with --verbose)

Is this a fair summary of the requirements or is there something wrong or missing?

If so, then I think we should focus on the traffic target list and present the items found there, indicating when it's a "LatestRevision" marked target (and resolve the latest revision at that point).

In addition, we can have a dedicated entry for the latest revision outside of this list (it's the answer to "what is the latest revisions" and is not necessarily connected to the traffic distribution). This will be shown independently of whether the latest revision is part of the traffic list or not (alternatively we could omit this separate special entry when latest is part of the traffic list anyway which would then mark. But this could be confusing, too)

Happy to discuss this of course (online or in a call)

Let me try to make up the examples I've shown with this suggestion in the following comments.

sixolet:
I think if what you're seeing is revisions, then the latest created and latest ready are always important, along with your routed revisions. Verbose or not. If what you're seeing is traffic, then traffic percent and tags is what's important. This view combines them. Not combining them is a valid option, but then the user has to go and do the join in their head.

rhuss:
> I think if what you're seeing is revisions, then the latest created and latest ready are always important, along with your routed revisions. Verbose or not. If what you're seeing is traffic, then traffic percent and tags is what's important. This view combines them. Not combining them is a valid option, but then the user has to go and do the join in their head.

I agree that the latest ready version is an important use case, so I'm totally fine that this is shown all the time. Not so sure about the `latest created`. IMO the default traffic list should only show the current traffic targets (reflecting the status quo), with --verbose showing the full list.

However, there is also the use with no `TrafficTarget.latestRevision == true`, so traffic distribution explicitly managed and pinned solely by the user. In that use case, the latest revision is not important.

I'm okish with having latest as part of the traffic list (undconditionally maybe, even when the user doesn't really use that "latest" feature, but I really would love to have

* a uniform list with entries looking the same, maybe with some optional fields
* separating the traffic percentage from the revision status by e.g. introducing an extra column for the status (e.g. I think this PRs semantics doesn't cover not-ready revision which a traffic targeted, i.e. the traffic percentage wouldn't show up, so don't adding up to 100%)
* all traffic percentages shown add up always to 100% 
* no unnecessary repetition of revisions


We should also decide whether to focus on the traffic split (and then call that section "Traffic:") or keep the "Revision:" section and intersperse the traffic information.

Or what's about that suggestion to have both sections like in

```
Traffic:
    5%  +@  [4] svc-cswmp-5 #v2
   95%  +   [1] svc-lmkbw-1 #v1

Revisions:
! [5] svc-cswmp-5 #v2 (30m)
      Image:  k8spatterns/random-generator:2.0@abcded
      Scale:  2 ... 10
      Concurrency:
        Limit:   10
        Target:  5
      Memory:  100M ... 200M
      CPU:     400m ... 1
+ [4] svc-cswmp-5 #v2 [4] (30m)
      Image:  k8spatterns/random-generator:2.0@5ceb6f
      Scale:  2 ... 10
        Concurrency:
        Limit:   10
        Target:  5
      Memory:  100M ... 200M
      CPU:     400m ... 1
+ [1] svc-lmkbw-1 [1] #v1, #anothertag (1h)
      Image:  k8spatterns/random-generator:1.0@ca2bb6
      Scale:  2 ... 10
      Concurrency:
        Limit:   10
        Target:  5
      Memory:  100M ... 200M
      CPU:     400m ... 1
```

This would have the benefit to see immediately the traffic split (if this is the question you ask), but have a full list of revisions (without --verbose only routed revisions will be shown, with --verbose all revisions ordered after creation).

Also, the latest created revision always is the first entry and the latest ready revision is the first in the list with a 'ready' status.

The traffic list that would show all entries from `Service.traffic` and indicate whether it's routed to the latest (via @ or another indicator in the status line).

This is just a first made up, the important point is to make a short concise single-line entry traffic lists without large, separated entries (which might even grow).

sixolet:
I think we need to meet. How's tomorrow morning?

rhuss:
Thanks a lot of for the help- and fruitful discussion on the WG call.

Let me summarize my remaining concerns:

* Please skip the two space at the beginning of every line as this not used anywhere else in `kn` and just wastes screen real estate.
* Reducing the list of revisions to a minimum (with a bit more maybe on `--verbose`). E.g also not sure if the environment variables should be shown at all as this can easily lead to to 10-20 more lines per revision entry
* If we can trim down each revision entry to two lines, I would remove the label alltogether and align the entries again. Repeating labels or fixed texts like "(at ...)" in lists also doesn't make good use of screen space (such thinks should go int column headers in list).
* Instead of "(at ..)" or "(pinned ..)" I suggest to use the (shortened) diggest directly on the image name for pinned images and the current digest in parentheses after the image name:
  - `myimage:1.2 (abce345)` for a non pinned image
  - `myimage:1.2@abc345 (abc345)` for a pinned image
  - `myimage@abc345 (abc345)` for a pinned image without tag

  With `--verbose` we could replace the parenthesed, shortened digest with the full sha256 sum.

-------------
Some additional suggestions (which we might consider to discuss later, after this PR has been landed would be):

* To show in a separate section the content of the `Configuration` (== latest created revision) with more details. This would help in removing the distinction between "latest created" and "latest routed" revision in the revision list, which I find confusing.

* And still I believe that status and percentage should be separated in the display. This would allow us to present both kind of splits, the configured and the actual one (if they differ). Eg. for an error situation

```
     (90%) @ !!  [5] mysvc-rev-2sds ...
                     myimage:v2 ....        
100% (10%)   ++  [4] @mysvc-rev-1axd ....
                     myimage:v1 ...
```

which would look like this in the "good" situation

```
 90% ++ @ [5] mysvc-rev-2sds ...
              myimage:v2 ....        
 10% ++   [4] mysvc-rev-1axd ....
              myimage:v1 ...
```

(i.e. showing both splits only when they differ). `@` is the marker for that this traffic entry is marked with `latestRevision` (there can be multiple of such entries in the traffic list). If splitting is as configured we could even leave out the status symbols (++) as this only happens when everything of the split items is up and running.

_The main idea here is, that the list presented directly reflects 1:1 the configured traffic split (including an indicator for 'revisionLatest') and show the latest created revision separately (only one entry)_

This mean directly mirroring the traffic split list, merged with the actual traffic split. E.g. if there is no split and everything goes to `latestRevision` we could have outpus like:

* Still deploying, not swiched over

```
    (100%) @ ??  [5] mysvc-rev-2sds ...
                     myimage:v2 ....        
100%         ++  [4] @mysvc-rev-1axd ....
                     myimage:v1 ...
```

* Ready deployed

which eventually would look like

```
100% @ ++ [5] mysvc-rev-2sds ...
              myimage:v2 ....        
```

rhuss:
@sixolet thanks a lot for the update, really like it :) There is one gotcha that it always shows "(pinned ...)" although I only used plain tags (so would expect a "(at ...)".

sixolet:
Currently works as designed — the `pinned` vs `at` choice is based on whether your user-image annotation is set, aka "if you deploy again without changing the image, will it re-pull?". `pinned` if it will not re-pull, `at` if it will.

pkg/kn/commands/service/describe_test.go
```diff
@@ -63,14 +64,177 @@ func TestServiceDescribeBasic(t *testing.T) {
...
+	output, err := executeServiceCommand(client, "describe", "foo")
+	assert.NilError(t, err)
+	validateServiceOutput(t, "foo", output)
+	assert.Assert(t, util.ContainsAll(output, "!!", "Ready"))
+
+	r.Validate()
+}
+
+func TestServiceDescribeLatest(t *testing.T) {
+
```
maximilien: nit... extra line. But can be improved after

Issue #358, opened 2019-08-09
Refactor service unit tests to use KnClient mock framework


rhuss:
## Describe the feature:

In order to simplify the maintenance of our tests, existing tests in services should be refactored to use the `MockKnClient` support class. An example of such a test can be found in [`service_create_mock_test.go`](https://github.com/knative/client/blob/master/pkg/kn/commands/service/service_create_mock_test.go)

Ideally all the test classes

* service_create_test.go
* service_list_test.go
* service_describe_test.go
* service_update_test.go

should be rewritten (and the mock sample integrated into service_create_test)

Maybe also add as service_delete_test.go which seems to be missing.

itsmurugappan:
can I help with this ?

rhuss:
@itsmurugappan yes, sure ! Any help is appreciated. Just pick an existing test and create a test in parallel (with a `_mock` in the filename, like in `servuce_update_mock_test`) and port the existing tests over. We will happily review the PR and if approved, we can remove the original test in an additional pull request.

Thanks !

itsmurugappan:
Sure. I will start with it. Thank you!

Issue #30, opened 2019-03-02
Gitops mode


sixolet:
Let's have a mode where you can provide a directory instead of a server to connect to. Instead of modifying objects on the server, `kn` will modify yaml files in the directory instead, for the `kubectl apply -f` pleasure of your friendly local CI/CD system.

sixolet:
Considerations:
* Downloading the right objects from etcd to a directory for the first time may another gitops usecase.
* Anything that reads status and uses that information becomes harder or impossible (ex, locking revisions to a container digest)
* Keeping only the "interesting" fields might make the directory easier to use.

cppforlife:
> Let's have a mode where you can provide a directory instead of a server to connect to

im somewhat skeptical that this would be that useful if users are using gitops workflow (== more familiar with yamls). what commands do you think will benefit from this?

nimakaviani:
throwing this here since it was discussed during today's call:

knative yaml generator
https://github.com/julz/knightrider

itsmurugappan:
@rhuss @navidshaikh i think this is the issue we discussed in the last wg. For implementing this am planning to take the below approach, please let me know if this is good

- Implement knclientinterface to work on local directory
- change this line in all commands to return the appropriate client based on cmd flags
```client, err := p.NewServingClient(namespace)```

**Comment by rhuss** [comment]
@itsmurugappan sorry for the late reply, but this sounds like a good strategy. I don't think you have to implement every method, only those that are needed for this functionality. Also the file name should then be part of that new implementation's struct. 

Also I learned this trick today (thanks to @markusthoemmes ):

> If you have a big interface to implement but only call a few functions, you can define the struct like this:
```
type mock struct {
  my.Interface
}
```
It'll implement all functions with nil-pointer function calls. Then you only have to implement the functions your test actually calls

rhuss:
(sorry closed by accident)

PR #389, opened 2019-08-25
feature(service): Register list of EnvVars in the same order as user input

toVersus:
## Proposed Changes

In [current logic](https://github.com/knative/client/blob/699ac5e7cbcfc994c42a3c717b5c5d700e03579c/pkg/serving/config_changes.go#L32) of handling user input of Environment variables, list of EnvVars are registered in random order because we convert user input string to Golang map and append them into slice of corev1.EnvVars, which has no guarantee of stable order as follows:

```yaml
$ ./kn service create hello --image="gcr.io/knative-samples/helloworld-go" -e FOO=foo -e BAR=bar -e BAZ=baz -e QUX=qux -e EMPTY=
Service 'hello' successfully created in namespace 'default'.
Waiting for service 'hello' to become ready ... OK

Service URL:
http://hello-default.example.com

$ ./kn service describe hello -o yaml
(...)
    spec:
      containers:
      - env:
        - name: BAZ
          value: baz
        - name: QUX
          value: qux
        - name: EMPTY
        - name: FOO
          value: foo
        - name: BAR
          value: bar
```

We need to fix this for passing mock test without flakiness as well as GitOps deployment without causing any bothering diffs.

* Register list of EnvVars in the same order as user input when creating new kservice
* Keep the original order of EnvVars when updating existing kservice
* Add mock test for create/update service with EnvVars

After:

```yaml
$ ./kn service create hello --image="gcr.io/knative-samples/helloworld-go" -e FOO=foo -e BAR=bar -e BAZ=baz -e QUX=qux -e EMPTY=
Service 'hello' successfully created in namespace 'default'.
Waiting for service 'hello' to become ready ... OK

Service URL:
http://hello-default.example.com

$ ./kn service describe hello -o yaml
(...)
    spec:
      containers:
      - env:
        - name: FOO
          value: foo
        - name: BAR
          value: bar
        - name: BAZ
          value: baz
        - name: QUX
          value: qux
        - name: EMPTY
(...)

$ ./kn service update hello -e FOO=foov2 -e BAR-
Waiting for service 'hello' to become ready ... OK
Service 'hello' updated in namespace 'default'.

$ ./kn service describe hello -o yaml
(...)
    spec:
      containers:
      - env:
        - name: FOO
          value: foov2
        - name: BAZ
          value: baz
        - name: QUX
          value: qux
        - name: EMPTY
(...)
```

Part of: #358
/lint


navidshaikh:
> We need to fix this for passing mock test without flakiness 

Is there a flakiness in tests due to order of env?

> GitOps deployment without causing any bothering diffs.

Interesting, are resource template files stored using `-o yaml` ?

toVersus:
> Is there a flakiness in tests due to order of env?

Sorry, currently not in tests. However, when I try to add [TestServiceCreateEnvMock](https://github.com/knative/client/blob/f5237d32b61584c0e66816f58dcfc40f4b42a6d2/pkg/kn/commands/service/service_create_mock_test.go#L59) mock test, it would emerge due to this [assert.DeepEqual](https://github.com/knative/client/blob/f5237d32b61584c0e66816f58dcfc40f4b42a6d2/pkg/serving/v1alpha1/client_mock.go#L292) as follows:

```yaml
=== RUN   TestServiceCreateEnvMock
--- FAIL: TestServiceCreateEnvMock (0.00s)
    client_mock.go:292: assertion failed: 
        --- call.args[i]
        +++ arg
            &v1alpha1.Service{
            TypeMeta:   v1.TypeMeta{},
          	ObjectMeta: v1.ObjectMeta{Name: "foo", Namespace: "default"},
          	Spec: v1alpha1.ServiceSpec{
          		... // 3 identical fields
          		DeprecatedManual:  nil,
          		DeprecatedRelease: nil,
          		ConfigurationSpec: v1alpha1.ConfigurationSpec{
          			DeprecatedGeneration:       0,
          			DeprecatedBuild:            nil,
          			DeprecatedRevisionTemplate: nil,
          			Template: &v1alpha1.RevisionTemplateSpec{
          				ObjectMeta: v1.ObjectMeta{},
          				Spec: v1alpha1.RevisionSpec{
          					RevisionSpec: v1beta1.RevisionSpec{
          						PodSpec: v1.PodSpec{
          							Volumes:        nil,
          							InitContainers: nil,
          							Containers: []v1.Container{
          								{
          									... // 5 identical fields
          									Ports:   nil,
          									EnvFrom: nil,
          									Env: []v1.EnvVar{
        - 										{Name: "a", Value: "mouse"},
          										{Name: "b", Value: "cookie"},
          										{Name: "empty"},
        + 										{Name: "a", Value: "mouse"},
          									},
          									Resources:    v1.ResourceRequirements{Limits: v1.ResourceList{}, Requests: v1.ResourceList{}},
          									VolumeMounts: nil,
          									... // 11 identical fields
          								},
          							},
          							RestartPolicy:                 "",
          							TerminationGracePeriodSeconds: nil,
          							... // 24 identical fields
          						},
          						ContainerConcurrency: 0,
          						TimeoutSeconds:       nil,
          					},
          					DeprecatedGeneration:   0,
          					DeprecatedServingState: "",
          					... // 4 identical fields
          				},
          			},
          		},
          		RouteSpec: v1alpha1.RouteSpec{},
          	},
          	Status: v1alpha1.ServiceStatus{},
          }
```

> Interesting, are resource template files stored using -o yaml ?

Yeah, I assume to use `kn` in that way. Also, I guess it would become an issue when implementing [GitOps mode](https://github.com/knative/client/issues/30).

toVersus:
@maximilien, sorry for confusing you. The tests included in this PR are always green without a flakiness from the beginning.

OK, I summarize. When I first introduced the mock test for service create with env vars, I encountered the error described above. Therefore, I fixed some internal logic and made mock test green, then raised this PR.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -266,3 +269,33 @@ func (p *ConfigurationEditFlags) AnyMutation(cmd *cobra.Command) bool {
 	}
 	return false
 }
+
+func envVarsFromSlice(envFlags []string, delimiter string) ([]corev1.EnvVar, error) {
+	envDict := make(map[string]string, len(envFlags))
```
navidshaikh: `s/envDict/envMap/g`

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -266,3 +269,33 @@ func (p *ConfigurationEditFlags) AnyMutation(cmd *cobra.Command) bool {
...
+	envFlagsUpdated := envFlags[:0]
```
navidshaikh: may be
`s/envFlagsUpdated/envOrdered/g` ?

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -266,3 +269,33 @@ func (p *ConfigurationEditFlags) AnyMutation(cmd *cobra.Command) bool {
...
+	for _, flag := range envFlags {
+		envPair := strings.SplitN(flag, delimiter, 2)
+		switch len(envPair) {
+		case 0:
+			return nil, fmt.Errorf("argument requires a value that contains the %q character; got %q", delimiter, envPair)
+		case 1:
+			if _, ok := envDict[envPair[0]]; !ok {
```
navidshaikh: IIUC, you added this check to ensure if user gave same env var multiple times.
The envDict though uses the value of the env parsed last.

pkg/serving/config_changes.go
```diff
@@ -187,37 +187,41 @@ func UpdateLabels(service *servingv1alpha1.Service, template *servingv1alpha1.Re
...
-			if envVar.Name == name {
-				env = append(env[:i], env[i+1:]...)
-				break
-			}
+	// merged EnvVars keep the original order of equal elements
+	merged := []corev1.EnvVar{}
+	for _, v := range srcUpdated {
+		if _, ok := dict[v.Name]; ok {
+			merged = append(merged, corev1.EnvVar{Name: v.Name, Value: dict[v.Name]})
+			// Delete dict item for avoiding duplicated entries
```
navidshaikh: wondering if we could avoid this duplication, need to recheck this part.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -126,18 +127,20 @@ func (p *ConfigurationEditFlags) Apply(
 	}
 
 	if cmd.Flags().Changed("env") {
-		envMap, err := util.MapFromArrayAllowingSingles(p.Env, "=")
+		envVars, err := envVarsFromSlice(p.Env, "=")
```
navidshaikh: Can we move all the env var processing to util ? 

This helps re-using the ordered processing of pairs separated by delimiter, which can be re-used for other configs apart from env vars.

pkg/kn/commands/service/configuration_edit_flags.go
toVersus: > The envDict though uses the value of the env parsed last.

I would like to use the value of last parsed env vars, however regarding the order it makes sense to use the first timing to be appeared in the user input, because I have an image of updating the value with the same key multiple times in case that the user specified duplicate pair of env vars. In addition, it becomes a little complicated to update the order based on the count of duplicate entry. WDYT?

navidshaikh:
@toVersus : During last WG PR rodeo, its discussed that simply sort the env vars before updating the service.

toVersus:
@navidshaikh Ah okay, but should we sort env vars in order of user input or env key name?

sixolet:
Env key name. User input is a lot harder to get consistent

toVersus:
@sixolet, Thanks for clarifying, I fixed! (Also updated both the title and description of this PR.)

pkg/serving/config_changes.go
```diff
@@ -31,14 +32,13 @@ var UserImageAnnotationKey = "client.knative.dev/user-image"
...
-func UpdateEnvVars(template *servingv1alpha1.RevisionTemplateSpec, toUpdate map[string]string, toRemove []string) error {
+// new env vars and change the values of existing ones, then sort by environment key name.
+func UpdateEnvVars(template *servingv1alpha1.RevisionTemplateSpec, toUpdate map[string]string, toRemove map[string]bool) error {
 	container, err := ContainerOfRevisionTemplate(template)
 	if err != nil {
 		return err
 	}
-	envVars := updateEnvVarsFromMap(container.Env, toUpdate)
-	container.Env = removeEnvVars(envVars, toRemove)
+	container.Env = mergeEnvVarsWithStableOrder(container.Env, toUpdate, toRemove)
```
sixolet: I'd rather keep the structure where we update them, we remove some, and then we sort them at the end.  Was there a reason you didn't structure it that way?

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -139,14 +139,15 @@ func (p *ConfigurationEditFlags) Apply(
 		if err != nil {
 			return errors.Wrap(err, "Invalid --env")
 		}
-		envToRemove := []string{}
-		for name := range envMap {
-			if strings.HasSuffix(name, "-") {
-				envToRemove = append(envToRemove, name[:len(name)-1])
-				delete(envMap, name)
+		envMapToRemove := make(map[string]bool)
```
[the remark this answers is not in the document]
toVersus: Oh, nice! Very useful.

pkg/serving/config_changes.go
toVersus: I put together the steps to update and remove EnvVars because I feel it meaningless to append them into a slice and then remove elements from the same slice by using this [slice trick](https://github.com/golang/go/wiki/SliceTricks#delete). I think it's better to handle by appending items unless found in items to remove.

Otherwise, I cut out all the steps into mergeEnvVarsWithStableOrder function for testing purpose. I would like to create a dedicated unit test for detecting the order violation of EnvVars because we have never noticed even this kind of small thing.

Does it make sense?

toVersus: Talking to Navid in Slack, I will only add the ordering logic and keep the structure of updating and removing items as suggested by @sixolet in this PR.

toVersus:
It's ready for review :)

rhuss:
>     * Register list of EnvVars in alphabetical order of Env key name when creating/updating new kservice

I agree, that env vars should not be listed in random order, but why should the order by alphabetically and not the order as defined in the Service (env vars are registered as an array) ?

toVersus:
In my initial proposal below, I tried to sort the env vars in the order of user input.

> - Register list of EnvVars in the same order as user input when creating new kservice
> - Keep the original order of EnvVars when updating existing kservice

However, the ordering logic became complicated given the consistency in the order when updating/deleting existing ones. See https://github.com/knative/client/pull/389#discussion_r318072106.

https://github.com/knative/client/blob/2ca3980b3ccc7c37335ab3c306b2189053a8c153/pkg/kn/commands/service/configuration_edit_flags.go#L298-L326

https://github.com/knative/client/blob/2ca3980b3ccc7c37335ab3c306b2189053a8c153/pkg/serving/config_changes.go#L238-L275

I'm not sure what was discussed in the last WG PR rodeo, but they decided to register the env vars alphabetically. See https://github.com/knative/client/pull/389#issuecomment-526054170.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -139,14 +139,15 @@ func (p *ConfigurationEditFlags) Apply(
...
 			return errors.Wrap(err, "Invalid --env")
 		}
-		envToRemove := []string{}
-		for name := range envMap {
-			if strings.HasSuffix(name, "-") {
-				envToRemove = append(envToRemove, name[:len(name)-1])
-				delete(envMap, name)
+		envKeyToRemove := []string{}
+		for key := range envMap {
+			if !strings.HasSuffix(key, "-") {
```
navidshaikh: Why we did negative check than positive one above? There is no change in the behavior.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -139,14 +139,15 @@ func (p *ConfigurationEditFlags) Apply(
...
+				continue
 			}
+			envKeyToRemove = append(envKeyToRemove, key[:len(key)-1])
+			delete(envMap, key)
 		}
-		err = servinglib.UpdateEnvVars(template, envMap, envToRemove)
+		err = servinglib.UpdateEnvVars(template, envMap, envKeyToRemove)
```
navidshaikh: envToRemove is better representation IMO.

pkg/serving/config_changes.go
```diff
@@ -235,17 +242,17 @@ func UpdateLabels(service *servingv1alpha1.Service, template *servingv1alpha1.Re
 
 // =======================================================================================
 
-func updateEnvVarsFromMap(env []corev1.EnvVar, vars map[string]string) []corev1.EnvVar {
+func updateEnvVars(env []corev1.EnvVar, toUpdate map[string]string) []corev1.EnvVar {
```
navidshaikh: I'd say we revert this change, `updateEnvVarsFromMap` is explicit. Also we have exported method 'UpdateEnvVars' (same name), IMO we should avoid using same function names.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -17,13 +17,13 @@ package service
...
 
-	errors "github.com/pkg/errors"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"knative.dev/client/pkg/kn/flags"
 	servinglib "knative.dev/client/pkg/serving"
-	util "knative.dev/client/pkg/util"
+	"knative.dev/client/pkg/util"
```
navidshaikh: :+1:

pkg/serving/config_changes.go
```diff
@@ -31,14 +32,20 @@ var UserImageAnnotationKey = "client.knative.dev/user-image"
...
+// new env vars and change the values of existing ones, then sort by env key name.
 func UpdateEnvVars(template *servingv1alpha1.RevisionTemplateSpec, toUpdate map[string]string, toRemove []string) error {
 	container, err := ContainerOfRevisionTemplate(template)
 	if err != nil {
 		return err
 	}
-	envVars := updateEnvVarsFromMap(container.Env, toUpdate)
-	container.Env = removeEnvVars(envVars, toRemove)
+	updated := updateEnvVars(container.Env, toUpdate)
+	removed := removeEnvVars(updated, toRemove)
```
navidshaikh: removeEnvVars returning the final set of env vars to be updated, isnt using 'removed' var to represent that seems odd?

I'd use `envVars` on the left side of `=` for both `updateEnvVars` and `removeEnvVars`.

pkg/kn/commands/service/configuration_edit_flags.go
toVersus: No strong opinion is in here. It's just a matter of preference for readability, so I reverted it back.

toVersus:
@navidshaikh I think Prow robot will squash the commits when a PR is merged. Should I manually squash the commits?

https://github.com/knative/test-infra/blob/816123d5e71fb88533d1564c331443387967e1dc/ci/prow/config.yaml#L95-L96

navidshaikh:
@toVersus : There are commits in this PR for the initial approach taken, then we changed to alphabetical order, I asked explicitly to squash to reflect the approach we're taking.

toVersus:
@navidshaikh I'm not sure I could take the action you asked me to do, but I removed the commit messages regarding the initial approach from this PR.

navidshaikh:
@rhuss : 
>I agree, that env vars should not be listed in random order, but why should the order be alphabetically and not the order as defined in the Service (env vars are registered as an array) ?

The actual order of env doesn't impact and maintaining the alphabetically order is simpler and consistent than maintaining the user provided env order.

rhuss:
ok, convinced. let's go with the alphabetical order (assuming that duplicate keys in the env-list is already treated properly by the env handling as the later key overrides a former key)

PR #392, opened 2019-08-28
Supporting `envFrom`, `volumeMount`, `serviceAccountName`, and `imagePullPolicy` feature

igsong:
## Proposed Changes

To use NOT-YET supported expressions of ksvc, following flags, which can be set in `service create` or `service update`, are added. 

* Support `envFrom` expression. By giving the flag `--env-from (config-map | secret):CONFIG_MAP_OR_SECRET_NAME`, a config map or secret is embedded using `envFrom`. To unset, a `-` suffix can be used like `--env` or `--label`.
* Support `volumeMount` expression. By giving the flag `--volume-mount VOLUME_NAME=(config-map|secret):CONFIG_MAP_OR_SECRET_NAME@/mount/here`, a config map or secret is mounted on the user-container. To unset, a `-` suffix can be used like `--env` or `--label`.
* Support `serviceAccountName` expression. By giving the flag `--service-account-name=SERVICE_ACCOUNT_NAME`, a user can set a service account for a ksvc. 
* Support `imagePullPolicy` expression. By giving the flag `--image-pull-policy=POLICY`, a user can set a image pull policy for a ksvc. POLICY should be one of `always`, `if-not-present`, `never` or `-` which can be used for unsetting imagePullPolicy.

**Release Note**

```release-note
NONE
```
/lint

### DIFF
--- a/docs/cmd/kn_service_create.md
+++ b/docs/cmd/kn_service_create.md
@@ -49,9 +49,11 @@ kn service create NAME --image IMAGE [flags]
   -l, --label stringArray        Service label to set. name=value; you may provide this flag any number of times to set multiple labels. To unset, specify the label name followed by a "-" (e.g., name-).
       --limits-cpu string        The limits on the requested CPU (e.g., 1000m).
       --limits-memory string     The limits on the requested memory (e.g., 1024Mi).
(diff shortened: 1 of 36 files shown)


drnic:
Why did this PR get closed/not merged? Why all on the same day it was opened and without discussion?

drnic:
I found the subsequent PR, sorry https://github.com/knative/client/pull/392

PR #393, opened 2019-08-28
Add support for envFrom and volumeMounts

igsong:
## Proposed Changes

To use NOT-YET supported expressions of ksvc, following flags, which can be set in `service create` or `service update`, are added. 

* Support `envFrom` expression. By giving the flag `--env-from (config-map | secret):CONFIG_MAP_OR_SECRET_NAME`, a config map or secret is embedded using `envFrom`. To unset, a `-` suffix can be used like `--env` or `--label`.
* Support `volumeMount` expression. By giving the flag `--volume-mount VOLUME_NAME=(config-map|secret):CONFIG_MAP_OR_SECRET_NAME@/mount/here`, a config map or secret is mounted on the user-container. To unset, a `-` suffix can be used like `--env` or `--label`.

> Unlike initial proposal, the support for serviceAccountName will be shown another PR and imagePullPolicy has removed because Knative Serving does not support the feature.

**Release Note**

```release-note
NONE
```
/lint


navidshaikh:
> Support imagePullPolicy expression. By giving the flag --image-pull-policy=POLICY, a user can set a image pull policy for a ksvc. POLICY should be one of always, if-not-present, never or - which can be used for unsetting imagePullPolicy.

We have `lock-to-digest` and `no-lock-to-digest` flags, which users can use to achieve desired state.

igsong:
> > 
> 
> We have `lock-to-digest` and `no-lock-to-digest` flags, which users can use to achieve desired state.

Thank you for comments. But, I think that the `lock-to-digest` and `no-lock-to-digest` flags are different with `image-pull-policy` because `image-pull-policy` is a kind of caching policy of the docker image on each k8s worker node, while the `lock-to-digest`  flag automatically alternates `image` to use a digest form of image URL. Of course, when a user is using a static tag or a digest to specify an image, it is not so useful, but when using a dynamic tag, setting `imagePullPolicy`  may be needed.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -103,11 +121,15 @@ func (p *ConfigurationEditFlags) addSharedFlags(command *cobra.Command) {
 			"Accepts golang templates, allowing {{.Service}} for the service name, "+
 			"{{.Generation}} for the generation, and {{.Random [n]}} for n random consonants.")
 	p.markFlagMakesRevision("revision-name")
+	command.Flags().StringVar(&p.ServiceAccountName, "service-account-name", "-", "Service account name. To unset, specify \"-\".")
+	p.markFlagMakesRevision("service-account-name")
+	command.Flags().StringVar(&p.ImagePullPolicy, "image-pull-policy", "-", "Image pull policy (e.g., always, if-not-present, or never). To unset, specify \"-\".")
```
sixolet: `image-pull-policy` is not supported by Knative serving by design — revisions are always constant images by hash.

pkg/serving/config_changes.go
```diff
@@ -27,6 +27,15 @@ import (
 	servingv1beta1 "knative.dev/serving/pkg/apis/serving/v1beta1"
 )
 
+// VolumeSourceType is a type standing for enumeration of ConfigMap and Secret
+type VolumeSourceType int
+
+// Enumeration list of Kay-Value sources: ConfigMap or Secret
```
sixolet: Key-Value?

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -254,6 +304,20 @@ func (p *ConfigurationEditFlags) Apply(
 		}
 	}
 
+	if cmd.Flags().Changed("service-account-name") {
```
sixolet: Can you make a separate PR for this please? It's good to keep PRs supporting individual features.

pkg/serving/config_changes.go
```diff
@@ -269,3 +344,200 @@ func removeEnvVars(env []corev1.EnvVar, toRemove []string) []corev1.EnvVar {
...
+	case SecretVolumeSourceType:
+		volume.ConfigMap = nil
+		volume.Secret = &corev1.SecretVolumeSource{SecretName: info.volumeSourceName}
+	default:
+		return fmt.Errorf("Invalid VolumeSourceType")
+	}
+	return nil
+}
+
+func updateVolumeMountsFromMap(volumes []corev1.Volume, volumeMounts []corev1.VolumeMount, mounts map[string]string) ([]corev1.Volume, []corev1.VolumeMount, error) {
```
sixolet: This is long and complicated, please include a doc comment.

pkg/serving/config_changes.go
```diff
@@ -42,6 +51,36 @@ func UpdateEnvVars(template *servingv1alpha1.RevisionTemplateSpec, toUpdate map[
...
+	}
+
+	volumes, volumeMounts, err := updateVolumeMountsFromMap(template.Spec.Volumes, container.VolumeMounts, toUpdate)
+	if err != nil {
+		return err
+	}
+
+	template.Spec.Volumes, container.VolumeMounts = removeVolumeMounts(volumes, volumeMounts, toRemove)
+	return nil
+}
```
sixolet: Can we sort 'em too?

pkg/serving/config_changes.go
```diff
@@ -269,3 +344,200 @@ func removeEnvVars(env []corev1.EnvVar, toRemove []string) []corev1.EnvVar {
...
+		return -1, "", fmt.Errorf("Argument requires a value that contains the : character; got %q", s)
+	}
+
+	if strings.TrimSpace(slices[0]) == "config-map" {
+		return ConfigMapVolumeSourceType, strings.TrimSpace(slices[1]), nil
+	} else if strings.TrimSpace(slices[0]) == "secret" {
+		return SecretVolumeSourceType, strings.TrimSpace(slices[1]), nil
+	}
+
+	return -1, "", fmt.Errorf("Not allowed key-value source is used. Currently config-map and secret can be used; got %q", s)
```
sixolet: "Unsupported volume source type "asdf"; supported volume source types are "config-map" and "secret"

pkg/serving/config_changes.go
sixolet: Maybe change `mounts` to "mountsToUpdate`?

pkg/serving/config_changes.go
```diff
@@ -269,3 +344,200 @@ func removeEnvVars(env []corev1.EnvVar, toRemove []string) []corev1.EnvVar {
...
+	default:
+		return fmt.Errorf("Invalid VolumeSourceType")
+	}
+	return nil
+}
+
+func updateVolumeMountsFromMap(volumes []corev1.Volume, volumeMounts []corev1.VolumeMount, mounts map[string]string) ([]corev1.Volume, []corev1.VolumeMount, error) {
+	updatedInVolumes := make(map[string]bool)
+	updatedInVolumeMounts := make(map[string]bool)
+	parsedMap := make(map[string]*volumeMountInfo)
```
sixolet: `parsedMap` might be better named `mountInfoByName`

pkg/serving/config_changes.go
```diff
@@ -269,3 +344,200 @@ func removeEnvVars(env []corev1.EnvVar, toRemove []string) []corev1.EnvVar {
...
+		if present {
+			err := updateVolume(volume, info)
+			if err != nil {
+				return nil, nil, err
+			}
+			updatedInVolumes[volume.Name] = true
+		}
+	}
+
+	for i := range volumeMounts {
```
sixolet: This seems like this code may not handle mounting the same secret (or whatever volume) at two different paths, or handling a message where the same secret (or whatever volume) was already mounted at two different paths. There are a lot of corner cases here.

I agree with the decision not to make the user handle all the minutae of the difference between volume mounts and volumes if they do not have to, though.

pkg/kn/commands/service/configuration_edit_flags.go
igsong: Oh, I understood. I will remove the option.

pkg/serving/config_changes.go
igsong: Could you explain which parts should be sorted in more details?

pkg/serving/config_changes.go
igsong: I agree!

pkg/serving/config_changes.go
igsong: OK, then I think that it would be better to have two flags instead of current one, like `--volume-mount VOLUME_NAME=/mount/here` and `--volume VOLUME_NAME=(config-map|secret):CONFIG_MAP_OR_SECRET_NAME`. How do you think of it?

pkg/serving/config_changes.go
igsong: It will be changed as your comments

pkg/kn/commands/service/configuration_edit_flags.go
igsong: Yes, I will

pkg/serving/config_changes.go
igsong: Yes. Key-Value. But, I've changed the entire comments to clarify it.

pkg/serving/config_changes.go
igsong: I split up the flag `--volume-mount` into `--volume` and `--volume-mount` to cope with the case that @sixolet mentioned

pkg/serving/config_changes.go
igsong: While splitting the flag `--volume-mount` into `--volume` and `--volume-mount`, this part is disappeared.

pkg/serving/config_changes.go
igsong: This part also seems gone while splitting the flags.

docs/cmd/kn_service_update.md
```diff
@@ -38,28 +38,31 @@ kn service update NAME [flags]
...
+  -h, --help                       help for update
+      --image string               Image to run.
+  -l, --label stringArray          Service label to set. name=value; you may provide this flag any number of times to set multiple labels. To unset, specify the label name followed by a "-" (e.g., name-).
+      --limits-cpu string          The limits on the requested CPU (e.g., 1000m).
+      --limits-memory string       The limits on the requested memory (e.g., 1024Mi).
+      --lock-to-digest             keep the running image for the service constant when not explicitly specifying the image. (--no-lock-to-digest pulls the image tag afresh with each new revision) (default true)
+      --max-scale int              Maximal number of replicas.
+      --min-scale int              Minimal number of replicas.
+  -n, --namespace string           List the requested object(s) in given namespace.
+      --no-lock-to-digest          do not keep the running image for the service constant when not explicitly specifying the image. (--no-lock-to-digest pulls the image tag afresh with each new revision)
```
maximilien: Nit ... Start with capital to be consistent with others "Do not ..."

docs/cmd/kn_service_update.md
```diff
@@ -38,28 +38,31 @@ kn service update NAME [flags]
...
+      --env-from stringArray       Config a envfrom with a config map or secret. (config-map | secret):CONFIG_MAP_OR_SECRET_NAME you may provide this flag any number of times. To unset, specify the config map name followed by a "-" (e.g., config-map:CONFIG_MAP_NAME- or secret:SECRET_NAME-).
+  -h, --help                       help for update
+      --image string               Image to run.
+  -l, --label stringArray          Service label to set. name=value; you may provide this flag any number of times to set multiple labels. To unset, specify the label name followed by a "-" (e.g., name-).
+      --limits-cpu string          The limits on the requested CPU (e.g., 1000m).
+      --limits-memory string       The limits on the requested memory (e.g., 1024Mi).
+      --lock-to-digest             keep the running image for the service constant when not explicitly specifying the image. (--no-lock-to-digest pulls the image tag afresh with each new revision) (default true)
```
maximilien: Nit ... Start with capital to be consistent with others => "Keep the ..."

maximilien:
You might need to rebase:

```bash
 pkg/kn/commands/service/service_create_test.go
pkg/kn/commands/service/service_update_test.go 
```

navidshaikh:
@igsong : ping! let's rebase and get this in queue.

igsong:
> @igsong : ping! let's rebase and get this in queue.

Finally, I've rebased.

maximilien:
You might need to make sure and include the newly generated docs.

drnic:
👍  I look forward to demoing some features in this PR at springone on Thursday. I'll build from PR branch if its not on master by then.

rhuss:
@drnic I think we can get it in soon and maybe even before we release 0.8.0 which will land possibly also this week.

docs/cmd/kn_service_update.md
igsong: Unfortunately, it is not written by me. BTW, "do not" is generated by a library (Corbra?).

docs/cmd/kn_service_update.md
igsong: Similar with previous comment, since `--no-lock-to-digest` is automatically generated, if changing "keep" to "Keep", the description of `--no-lock-to-digest` will be "do not Keep ...".

pkg/serving/config_changes.go
```diff
@@ -305,3 +361,220 @@ func removeEnvVars(env []corev1.EnvVar, toRemove []string) []corev1.EnvVar {
...
+
+type volumeSourceInfo struct {
+	volumeSourceType VolumeSourceType
+	volumeSourceName string
+}
+
+func parseVolumeSourceString(s string) (*volumeSourceInfo, error) {
+	slices := strings.SplitN(s, ":", 2)
+	if len(slices) != 2 {
+		return nil, fmt.Errorf("Argument requires a value that contains the : character; got %q", s)
```
rhuss: error starting by lowercase (by golang convention)

docs/cmd/kn_service_create.md
```diff
@@ -39,26 +39,29 @@ kn service create NAME --image IMAGE [flags]
...
+  -p, --port int32                 The port where application listens on.
+      --requests-cpu string        The requested CPU (e.g., 250m).
+      --requests-memory string     The requested memory (e.g., 64Mi).
+      --revision-name string       The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
+      --volume stringArray         Config a volume with a config map or secret. VOLUME_NAME=(config-map|secret):CONFIG_MAP_OR_SECRET_NAME ; you may provide this flag any number of times to define multiple volumes. To unset, specify the volume name followed by a "-" (e.g., /mount/path-).
+      --volume-mount stringArray   Config a volume mount. /mount/path=VOLUME_MOUNT ; you may provide this flag any number of times to mount multiple volumes. To unset, specify the mount path followed by a "-" (e.g., /mount/path-).
```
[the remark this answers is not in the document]
igsong: Since the flags defined in `pkg/kn/commands/service/configuration_edit_flags.go` are shared for both of `kn service create` and `kn service update`, this seems inevitable unless the structure is changed. From `--env` or `--label`, the similar description can be found.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -72,6 +76,25 @@ func (p *ConfigurationEditFlags) addSharedFlags(command *cobra.Command) {
 			"any number of times to set multiple environment variables. "+
 			"To unset, specify the environment variable name followed by a \"-\" (e.g., NAME-).")
 	p.markFlagMakesRevision("env")
+
+	command.Flags().StringArrayVarP(&p.EnvFrom, "env-from", "", []string{},
+		"Config a envfrom with a config map or secret. (config-map | secret):CONFIG_MAP_OR_SECRET_NAME "+
```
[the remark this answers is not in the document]
igsong: Your suggestion seems great. I'll change it like that. In addition, since the abbreviation `cm` looks good, I'll add it as well.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -72,6 +76,25 @@ func (p *ConfigurationEditFlags) addSharedFlags(command *cobra.Command) {
...
+		"Config a volume mount. /mount/path=VOLUME_MOUNT ; "+
+			"you may provide this flag any number of times to mount multiple volumes. "+
+			"To unset, specify the mount path followed by a \"-\" (e.g., /mount/path-).")
+	p.markFlagMakesRevision("volume-mount")
+
+	command.Flags().StringArrayVarP(&p.Volume, "volume", "", []string{},
+		"Config a volume with a config map or secret. VOLUME_NAME=(config-map|secret):CONFIG_MAP_OR_SECRET_NAME ; "+
+			"you may provide this flag any number of times to define multiple volumes. "+
+			"To unset, specify the volume name followed by a \"-\" (e.g., /mount/path-).")
+	p.markFlagMakesRevision("volume")
```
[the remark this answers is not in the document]
igsong: My original proposal was using the combined flag as you suggested. However, as you can see in https://github.com/knative/client/pull/393#discussion_r320467050, there was other opinion about giving more freedom to user for mounting a volume.

Basically, I think that, for this kind of CLI, it is very important to provide all the functionalities without assumptions as well as to provide convenience.  So, I would like to suggest to provide all the flags: `--volume-mount /data=mydata --volume mydata=cm:myconfigmap` and `--mount /data=mydata:cm:myconfigmap`.  How do you think about it?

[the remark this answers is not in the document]
rhuss: @igsong thanks for the comment reference, I missed that. However, if I understand @sixolet correctly, she also opts for ease of use over completeness and covering every corner case.

I agree, if we can support both, a simple, single option use case with creating both entries (volume + mount) in one go and for more advanced uses case a second option to allow volumes to be reused, then I'm ok (but as said, we definitely should make it as easy as possible to use volumes). 

I like your suggestion, but I wonder whether we can make it even easier for the user as the volume name itself is actually an implementation detail (the user never will see this name again when using `kn` only). So I like the idea to have it autogenerated (defaulting to the cm/secret name, adding suffixes if not unique).

So my suggestion would be for the mount option, having multiple parts seperate by `:`:

* 1 part ("mydata") --> consider it a volume name
* 2 parts ("cm:mycm", "secret:mysecret") -> mount the cm or secret and create a volume name implicitely (here: configmap with name "mycm" creates a volume with name "mycm")
* 3 parts ("cm:mycm:mydata") --> mount a configmap and create a volume of the given name (here: configmap "mycm" is mounted by creating a volume "mydata")

wdyt ?

docs/cmd/kn_service_create.md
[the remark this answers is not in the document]
rhuss: ok, let's postpone this cleanup for a later release.

pkg/kn/commands/service/configuration_edit_flags.go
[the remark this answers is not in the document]
igsong: Thank you for comments. I agree with that most of users don't need to specify an explicit volume name, but, if we provide 3 kinds of data format for one flag and they are distinguished just by `:` , it may cause confusion or looks like too complicated. So my refined suggestion would be following.

* When the user want to mount a existing volume, `--mount /mydir=volume:myvolume`
* When the user want to mount a configmap or secret(a volume is automatically generated),  `--mount /mydir=(cm:myconfigmap|secret:mysecret)`
* When the user want to define a volume explicitly and mount the volume `--volume myvolume=(cm:myconfigmap|secret:mysecret) --mount /mydir=volume:myvolume`

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -160,6 +184,40 @@ func (p *ConfigurationEditFlags) Apply(
 		}
 	}
 
+	if cmd.Flags().Changed("env-from") {
+		envFromSourceToUpdate := []string{}
+		envFromSourceToRemove := []string{}
+		for _, name := range p.EnvFrom {
+			if strings.HasSuffix(name, "-") {
```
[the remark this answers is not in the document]
igsong: OK. I'll take care about that.

pkg/kn/commands/service/create_test.go
```diff
@@ -493,3 +493,111 @@ func TestServiceCreateWithServiceAccountName(t *testing.T) {
 		t.Fatalf("wrong service account name:%v", template.Spec.ServiceAccountName)
 	}
 }
+
+func TestServiceCreateWithEnvFromConfigMap(t *testing.T) {
+	action, created, _, err := fakeServiceCreate([]string{
+		"service", "create", "foo", "--image", "gcr.io/foo/bar:baz",
+		"--env-from", "config-map:config-map-name",
```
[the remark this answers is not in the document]
igsong: OK, I understood. 
I'll add a test case for such case.

pkg/serving/config_changes.go
```diff
@@ -305,3 +361,220 @@ func removeEnvVars(env []corev1.EnvVar, toRemove []string) []corev1.EnvVar {
...
+	}
+
+	var volumeSourceType VolumeSourceType
+	volumeSourceName := strings.TrimSpace(slices[1])
+	if strings.TrimSpace(slices[0]) == "config-map" {
```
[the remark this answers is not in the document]
igsong: Yes, it will be done

pkg/serving/config_changes.go
```diff
@@ -305,3 +361,220 @@ func removeEnvVars(env []corev1.EnvVar, toRemove []string) []corev1.EnvVar {
...
+		volumeSourceName: volumeSourceName,
+	}, nil
+}
+
+func updateEnvFrom(envFromSources []corev1.EnvFromSource, toUpdate []string) ([]corev1.EnvFromSource, error) {
+	insertConfigMapSet := make(map[string]bool)
+	insertSecretSet := make(map[string]bool)
+
+	for _, s := range toUpdate {
+		info, err := parseVolumeSourceString(s)
```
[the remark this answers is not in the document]
igsong: I'll change it to `parseConfigMapOrSecretSpecString`

pkg/kn/commands/service/update_test.go
```diff
@@ -669,6 +669,471 @@ func TestServiceUpdateLabelExisting(t *testing.T) {
...
+func TestServiceUpdateEnvFromAddingWithConfigMap(t *testing.T) {
+	original := newEmptyService()
+	originalTemplate, _ := servinglib.RevisionTemplateOfService(original)
+	originalTemplate.Spec.Containers[0].EnvFrom = append(originalTemplate.Spec.Containers[0].EnvFrom, corev1.EnvFromSource{
+		ConfigMapRef: &corev1.ConfigMapEnvSource{
+			LocalObjectReference: corev1.LocalObjectReference{
+				Name: "existing-name",
+			}}})
+
+	action, updated, _, err := fakeServiceUpdate(original, []string{
```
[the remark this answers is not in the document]
igsong: Yes. I understood.

pkg/serving/config_changes.go
```diff
@@ -305,3 +361,220 @@ func removeEnvVars(env []corev1.EnvVar, toRemove []string) []corev1.EnvVar {
...
+		case ConfigMapVolumeSourceType:
+			insertConfigMapSet[info.volumeSourceName] = false
+		case SecretVolumeSourceType:
+			insertSecretSet[info.volumeSourceName] = false
+		}
+	}
+
+	for i := range envFromSources {
+		envSrc := &envFromSources[i]
+		if envSrc.ConfigMapRef != nil {
```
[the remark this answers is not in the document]
igsong: Yes, right. it prevents duplication.

pkg/kn/commands/service/create_test.go
```diff
@@ -493,3 +493,111 @@ func TestServiceCreateWithServiceAccountName(t *testing.T) {
 		t.Fatalf("wrong service account name:%v", template.Spec.ServiceAccountName)
 	}
 }
+
+func TestServiceCreateWithEnvFromConfigMap(t *testing.T) {
```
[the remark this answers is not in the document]
igsong: Sure.

pkg/serving/config_changes.go
igsong: OK

pkg/kn/commands/service/configuration_edit_flags.go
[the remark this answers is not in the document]
rhuss: Sounds good to me, with maybe the refined to omit the `volume:` prefix for volumes, and use a prefix only for `cm` (`configmap`) and `secret`. I think this is still unique and imo easier to understand. (as you would use the volume name only in `--volume myvolume=cm:myconfig` instead of `--volume volume:myvolume=cm:myconfig`)

igsong:
@rhuss I'm now working on it. If there are no special events, I can commit all the changes before next Tuesday.

pkg/kn/commands/service/configuration_edit_flags.go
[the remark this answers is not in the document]
igsong: OK, I've changed as you mentioned.

pkg/serving/config_changes.go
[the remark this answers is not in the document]
igsong: It is changed into `newVolumeSourceInfoWithSpecString` at my last commit.

pkg/serving/config_changes.go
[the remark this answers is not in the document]
igsong: Basically, I've totally restructured the algorithm. Could you please review again?

pkg/serving/config_changes.go
```diff
@@ -305,3 +361,220 @@ func removeEnvVars(env []corev1.EnvVar, toRemove []string) []corev1.EnvVar {
...
+			if _, ok := insertConfigMapSet[envSrc.ConfigMapRef.Name]; ok {
+				insertConfigMapSet[envSrc.ConfigMapRef.Name] = true
```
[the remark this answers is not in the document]
igsong: As I mentioned above, I've totally restructured that part.

pkg/serving/config_changes.go
```diff
@@ -305,3 +361,220 @@ func removeEnvVars(env []corev1.EnvVar, toRemove []string) []corev1.EnvVar {
...
+	}
+
+	return &volumeSourceInfo{
+		volumeSourceType: volumeSourceType,
+		volumeSourceName: volumeSourceName,
+	}, nil
+}
+
+func updateEnvFrom(envFromSources []corev1.EnvFromSource, toUpdate []string) ([]corev1.EnvFromSource, error) {
+	insertConfigMapSet := make(map[string]bool)
```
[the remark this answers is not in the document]
igsong: Same as above comments, I've re-programmed that part. Now the order will be kept.

igsong:
I've done to apply all the suggested and discussed changes. 

One thing I want to notice is that `OrderedMap` is newly adopted to make behaviors caused by `--volume` and `--mount` deterministic.

docs/cmd/kn_service_create.md
```diff
@@ -56,13 +57,15 @@ kn service create NAME --image IMAGE [flags]
       --lock-to-digest           keep the running image for the service constant when not explicitly specifying the image. (--no-lock-to-digest pulls the image tag afresh with each new revision) (default true)
       --max-scale int            Maximal number of replicas.
       --min-scale int            Minimal number of replicas.
+      --mount stringArray        Mount a ConfigMap (prefix cm: or config-map:), a Secret (prefix secret: or sc:), or an existing Volume (without any prefix) on the specified directory. Example: --mount /mydir=cm:myconfigmap, --mount /mydir=secret:mysecret, or --mount /mydir=myvolume. When a configmap or a secret is specified, a corresponding volume is automatically generated. You can use this flag multiple times. To unmount, append "-" to the mounted directory, e.g. --volume /mydir-. When unmounting, the automatically generated volume is removed as well.
```
rhuss: In the last part, you mean `--mount /mydir`, not `--volume /mydir`, or ?

I would rephrase that last sentence to:

> For unmounting a directory append "-" e.g. --mount /mydir- which also removes any auto-generated volume.

(the last sentence is a bit of a simplication as it only removes the auto-generated volume for this mount, but I think this is clear enough).

pkg/serving/config_changes.go
```diff
@@ -304,3 +425,255 @@ func removeEnvVars(env []corev1.EnvVar, toRemove []string) []corev1.EnvVar {
...
+	for _, volumeMount := range volumeMounts {
+		if volumeMount.Name == volumeName {
+			return true
+		}
+	}
+	return false
+}
+
+func generateVolumeName(path string) string {
+	return fmt.Sprintf("kn-managed-volume:%s", path)
```
rhuss: This is too simplistic, as you have to sanitize the path that it's a well-formed kubernetes identifier. As the path has to start with `/` (as you can only mount absolute paths) you will always get an invalid name here. Unfortunately, such things can be only detected in integration tests, not unit tests, as unit tests don't do any validation.

Also, I would either don't use a prefix or, if required e.g. when for auto-deleting volumes, then just use `kn:`, which is just shorter and easier to read when examing the k8s files.

docs/cmd/kn_service_create.md
igsong: Oh, my mistake. I fixed it as your suggestion.

pkg/serving/config_changes.go
igsong: I've changed the part to use a sanitized path string + checksum of original path string without any prefix.

rhuss: Thanks, works for me now !

I'm going to go over the rest of the PR quickly and lets get it merged today so that we get it into 0.10.0 (to be released today). We can then fix any outstanding issues (or add some real integrations tests) later.

rhuss:
The CI test error is a strange beast. See slack channel for the discussion about it. Hopefully its only a hickup.

PR #401, opened 2019-09-04
Add --service-account-name flag

igsong:
## Proposed Changes

* Add a flag `--service-account-name` to set serviceAccountName property of ksvc 

**Release Note**

```release-note
Add --service-account flag
```


pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -107,7 +108,8 @@ func (p *ConfigurationEditFlags) addSharedFlags(command *cobra.Command) {
 		"keep the running image for the service constant when not explicitly specifying "+
 			"the image. (--no-lock-to-digest pulls the image tag afresh with each new revision)")
 	// Don't mark as changing the revision.
-
+	command.Flags().StringVar(&p.ServiceAccountName, "service-account-name", "-", "Service account name. To unset, specify \"-\".")
```
navidshaikh: any particular reason we are not using default as `-` instead of `nil`?

navidshaikh:
Addition of new flag should be highlighted in release notes, can you please append a line in https://github.com/knative/client/blob/master/CHANGELOG.adoc ? 
rebase needed.

docs/cmd/kn_service_create.md
```diff
@@ -39,26 +39,27 @@ kn service create NAME --image IMAGE [flags]
...
+      --max-scale int                 Maximal number of replicas.
+      --min-scale int                 Minimal number of replicas.
+  -n, --namespace string              List the requested object(s) in given namespace.
+      --no-lock-to-digest             do not keep the running image for the service constant when not explicitly specifying the image. (--no-lock-to-digest pulls the image tag afresh with each new revision)
+  -p, --port int32                    The port where application listens on.
+      --requests-cpu string           The requested CPU (e.g., 250m).
+      --requests-memory string        The requested memory (e.g., 64Mi).
+      --revision-name string          The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
+      --service-account-name string   Service account name. To unset, specify "-". (default "-")
```
maximilien: Let’s make clearer with “To unset specify “-“ as the value”

Also would this be clearer with “” instead of “-“?

pkg/serving/config_changes.go
```diff
@@ -233,6 +233,17 @@ func UpdateLabels(service *servingv1alpha1.Service, template *servingv1alpha1.Re
 	return nil
 }
 
+// UpdateServiceAccountName updates the service account name used for the corresponding knative service
+func UpdateServiceAccountName(template *servingv1alpha1.RevisionTemplateSpec, serviceAccountName string) error {
+	serviceAccountName = strings.TrimSpace(serviceAccountName)
+	if serviceAccountName == "-" {
```
maximilien: Related to my comment above in description. Empty string for unsettling would make this read better

docs/cmd/kn_service_create.md
igsong: I agree. I will make the change.

rhuss: How can you unset something when you create a service ? Isn't it just good enough to leave out the option if you don't want to create a service account being set ?

docs/cmd/kn_service_update.md
```diff
@@ -38,28 +38,29 @@ kn service update NAME [flags]
...
+      --max-scale int                 Maximal number of replicas.
+      --min-scale int                 Minimal number of replicas.
+  -n, --namespace string              List the requested object(s) in given namespace.
+      --no-lock-to-digest             do not keep the running image for the service constant when not explicitly specifying the image. (--no-lock-to-digest pulls the image tag afresh with each new revision)
+  -p, --port int32                    The port where application listens on.
+      --requests-cpu string           The requested CPU (e.g., 250m).
+      --requests-memory string        The requested memory (e.g., 64Mi).
+      --revision-name string          The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
+      --service-account-name string   Service account name. To unset, specify "-". (default "-")
```
rhuss: Does this mean, that when you use `kn service update` without this option, that the service account will be always removed because the default is `-` ? I would expect that there is no default and service account name is left untouched when this option is not given.

rhuss: For `kn service update` I would prefer:

* `--service-account new-name` : Update the service account with the new name
* `--service-account` : Clear the service account
* Option not given : Don't touch service account

docs/cmd/kn_service_create.md
igsong: Oh, I see. I'll make it optional.

pkg/kn/commands/service/configuration_edit_flags.go
igsong: I've changed the default value to an empty string.

docs/cmd/kn_service_create.md
igsong: I've changed the description of the flag with the change of the name of the flag.

pkg/serving/config_changes.go
igsong: I've changed the default value to an empty string.

docs/cmd/kn_service_update.md
igsong: @rhuss I've made a change according to your comments. Please check it out and, if there are additional comments, please let me know.

pkg/kn/commands/service/configuration_edit_flags.go
navidshaikh: thanks!

navidshaikh:
We'll need `service describe` to list the service account info if present.

docs/cmd/kn_service_create.md
```diff
@@ -58,6 +58,7 @@ kn service create NAME --image IMAGE [flags]
       --requests-cpu string      The requested CPU (e.g., 250m).
       --requests-memory string   The requested memory (e.g., 64Mi).
       --revision-name string     The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
+      --service-account string   Service account name to set. Empty service account name will result to clear the service account.
```
rhuss: I know, it's a bit nit-picky, but using a `--service-account` without arg during a `kn service create` is a no-op and it won't clear the service account (as there is one). I know that this option is shared, but maybe we can use still different help messages for create & update ?

happy to get this PR merged now, but maybe we can adjust this in a later PR.

Issue #396, opened 2019-08-30
Propose - migration Knative resources from one cluster to another by kn plugin


ZhuangYuZY:
## Describe the feature:

Describe the feature:

  Today is multiple clouds world, customer may work on multiple k8s clusters from multiple cloud providers to run Knative applications, like IBM Cloud, AWS, GCP. Even in one cloud provider, client may also have multiple k8s clusters which running Knative applications. So there is a requirement that customer wants to clone and migrate Knative resources from one cluster and another cluster, instead of re-deploy Knative application multiple times. It will give benefit that keeping Knative configuration and revision consistent in multiple clusters, it will be very important because some traffic control rules depend on revisions. Beside clone and migration, backup and recovery are also the important requirement for Knative users to handle risk of cluster broken, as deploy Knative application again can not recover revision history.

Proposal to do:

  1. Export all Knative resources. It can generate all of json/yml files for Knative service, revision, and route/configuration.
  2. Import all Knative resources to another cluster and namespace. It will give the capability to recreate the Knative service with all revision history. 
  3. With export and import, we can migrate or clone Knative resources from one cluster to another cluster.

  Implement plugin "migration" for kn client with 3 sub-commands as below:
```
  1. kn migration export --source-namespace ns-source -o knative-resource.yml
  2. kn migration import --destination-namespace ns-dest -i knative-resource.yml  --force
  3. kn migration run --destination-namespace ns-dest --destination-kubeconfig xxx  --source-namespace ns-source --source-kubeconfig xxx  --force
```

zhangtbj:
This is what we did for the proposal first:
```
jordanzt@zhangtaodembp:~/Works/workspaces/gopath/src/github.ibm.com/BlueMix-Fabric/kn-migration-tool$ ./kn-migration-tool migrate --help
Migrate knative services from source cluster to destination cluster

Usage:
  knmt migrate [flags]

Flags:
  -d, --delete                          Force delete the knative resource after migration on source cluster
  -t, --destination-kubeconfig string   The kubeconfig of the destination knative resources (default is KUBECONFIG2 from ENV property)
  -m, --destination-namespace string    The namespace of the destination knative resources (default "default")
  -f, --force                           Force update if the knative resource exists on destination cluster
  -h, --help                            help for migrate
  -s, --source-kubeconfig string        The kubeconfig of the source knative resources (default is KUBECONFIG from ENV property)
  -n, --source-namespace string         The namespace of the source knative resources (default "default")

Global Flags:
[...]

PR #406, opened 2019-09-11
feature(kn migration): Migrate kn services from source to destination cluster

zhangtbj:
This is the implementation of the proposal:
https://github.com/knative/client/issues/396

We took some time to design workflow and make sure:
1, Service configuration and traffic splitting percents are correct after migration
2, Revision names are same after migration
3, Revision history and generation version are same after migration
Then we can make sure the customer won't notice this migration and all kn services work as before correctly.
 
 
**Here is the implementation logic**:
1, Get kn service list and migrate this list one by one
2, Get revision list of this service
3, Migrate the service then create and link the first revision by LastReadyRevision and change its generation version to the correct version
4, Check if provides --force option, if service exists, delete and recreate on the destination cluster
5, Migrate all other revisions of this service and link them to the service/configuration
6, Continue migrating other services
7, Check if all services are migrated correctly
8, Check if provides --delete option, if yes, delete kn resources on source cluster in the end
 

Here are the screenshots for the new command in kn client:
 
migrate sub command in kn:
kn_migration

kn migrate command help:
kn_migrate_help

I can also paste the migrate process log here for review if needed.

 Please review and thanks a lot!


pkg/contrib/plugins/kn-migration/migrate.go
```diff
@@ -0,0 +1,326 @@
+// Copyright © 2019 The Knative Authors
```
maximilien: Have you considered breaking this into smaller components or files?

rhuss:
Haven't looled at the PR yet in detail, but shouldn't it be `kn service migrate` instead of only `kn migrate` ? Remember, our general command scheme is `kn <noun> <verb>` and afaiu its about service migration, right ?

rhuss:
So you could name the plugin `kn-service-migrate` ....

toVersus:
> Then we can make sure the customer won't notice this migration and all kn services work as before correctly.

To be precise, this is not correct.

- The autoscaler in destination cluster doesn't have any scaling metrics (like desired Pod counts), so the imported kservice won't be scaled to the same numbers of Pod in source cluster. This becomes a problem depending on what you expect to `kn migration`. If the customer wants to sync the current numbers of Pod, should you set it to the `autoscaling.knative.dev/minScale` annotation?

- The Revision GC timing will be broken because the Revision GC controller decides to delete a Revision [based on its CreationTimestamp](https://github.com/knative/serving/blob/a6a69f05dddad562f81539d6ef78b52cebc44226/pkg/reconciler/gc/gc.go#L128-L131) and [stale-revision-create-delay parameter](https://github.com/knative/serving/blob/8e3c98329c7b983b351644b0b7d5697353ec4db0/config/config-gc.yaml#L40-L41).

zhangtbj:
Hi @toVersus ,

Thanks for the comment.

1, for the annotation can be migrated too. I have verified, like the annotation `autoscaling.knative.dev/minScale` will be kept after migration, so it is safe:
```
  annotations:
    autoscaling.knative.dev/minScale: "1"
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"serving.knative.dev/v1alpha1","kind":"Service","metadata":{"annotations":{},"creationTimestamp":"2019-06-20T07:54:15Z","generation":3,"name":"helloworld","namespace":"default"},"spec":{"runLatest":{"configuration":{"revisionTemplate":{"metadata":{"creationTimestamp":null},"spec":{"container":{"image":"us.icr.io/knative_jordan/helloworld:2.0","name":"","resources":{}},"timeoutSeconds":300}}}}}}
    serving.knative.dev/creator: IAM#zhangtbj@cn.ibm.com
    serving.knative.dev/lastModifier: IAM#zhangtbj@cn.ibm.com
```

2, Yes, I saw the revisions will be cleared up after some time. But I also saw the revisions history will be cleared up too, so we never see them in the source cluster too.

So I think the current status(cleared revision history) in source cluster be migrated to destination cluster is safe too.

Thanks!

zhangtbj:
Hi @rhuss ,
Thanks for the comment, I have put the `migrate` under service command:
https://github.com/knative/client/pull/406/commits/0a32eec42a49c1c3b7587d7810166f7ea402b14f

But I also put the code under service package because:
1, Avoid the cycle import issue for test and AddCommand
2, Share same functions with service

Pls review and thanks!

pkg/contrib/plugins/kn-migration/migrate.go
```diff
@@ -0,0 +1,326 @@
...
+
+	serviceMigrateCommand := &cobra.Command{
+		Use:   "migrate",
+		Short: "Migrate Knative services from source cluster to destination cluster",
+		Example: `
+  # Migrate Knative services from source cluster to destination cluster by export KUBECONFIG as environment variables
+  kn migrate --source-namespace default --destination-namespace default
+
+  # Migrate Knative services from source cluster to destination cluster by set kubeconfig as parameters
+  kn migrate --source-namespace default --destination-namespace default --source-kubeconfig /Users/jordan/.kube/config/source-cluster-config.yml --destination-kubeconfig /Users/jordan/.kube/config/destination-cluster-config.yml
```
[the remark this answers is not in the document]
zhangtbj: fixed it https://github.com/knative/client/pull/406/commits/0a32eec42a49c1c3b7587d7810166f7ea402b14f thanks!

pkg/contrib/plugins/kn-migration/migrate.go
```diff
@@ -0,0 +1,326 @@
...
+			} else {
+				namespaceD = migrateFlags.DestinationNamespace
+			}
+
+			kubeconfigS := migrateFlags.SourceKubeconfig
+			if kubeconfigS == "" {
+				kubeconfigS = os.Getenv("KUBECONFIG")
+			}
+			if kubeconfigS == "" {
+				return fmt.Errorf("cannot get source cluster kube config, please use --destination-kubeconfig or export env KUBECONFIG2 to set")
```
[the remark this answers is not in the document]
zhangtbj: fixed in https://github.com/knative/client/pull/406/commits/0a32eec42a49c1c3b7587d7810166f7ea402b14f thanks!

pkg/contrib/plugins/kn-migration/migrate.go
```diff
@@ -0,0 +1,326 @@
...
+			}
+			namespaceExists, err := namespaceExists(*clientset, namespaceD)
+			if err != nil {
+				return err
+			}
+
+			if !namespaceExists {
+				fmt.Println("Create namespace", color.BlueString(namespaceD), "in destination cluster")
+				nsSpec := &apiv1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespaceD}}
+				_, err = clientset.CoreV1().Namespaces().Create(nsSpec)
```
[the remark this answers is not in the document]
zhangtbj: fixed in https://github.com/knative/client/pull/406/commits/0a32eec42a49c1c3b7587d7810166f7ea402b14f thanks!

pkg/contrib/plugins/kn-migration/migrate.go
```diff
@@ -0,0 +1,326 @@
...
+
+				if serviceExists {
+					if !migrateFlags.ForceReplace {
+						fmt.Println("\n[Error] Cannot kn-migration service", color.CyanString(service_s.Name), "in namespace", color.BlueString(namespaceS),
+							"because the service already exists and no --force option was given")
+						os.Exit(1)
+					}
+					fmt.Println("Deleting service", color.CyanString(service_s.Name), "from the destination cluster and recreate as replacement")
+					client_d.DeleteService(service_s.Name)
+					if err != nil {
```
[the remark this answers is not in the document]
zhangtbj: fixed in https://github.com/knative/client/pull/406/commits/0a32eec42a49c1c3b7587d7810166f7ea402b14f thanks!

pkg/contrib/plugins/kn-migration/migrate.go
```diff
@@ -0,0 +1,326 @@
...
+	}
+	if err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+// Create service struct from provided options
+func constructService(originalservice serving_v1alpha1_api.Service, namespace string) *serving_v1alpha1_api.Service {
+
```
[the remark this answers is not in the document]
zhangtbj: fixed thanks!

pkg/contrib/plugins/kn-migration/migrate_test.go
```diff
@@ -0,0 +1,168 @@
...
+		t.Fatalf("wrong stdout message: %v", output)
+	}
+	if !strings.Contains(output, "OK") || !strings.Contains(output, "Waiting") {
+		t.Fatalf("not running in sync mode")
+	}
+
+	_, _, _, err = fakeMigrate([]string{
+		"migrate", "--source-namespace", "default", "--destination-namespace", "default", "--force", "--delete"}, false, false)
+	if err != nil {
+		//t.Fatal(err)
```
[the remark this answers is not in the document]
zhangtbj: Yes, I didn't prepare the test first, so it is just a fake test. I would like to see if the implementation is correct and agreed. If all agree, I will finish the test soon.

Or I have to do many refactor based on the comments. thanks!

rhuss:
@zhanggbj thanks ! I will have the chance to give it a look and review tomorrow (sorry, busy otherwise today).

docs/cmd/kn_service_migrate.md
```diff
@@ -0,0 +1,53 @@
...
+
+  # Migrate Knative services from source cluster to destination cluster and delete the service in source cluster
+  kn migrate --source-namespace default --destination-namespace default --force --delete
+```
+
+### Options
+
+```
+      --delete                          Delete all Knative resources after kn-migration from source cluster
+      --destination-kubeconfig string   The kubeconfig of the destination Knative resources (default is KUBECONFIG2 from ENV property)
```
rhuss: Why would one prefer to different kubeconfig files instead of two different Kubernetes contexts ?. I.e. a kubeconfig file can contain many contexts, and the source and destination should map directly to a context not to kubeconfig's default context only.

I.e. you could also want to migrate from one to another context stored in the _same_ kubeconfig file. It's not the file which defines a Kubernetes cluster (but the context within the config file)

docs/cmd/kn_service_migrate.md
```diff
@@ -0,0 +1,53 @@
...
+      --destination-namespace string    The namespace of the destination Knative resources
+      --force                           Migrate service forcefully, replaces existing service if any.
+  -h, --help                            help for migrate
+      --source-kubeconfig string        The kubeconfig of the source Knative resources (default is KUBECONFIG2 from ENV property)
+      --source-namespace string         The namespace of the source Knative resources
```
rhuss: Can we use `--namespace` as this how the other commands use the name to operate on (which as default would be the current namespace).

docs/cmd/kn_service_migrate.md
rhuss: My suggestion would be:

* Allow multiple kubeconfig files like we already do by supporting a path like structure for the env var (`KUBECONFIG=firstconfig:secondconfig`)
* For specifying source and target, we by default both the default context, but one can override this with `--source-context` and `--target-context`, respectively.

That way we would stay consistent with the current handling of the kubeconfi configuration.

docs/cmd/kn_service_migrate.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+### Options inherited from parent commands
+
+```
+      --config string       kn config file (default is $HOME/.kn/config.yaml)
+      --kubeconfig string   kubectl config file (default is $HOME/.kube/config)
```
rhuss: See here where we already us a `--kubeconfig` (in addition to the other config files). I just would keep that and would work with Kubernetes contexts.

pkg/kn/commands/service/migrate_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+func fakeMigrate(args []string, withExistingService bool, sync bool) (
+	action client_testing.Action,
+	service *v1alpha1.Service,
+	output string,
+	err error) {
+	knParams := &commands.KnParams{}
+	nrGetCalled := 0
+	cmd, fakeServing, buf := commands.CreateTestKnCommand(NewServiceCommand(knParams), knParams)
+	fakeServing.AddReactor("get", "services",
```
rhuss: Woudl it be possible to switch to the KnClient Mock testing framework as we use it in https://github.com/knative/client/blob/master/pkg/kn/commands/service/describe_test.go 

See also https://github.com/knative/client/issues/358 with the plan to migrate existing tests to the Mock Tests which would also ensure that Knative is accessed only via the `KnClient` abstraction. New code should already adapt to this "new" way of testing to save some work later on.

go.mod
```diff
@@ -1,13 +1,16 @@
 module knative.dev/client
 
 require (
+	github.com/fatih/color v1.7.0
```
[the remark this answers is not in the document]
zhangtbj: Sure, I have remove the `color` import and format, thanks!
https://github.com/knative/client/pull/406/commits/8f44d46bd4ebf913a9df4b225e3b792569e68a6a

docs/cmd/kn_service_migrate.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Options
+
+```
+      --delete                          Delete all Knative resources after kn-migration from source cluster
+      --destination-kubeconfig string   The kubeconfig of the destination Knative resources (default is KUBECONFIG2 from ENV property)
+      --destination-namespace string    The namespace of the destination Knative resources
+      --force                           Migrate service forcefully, replaces existing service if any.
+  -h, --help                            help for migrate
+      --source-kubeconfig string        The kubeconfig of the source Knative resources (default is KUBECONFIG2 from ENV property)
```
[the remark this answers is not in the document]
zhangtbj: - Yes, it is environment variable, and I have change to `KUBECONFIG_DESTINATION` in https://github.com/knative/client/pull/406/commits/8f44d46bd4ebf913a9df4b225e3b792569e68a6a

docs/cmd/kn_service_migrate.md
zhangtbj: Yes, changed to use `--namespace` like other commands:
https://github.com/knative/client/pull/406/commits/8f44d46bd4ebf913a9df4b225e3b792569e68a6a

docs/cmd/kn_service_migrate.md
zhangtbj: Yes, change to use `--kubeconfig` like other commands:
https://github.com/knative/client/pull/406/commits/8f44d46bd4ebf913a9df4b225e3b792569e68a6a

pkg/kn/commands/service/migrate_test.go
zhangtbj: Yes, change to use KnClient Mock testing framework

docs/cmd/kn_service_migrate.md
zhangtbj: Hi @rhuss ,

I think if it is a migration case, people won't like to save two kubeconfig in one file.

In most of the cases, after migration, the source cluster may be removed or never be used.

So I prefer to use `--kubeconfig` as other commands for source cluster and use `--destination-kubeconfig` for target cluster. then user doesn't need to copy kube confi to the one file manually.

Pls let me know if you have other comment, thanks a lot!

toVersus: I concur with rhuss's suggestion. Should we open a knob for migrating ksvcs using cluster context?

> user doesn't need to copy kube confi to the one file manually.

You can use a single kubeconfig file for managing multiple clusters regardless of the provider. So you just add new context for the destination cluster and delete it after migration if you would like to do by using kubectl or other CLI tools. Some CLI tools (`gcloud`, `eksctl`, ...) automatically add a context to `~/.kube/config` file (or where you want to put it in) and set it to current context when creating a new cluster.

1. An cluster admin creates first K8S cluster and saves its context in `~/.kube/config` file with `first-cluster`

1. An cluster admin creates a ksvc in first K8S cluster using current context (`first-cluster`)

    ```bash
    kn service create hello --image "gcr.io/knative-samples/helloworld-go"
    ```

1. An cluster admin creates second K8S cluster and saves its context in `~/.kube/config` file with `second-cluster`

1. An cluster admin migrates existing ksvc into new K8S cluster

    ```bash
    # If a current context is set to target cluster:
    kn service migrate hello --source-context first-cluster
    # or a current context is set to source cluster:
    kn service migrate hello --target-context second-cluster
    ```

1. An cluster admin delete source context from `~/.kube/config` file

    ```bash
    kubectl config delete-cluster <first cluster name>
    kubectl config delete-cluster first-cluster
    kubectl config  unset users.<user name>
    ```

pkg/kn/commands/service/migrate.go
```diff
@@ -152,11 +152,11 @@ func NewServiceMigrateCommand(p *commands.KnParams) *cobra.Command {
 
 				if serviceExists {
 					if !migrateFlags.ForceReplace {
-						fmt.Println("\n[Error] Cannot kn-migration service", color.CyanString(service_s.Name), "in namespace", color.BlueString(namespaceS),
+						fmt.Println("\n[Error] Cannot kn-migration service", service_s.Name, "in namespace", namespaceS,
```
maximilien: Change to “... Cannot migrate service ...”

pkg/kn/commands/service/migrate.go
```diff
@@ -167,7 +167,7 @@ func NewServiceMigrateCommand(p *commands.KnParams) *cobra.Command {
 				if err != nil {
 					return err
 				}
-				fmt.Println("Migrate service", color.CyanString(service_s.Name), "Successfully")
+				fmt.Println("Migrate service", service_s.Name, "Successfully")
```
maximilien: Change to “Migrated service”

pkg/kn/commands/service/migrate.go
```diff
@@ -195,7 +195,7 @@ func NewServiceMigrateCommand(p *commands.KnParams) *cobra.Command {
 						if err != nil {
 							return err
 						}
-						fmt.Println("Migrate revision", color.CyanString(revision.Name), "successfully")
+						fmt.Println("Migrate revision", revision.Name, "successfully")
```
maximilien: Change to “Migrated revision”

cppforlife:
i think this functionality would be great as a plugin instead of being placed into core.

zhangtbj:
Hi @cppforlife , as we discussed before, we plan to put this in the kn client plugin folder: `contrib/plugins/kn-service-migrate`

pkg/kn/commands/service/migrate.go
zhangtbj: Fixed in https://github.com/knative/client/pull/406/commits/bb3fd28369fc262e8757bce812b525effe74fe3d, thanks!

pkg/kn/commands/service/migrate.go
zhangtbj: Fixed in https://github.com/knative/client/pull/406/commits/bb3fd28369fc262e8757bce812b525effe74fe3d thanks!

pkg/kn/commands/service/migrate.go
zhangtbj: Fixed in https://github.com/knative/client/pull/406/commits/bb3fd28369fc262e8757bce812b525effe74fe3d thanks!

docs/cmd/kn_service_migrate.md
```diff
@@ -0,0 +1,52 @@
...
+
+```
+kn service migrate [flags]
+```
+
+### Examples
+
+```
+
+  # Migrate Knative services from source cluster to destination cluster by export KUBECONFIG and KUBECONFIG_DESTINATION as environment variables
```
sixolet: Let's remove the KUBECONFIG_DESTINATION env var, as it's not really something you want to set-and-forget

docs/cmd/kn_service_migrate.md
```diff
@@ -0,0 +1,52 @@
...
+
+  # Migrate Knative services from source cluster to destination cluster and delete the service in source cluster
+  kn migrate --namespace default --destination-namespace default --force --delete
+```
+
+### Options
+
+```
+      --delete                          Delete all Knative resources after kn-migration from source cluster
+      --destination-kubeconfig string   The kubeconfig of the destination Knative resources (default is KUBECONFIG_DESTINATION from environment variable)
```
sixolet: Can we also use contexts within the same kubeconfig? I'd like to.

pkg/kn/commands/service/migrate.go
```diff
@@ -0,0 +1,320 @@
...
+					return err
+				}
+
+				if serviceExists {
+					if !migrateFlags.ForceReplace {
+						fmt.Println("\n[Error] Cannot migrate service", service_s.Name, "in namespace", namespaceS,
+							"because the service already exists and no --force option was given")
+						os.Exit(1)
+					}
+					fmt.Println("Deleting service", service_s.Name, "from the destination cluster and recreate as replacement")
```
sixolet: "and replacing with migrated service"

pkg/kn/commands/service/migrate.go
```diff
@@ -0,0 +1,320 @@
...
+}
+
+// Create revision struct from provided options
+func constructRevision(originalrevision serving_v1alpha1_api.Revision, config_uuid types.UID, namespace string) *serving_v1alpha1_api.Revision {
+
+	revision := serving_v1alpha1_api.Revision{
+		ObjectMeta: originalrevision.ObjectMeta,
+	}
+
+	revision.ObjectMeta.Namespace = namespace
```
sixolet: Set fields directly on revision

pkg/kn/commands/service/migrate.go
```diff
@@ -0,0 +1,320 @@
...
+	return nil
+}
+
+// Create service struct from provided options
+func constructMigratedService(originalservice serving_v1alpha1_api.Service, namespace string) *serving_v1alpha1_api.Service {
+	service := serving_v1alpha1_api.Service{
+		ObjectMeta: originalservice.ObjectMeta,
+	}
+
+	service.ObjectMeta.Namespace = namespace
```
sixolet: set ObjectMeta fields directly on service

pkg/kn/commands/service/migrate.go
```diff
@@ -0,0 +1,320 @@
...
+	service.Spec = originalservice.Spec
+	service.Spec.Template.ObjectMeta.Name = originalservice.Status.LatestCreatedRevisionName
+	service.ObjectMeta.ResourceVersion = ""
+	return &service
+}
+
+// Create revision struct from provided options
+func constructRevision(originalrevision serving_v1alpha1_api.Revision, config_uuid types.UID, namespace string) *serving_v1alpha1_api.Revision {
```
sixolet: Suggest returning RevisionTemplate instead of revision

pkg/kn/commands/service/migrate.go
```diff
@@ -0,0 +1,320 @@
...
+					fmt.Errorf(err.Error())
+				}
+				servingclient_d, err := dp.GetConfig()
+				if err != nil {
+					return err
+				}
+				for i := 0; i < len(revisions_s.Items); i++ {
+					revision_s := revisions_s.Items[i]
+					if revision_s.Name != service_s.Status.LatestReadyRevisionName {
+						revision, err := servingclient_d.Revisions(namespaceD).Create(constructRevision(revision_s, config_uuid, namespaceD))
```
sixolet: Don't directly create revisions

pkg/kn/commands/service/migrate.go
```diff
@@ -0,0 +1,320 @@
...
+						retries := 0
+						for {
+							revision, err := client_d.GetRevision(revision_s.Name)
+							if err != nil {
+								return err
+							}
+
+							source_revision_generation := revision_s.ObjectMeta.Labels["serving.knative.dev/configurationGeneration"]
+							revision.ObjectMeta.Labels["serving.knative.dev/configurationGeneration"] = source_revision_generation
+							_, err = servingclient_d.Revisions(namespaceD).Update(revision)
```
sixolet: Don't directly update revisions

maximilien:
@zhangtbj you have some conflicts that require rebase and more importantly you need to add more tests to satisfy [coverage](https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_client/406/pull-knative-client-go-coverage/1174952502114652160).

zhangtbj:
Hi Max and all,

I resolved the conflict today. I think after we confirm the FINAL implementation, I can add the test and coverage soon.

Right now, I think the big question is if we need to add kubeconfig context support in this PR.

I have some thought about that:
1, This is a migration case, it mean user or automation will migrate all knative resources from one cluster to another. The second cluster maybe new cluster or auto-creation. soe the second cluster may not under the KUBECONFIG context.
So use the multiple contexts in ONE KUBECONFIG is not good for the new cluster case, it is not good to allow use to manually copy the cluster config into KUBECONFIG before doing this migration.

2, The current kn client doesn't support get kube config from context:
https://github.com/knative/client/blob/master/pkg/kn/commands/types.go
ALL clients and functions use `GetClientConfig()` to get config, if we want to support context, I need to add a new `GetClientConfig(context string)` function and new newServingClient(context string) for the support. Do we also want to change and add here?

In the review meeting, we agreed to use `KUBECONFIG` and `KUBECONFIG_DESTINATION` way to do this implementation.
If we want to add context way support, I think another PR later to enhance this function is better.

Do we all agree?
If we want to use the `KUBECONFIG` and `KUBECONFIG_DESTINATION` way, I can add test and re-submit the PR

If we want to use context way, I need more time to add the logic...

Thanks all!

rhuss:
@zhangtbj thanks for resolving the conflicts. I'm just back from PTO and hope to revisit this PR soonish. 

In the meantime, would you mind to have a look at why the unit tests are failing ?

maximilien:
@zhangtbj please close this since you now have an equivalent [PR in client-contrib](https://github.com/knative/client-contrib/pull/6) which is also a better location for this.

/close

Issue #350, opened 2019-08-08
I want to see tags in 'kn route list' output


navidshaikh:
> /kind feature
> /kind proposal

## Describe the feature:
Current `kn route list` output
```
./kn route list               
NAME   URL                                   AGE   CONDITIONS   TRAFFIC
svc    http://svc.default.apps-crc.testing   9h    3 OK / 3     0% -> svc-8cgjh, 50% -> svc-8cgjh, 50% -> svc-dw8kn
```
I want to know, what if my latest revision is receiving any traffic? OR which tag is receiving what portion of traffic?

If we could add tag along with revision name, like
```
./kn route list               
NAME   URL                                   AGE   CONDITIONS   TRAFFIC
svc    http://svc.default.apps-crc.testing   9h    3 OK / 3     0% -> svc-8cgjh (old), 50% -> svc-8cgjh (current), 50% -> svc-dw8kn (stable)
```
revision without tags will be printed without ().

navidshaikh:
The traffic column value is going too long, we might want to consider 
```
./kn route list               
NAME   URL                                   AGE   CONDITIONS   TRAFFIC
svc    http://svc.default.apps-crc.testing   9h    3 OK / 3     (old)svc-8cgjh=0, (current)svc-8cgjh=50, (stable)svc-dw8kn=50
```
replace ` -> ` with `=` and remove `%`

zhanggbj:
@navidshaikh 
Also feeling the traffic column is too long,  and are you working on this? If not, I would like to take it.

zhanggbj:
@navidshaikh 
How about the output in this format `revisionName(tag)=percent`?
```
[zhanggong@192 client (refine_router_list ✗)]$ ./kn route list                                                                           
NAME            URL                                                   AGE     CONDITIONS   TRAFFIC
helloworld      http://helloworld.default.zhanggbj-knative.cloud      4m24s   3 OK / 3     helloworld-kp2pk=50%, helloworld-fw5zj=50%
helloworld-go   http://helloworld-go.default.zhanggbj-knative.cloud   35d     1 OK / 3     helloworld-go-6ttck(current)=100%
```

toVersus:
> replace -> with = and remove %

I agree with Navid's idea to remove % from traffic percentage. WDYH?

navidshaikh:
@zhanggbj :

> How about the output in this format revisionName(tag)=percent?

[...]

PR #407, opened 2019-09-12
Refine route list output

zhanggbj:
Fixes #
[issue 350](https://github.com/knative/client/issues/350)

## Proposed Changes

* Show tag for revisionName if any 
* Make the traffic column shorter

With this PR, output of route list will be
```
$ ./kn route list
NAME          URL                                                                                AGE    CONDITIONS   TRAFFIC
gracecanary   http://gracecanary.default.zhanggbj-knative.us-south.containers.appdomain.cloud   5d1h   3 OK / 3     (current)gracecanary-blue=90, gracecanary-green=10
helloworld    http://helloworld.default.zhanggbj-knative.us-south.containers.appdomain.cloud    27d    3 OK / 3     (old)helloworld-kp2pk=50, (new)helloworld-fw5zj=50
```

**Release Note**

```release-note

```

### DIFF
--- a/pkg/kn/commands/route/human_readable_flags.go
+++ b/pkg/kn/commands/route/human_readable_flags.go
@@ -15,8 +15,6 @@
 package route
 
 import (
-	"fmt"
-
 	metav1beta1 "k8s.io/apimachinery/pkg/apis/meta/v1beta1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"knative.dev/client/pkg/kn/commands"
@@ -29,9 +27,7 @@ func RouteListHandlers(h hprinters.PrintHandler) {
 	kRouteColumnDefinitions := []metav1beta1.TableColumnDefinition{
 		{Name: "Name", Type: "string", Description: "Name of the Knative route.", Priority: 1},
 		{Name: "URL", Type: "string", Description: "URL of the Knative route.", Priority: 1},
-		{Name: "Age", Type: "string", Description: "Age of the Knative route.", Priority: 1},
-		{Name: "Conditions", Type: "string", Description: "Conditions describing statuses of route components.", Priority: 1},
-		{Name: "Traffic", Type: "integer", Description: "Traffic configured for route.", Priority: 1},
+		{Name: "READY", Type: "string", Description: "Ready condition status of the Knative route.", Priority: 1},
 	}
 	h.TableHandler(kRouteColumnDefinitions, printRoute)
 	h.TableHandler(kRouteColumnDefinitions, printKRouteList)
@@ -54,29 +50,13 @@ func printKRouteList(kRouteList *servingv1alpha1.RouteList, options hprinters.Pr
 func printRoute(route *servingv1alpha1.Route, options hprinters.PrintOptions) ([]metav1beta1.TableRow, error) {
 	name := route.Name
 	url := route.Status.URL
-	age := commands.TranslateTimestampSince(route.CreationTimestamp)
-	conditions := commands.ConditionsValue(route.Status.Conditions)
-	traffic := calculateTraffic(route.Status.Traffic)
+	ready := commands.ReadyCondition(route.Status.Conditions)
 	row := metav1beta1.TableRow{
 		Object: runtime.RawExtension{Object: route},
 	}
 	row.Cells = append(row.Cells,
 		name,
 		url,
-		age,
-		conditions,
-		traffic)
+		ready)
 	return []metav1beta1.TableRow{row}, nil
-}
-
-func calculateTraffic(targets []servingv1alpha1.TrafficTarget) string {
-	var traffic string
-	for _, target := range targets {
-		if len(traffic) > 0 {
-			traffic = fmt.Sprintf("%s, %d%% -> %s", traffic, target.Percent, target.RevisionName)
-		} else {
-			traffic = fmt.Sprintf("%d%% -> %s", target.Percent, target.RevisionName)
-		}
-	}
-	return traffic
 }
--- a/pkg/kn/commands/route/list_test.go
+++ b/pkg/kn/commands/route/list_test.go
@@ -67,9 +67,9 @@ func TestRouteListDefaultOutput(t *testing.T) {
 	} else if !action.Matches("list", "routes") {
 		t.Errorf("Bad action %v", action)
 	}
-	assert.Check(t, util.ContainsAll(output[0], "NAME", "URL", "AGE", "CONDITIONS", "TRAFFIC"))
-	assert.Check(t, util.ContainsAll(output[1], "foo", "100% -> foo-01234"))
-	assert.Check(t, util.ContainsAll(output[2], "bar", "100% -> bar-98765"))
+	assert.Check(t, util.ContainsAll(output[0], "NAME", "URL", "READY"))
+	assert.Check(t, util.ContainsAll(output[1], "foo"))
+	assert.Check(t, util.ContainsAll(output[2], "bar"))
 }
 
 func TestRouteListDefaultOutputNoHeaders(t *testing.T) {
@@ -84,9 +84,9 @@ func TestRouteListDefaultOutputNoHeaders(t *testing.T) {
 		t.Errorf("Bad action %v", action)
 	}
 
-	assert.Check(t, util.ContainsNone(output[0], "NAME", "URL", "GENERATION", "AGE", "CONDITIONS", "READY", "REASON"))
-	assert.Check(t, util.ContainsAll(output[0], "foo", "100% -> foo-01234"))
-	assert.Check(t, util.ContainsAll(output[1], "bar", "100% -> bar-98765"))
+	assert.Check(t, util.ContainsNone(output[0], "NAME", "URL", "READY"))
+	assert.Check(t, util.ContainsAll(output[0], "foo"))
+	assert.Check(t, util.ContainsAll(output[1], "bar"))
 
 }
 
@@ -100,8 +100,8 @@ func TestRouteListWithTwoTargetsOutput(t *testing.T) {
 	} else if !action.Matches("list", "routes") {
 		t.Errorf("Bad action %v", action)
 	}
-	assert.Check(t, util.ContainsAll(output[0], "NAME", "URL", "AGE", "CONDITIONS", "TRAFFIC"))
-	assert.Check(t, util.ContainsAll(output[1], "foo", "20% -> foo-01234, 80% -> foo-98765"))
+	assert.Check(t, util.ContainsAll(output[0], "NAME", "URL", "READY"))
+	assert.Check(t, util.ContainsAll(output[1], "foo"))
 }
 
 func createMockRouteMeta(name string) *v1alpha1.Route {
--- a/test/e2e/route_test.go
+++ b/test/e2e/route_test.go
@@ -64,7 +64,7 @@ func (test *e2eTest) routeList(t *testing.T) {
 	out, err := test.kn.RunWithOpts([]string{"route", "list"}, runOpts{})
 	assert.NilError(t, err)
 
-	expectedHeaders := []string{"NAME", "URL", "AGE", "CONDITIONS", "TRAFFIC"}
+	expectedHeaders := []string{"NAME", "URL", "READY"}
 	assert.Check(t, util.ContainsAll(out, expectedHeaders...))
 }
 
@@ -72,8 +72,7 @@ func (test *e2eTest) routeListWithArgument(t *testing.T, routeName string) {
 	out, err := test.kn.RunWithOpts([]string{"route", "list", routeName}, runOpts{})
 	assert.NilError(t, err)
 
-	expectedOutput := fmt.Sprintf("100%% -> %s", routeName)
-	assert.Check(t, util.ContainsAll(out, routeName, expectedOutput))
+	assert.Check(t, util.ContainsAll(out, routeName))
 }
 
 func (test *e2eTest) routeDescribe(t *testing.T, routeName string) {


pkg/kn/commands/route/list_test.go
```diff
@@ -113,26 +116,27 @@ func createMockRouteMeta(name string) *v1alpha1.Route {
 	return route
 }
 
-func createMockTrafficTarget(revision string, percent int) *v1alpha1.TrafficTarget {
+func createMockTrafficTarget(revision string, tag string, percent int) *v1alpha1.TrafficTarget {
```
toVersus: I think it would be better to separate the function for tagging to the traffic target like `createMockTrafficTargetWithTag`.

pkg/kn/commands/route/human_readable_flags.go
```diff
@@ -72,10 +72,14 @@ func printRoute(route *servingv1alpha1.Route, options hprinters.PrintOptions) ([
 func calculateTraffic(targets []servingv1alpha1.TrafficTarget) string {
 	var traffic string
 	for _, target := range targets {
+		var tag string
+		if target.Tag != "" {
+			tag = fmt.Sprintf("(%s)", target.Tag)
+		}
 		if len(traffic) > 0 {
```
toVersus: It's just a matter of taste, but I prefer to remove duplicate effort like this:

```go
		format := fmt.Sprintf("%s%s=%d", tag, target.RevisionName, target.Percent)
		if len(traffic) > 0 {
			traffic += ", " + format
		} else {
			traffic = format
		}
```

zhanggbj: @toVersus 
Refined it in next patch. Thanks!

pkg/kn/commands/route/list_test.go
zhanggbj: That's good, but I'm afraid there will be too many duplicated code in this test file, as the test scenarios are very similar just with/without tag.

navidshaikh:
Are `AGE` and `CONDITIONS` columns useful here ? 
The readiness of a route can be seen in service status as well, can we compact the route list output to show the most relevant info ?

rhuss:
I even question whether we want to have the detailed traffic splitting in a list output. This column can be unbounded and does not nicely fit on a single line in a general case. And if crammed into a single line it easily gets confusing.

IMO this kind of information belongs to `kn route describe` like it is for `kn service describe`, where `kn service list` does not reveal the traffic split (but 'describe' does).

rhuss:
In general I would like to impose the rule not to show array/list-value on an overview view in `kn ... list`. It's really hard to keep that confined to a line's max lengths, without truncating. And if truncating happens the value of the column is very limited. So I would reserve such output for `kn ... describe` where we can easily print out list output via multiple rows (so it can be quite length as the number of lines is not so critical as the length of a line).

Said that we could add something like the *main* revision + the portion of traffic it gets if its not 100%


```
$ ./kn route list
NAME          URL                                                                                AGE    CONDITIONS   REVISION
gracecanary   http://gracecanary.default.zhanggbj-knative.us-south.containers.appdomain.cloud   5d1h   3 OK / 3     gracecanary-blue (90%) helloworld    http://helloworld.default.zhanggbj-knative.us-south.containers.appdomain.cloud    27d    3 OK / 3    helloworld-kp2pk
```

Still not sure whether this is very helpful, but at least the length of the last column is then fixed to a single revision name. The tag name I would reserve as a detail for `kn route describe`.

pkg/kn/commands/route/human_readable_flags.go
```diff
@@ -70,13 +71,14 @@ func printRoute(route *servingv1alpha1.Route, options hprinters.PrintOptions) ([
...
 func calculateTraffic(targets []servingv1alpha1.TrafficTarget) string {
-	var traffic string
+	var traffics []string
 	for _, target := range targets {
-		if len(traffic) > 0 {
-			traffic = fmt.Sprintf("%s, %d%% -> %s", traffic, target.Percent, target.RevisionName)
-		} else {
-			traffic = fmt.Sprintf("%d%% -> %s", target.Percent, target.RevisionName)
+		var tag string
+		if target.Tag != "" {
```
maximilien: I wonder if we should explicitly deal with cases where the API returns empty tag? Perhaps have a token like NOTAG or similar?

rhuss: I don't think that is necessary as an empty tag is not distinguishable from having no tag ("" is the empty string value, too, there is no `nil` string in go).

And when there is no tag there is no need to add something.

However, that tag should be added with a space between revision name and opening parentheses.

maximilien: I disagree. But I can live with it.

zhanggbj:
Vote 1:
a): Keep `AGE` and `CONDITIONS` in route list

zhanggbj:
Vote 2b):
just show "main revision + the portion of traffic it gets if its not 100%" and leave details in `route describe`
```
$ ./kn route list
NAME          URL                                                                               AGE    CONDITIONS   REVISION
gracecanary   http://gracecanary.default.zhanggbj-knative.us-south.containers.appdomain.cloud   5d1h   3 OK / 3     gracecanary-blue (90%)
helloworld    http://helloworld.default.zhanggbj-knative.us-south.containers.appdomain.cloud    27d    3 OK / 3     helloworld-kp2pk
```

sixolet:
Show the most-traffic revision and relabel it "MAIN REVISION" or "MOST TRAFFIC" or something like that.

sixolet:
> **Vote 1b):**
> Remove `AGE` and `CONDITIONS` in route list

Just show READY

navidshaikh:
We might also want to rename the column `NAME` to `SERVICE`, relates better.

rhuss:
> We might also want to rename the column `NAME` to `SERVICE`, relates better.

What if there are service independent routes ? Is it relally that the name of a `Route` always corresponds to a `Service` ?

navidshaikh:
>  Is it relally that the name of a Route always corresponds to a Service ?

AFAIK yes. These routes we list are owned by services and we dont support managing these routes directly (also not [recommended](https://github.com/knative/docs/blob/master/docs/serving/spec/knative-api-specification-1.0.md#route) to update such routes).

navidshaikh:
> just show "main revision + the portion of traffic it gets if it's not 100%" and leave details in route describe

Showing only "main revision" seems like listing incomplete information, we should instead take off the revision-portion information altogether from route list. `route describe` should be a better place to list that info. Since looking at (hard to parse) revision name and its portion isn't that useful unless user gets to see the associated configurations with revision (like image, envs etc) which user sees in describe.
So 3 columns, `SERVICE`, `ROUTE` and `READY`, the details go to `* describe` output.

rhuss:
> AFAIK yes. These routes we list are owned by services and we dont support managing these routes directly (also not [recommended](https://github.com/knative/docs/blob/master/docs/serving/spec/knative-api-specification-1.0.md#route) to update such routes).

> So 3 columns, `SERVICE`, `ROUTE` and `READY`, the details go to `* describe` output.

I just wonder, what is the value-add of a `kn route list` over a `kn service list`  if this is a 1:1 relation and all the info is also present in `kn service list` ?  Also what would `kn route describe` offer over a `kn service describe`'s traffic section ? 

Playing devil's advocate here: Is route management in `kn` required at all ? It's different than for revisions as there is a 1:N relationship. However it's similar to `Configuration` for which also don't offer any dedicated commands.

Also, we follow the convention of `kubectl` that we always start with the `NAME` of the domain object that we list, except when showing for all namespaces, then the namespace is in the first column.

navidshaikh:
> I just wonder, what is the value-add of a kn route list over a kn service list if this is a 1:1 relation and all the info is also present in kn service list ? Also what would kn route describe offer over a kn service describe's traffic section ?

True!
Just the status conditions for service and route differs here (though we aren't listing the reason route isnt `READY` or `REASON` column here).

zhanggbj:
@sixolet @rhuss @navidshaikh
Talking about the difference of `kn service list` and `kn route list`, I think we should provide more details about `routing` in `kn route list`, as route is about the endpoints to different revisions.

In the URL or ROUTE column, we also provide `custom url` if there's a tag to a revision, as developers may use the `custom url` directly to debugging for one specific revision, instead of only print the URL same in `kn service list`

And based on the above comments, it would be something as below, any thoughts? 👍 or 👎 ?
```
# kn route list
NAME          ROUTE                                                                              READY
gracecanary   http://gracecanary.default.zhanggbj-knative.containers.appdomain.cloud             true
              http://current-gracecanary.default.zhanggbj-knative.containers.appdomain.cloud
              http://candidate-gracecanary.default.zhanggbj-knative.containers.appdomain.cloud
helloworld    http://helloworld.default.zhanggbj-knative.us-south.containers.appdomain.cloud     true
```

Compared with `kn service list`
```
# kn service list
NAME          URL                                                                                                                     GENERATION   AGE   CONDITIONS   READY   REASOITIONS   READY   REASON
gracecanary   http://gracecanary.default.zhanggbj-knative.us-south.containers.appdomain.cloud   7            18d   3 OK / 3     True    
helloworld    http://helloworld.default.zhanggbj-knative.us-south.containers.appdomain.cloud    8            40d   3 OK / 3     True    
```

toVersus:
I think `kn <resource> list` should print a list of summary information of resources **in a row**. Something like custom url or detailed traffic splitting are informative so they might be placed in the output of `kn service/route describe`. TBH, it's OK for me to remove the URL column because it is duplicated with the output of `kn service list`.

```
$ ./kn route list
NAME          REVISION                 TAG    READY
gracecanary   gracecanary-blue (90%)   blue   True
helloworld    helloworld-kp2pk                True
```

Otherwise, merge the functionality of `kn route list` into `kn service describe`.

rhuss:
> Talking about the difference of `kn service list` and `kn route list`, I think we should provide more details about `routing` in `kn route list`, as route is about the endpoints to different revisions.

The question is, whether you need to see that URL details on the list level. My feeling is that people start with a specific service in mind and want to see it traffics split (which she can do via `kn service describe`). 

What actually is the use case to see all URLs across servives ? Maybe to find out the service for a given URL ?

Also, a list should not have multi-line output per entry (because that is very hard to script against). That's what describe is for IMO.

rhuss:
> TBH, it's OK for me to remove the URL column because it is duplicated with the output of `kn service list`.

What would be the value add for `kn route list` then ? Especially as routes are typically associated with URLs, not revision names.

rhuss:
Maybe we can list a set of use cases for `kn route list` and `kn route describe` which doesn't fit to `kn service list` or  `kn service describe`. We can then decide whether its worth to introduce a high-level command like 'kn route' or whether we can find other options. 

To be clear, *if* we would have a `kn route create` for creating routes independently from services, then listing of routes makes total sense. But if this is supposed to be **always a 1:1** relation I don't see the need for an additional command (everything could be covered by the `service` command, too).

Here is a use case, I could see for `kn route list`.

* As a user, for a given URL I want to find out which service is responsible for.
  Solution: `kn route list | grep $url`, but maybe also `kn service describe --url $url` ? (if this actually is an importan use case to cover. not sure about this)

For `kn route describe` I don't see anything which is not already covered by `kn service describe`.

Any other use case you can imaging for `kn route` commands ?

rhuss:
@zhanggbj @toVersus @navidshaikh @sixolet I put the question whether we need `kn route list` and `kn route describe` at all on tomorrows WG agenda. 

@zhanggbj sorry, I know it's very late for you (probably too late). We can also chat about it tomorrow on slack #cli (knative.slack.com). I will be there starting from 8am CEST.

zhanggbj:
@rhuss @toVersus 
I think for basic scenarios, Service has a 1:1 relation with Route, so we didn't see much value of output `kn route list`, but IMHO in some advanced scenarios as below, its **not always 1:1 relation**, so I think it's better to keep `kn route list` but refine the output to print more `valuable and route related info`.

**1)  Use custom url to access one Revision directly**
As described in https://github.com/knative/client/pull/407#issuecomment-536548161

**2)  Routing across multiple Knative services**
Details: https://knative.dev/docs/serving/samples/knative-routing-go/
In short, we can setup routing for different services using domain url as below

knative-routing-sample-flow


So in `kn route list` we may have
```
# kn route list
NAME          ROUTE                                                                              READY
gracecanary   http://gracecanary.default.zhanggbj-knative.containers.appdomain.cloud             true
gracecanary   http://zhanggbj-knative.containers.appdomain.cloud/login                           true
helloworld    http://helloworld.default.zhanggbj-knative.us-south.containers.appdomain.cloud     true
helloworld    http://zhanggbj-knative.us-south.containers.appdomain.cloud/test                   true
```

**3)  http-based routing for different Revisions of one Service**
For now, Serving only supports routing based on weighted traffic, but I think in production users may have various routing requirements based on the incoming request, for e.g. based on path/user indentity/GET or Post request, So it's valuable for `kn route list` to show complex route rules to different Revisions of one service. Detailed discussion in https://github.com/knative/serving/issues/4736, Knative Serving may support it after 1.0. 

So in `kn route list` we may have
```
# kn route list
NAME          ROUTE                                                                                       REVISION                READY
gracecanary   http://gracecanary.default.zhanggbj-knative.containers.appdomain.cloud                      current-gracecanary     true
gracecanary   http://gracecanary.default.zhanggbj-knative.containers.appdomain.cloud Header=enduser:json  current-gracecanary     true
gracecanary   http://gracecanary.default.zhanggbj-knative.containers.appdomain.cloud Header=enduser:mike  candidate-gracecanary   true       
```

@rhuss It's great to discuss in WG on this, I may not able to attend it :-) And yes, we can discuss later on Slack. Thanks!

rhuss:
@zhanggbj thanks a lot for your detailed exploration of possible route use cases. My comments on them following your enumeration above:

1. I see the point but you get the very same information also with `kn service describe gracecanary` and `kn service helloworld`, even much more important information like to which revision and especially to which image is an URL pointing to and how the traffic is split. The question I really had is important to see all URL from all Service at once (like in `kn route list`) ? My feeling this case 1. is already nicely covered with the `kn service` commands.

2. I don't think that this use case applies to `Route` resources at all. It's about how to configure an Istio `VirtualService` as in this [routing.yaml](https://github.com/knative/docs/blob/master/docs/serving/samples/knative-routing-go/routing.yaml). So, the gateway would not be managed by `kn` anyway (but by Istio or generic Kubernetes tooling).

3. Header based routing might a future use case for kn when it is available in Serving. It could be managed equally well with `kn service` commands like the traffics split. But more importantly, this is something in the future and its not clear to me how Serving is going to implement that. IMO we should wait with a client implementation (or user visible preparation like adding new commands) until this feature has landed server side.

rhuss:
I still think that the decision whether to implement `kn routes list` boils down to two questions:

* Has the collective displays of all routes a substantial value which justifies the introduction of a new top-level noun (routes)?
* Do we want to allow to manage routes independently from Services, breaking the 1:1 relationship? If so, then we should add also `create` and `update` to `kn route` to make it complete.

zhanggbj:
Hi All,
Thanks for all the discussions, based on above comments, I think another alternative option is we still keep the `route list` for now but with only NAME/URL/READY as below. And leave the desition to completely remove route list until we have more users and more scenarios, if they think `route list` is not meaningful, we can remove it completely then. Thx!
```
$ ./kn route list   
NAME                URL                                                                                                                           READY
gracecanary         http://gracecanary.default.zhanggbj-knative.us-south.containers.appdomain.cloud         True
helloworld          http://helloworld.default.zhanggbj-knative.us-south.containers.appdomain.cloud          True
```

PR #411, opened 2019-09-20
Fixes issue #338 by adding documentation for the Kn config

### LINKED ISSUE #338: Add docs about the ~/.kn/config config file
## In what area(s)?
/kind good-first-issue
/kind doc

https://github.com/knative/client/blob/master/docs/README.md mentions the `kn.yaml`
config file but there are no docs about it. We should document:
- where it needs to live
- it's format (yaml obviously)
- what are the available config flags people can set

maximilien:
Fixes #338

## Proposed Changes

* documentation for Kn config
* details on Kn plugins configuration parameters

Release Note:

- 🎁 docs for Kn config


docs/README.md
```diff
@@ -34,6 +34,55 @@ You'll need a `kubectl`-style config file to connect to your cluster.
 `$HOME/.kube/config`. You can specify an alternate kubeconfig connection file
 with `--kubeconfig`, or the env var `$KUBECONFIG`, for any command.
 
+## Kn Config
+
+To better customize Kn there is a small set of configuration parameters you are able to setup. In particular you can specify where your Kn plugins are located and how they are found. The Kn configuration file is meant to capture these limited configuration options. Let's explore this file, it's location, and the current options you are able to change with it.
```
rhuss: "limited" compared to what ? 

'would avoid any size qualification as it is (a) not important and might soon be not true anymore. Also its not really important.

docs/README.md
```diff
@@ -34,6 +34,55 @@ You'll need a `kubectl`-style config file to connect to your cluster.
...
+
+### Location
+
+By default the Kn config file is located in the user's home directory under `.kn` hidden subdirectory and is named `config.yaml`. So in `~/.kn/config.yaml`. It is not created for you by default. You can easily create one with:
+
+```bash
+mkdir ~/.kn
+touch ~/.kn/config.yaml
```
rhuss: Why would you ever do this ? Wouldn't you just create one with content when needed ?

We could also create one by default if not existing, with some commented out sample configuration. That is what many other tools do as well and is the best way to ramp up.

docs/README.md
```diff
@@ -34,6 +34,55 @@ You'll need a `kubectl`-style config file to connect to your cluster.
 `$HOME/.kube/config`. You can specify an alternate kubeconfig connection file
 with `--kubeconfig`, or the env var `$KUBECONFIG`, for any command.
 
+## Kn Config
+
+To better customize Kn there is a small set of configuration parameters you are able to setup. In particular you can specify where your Kn plugins are located and how they are found. The Kn configuration file is meant to capture these limited configuration options. Let's explore this file, it's location, and the current options you are able to change with it.
+
+### Location
+
+By default the Kn config file is located in the user's home directory under `.kn` hidden subdirectory and is named `config.yaml`. So in `~/.kn/config.yaml`. It is not created for you by default. You can easily create one with:
```
rhuss: Can we use "`kn`" instead of "Kn" as this is the name how it should be typed in the CLI ? AFAIK nowhere we refer to kn as Kn so far.

docs/README.md
```diff
@@ -34,6 +34,55 @@ You'll need a `kubectl`-style config file to connect to your cluster.
...
+./kn plugin list
+Using kn config file: /home/maximilien/.kn/config.yaml
+No plugins found.
+```
+
+To see the values in the Kn config file you can do:
+
+```bash
+cat ~/.kn/config.yaml 
+lookuppluginsinpath: false
```
rhuss: isn't it in camel case ? (should be imo)

docs/README.md
```diff
@@ -34,6 +34,55 @@ You'll need a `kubectl`-style config file to connect to your cluster.
...
+pluginsdir: ~/.kn/plugins
```
rhuss: camel case ?

docs/README.md
```diff
@@ -34,6 +34,55 @@ You'll need a `kubectl`-style config file to connect to your cluster.
...
+```
+
+You can now modify these values by editing the file. To specify a different config file then use the `--config` persistent flag when invoking Kn.
```
rhuss: only now ?

docs/README.md
```diff
@@ -34,6 +34,55 @@ You'll need a `kubectl`-style config file to connect to your cluster.
...
+
+### Options
+
+Currently there are two options you can specific in the Kn config file and they are related to how Kn locates plugins.
```
rhuss: remove "currently" please (or any other temporal expression as it is supposed to be a reference guide which always describes the current state of certain version)

docs/README.md
```diff
@@ -34,6 +34,55 @@ You'll need a `kubectl`-style config file to connect to your cluster.
...
+
+1. `pluginsDir` or `pluginsdir` which is the same as the persistent flag `--plugins-dir` and  specifies the Kn plugins directory. It defaults to: `~/.kn/plugins`. By using the persistent flag (when you issue a command) or by specifying the value in the Kn config, a user can selects which directory to find Kn plugins. It can be any directory that is visible to the user.
```
rhuss: I always would teach to use the camelcase varian (`pluginsDir` here) that is might work all lowercase, too (and maybe even all UPPERCASE ?) is not worth mentioning imo.

docs/README.md
```diff
@@ -34,6 +34,55 @@ You'll need a `kubectl`-style config file to connect to your cluster.
...
+
+For example, the following Kn config will look for Kn plugins in the user's `PATH`.
+
+```bash
+cat ~/.kn/config.yaml
+lookupPluginsInPath: true
+pluginsdir: ~/.kn/plugins
+```
+
+Note that in the configuration above, you need to add `~/.kn/plugins` directory in your `PATH` variable if you want plugins located in that directory to be seen and be usable with Kn.
```
rhuss: Is this really true ? I read the configuration above as lookup in `~/.kn/plugins` *AND* in the PATH. So no need to put `~/.kn/plugins` also into the path. At least that is what I would suppose as a user (because otherwise there is no need to add `pluginsDir` at all).

docs/README.md
maximilien: Either way, if you do above and run Kn you will have the default settings. I was documenting current behavior. I am OK to auto create if not present everytime we run Kn.

docs/README.md
maximilien: I use `kn` when referring to the binary and Kn when referring to the client. Are you suggesting to use `kn` everywhere?

docs/README.md
maximilien: Not in the file. The way Viper works is to make it case insensitive--as far as I can tell. Not my first choice.

I've been meaning to open an issue on this and investigate if we can do something about it. I agree it would be better if if was case sensitive and camel cased in the config file.

docs/README.md
maximilien: See above

docs/README.md
maximilien: I'll remove "now" not needed.

docs/README.md
maximilien: Yeah, that's fair. Hopefully we can enforce this

docs/README.md
maximilien: I don't think so. Will verify.

docs/README.md
maximilien: See issue #414

docs/README.md
rhuss: I'm fine also with just kn (instead of `kn`)

docs/README.md
maximilien: Resolving this. We can open an issue to add the default config as needed and add a line saying that the config file will be created. Either way this current behavior would be correct.

docs/README.md
maximilien: OK changing to kn

docs/README.md
maximilien: Submitting this as is and when #414 is resolved I'll make necessary changes that matches resolution. Hopefully camel case.

docs/README.md
maximilien: Actually correct. It does find and execute plugins in `~/.kn/plugins` and I think behavior I am seeing is based on issue #417. My Linux system is running Ubuntu but I doubt that's the issue. I need to investigate what's going on. But will do that when I address #417. 

Fixing now since the behavior you mention is ideal.

maximilien: Also the overall behavior is broken (independently of #417) since on a Mac OS X system I am seeing that if I have a `~/.kn/config.yaml` it does get populated but if I change `lookupPluginsInPath: true` and combination of `pluginsDir` I am not getting the right thing. 

I need to get add some integration / e2e tests for plugins and get to the bottom of this. 

OK updating this PR but lots to get done to have this doc be what the code is doing. Feel free to lgtm if all otherwise look good since it reflects what we want.

docs/README.md
```diff
@@ -34,6 +34,54 @@ You'll need a `kubectl`-style config file to connect to your cluster.
...
+cat ~/.kn/config.yaml 
+lookupPluginsInPath: false
+pluginsDir: ~/.kn/plugins
+```
+
+You can modify these values by editing the file. To specify a different config file then use the `--config` persistent flag when invoking kn.
+
+### Options
+
+There are two options you can specific in the kn config file and they are related to how kn locates plugins.
```
navidshaikh: *specify

docs/README.md
```diff
@@ -34,6 +34,54 @@ You'll need a `kubectl`-style config file to connect to your cluster.
...
+
+1. `pluginsDir` which is the same as the persistent flag `--plugins-dir` and  specifies the kn plugins directory. It defaults to: `~/.kn/plugins`. By using the persistent flag (when you issue a command) or by specifying the value in the kn config, a user can selects which directory to find kn plugins. It can be any directory that is visible to the user.
```
navidshaikh: extra space between `and  specifies`

navidshaikh: *a use can select which directory

docs/README.md
```diff
@@ -34,6 +34,54 @@ You'll need a `kubectl`-style config file to connect to your cluster.
...
+
+2. `lookupPluginsInPath` which is the same as the persistent flag `--lookup-plugins-in-path` and specficies if kn should look for plugins anywhere in the specified `PATH` shell variable. The value for this configuration parameter is `true` or `false`.
```
navidshaikh: *in the specified `PATH` environment variable ?

navidshaikh: *This is a boolean configuration option and the default value is `false`.

docs/README.md
```diff
@@ -34,6 +34,54 @@ You'll need a `kubectl`-style config file to connect to your cluster.
 `$HOME/.kube/config`. You can specify an alternate kubeconfig connection file
 with `--kubeconfig`, or the env var `$KUBECONFIG`, for any command.
 
+## Kn Config
+
+To better customize kn there is a set of configuration parameters you are able to setup and customize. In particular you can specify where your kn plugins are located and how they are found. The kn configuration file is meant to capture these configuration options. Let's explore this file, it's location, and the options you are able to change with it.
+
+### Location
+
+By default the kn config file is located in the user's home directory under `.kn` hidden subdirectory and is named `config.yaml`. So in `~/.kn/config.yaml`. It is not created for you by default. You can easily create one with:
```
navidshaikh: The default location kn looks for config is under the home directory of user at `$HOME/.kn/config.yaml`. It is not created for you part of Kn installation. You can create this file elsewhere and use `--config` flag to specify its path.

navidshaikh: I'd just mention the location description and let `options` section describe the contents of it, the rest of the below contents in `### Location` section can be removed IMO.

maximilien: OK I simplified it. Kept the location but reduced. See latest version if it reads better.

PR #419, opened 2019-09-27
Update the description of flag namespace

### LINKED ISSUE #418: The description of flag `--namespace` shall be updated
## In what area(s)?

/kind good-first-issue

latest

0.8.x

## Expected Behavior

When I try `kn service create --help`, the flag `--namespace` description shall be something like `the namespace scope for this CLI request`.

## Actual Behavior

The `--namespace` description is `List the requested object(s) in given namespace.` even in the create command.

Run command `kn service create --help`
```
$ kn service create --help
Create a service.

Usage:
  kn service create NAME --image IMAGE [flags]

Examples:
  ...

Flags:
  ...
  -n, --namespace string         List the requested object(s) in
                                 given namespace.
```

daisy-ycguo:
Fixes #418

## Proposed Changes

* Change the description of flag `namespace` because it will serve for other actions besides `list`, e.g. create, update and etc.

### DIFF
--- a/pkg/kn/commands/namespaced.go
+++ b/pkg/kn/commands/namespaced.go
@@ -28,7 +28,7 @@ func AddNamespaceFlags(flags *pflag.FlagSet, allowAll bool) {
 		"namespace",
 		"n",
 		"",
-		"List the requested object(s) in given namespace.",
+		"Specify the namespace to operate in.",
 	)
 
 	if allowAll {
--- a/docs/cmd/kn_revision_delete.md
+++ b/docs/cmd/kn_revision_delete.md
@@ -22,7 +22,7 @@ kn revision delete NAME [flags]
 
 ```
   -h, --help               help for delete
-  -n, --namespace string   List the requested object(s) in given namespace.
+  -n, --namespace string   Specify the namespace to operate in.
 ```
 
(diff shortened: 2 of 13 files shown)


pkg/kn/commands/namespaced.go
```diff
@@ -28,7 +28,7 @@ func AddNamespaceFlags(flags *pflag.FlagSet, allowAll bool) {
 		"namespace",
 		"n",
 		"",
-		"List the requested object(s) in given namespace.",
+		"If present, set the namespace scope for this CLI request",
```
rhuss: Suggestion: "Specify the namespace to operate in" (as it is self-evident that the option only has an influence 'if present'  and 'this CLI request' is so generic that it can be left out)

navidshaikh: +1 to suggestion above + nit: add a dot (.) at the end of the statement to align with other flags' descriptions.
I see, kubectl is using this help text
```
 kubectl options | grep namespace
  -n, --namespace='': If present, the namespace scope for this CLI request
```

daisy-ycguo: Fixed in the new version.

PR #430, opened 2019-10-04
Refactor describe: printing levels and shared kn code

sixolet:
This is the first PR in a chain that will implement revision describe. It moves a lot of logic that wants to be shared by describe commands into:
* The printer, to keep track of indentation level and columns (so it needn't be done statically), and write attributes
* A new file in the shared commands package, which has useful methods to write metadata and do some other Kn-specific formatting.

### DIFF
--- a/pkg/kn/commands/describe.go
+++ b/pkg/kn/commands/describe.go
@@ -1,0 +1,198 @@
+// Copyright © 2019 The Knative Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package commands
+
+import (
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/duration"
+	"knative.dev/client/pkg/printers"
+	"knative.dev/pkg/apis"
+)
+
+// Max length When to truncate long strings (when not "all" mode switched on)
+const TruncateAt = 100
+
+func WriteMetadata(dw printers.PrefixWriter, m *metav1.ObjectMeta, printDetails bool) {
+	dw.WriteAttribute("Name", m.Name)
+	dw.WriteAttribute("Namespace", m.Namespace)
+	WriteMapDesc(dw, m.Labels, l("Labels"), "", printDetails)
+	WriteMapDesc(dw, m.Annotations, l("Annotations"), "", printDetails)
+	dw.WriteAttribute("Age", Age(m.CreationTimestamp.Time))
+
+}
+
+var boringDomains = map[string]bool{
+	"serving.knative.dev":   true,
+	"client.knative.dev":    true,
+	"kubectl.kubernetes.io": true,
+}
+
+// Write a map either compact in a single line (possibly truncated) or, if printDetails is set,
+// over multiple line, one line per key-value pair. The output is sorted by keys.
+func WriteMapDesc(dw printers.PrefixWriter, m map[string]string, label string, labelPrefix string, details bool) {
+	if len(m) == 0 {
+		return
+	}
+
+	var keys []string
+	for k := range m {
+		parts := strings.Split(k, "/")
+		if details || len(parts) <= 1 || !boringDomains[parts[0]] {
+			keys = append(keys, k)
+		}
+	}
+	if len(keys) == 0 {
+		return
+	}
+	sort.Strings(keys)
+
+	if details {
+		l := labelPrefix + label
+
+		for _, key := range keys {
+			dw.WriteColsLn(l, key+"="+m[key])
+			l = labelPrefix
+		}
+		return
+	}
+
+	dw.WriteColsLn(label, joinAndTruncate(keys, m))
+}
+
+func Age(t time.Time) string {
+	if t.IsZero() {
+		return ""
+	}
+	return duration.ShortHumanDuration(time.Now().Sub(t))
+}
+
+// Color the type of the conditions
+func formatConditionType(condition apis.Condition) string {
+	return string(condition.Type)
+}
+
+// Status in ASCII format
+func formatStatus(c apis.Condition) string {
+	switch c.Status {
+	case corev1.ConditionTrue:
+		return "++"
+	case corev1.ConditionFalse:
+		switch c.Severity {
+		case apis.ConditionSeverityError:
+			return "!!"
+		case apis.ConditionSeverityWarning:
+			return " W"
+		case apis.ConditionSeverityInfo:
+			return " I"
+		default:
+			return " !"
+		}
+	default:
+		return "??"
+	}
+}
+
+// Used for conditions table to do own formatting for the table,
+// as the tabbed writer doesn't work nicely with colors
+func getMaxTypeLen(conditions []apis.Condition) int {
+	max := 0
+	for _, condition := range conditions {
+		if len(condition.Type) > max {
+			max = len(condition.Type)
+		}
+	}
+	return max
+}
+
+// Sort conditions: Ready first, followed by error, then Warning, then Info
+func sortConditions(conditions []apis.Condition) []apis.Condition {
+	// Don't change the orig slice
+	ret := make([]apis.Condition, 0, len(conditions))
+	for _, c := range conditions {
+		ret = append(ret, c)
+	}
+	sort.SliceStable(ret, func(i, j int) bool {
+		ic := &ret[i]
+		jc := &ret[j]
+		// Ready first
+		if ic.Type == apis.ConditionReady {
+			return jc.Type != apis.ConditionReady
+		}
+		// Among conditions of the same Severity, sort by Type
+		if ic.Severity == jc.Severity {
+			return ic.Type < jc.Type
+		}
+		// Error < Warning < Info
+		switch ic.Severity {
+		case apis.ConditionSeverityError:
+			return true
+		case apis.ConditionSeverityWarning:
+			return jc.Severity == apis.ConditionSeverityInfo
+		case apis.ConditionSeverityInfo:
+			return false
+		default:
+			return false
+		}
+		return false
+	})
+	return ret
+}
+
+// Print out a table with conditions. Use green for 'ok', and red for 'nok' if color is enabled
+func WriteConditions(dw printers.PrefixWriter, conditions []apis.Condition, printMessage bool) {
+	section := dw.WriteAttribute("Conditions", "")
+	conditions = sortConditions(conditions)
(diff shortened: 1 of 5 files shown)


pkg/kn/commands/describe.go
```diff
@@ -0,0 +1,198 @@
...
+// Max length When to truncate long strings (when not "all" mode switched on)
+const TruncateAt = 100
+
+func WriteMetadata(dw printers.PrefixWriter, m *metav1.ObjectMeta, printDetails bool) {
+	dw.WriteAttribute("Name", m.Name)
+	dw.WriteAttribute("Namespace", m.Namespace)
+	WriteMapDesc(dw, m.Labels, l("Labels"), "", printDetails)
+	WriteMapDesc(dw, m.Annotations, l("Annotations"), "", printDetails)
+	dw.WriteAttribute("Age", Age(m.CreationTimestamp.Time))
+
```
maximilien: *new line

pkg/kn/commands/describe.go
```diff
@@ -0,0 +1,198 @@
...
+// over multiple line, one line per key-value pair. The output is sorted by keys.
+func WriteMapDesc(dw printers.PrefixWriter, m map[string]string, label string, labelPrefix string, details bool) {
+	if len(m) == 0 {
+		return
+	}
+
+	var keys []string
+	for k := range m {
+		parts := strings.Split(k, "/")
+		if details || len(parts) <= 1 || !boringDomains[parts[0]] {
```
maximilien: I wonder if this condition could be simplified or split into two to help legibility?

pkg/kn/commands/describe.go
```diff
@@ -0,0 +1,198 @@
...
+			keys = append(keys, k)
+		}
+	}
```
maximilien: Maybe all of this is moved to a helper that searches for the keys?

pkg/kn/commands/describe.go
```diff
@@ -0,0 +1,198 @@
...
+			dw.WriteColsLn(l, key+"="+m[key])
+			l = labelPrefix
+		}
+		return
+	}
+
+	dw.WriteColsLn(label, joinAndTruncate(keys, m))
+}
+
+func Age(t time.Time) string {
```
maximilien: why are we exporting this? Seems like it's helper private function. Or you intend to use elsewhere should it be moved to a utils or helper file?

pkg/kn/commands/describe.go
```diff
@@ -0,0 +1,198 @@
...
+			return false
+		default:
+			return false
+		}
+		return false
+	})
+	return ret
+}
+
+// Print out a table with conditions. Use green for 'ok', and red for 'nok' if color is enabled
```
maximilien: I love the idea of using colors but don't see where that's being applied here. Unless in the printer? I was expecting you would import `github.com/fatih/color` or similar?

sixolet: This is preserving something Roland planned for later use. I can delete the comment.

pkg/kn/commands/describe.go
sixolet: Helper to decide if key is boring.

pkg/kn/commands/describe.go
sixolet: I think it's now reasonably simple and clear.

pkg/kn/commands/describe.go
sixolet: It gets used elsewhere later, I think.

pkg/kn/commands/describe.go
```diff
@@ -0,0 +1,203 @@
...
+	case corev1.ConditionTrue:
+		return "++"
+	case corev1.ConditionFalse:
+		switch c.Severity {
+		case apis.ConditionSeverityError:
+			return "!!"
+		case apis.ConditionSeverityWarning:
+			return " W"
+		case apis.ConditionSeverityInfo:
+			return " I"
```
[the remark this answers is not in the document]
rhuss: As alternative compromise I would suggest to at least use the first char to indicat falseness, 

i.e.

```
!! == False, Error   (alt: !E)
!I == False, Info    (alt: !+)
!W == False, Warning (alt: !~)
```

pkg/kn/commands/describe.go
```diff
@@ -0,0 +1,203 @@
...
+	"knative.dev/pkg/apis"
+)
+
+// Max length When to truncate long strings (when not "all" mode switched on)
+const TruncateAt = 100
+
+func WriteMetadata(dw printers.PrefixWriter, m *metav1.ObjectMeta, printDetails bool) {
+	dw.WriteAttribute("Name", m.Name)
+	dw.WriteAttribute("Namespace", m.Namespace)
+	WriteMapDesc(dw, m.Labels, l("Labels"), "", printDetails)
```
rhuss: I suggest to move this function to `dw`, too (and the remove the `l(...)`).

At least, `l(...)` should be replaced by `printers.Label(...)` (as the whole purpose of this function is to put the label mangling at a single place so that it's easier to add a different label handling (like adding colors).

pkg/kn/commands/describe.go
```diff
@@ -0,0 +1,203 @@
...
+	WriteMapDesc(dw, m.Annotations, l("Annotations"), "", printDetails)
+	dw.WriteAttribute("Age", Age(m.CreationTimestamp.Time))
+}
+
+var boringDomains = map[string]bool{
+	"serving.knative.dev":   true,
+	"client.knative.dev":    true,
+	"kubectl.kubernetes.io": true,
```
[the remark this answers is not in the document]
sixolet: Let's, but in a PR that isn't focused on refactoring.

pkg/kn/commands/describe.go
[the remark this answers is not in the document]
sixolet: `??` is "unknown" right now.

How about add a couple columns and we use the strings `ok` (True), `pend` (Unknown), `err`, `warn` and `info`?

pkg/kn/commands/describe.go
```diff
@@ -0,0 +1,203 @@
...
+func formatStatus(c apis.Condition) string {
+	switch c.Status {
+	case corev1.ConditionTrue:
+		return "++"
+	case corev1.ConditionFalse:
+		switch c.Severity {
+		case apis.ConditionSeverityError:
+			return "!!"
+		case apis.ConditionSeverityWarning:
+			return " W"
```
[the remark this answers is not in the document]
sixolet: Right?

pkg/kn/commands/describe.go
sixolet: Wait for the next PR...

pkg/kn/commands/describe.go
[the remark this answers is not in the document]
rhuss: Yes, that would be ok for me, **if** the combinations `(status, severity) == (true, warn), (true, info), (true, error) ` are never possible. Tbh, I would expect `(true, info)` to be more likely than a `(false, info)`. Also, `(true, warn)` and `(false, warn)` bith make sense for me, too. Are the possible combination specified anywhere ?

To be on the sure side and make it easier to map to the status without any client-side logic, I still opt for two columns, one for status and one for severity.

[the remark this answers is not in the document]
rhuss: Or to put it he other way round: If we print only "info" is it clear to the user that the status is "false" and the e.g. the service is not ready (if this is for the "Ready" condition) ? I don't think so.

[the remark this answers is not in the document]
rhuss: Reading the spec at https://github.com/knative/docs/blob/master/docs/serving/spec/knative-api-specification-1.0.md#error-signalling I think we should be fine with "True", "False", "Info", "Warn" as four levels to print (or as "+++", "!!!", "???", "INF", "WRN" as below) as "Info" and "Warn" are supposed to be _additional_ info only without any other semantic meaning (e.g. how to influence the `Ready` condition). Severity for `Ready` condition is always "", and for other conditions only the status of conditions with a severity "" have a meaning.

So in pseudo-code I would change the handling to follow the spec more closely (which discriminates on severity = "" first) :

```

if severity == ""
  if status == "True"
     return "+++"
  else if status == "False"
     return "!!!"
  else
     return "???"
  fi    
else if severity ==" Info"
   return "INF"
else 
   return "WRN"
fi
```

as it's really that for any semantic meaning (true/false/unknown) the severity has to be empty (""), and only for additional info the severity is set.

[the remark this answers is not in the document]
rhuss: Using "INF" and "WRN" (severity) as acronyms and "+++", "!!!" and "???" (status) as symbols makes for a visual distinction between additional info, and status fields.

rhuss:
How do we want to proceed with this PR ? Would love to merge it soon, and I wouldn't mind that we merge it now and do a refinement later. Maybe its easier to make improvements suggestions via code.

@sixolet @maximilien @navidshaikh wdyt ?

navidshaikh:
> Maybe its easier to make improvements suggestions via code.

+1

sixolet:
@rhuss I think it'll be best to merge this for now, and adjust a bit later — We can work through details on a call, once you can join calls again.

Next comment is more detailed opinion about displaying conditions.

sixolet:
For example, I want to be careful about saying warning conditions should say Warning when True — reading the spec the way I do (and I asked for the spec this way), a False indicates that there's something to warn about, and a True indicates that something is fine (a fine thing that would be a warning if it were False)

Example: Let's say we want to base a condition on OOM kills. The normative case is "not very many", so we'll call the condition StableMemory, and StableMemory: True means few OOM kills, and StableMemory: False means lots. Now, a service can still function in the presence of OOM kills, probably, so this condition is a WARN

When True:
✔️ StableMemory

When False:
WARN StableMemory 6 pods died from running out of memory in the last hour

We don't want this when the condition is true indicating a lack of a problem:
WARN StableMemory

sixolet:
In general, *all* conditions should be designed to mean that True is Stable, Good, Operational, Active, or any of those "happy" things. False is Unstable, Crashing, Bad, Inactive, etc. 

Now, "not active" might not be a problem, in which case it's Info, or it might be a minor problem, in which case Warning, or a major problem, in which case "", but either way it'll have a meaningful True/False status.

rhuss:
Ok, I see your use case. But this is then even more an argument to separate severity (Error/Warn/Info) and status (True/False/Unknown) into two separate fields/columns (otherwise you would have to use at max 9 different symbol (but some combinations are not really possible). 

IIn the code I still think we should make the conditional in that order:

* Is the condition "Ready" or "Suceeded" ?
* Is it an error severity (severity == "") for non "Ready" conditions ?
* Is is an info/warning severity for non "Ready" condition ?

This directly reflects also the way how the decision tree in https://github.com/knative/docs/blob/master/docs/serving/spec/knative-api-specification-1.0.md#error-signalling is built up and makes it easier to verify the way how we present the status.

rhuss:
@sixolet there is one conflict remaining. Please resolve it and then we can merge the PR.

sixolet:
@rhuss Conflict resolved.

sixolet:
Re: the flow chart, the point I was making is that for True, it doesn't matter what the severity is — the user doesn't need to see "it's fine and it would be a warning if it weren't" as any different from "it's fine and it'd be an error if it weren't", but "it's not fine and it's a warning" is different from "it's not fine and it's an error"

And specifically, any time you say "warning" the user will feel warned. So if you say "it's fine, but warning (if it weren't fine)" the user will hear "oh dang, WARNING"

rhuss:
@sixolet ok, agree that a warning should be only shown when `False` (when I would do some nitpicking I naively would think, that a 'this message with severity warning is False'  means no-warning (and warning, true == warning active. But I don't and understand that True == good, False == bad overrules here ;-)

/lgtm

Issue #423, opened 2019-10-01
latest release can't deploy service to v0.9 cluster


mchmarny:
## What version of Knative Client?

v0.2.0

## What version of Knative Serving running on your cluster?

0.9

## Expected Behavior

```shell
kn service create prime \
  --image gcr.io/cloudylabs-public/maxprime \
  --namespace demo
```

results in service deployed 

## Actual Behavior

Service 'prime' successfully created in namespace 'demo'.
Waiting for service 'prime' to become ready ...

With kn waiting for the service to become available forever 

## Steps to Reproduce the Problem

run the following command against v0.9 Knative Service cluster  

```shell
kn service create prime \
  --image gcr.io/cloudylabs-public/maxprime \
  --namespace demo
```

rhuss:
> With kn waiting for the service to become available forever

That's interesting as there is supposed to be a timeout after 60s (which actually works and is independent on any Knative serving version). When you say 'forever' how long is this :) ?

After you stopped the command, has a Knative service be created (e.g. `kubectl get ksvc`). If so, could you check the status section of the resource (`kubectl get ksvc prime -o yaml`) and post it here ?

If a Knative service has been created, can you use `kn service list` and `kn service describe prime`) ?

Also, could you please try to recreate the service with the option `--log-http` ?

drnic:
FWIW - latest nightly `kn` worked for me against knative-service 0.9.0/istio 1.3.0.

drnic:
I get the issue sometimes:

```
$ kn service create sample-app-nodejs --image starkandwayne/sample-app-nodejs:latest --force --log-http
===== REQUEST =====
GET /apis/serving.knative.dev/v1alpha1/namespaces/default/services/sample-app-nodejs HTTP/1.1
Host: 35.189.6.125
User-Agent: kn/v0.0.0 (darwin/amd64) kubernetes/$Format
Accept: application/json, */*
Authorization: *******************************************************************************************************************************************************
Accept-Encoding: gzip


===== RESPONSE =====
HTTP/2.0 200 OK
Content-Length: 1565
Audit-Id: 2ac204e2-c72f-4fd3-a562-29ba63aa5133
Content-Type: application/json
Date: Wed, 02 Oct 2019 05:10:02 GMT

[...]

Issue #290, opened 2019-07-24
Print the revision name after service create and update commands


navidshaikh:
> /kind feature

## Describe the feature:

When a user updates a service, we print the success message
```
➜ kn service update bluegreen --env TARGET=green
Waiting for service 'bluegreen' to become ready ... OK
Service 'bluegreen' updated in namespace 'default'.
```

however, we should also print the revision name generated after this update, which can be used to refer that revision for further operations, eg: traffic splitting. This can also be used for scripting use cases.

Update:
Implement same behavior for `service create` command as well.

duglin:
they should also be allowed to specify the new traffic split on the update cmd

navidshaikh:
> they should also be allowed to specify the new traffic split on the update cmd

sorry I didn't get that, could you please add some more detail  ?

duglin:
Sorry I really should have put that comment on the other issue - see: https://github.com/knative/client/pull/285#issuecomment-514602694

rhuss:
> Implement same behavior for `service create` command as well.

For `kn service create` we print the URL at the end. The service name does not make sense, as its provided as argument.

With having the URL on a single line as last argument you can do things like

```
curl $(kn service create hello --image ... | tail -1)
```

Not sure if this  a good thing when people will rely on that or if  a `kn service describe hello --url` is better suited (I tend to the latter)

navidshaikh:
Before printing the URL, we can add a line may be saying
```
Created revision 'svc-abcde'
[..]
```
keeping the URL line at the bottom only.
WDYT ?

duglin:
I agree with @rhuss on the `tail` thing - I tend to use stuff like that in bash-scripts all the time.

navidshaikh:
@duglin @rhuss : the `tail` use case is intact, I am saying to have a line in output (not the last line) printing revision name.. like
```
✗  kn service create hello --image docker.io/swordphilic/go-helloworld
Service 'hello' successfully created in namespace 'default'.
Revision 'hello-asdf' created..
Waiting for service 'hello' to become ready ... OK

Service URL:
http://hello.default.apps-crc.testing
```

```
✗  kn service update hello --env TARGET=v2                            
Revision hello-wxyz created ...
Waiting for service 'hello' to become ready ... OK
Service 'hello' updated in namespace 'default'.
```

WDYT ?

duglin:
@navidshaikh looks good - easily parse-able by scripts  :-) 

[...]

PR #431, opened 2019-10-07
chore(service): Improvements for waiting on service readiness

rhuss:
* Increased default timeout to 600s. This timeout will be triggered
  only when the Ready condition stays in UNKNOWN for that long. If its
  True or False then the command will return anyway sooner.
  So it makes sense to go for a much longer timeout than 60s
* Enhanced output to indicate the progress

This change needs some updates to the API and introduces a 'MessageCallback'
type which is calle for each intermediate event with the "Ready" condition message.

Example of the UX:

[asciicast](https://asciinema.org/a/5MwD9j3HJpLdBgosJFbFsDBdB)

### DIFF
--- a/docs/cmd/kn_service_create.md
+++ b/docs/cmd/kn_service_create.md
@@ -63,7 +63,7 @@ kn service create NAME --image IMAGE [flags]
       --requests-memory string   The requested memory (e.g., 64Mi).
       --revision-name string     The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
       --service-account string   Service account name to set. Empty service account name will result to clear the service account.
-      --wait-timeout int         Seconds to wait before giving up on waiting for service to be ready. (default 60)
+      --wait-timeout int         Seconds to wait before giving up on waiting for service to be ready. (default 600)
 ```
 
 ### Options inherited from parent commands
--- a/pkg/wait/wait_for_ready.go
+++ b/pkg/wait/wait_for_ready.go
@@ -16,6 +16,7 @@ package wait
 
 import (
 	"fmt"
+	"io"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -38,8 +39,8 @@ type waitForReadyConfig struct {
 type WaitForReady interface {
 
 	// Wait on resource the resource with this name until a given timeout
-	// and write status out on writer
-	Wait(name string, timeout time.Duration) error
+	// and write event messages for unknown event to the status writer
+	Wait(name string, timeout time.Duration, msgCallback MessageCallback) (error, time.Duration)
 }
 
 // Create watch which is used when waiting for Ready condition
@@ -47,6 +48,9 @@ type WatchFunc func(opts v1.ListOptions) (watch.Interface, error)
 
 // Extract conditions from a runtime object
 type ConditionsExtractor func(obj runtime.Object) (apis.Conditions, error)
+
+// Callback for event messages
+type MessageCallback func(durationSinceState time.Duration, message string)
 
 // Constructor with resource type specific configuration
 func NewWaitForReady(kind string, watchFunc WatchFunc, extractor ConditionsExtractor) WaitForReady {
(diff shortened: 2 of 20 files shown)


docs/cmd/kn_service_create.md
```diff
@@ -59,7 +59,7 @@ kn service create NAME --image IMAGE [flags]
       --requests-memory string   The requested memory (e.g., 64Mi).
       --revision-name string     The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
       --service-account string   Service account name to set. Empty service account name will result to clear the service account.
-      --wait-timeout int         Seconds to wait before giving up on waiting for service to be ready. (default 60)
+      --wait-timeout int         Seconds to wait before giving up on waiting for service to be ready. (default 600)
```
maximilien: Does it make sense to also have a `--no-wait`?

maximilien: 600 seconds == 10 minutes. Is that too much for a default? I feel like one minute (60 seconds) should be good enough. Thoughts?

pkg/wait/wait_for_ready.go
```diff
@@ -38,8 +39,8 @@ type waitForReadyConfig struct {
 type WaitForReady interface {
 
 	// Wait on resource the resource with this name until a given timeout
-	// and write status out on writer
-	Wait(name string, timeout time.Duration) error
+	// and write event messages for uknown event to the status writer
```
maximilien: *unknown

docs/cmd/kn_service_create.md
rhuss: We had 60s before, but that was often too short, especially when the created Pods still have to pull the image, which is often longer than one minute. In that case a "timeout error" occurs as in #423, which is confusing as the service itself is ready eventually.

As I'm assuming that as long the "Ready" condition is in the state "UNKNOWN" I rely on Knative that this eventually will be reconciled to either "TRUE" or "FALSE". If this is the case, the command will return. It will only run into a timeout when staying in UNKNOWN. So from my experience, we should rely on Knative serving to do that reconciliation and e.g. if it can't get the service up in a certain amount of time, will be put the "Ready" condition to "FALSE". Hence this large timeout, which could maybe even longer (not sure if there are image that large that it will take longer than 10 mins, but then a sync wait on the CLI without a real progress meter is tedious, too).

I added the detailed feedback on the event as a poor-men progress meter replacement, so that the user sees that something is going on even when the creation/update takes a bit longer.

That output is also helpful for scripting use cases as it helps in debugging timing issues.

Still, with a `-q` we should certainly not show that messages. 

One can argue that one migt move these event to `--verbose`, but then one would lose the simple progress indicator.

rhuss: `--no-wait` is achieved with `--async` (that was chosen over no-wait at the time when the feature was introduced)

maximilien: OK I can live with that. Thanks.

navidshaikh:
From e2e, `ContainsAll` comparisons need update.

navidshaikh:
@rhuss  Since we're updating the text printed after service create and update, how about considering https://github.com/knative/client/issues/290 part of this?

rhuss:
> @rhuss Since we're updating the text printed after service create and update, how about considering #290 part of this?

Yes, that makes sense for the sync case as the revision would be available at that point, too (we could just add it to the 'showUrl()' method). Or we could add it to the event list, so that the final entry could look like:

```
       12.342s Ready to serve (Revision: myservice-2-3bmz3)
```

Another idea would be to have a `kn service create ... -o yaml` which could output a stream of JSON objects, which then would be even easier to parse in a shell loop (like this kind of [shell-based controller](https://github.com/k8spatterns/examples/blob/ac82b5a875913e24ac08c569ffb38737f7af2321/advanced/Controller/config-watcher-controller.sh#L19-L39).

This would us also allow to not think about how to format human-readable output that it can be easily parsed (which I think is dangerous anyway in the long run as we can easily break things if we make the concrete output format part of a kind of "API")

rhuss:
BTW, one reason why I would love to have the URL on a separate line also for human readable output is highlighted in https://youtu.be/9rnrOK0Ifqs?t=1876 : As you can see its not so easy to select an URL in floating text than having a double (or tripple) mouse click on a single line.

Not sure whether we should do that for a revision, too.

Maye we can rephrase the last URL line as:

```
Sevice 'demo' with revision 'demo-1-abced' can be reached at
https://demo.example.com
```

navidshaikh:
@rhuss : The URL on a separate (last) line is good, +1 for that.

For,

> Sevice 'demo' with revision 'demo-1-abced' can be reached at

the service might not necessarily be serving from this single revision.

WDYT about this:

For service create:
```
Service 'demo-1' created in namespace 'default'.
Revision 'demo-1-abced' created in namespace 'default'.
[..]

URL:
http://demo.default.example.com
```
For update creating a new revision:
```
Service 'demo-1' updated in namespace 'default'.
Revision 'demo-1-wxyz' created in namespace 'default'.
[..]

URL:
http://demo.default.example.com
```

For update **not** creating a new revision:
```
Service 'demo-1' updated in namespace 'default'.
[..]

URL:
http://demo.default.example.com
```

rhuss:
> the service might not necessarily be serving from this single revision.

I thought for `kn service create` we will only have one revision ? (not sure about `kn service create --force though`)

rhuss:
The revision can be inferred best only after the service is ready, not when the Service CR is created.

The output for sync create looks like (with a revision added after the service reconciled):

```
Creating service 'svc' in namespace 'default':

    0.123 bla
    1.245 bla
   10.345 Ready to serve.

Revision: 'svc-1-sada'
URL:
https://demo.com
```

I don't think we need to repeat the namespace for the revision name.

rhuss:
@navidshaikh I added now the revision name (uncoditionally, but for `update` I could try to find out whether the revision has changed and add an "(unchanged)" if not.

I also updated the asciinema video in the issue comment to reflect this change.

rhuss:
@navidshaikh updated now with "(unchanged)" as suffix to a revision if that revision has not changed (also did an update of the video)

PR #432, opened 2019-10-08
chore(build.sh) Add indicator in watch mode after successful compilation

rhuss:


rhuss:
@navidshaikh a quick one, helps in `build.sh -w` to see when compilation finished.

PR #436, opened 2019-10-08
Add navigation for management section

joaedwar:


docs/operations/management.md
```diff
@@ -37,6 +43,7 @@ Examples:
 # Listing a service's revision
 
 kn revision list --service srvc # CHECK this since current command does not have --service flag
```
navidshaikh: @joaedwar : can we remove the following text pelase ?
```
# CHECK this since current command does not have --service flag
```

PR #439, opened 2019-10-10
removes zsh completion since it does not work (fixes #426)

### LINKED ISSUE #426: zsh completion gives errors
```
$ uname -a
Darwin starkair.lan 18.7.0 Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64 x86_64

$ brew list zsh
/usr/local/Cellar/zsh/5.7.1/bin/zsh
...

$ source <(kn completion --zsh)
_arguments:comparguments:319: can only be called from completion function
_arguments:comparguments:319: can only be called from completion function
```
/cc @maximilien from #132 - any pro tips?

maximilien:
Fixes #426

## Proposed Changes

* removes zsh completion as it the code it generates does not work

- 🧽 Update or clean up current behaviour

### DIFF
--- a/pkg/kn/commands/completion.go
+++ b/pkg/kn/commands/completion.go
@@ -20,26 +20,13 @@ import (
 	"github.com/spf13/cobra"
 )
 
-type CompletionFlags struct {
-	Zsh bool
-}
-
 func NewCompletionCommand(p *KnParams) *cobra.Command {
-	var completionFlags CompletionFlags
-
-	completionCmd := &cobra.Command{
+	return &cobra.Command{
 		Use:    "completion",
-		Short:  "Output shell completion code (default Bash)",
-		Hidden: true, // Don't show this in help listing.
+		Short:  "Output shell completion code (Bash)",
+		Hidden: true, // Don't show this in help listing
 		Run: func(cmd *cobra.Command, args []string) {
-			if completionFlags.Zsh {
-				cmd.Root().GenZshCompletion(os.Stdout)
-			} else {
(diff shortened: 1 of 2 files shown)


rhuss:
Looks good to me, but I wonder why would not remove all completions as I think bash completion is also not very helpful (for the reasons mentioned in https://github.com/knative/client/issues/426#issuecomment-537050451). Now it might be even more confusing and raising the question, why bash-completion and not zsh ?

navidshaikh:
This works for zsh https://github.com/knative/client/pull/443.

maximilien:
> Looks good to me, but I wonder why would not remove all completions as I think bash completion is also not very helpful (for the reasons mentioned in [#426 (comment)](https://github.com/knative/client/issues/426#issuecomment-537050451)). Now it might be even more confusing and raising the question, why bash-completion and not zsh ?

Fair. I don't care for bash completion either. However, I believe @sixolet added it with her first commit. Care to chime in? Keep or scrap BASH completion too?

maximilien:
@rhuss separating BASH completion removal discussions with issue #456.

So if this LGTY then please merge. Thanks.

rhuss:
#443 make zsh completion working again (if we don't use `[` in our help messages), but I would be happy to also remove it for now (so merging this PR, but still updating via #443)

@navidshaikh wdyt ?

navidshaikh:
@rhuss : Sure lets merge this PR since the feature is not usable now. We can conclude on #443 and add it again by #443 .

PR #446, opened 2019-10-14
fix(service): Set default timeout for update to the same value as for create

rhuss:
Also, added accidentally removed seperator lines during create/update


maximilien:
Seems like istio is flapping...

```bash
istio-system      7m8s        Warning   FailedMount                    pod/istio-security-post-install-1.1.12-5qq46                         MountVolume.SetUp failed for volume "istio-security-post-install-account-token-hz6fd" : couldn't propagate object cache: timed out waiting for the condition
istio-system      7m8s        Warning   FailedMount                    pod/istio-security-post-install-1.1.12-5qq46                         MountVolume.SetUp failed for volume "tmp-configmap-security" : couldn't propagate object cache: timed out waiting for the condition
```

PR #448, opened 2019-10-15
Updates YAML transitive dependency to v2.2.4

navidshaikh:
As per title, just to make sure we're not vulnerable to CVE-2019-11253.
 See kubernetes/kubernetes#83253.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,8 @@ require (
 	github.com/stretchr/testify v1.4.0 // indirect
 	golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2
 	golang.org/x/net v0.0.0-20190724013045-ca1201d0de80 // indirect
-	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
+	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
+	gopkg.in/yaml.v2 v2.2.4 // indirect
 	gotest.tools v2.2.0+incompatible
 	k8s.io/api v0.0.0-20190226173710-145d52631d00
 	k8s.io/apimachinery v0.0.0-20190221084156-01f179d85dbc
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -17,7 +17,8 @@ require (
 	github.com/stretchr/testify v1.4.0 // indirect
 	golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2
 	golang.org/x/net v0.0.0-20190724013045-ca1201d0de80 // indirect
-	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
+	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
+	gopkg.in/yaml.v2 v2.2.4 // indirect
```
rhuss: I'd recommend to remove the `// indirect` to indicate that it is an explicit, and not a transitive dependency now. Otherwise this will be overwritten again in  a future update by the dependency which is using that.

Do you know which of our dependencies is using those packages ? Maybe it makes more sense to update those, too (which then include this updated dependency ?) At least a PR should be raised on the upstream dependency which uses the old package.

navidshaikh: I think its coming from spf13/viper https://github.com/knative/client/blob/master/vendor/github.com/spf13/viper/go.mod#L23

We're using spf13/viper 1.3.1(latest available is 1.4.0, is there a reason we picked up the older version?).

@rhuss : PR to viper for updating YAML to 2.2.4 https://github.com/spf13/viper/pull/785

Will remove the //indirect in another PR.

navidshaikh: spf13/viper isn't vendoring any deps https://github.com/spf13/viper/blob/master/.gitignore#L29 , we updating the yaml keeps it using the one we vendored

navidshaikh: The PR to viper is in and 1.5.0 is expected to be released later this week.

PR #453, opened 2019-10-17
fix(version): Displays supported APIs and version

### LINKED ISSUE #404: `kn version` : Display supported API and version
## Describe the feature:

Currently the display of `kn version` is a bit misleading:

```
Version:      v20190910-local-a607b16
Build Date:   2019-09-10 09:41:55
Git Revision: a607b16
Support:
- Serving: v0.8.0  v0.7.1
- API(s):  v1alpha1
```

* Which is the API group ?
* Is v1alpha1 supported for both, v0.8.0 and v0.7.1
* What if we add eventing ? There will be differing API versios supported
* Also eventing as several API groups which might have also different versions.
* Does this mean that we can't run `kn` against a (yet) unknown 0.9.0 serving version which uses the same API group ?

My suggestion is to list the group and versions supported by kn (for serving and eventing) and only use the product version for alpha version as this are supposedly the only API which can change within the same version (e.g. v1alpha1 for serving 0.6 is different than v1alpha1 for serving 0.7). This is not the case for beta or full versions (if those change then also the version number has to be changed for every change in the API model), e.g. v1beta1, v1beta2, ...

My proposal for the version display would be

```
Version:      v20190910-local-a607b16
Build Date:   2019-09-10 09:41:55
Git Revision: a607b16
Supported APIs:
- serving.knative.dev/v1
- serving.knative.dev/v1beta1
- serving.knative.dev/v1alpha1 (knative-serving v0.8.0)
- eventing.knative.dev/v1alpha1 (knative-eventing v0.7.1)
- messaging.knative.dev/v1beta1
```

This would also imply if a new version of knative-serving which still supports `serving.knative.dev/v1beta1` this older client would also support this new knative serving version.

navidshaikh:
Fixes #404
/lint

## Proposed Changes
 - Add the API versions and component version vendored
 - Moves the version command under its package version
-  Removes ServingVersion feeding from ldflags

 ```
  ./kn version
Version:      v20191017-local-6328a73-dirty
Build Date:   2019-10-17 09:55:55
Git Revision: 6328a73
Supported APIs:
- serving.knative.dev/v1alpha1 (knative-serving v0.8.0)
 ```

### DIFF
--- a/pkg/kn/commands/version/version.go
+++ b/pkg/kn/commands/version/version.go
@@ -26,14 +26,12 @@ import (
 var Version string
 var BuildDate string
 var GitRevision string
-var ServingVersion string
 
 type SupportedAPIs []string
 
-// update this var as we increase the serving version in go.mod
-var knServingDep = "v0.8.0"
+// update this var as we add more deps or update
 var supportMatrix = map[string]*SupportedAPIs{
-	knServingDep: {"serving.knative.dev/v1alpha1 (knative-serving v0.8.0)"},
+	"v0.8.0": {"serving.knative.dev/v1alpha1 (knative-serving v0.8.0)"},
 }
 
 func (s SupportedAPIs) print(out io.Writer) {
@@ -47,20 +45,15 @@ func NewVersionCommand(p *commands.KnParams) *cobra.Command {
 	versionCmd := &cobra.Command{
 		Use:   "version",
 		Short: "Prints the client version",
-		RunE: func(cmd *cobra.Command, args []string) error {
+		Run: func(cmd *cobra.Command, args []string) {
 			out := cmd.OutOrStdout()
 			fmt.Fprintf(out, "Version:      %s\n", Version)
 			fmt.Fprintf(out, "Build Date:   %s\n", BuildDate)
 			fmt.Fprintf(out, "Git Revision: %s\n", GitRevision)
 			fmt.Fprintf(out, "Supported APIs:\n")
-			if apis, ok := supportMatrix[ServingVersion]; ok {
+			for _, apis := range supportMatrix {
 				apis.print(out)
-			} else {
-				// ensure the go build works when we update,
-				// but version command tests fails to prevent shipping
-				fmt.Fprintf(out, "- Serving: %s\n", ServingVersion)
 			}
-			return nil
 		},
 	}
 	return versionCmd
--- a/pkg/kn/commands/version/version.go
+++ b/pkg/kn/commands/version/version.go
@@ -12,11 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package commands
+package version
 
 import (
 	"fmt"
-	"strings"
+	"io"
+
+	"knative.dev/client/pkg/kn/commands"
 
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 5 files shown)


pkg/kn/commands/version/version.go
```diff
@@ -26,19 +28,22 @@ var BuildDate string
 var GitRevision string
 var ServingVersion string
```
rhuss: tbh, I don't it makes sense to have this variable fed in from the outside when we have to update the content of this source anyway for each serving update. I would just print out the supported and used the version for this client from a fixed string which might be updated over time.

Later on, we can query the server for the supported versions (from the CRD) and print this, too. Ideally we would also support multiple versions and depending on the server support choose one the version that we support, too (currently, it's just `v1alpha1` but we should jump on `v1beta1` or even `v1` soonish)

pkg/kn/commands/version/version.go
```diff
@@ -26,19 +28,22 @@ var BuildDate string
...
-	Versions, APIs []string
-}
+type SupportedAPIs []string
 
 // update this var as we increase the serving version in go.mod
 var knServingDep = "v0.8.0"
-var supportMatrix = map[string]*VersionsAPIs{
-	knServingDep: {[]string{"v0.8.0", "v0.7.1"}, []string{"v1alpha1"}},
+var supportMatrix = map[string]*SupportedAPIs{
+	knServingDep: {"serving.knative.dev/v1alpha1 (knative-serving v0.8.0)"},
```
maximilien: Could this be read from external file?

pkg/kn/commands/version/version.go
navidshaikh: I was inclined towards hardcoding this too for the reason that we have to update this file after serving dep bump. Thanks for the +1, will update.

pkg/kn/commands/version/version.go
navidshaikh: I think yes, but then we'd need to update the external file when we bump the version, so I think updating it directly here would save us introducing an additional file. Also, as per discussion above, I think it makes more sense to hard code it now as feeding the var in here doesn't add much value and one has to anyway edit this file part of release.

pkg/kn/commands/version/version.go
```diff
@@ -12,51 +12,48 @@
...
-			fmt.Fprintf(out, "Support:\n")
-			if m, ok := supportMatrix[ServingVersion]; ok {
-				fmt.Fprintf(out, "- Serving: %s\n", strings.Join(m.Versions, "  "))
-				fmt.Fprintf(out, "- API(s):  %s\n", strings.Join(m.APIs, " "))
-			} else {
-				// ensure the go build works when we update,
-				// but version command tests fails to prevent shipping
-				fmt.Fprintf(out, "- Serving: %s\n", ServingVersion)
+			fmt.Fprintf(out, "Supported APIs:\n")
+			for _, apis := range supportMatrix {
```
rhuss: Can we please just do a series of printf here for the versions we support and remove the map ? 

Actually the order of output  is not deterministic, and having a map here is not very helpful. You could just build up an array of the output strings and simply iterate over those here.

Like in 

```golang
apiVersions := []string {
    "serving.knative.dev/v1alpha1 (knative-serving v0.8.0)",
    "eventing.knative.dev/v1",
     ....
}
```

navidshaikh: Yup, I should've used the array, updated now.

PR #457, opened 2019-10-22
Add sources client to types and testing_helper

daisy-ycguo:
## Proposed Changes

* Add kn sources client to types and testing_helper, to support future kn eventing sources client functions, for easily review
* Add sources client dependencies


pkg/kn/commands/types_test.go
```diff
@@ -141,3 +141,53 @@ func TestGetClientConfig(t *testing.T) {
...
+			"",
+			false,
+		},
+		{ // Test that the cast to wrap the http client in a logger works
+			basic,
+			"",
+			true,
+		},
+	} {
+		p := &KnParams{
```
evankanderson: Put this part in a `t.Run()` and include a name of the scenario you are testing?

pkg/kn/commands/types.go
```diff
@@ -38,12 +40,13 @@ type Config struct {
...
-	ClientConfig clientcmd.ClientConfig
-	NewClient    func(namespace string) (serving_kn_v1alpha1.KnServingClient, error)
+	Output           io.Writer
+	KubeCfgPath      string
+	ClientConfig     clientcmd.ClientConfig
+	NewClient        func(namespace string) (serving_kn_v1alpha1.KnServingClient, error)
+	NewSourcesClient func(namespace string) (sources_kn_v1alpha1.KnSourcesClient, error)
 
 	// General global options
 	LogHTTP bool
```
maximilien: Interesting. Where is this being used?

daisy-ycguo:
@maximilien most of the files are dependencies, because Eventing client is added. They are auto included by `hack/build.sh`. 
The updated files by me are `pkg/kn/commands/testing_helper.go` and `pkg/kn/commands/types.go` and their test files. 
I also created a folder `pkg/sources/v1alpha1` to add a basic `knSourcesClient` as a wrapper to handle the logics of eventing sources.
All these are following the existing structure of serving client logic.
After this PR is merged, we can easily add features of sources client and easily review the following PRs.

@rhuss please suggest how to handle the dependency files. I will suggest to merge them with this basic PR, because this PR will not pass the tests without these dependencies.

rhuss:
> @rhuss please suggest how to handle the dependency files. I will suggest to merge them with this basic PR. Please consider that they are necessary dependencies. This PR will not pass the tests without these dependencies.

I think we should just merge eventing 0.9.0 dependencies with one PR and then build on top of it. Then the subsequent PRs are much easier to review.

However, I really would love to hear about the concrete use case for `KnSourceClient`. It might be useful for carrying shared code, but I don't see that yet.

pkg/kn/commands/testing_helper_test.go
```diff
@@ -54,3 +55,34 @@ func TestCreateTestKnCommand(t *testing.T) {
...
+		knParams := &KnParams{}
+		knCmd, serving, buffer = CreateSourcesTestKnCommand(&cobra.Command{Use: "fake"}, knParams)
+		assert.Assert(t, knCmd != nil)
+		assert.Assert(t, len(knCmd.Commands()) == 1)
+		assert.Assert(t, knCmd.Commands()[0].Use == "fake")
+		assert.Assert(t, serving != nil)
+		assert.Assert(t, buffer != nil)
+	}
+
+	t.Run("creates a new kn cobra.Command", func(t *testing.T) {
```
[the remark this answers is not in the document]
rhuss: I think the motivation was to have a description in the test output for the test, but I think the test's name alone is descriptive enough, so no need for a subtest.

pkg/kn/commands/types.go
daisy-ycguo: When we begin to add Eventing functions to client, we may hope to have all methods to interact with Eventing server APIs in a single file. Similar as [pkg/serving/v1alpha1/client.go](https://github.com/knative/client/blob/master/pkg/serving/v1alpha1/client.go), I created pkg/sources/v1alpha1/client.go correspondingly. NewSourcesClient will be used to create such a KnSourcesClient defined in pkg/sources/v1alpha1/client.go. It will be invoked in every command who needs to interact with eventing sources api.

daisy-ycguo:
> Dependency looks good, but I wonder why we need a generic `KnSourceClient` in the top-level packages. I think its better that each source has its own client, living in the source-specific package. E.g. like in https://github.com/rhuss/client/blob/pr/cron-importer/pkg/kn/commands/source/cron/cron_client.go for the `CronJobSource`.
> 
> Is there anything which should be shared between those Sources (which all have own CRDs to manage and hence own signatures)?
> 
> Having the source client independently has several advantages:
> 
> * Different release cycles/versions of the Source's API can be easily supported. Not so critical for built-in source which are hosted within Eventing, but even there the three supported sources can be at different versions. It doesn't make sense to put them into a single interface imo.
> * We don't need to work with an interface as there is supposed to be only one implementation (except maybe for testing purposes).
> 
> Or let me ask the other way around: What is the benefit of having a top-level `KnSourceClient` instead of specific `KnCronJobSourceClient` and `KnApiserverSourceClient` ? What can be shared and is it worth the coupling ?
> 
> Soon we will have a `KnEventingClient` for managing Eventing CRDs, but for source I don't see this yet.

daisy-ycguo:
misoperation. reopen it.

daisy-ycguo:
@rhuss suggests to have a single PR with only dependencies in order to ease the review. I will close this one and open a new one.

PR #460, opened 2019-10-28
fix(service describe): Clarify URL labels

### LINKED ISSUE #459: Change "Address" field name on service descibe
```
$ kn service describe echo1 

Name:       echo1
Namespace:  default
URL:        http://echo1.default.kndev.us-south.containers.appdomain.cloud
Address:    http://echo1.default.svc.cluster.local
...
```

I think the use of the terms "Address" and "URL" are going to be confusing to people because those words mean just about the same thing. I would like to suggest that we use something like "Internal URL" instead of "Address" because 1) I believe it really is for internal cluster usage, and 2) it would make it consistent with "URL" by reusing the word "URL" from the previous field.

rhuss:
Fixes #459.

### DIFF
--- a/pkg/kn/commands/service/describe.go
+++ b/pkg/kn/commands/service/describe.go
@@ -194,10 +194,10 @@ func describe(w io.Writer, service *v1alpha1.Service, revisions []*revisionDesc,
 // Write out main service information. Use colors for major items.
 func writeService(dw printers.PrefixWriter, service *v1alpha1.Service) {
 	commands.WriteMetadata(dw, &service.ObjectMeta, printDetails)
-	dw.WriteAttribute("Route URL", extractURL(service))
+	dw.WriteAttribute("URL", extractURL(service))
 	if service.Status.Address != nil {
 		url := service.Status.Address.GetURL()
-		dw.WriteAttribute("Cluster URL", url.String())
+		dw.WriteAttribute("Cluster", url.String())
 	}
 	if (service.Spec.Template != nil) && (service.Spec.Template.Spec.ServiceAccountName != "") {
 		dw.WriteAttribute("ServiceAccount", service.Spec.Template.Spec.ServiceAccountName)
--- a/pkg/kn/commands/service/describe.go
+++ b/pkg/kn/commands/service/describe.go
@@ -194,10 +194,10 @@ func describe(w io.Writer, service *v1alpha1.Service, revisions []*revisionDesc,
 // Write out main service information. Use colors for major items.
 func writeService(dw printers.PrefixWriter, service *v1alpha1.Service) {
 	commands.WriteMetadata(dw, &service.ObjectMeta, printDetails)
-	dw.WriteAttribute("URL", extractURL(service))
+	dw.WriteAttribute("Route URL", extractURL(service))
 	if service.Status.Address != nil {
 		url := service.Status.Address.GetURL()
-		dw.WriteAttribute("Address", url.String())
+		dw.WriteAttribute("Cluster URL", url.String())
 	}
 	if (service.Spec.Template != nil) && (service.Spec.Template.Spec.ServiceAccountName != "") {
 		dw.WriteAttribute("ServiceAccount", service.Spec.Template.Spec.ServiceAccountName)
--- a/pkg/kn/commands/service/describe_test.go
+++ b/pkg/kn/commands/service/describe_test.go
@@ -489,8 +489,8 @@ func TestServiceDescribeMachineReadable(t *testing.T) {
 func validateServiceOutput(t *testing.T, service string, output string) {
 	assert.Assert(t, cmp.Regexp("Name:\\s+"+service, output))
 	assert.Assert(t, cmp.Regexp("Namespace:\\s+default", output))
-	assert.Assert(t, cmp.Regexp("Address:\\s+http://"+service+".default.svc.cluster.local", output))
-	assert.Assert(t, cmp.Regexp("URL:\\s+"+service+".default.example.com", output))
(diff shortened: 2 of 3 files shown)


rhuss:
The actual naming should be discussed though: Is `Route URL` a good label for specifying an externally accessible URL or should this be better called `External URL` (on the expense of 3 additional chars for the left label column)

pkg/kn/commands/service/describe.go
```diff
@@ -202,10 +202,10 @@ func describe(w io.Writer, service *v1alpha1.Service, revisions []*revisionDesc,
 func writeService(dw printers.PrefixWriter, service *v1alpha1.Service) {
 	dw.WriteColsLn(printers.Level0, l("Name"), service.Name)
 	dw.WriteColsLn(printers.Level0, l("Namespace"), service.Namespace)
-	dw.WriteColsLn(printers.Level0, l("URL"), extractURL(service))
+	dw.WriteColsLn(printers.Level0, l("Route URL"), extractURL(service))
```
navidshaikh: IMO, lets keep this field intact (use `URL`).

rhuss: The only thing which bothers me a bit that it's assymetric as "Cluster URL" is-a "URL", too. So maybe "External URL" is better ? We could even skip the "URL" part from cluster its clear from the value that its a URL. E.g.

```
URL:      http://myservice.external.com
Cluster:  http://myservice.myns.service.cluster.local
```

(alternative would be to replace "URL:" with "External:" or "Route:")

rhuss: Some variations:

```
URL:      http://myservice.example.com
Internal: http://myservice.myns.service.cluster.local
```

rhuss:
```
External: http://myservice.example.com
Internal: http://myservice.myns.service.cluster.local
```

rhuss:
```
External URL: http://myservice.example.com
Cluster URL:  http://myservice.myns.service.cluster.local
```

rhuss:
```
URL:         http://myservice.example.com
Cluster URL: http://myservice.myns.service.cluster.local
```

n3wscott: How about:

```
External URL: http://myservice.example.com
Cluster URL:   http://myservice.myns.service.cluster.local
```

n3wscott: Oh, you already have that option :DD:D:D

rhuss: A final suggestion, to be clear about URL but reduce line width (as this is normally the biggest constraint, as URLs can become really long):

```
URL:
  External: http://myservice.example.com
  Cluster:  http://myservice.myns.service.cluster.local
```

bbrowning:
I also feel we should be consistent with how Serving labels this, which is `URL` for the `status.url` value (https://github.com/knative/serving/blob/release-0.10/config/300-service.yaml#L50-L52). You should not assume this is only an external value. In the case of cluster-local services, this will actually be an internal URL only accessible from inside the cluster. For example:
```
$ oc get ksvc -n bbrowning-sandbox
NAME                  URL                                                                           LATESTCREATED               LATESTREADY                 READY   REASON
helloworld-go         http://helloworld-go.bbrowning-sandbox.apps.myexternaldomain.org              helloworld-go-dfzjz         helloworld-go-dfzjz         True    
helloworld-go-local   http://helloworld-go-local.bbrowning-sandbox.svc.cluster.local                helloworld-go-local-s46cl   helloworld-go-local-s46cl   True 
```

rhuss:
> You should not assume this is only an external value. In the case of cluster-local services, this will actually be an internal URL only accessible from inside the cluster.

How would one then better describe the difference between `status.url` and `status.address.url` ? The latter is the implementation `Addressable` interface for being considered to be a target for eventing. Actually, the background for this PR is #459 where it is correctly stated that the difference of "Address" and "URL" is not intuitive for a user (especially as Address is *also* an URL).

But I get the point that URL is not necessarily reachable from the outside, actually, it doesn't make any assumption of its routability. I think we should highlight somewhat that "URL" is the user-accessible _address_ (either from outside or from another service inside), and that "Address" is always something internal, mostly used by eventing. Not sure how to phrase this in two words, though.

Maybe it would be even better to hide `status.address.url` behind the `--verbose` flag as for the casual use it's of not much value, and then, for `--verbose`, we could call address `Eventing Address`. That would have the benefit that by default only the "main" URL is shown.

pkg/kn/commands/service/describe.go
rhuss: So we have a draw between URL/Cluster and External URL/Cluster URL. Considering @bbrowning comment, that the 'external' URL is not always external, I updated the PR to the "URL/Cluster" combo.

In a followup PR I would suggest hiding the cluster URL behind `--verbose` as this is really only of interest in very rare circumstances (e.g. when used as a Destination in Eventing)

rhuss:
@bbrowning 

> In the case of cluster-local services, this will actually be an internal URL only accessible from inside the cluster.

Just out of curiosity (and sorry for my ignorance), how can a user indicate the creation of a cluster-local service instead of an exposed service. Or is this a cluster/installation wide option ?

rhuss:
@sixolet @maximilien I have updated the PR according to the vote. Please see https://github.com/knative/client/pull/460#discussion_r342759532 for the selected choice. 

If this is ok, please lgtm

pkg/kn/commands/service/describe.go
navidshaikh: logged here https://github.com/knative/client/issues/482

Issue #308, opened 2019-07-26
Kn plugins integration test


maximilien:
## In what area(s)?

Kn plugins

Classifications:

/kind feature

## Describe the feature:

Now that #249 is merged we need to have a separate integration test (IT) for that feature. There is a lot to test. As a start, let's add this:

1. simple plugin execution of a BASH and executable plugin in
  a. `--plugins-dir`
  b.  $PATH
2. finding plugin in both `--plugins-dir` and $PATH
3. verifying plugin overshadowing
4. verifying plugin cannot override a core command

---
There will be more to add to these ITs as plugins matures and more features are added.

maximilien:
I’ll push this once #313 is merged since I need to change some of the output tests.

navidshaikh:
@maximilien +1

maximilien:
Need some integration tests. Latest issues found in #411 showed this is needed. Making this my top priority with #414 and then #320

/cc @rhuss

maximilien:
Finally, getting back to this after our long unforseen delays. Fix one issue #468 which was blocking my WIP on this.

maximilien:
Now that #468 is merged (yeah). Getting back to this to submit it soon. Finally.

PR #468, opened 2019-10-31
load config values from viper if present in config

### LINKED ISSUE #467: `kn` not loading config values on start
## In what area(s)?

Kn config and viper binding

> /kind bug

```bash
./hack/build.sh
🕸️  Update
🧽  Format
⚖️  License
📖 Docs
🚧 Compile
🧪  Test
────────────────────────────────────────────
Version:      v20191031-local-a269d29
Build Date:   2019-10-31 20:48:01
Git Revision: a269d29
Supported APIs:
- serving.knative.dev/v1alpha1 (knative-serving v0.9.0)
```

> 0.8.x

## Expected Behavior

When changing the content of the config file `~/.kn/config.yaml` I expect that every subsequent invocations of `kn` (w/o overrides to the config values) to use the `config.yaml` values as its default. 

Since we have two configs now: `lookup-plugins` and `plugins-dir`. If I change `lookup-plugins` to true (non-default value). I expect `kn` to look for plugins in my `$PATH`. That does not happen.

## Actual Behavior

Currently, `kn` does not load the values in the `~/.kn/config.yaml` to initialize the configs mentioned above. You can change the values of these config by passing them as flags in your invocation, e.g., `./kn ... --lookup-plugings=true` and the behavior will be correct. However, changing the value in the config file does not work.

```bash
➜  client git:(master) ✗ cat ~/.kn/config.yaml
lookup-plugins: true
plugins-dir: ~/.kn/plugins
➜  client git:(master) ✗ ./kn plugin list
/Users/maximilien/.kn/plugins/kn-hello
➜  client git:(master) ✗ ./kn plugin list --lookup-plugins
/Users/maximilien/.kn/plugins/kn-hello
/Users/maximilien/bin/kn-hello2
/usr/local/bin/kn-hello2
WARNING:
  - /usr/local/bin/kn-hello2 is ignored because it is shadowed by a equally named plugin: /Users/maximilien/bin/kn-hello2.
```

maximilien:
Fixes #467 which details how to reproduce.

The gist of the issue is that viper binding will not auto load the config values (`lookup-plugins` and `plugins-dir`). This change checks and loads if the config values are present.

## Proposed Changes

* add `initConfigFlags` call in root's command
* set all Kn's config values to loaded values from Viper when ""

Release Note:

- 🐛 Fix bug #467 by loading config values from config file (when it exists)


pkg/kn/core/root.go
```diff
@@ -212,6 +213,16 @@ func initConfig() {
...
 }
 
+func initConfigFlags() {
+	if commands.Cfg.PluginsDir == "" {
+		commands.Cfg.PluginsDir = viper.GetString("plugins-dir")
+	}
+
+	if commands.Cfg.LookupPlugins != viper.GetBool("lookup-plugins") {
+		commands.Cfg.LookupPlugins = viper.GetBool("lookup-plugins")
+	}
```
rhuss: You can as well remove the conditional, because when config == option stting it will be a no-op. In any case, you will override the configuration.

E.g. even when `--lookup-plugins` is not-set it defaults to `false`. That means that a `cfg.LookupPlugins == true` will be overwritten with `false` even when no `--lookup-plugins` is  given. I don't think that is what you want. For this to work, the flag needs to be a pointer with a `nil` value when not set.

maximilien: Yeah, I thought of that but not using a pointer. Perhaps makes sense. Let me see.

For tests, I am adding comprehensive tests in #308 since it's not so easy to do this test without more setup work to get config and load and call `kn`.

So if it's OK please approve this (after I change to pointer) so I can get the tests for #308 done.

maximilien: Actually @rhuss there is really no way to do this nicely for boolean flags. I changed it to pointer to a bool and we essentially end up with same code. 

The reason above works is that no `lookup-plugins` in the config will result in viper returning false which is the default value. And since this code will execute `PreRun` it will set the value to true if viper has that value. Naturally if the user passes a flag then that will override the flag.

I suggest we leave it as is. If you want to see the pointer code I can push it but it changes lots of files and ends up with the `initConfigFlags` as below:

```go
func initConfigFlags() {
	if commands.Cfg.PluginsDir == "" {
		commands.Cfg.PluginsDir = viper.GetString("plugins-dir")
	}

	var aBool bool
	// This condition if (...) never actually executes since setting 
	// commands.Cfg.LookupPlugins to nil ends up with SEG fault in adding the flag
	// to the root command so it has to be set to a pointer to a false bool
	if commands.Cfg.LookupPlugins == nil {
		aBool = viper.GetBool("lookup-plugins")
		commands.Cfg.LookupPlugins = &aBool
	} else {
		if viper.GetBool("lookup-plugins") {
			aBool = viper.GetBool("lookup-plugins")
			commands.Cfg.LookupPlugins = &aBool	
		} else {
			aBool = false
			commands.Cfg.LookupPlugins = &aBool
		}
	}
}
```

And that's above the ugly bool code we have to do since golang won't allow `&false` or `&true`.

maximilien: And of course as I mentioned, integration tests for #308 will cover this in details.

I changed the description and title as @sixolet mentions so I would say this is good to go IMO.

maximilien: > For this to work, the flag needs to be a pointer with a `nil` value when not set.

Not possible since that will cause a segmentation fault when adding the flag to the `rootCmd` in `plugins.go` function:

```go
func AddPluginFlags(cmd *cobra.Command) {
	cmd.Flags().StringVar(&commands.Cfg.PluginsDir, "plugins-dir", "~/.kn/plugins", "kn plugins directory")
	cmd.Flags().BoolVar(commands.Cfg.LookupPlugins, "lookup-plugins", false, "look for kn plugins in $PATH")
}
```

rhuss: > The reason above works is that no `lookup-plugins` in the config will result in viper returning false which is the default value. And since this code will execute `PreRun` it will set the value to true if viper has that value. Naturally if the user passes a flag then that will override the flag.

Sorry, I'm still not buying this. Let's assume that you don't set a configuration value for `lookup-plugins` but you set `--lookup-plugins` to true from the command line. What will `commands.Cfg.LookupPlugins` end up with ? Where have you seen that the values from command lines flags are injected *after* `PreRun` ?

With _using pointers_ I meant to use `cmd.Flags.BoolVarP()` variant to point to the address of `cfg.LookupPlugins`, not to change `cfg.LookupPlugins` to a pointer.

But again, you should bind the viper config values directly to the configuration as described in https://github.com/spf13/cobra#bind-flags-with-config and https://github.com/spf13/viper#working-with-flags . There is no need to deal with this low level stuff on our own.

rhuss: Hmm, maybe I really don't understand the issue at all, as we are using allready `BindPluginsFlagToViper()` for this kind of binding, so I'm not sure why we need a special treatment (and what the bug is actually).  

Could be have an issue with the binding because we introduced an own reconciliation ? And if we do an own handling, do we still need to call to `BindPluginsFlagForViper()` ?

maximilien: The binding is sort of one way Viper to file and file to Viper. But AFAIK does not load from Viper to variable. This is what I am doing in my change. Try it and see for yourself. The issue #467 shows how to repro.

sixolet: Can we solve this by conditioning on flags.IsSet()?

maximilien: Let me look into it @sixolet, interesting idea. Give me a day or so to investigate, trying to complete some things before Thxgiving (if possible).

maximilien: @sixolet the logic is simplified a bit but the result is the same. We still need to implicitly init the values from viper.

This works. I suggest we merge this in and move on. In future if someone wants to dig into how to save 10 lines of code via some Viper magic I am missing, I'd be happy to see it.

PR #472, opened 2019-11-04
Add service account info into service describe

### LINKED ISSUE #413: Add service account info in service describe
### Classifications:

/kind good-first-issue
/kind feature
/kind proposal

### Describe the feature:
`service describe` default output should list the configured service account for the service if present.

daisy-ycguo:
Fixes #413 

## Proposed Changes

* Add service account info to `service describe` if present.

Release Note:
- 🧽 Add service account info to `service describe` if present.

### DIFF
--- a/pkg/kn/commands/service/describe_test.go
+++ b/pkg/kn/commands/service/describe_test.go
@@ -52,7 +52,7 @@ func TestServiceDescribeBasic(t *testing.T) {
 	// Recording:
 	r := client.Recorder()
 	// Prepare service
-	expectedService := createTestService("foo", []string{"rev1"}, goodConditions())
+	expectedService := createTestServiceWithServiceAccount("foo", []string{"rev1"}, "default-sa", goodConditions())
 
 	// Get service & revision
 	r.GetService("foo", &expectedService, nil)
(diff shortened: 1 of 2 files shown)


pkg/kn/commands/service/describe_test.go
```diff
@@ -495,7 +496,7 @@ func validateServiceOutput(t *testing.T, service string, output string) {
 	assert.Assert(t, util.ContainsAll(output, "Ready", "RoutesReady", "OK", "TYPE", "AGE", "REASON"))
 }
 
-func createTestService(name string, revisionNames []string, conditions duckv1.Conditions) v1alpha1.Service {
+func createTestService(name string, revisionNames []string, serviceaccount string, conditions duckv1.Conditions) v1alpha1.Service {
```
navidshaikh: Can we define `createTestServiceWithServiceAccount` function and have it called `createTestService` + code for adding ServiceAccount ?  
Most of the invocation to createTestService are with empty strings otherwise.

navidshaikh: *serviceAccount

pkg/kn/commands/service/describe_test.go
```diff
@@ -546,6 +547,20 @@ func createTestService(name string, revisionNames []string, conditions duckv1.Co
 		}
 		service.Status.Traffic = trafficTargets
 	}
+
+	if len(serviceaccount) > 0 {
```
maximilien: This check reads odd... if `serviceaccount` is suppose to be the `serviceAccountName` as it's set in the `PodSpec` can we call it that and also check for `!= ""` instead?

daisy-ycguo: Fixed. Thanks.

pkg/kn/commands/service/describe_test.go
daisy-ycguo: Fixed. Thanks.

daisy-ycguo:
@maximilien @navidshaikh It's ready for a second review. Thank you.

PR #476, opened 2019-11-05
Adds zsh completion

### LINKED ISSUE #456: Is BASH completion needed and useful?
## In what area(s)?

`completion` command

Classifications:

> /kind usage

## Ask your question here:

We added `zsh` completion but it was useless so it's removed in #439 and @rhuss raised the [question](https://github.com/knative/client/pull/439#issuecomment-540977753) if we should also remove BASH completion since it's not also too useful.

Separating the two concerns with this issue. We can discuss BASH completion's merits or lack of here and allow #439 to be merged independently.

navidshaikh:
Fixes #456 

## Proposed Changes
 - Adds back zsh completion after vendoring cobra library with a fix
 - Updates the command definition to accept argument either of ['bash', 'zsh']
 - Moves the completion command definition under `pkg/kn/commands/completion`
 - Updates docs
 - Updates CHANGELOG

/lint


rhuss:
Looks good to me.

I added https://github.com/knative/client/issues/477 to not forget that we should reintroduce the completion command to the help message, too (or at least discuss this).

/lgtm

PR #478, opened 2019-11-05
Add completion to help message

### LINKED ISSUE #477: Add "completion" to help message (again)
## In what area(s)?

`kn help`

## Describe the feature:

Re-add `kn completion` to top-level help message as this is missing currently. Not sure why we removed it, so let's first discucss/decide this here.

navidshaikh:
Fixes #477

### DIFF
--- a/docs/cmd/kn_completion.md
+++ b/docs/cmd/kn_completion.md
@@ -1,0 +1,47 @@
+## kn completion
+
+Output shell completion code
+
+### Synopsis
+
+
+This command prints shell completion code which needs to be evaluated
+to provide interactive completion
+
+Supported Shells:
+ - bash
+ - zsh
+
+```
+kn completion [SHELL] [flags]
+```
+
+### Examples
+
+```
+
+ # Generate completion code for bash
+ source <(kn completion bash)
+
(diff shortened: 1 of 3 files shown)


rhuss:
Thanks! Looks good to me, but please let's check whether we should do it as we intentionally removed it with https://github.com/knative/client/pull/263

How do folks feel to include `kn completion` again in the top-level `kn help` output? 

My take is, that we should add it with this PR and not treat it as a _magic gem_ (especially as we are advertising it on our main README, which is another topic on its own). Since we have now more top-level commands (and more to come with `source` and `trigger`), I think it's not so disturbing anymore.

docs/cmd/kn_completion.md
```diff
@@ -0,0 +1,47 @@
+## kn completion
+
+Output shell completion code
+
+### Synopsis
+
+
+This command prints shell completion code which need to be evaluated
```
maximilien: *"which needs"

rhuss:
```suggestion
This command prints shell completion code whichs need to be evaluated
```

PR #481, opened 2019-11-05
removes remaining references to 'lookup-plugins-in-path'

maximilien:
Removes remaining references to 'lookup-plugins-in-path' which is now renamed to the simpler 'lookup-plugins'

## Proposed Changes

* change references of 'lookup-plugins-in-path' to 'lookup-plugins' (new name for flag)

### DIFF
--- a/pkg/kn/commands/plugin/handler_test.go
+++ b/pkg/kn/commands/plugin/handler_test.go
@@ -90,7 +90,7 @@ func TestPluginHandler(t *testing.T) {
 		})
 
 		t.Run("when plugin is in $PATH", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pkg/kn/commands/plugin/handler_test.go
```diff
@@ -90,7 +90,7 @@ func TestPluginHandler(t *testing.T) {
 		})
 
 		t.Run("when plugin is in $PATH", func(t *testing.T) {
-			t.Run("--lookup-plugins-in-path=true", func(t *testing.T) {
+			t.Run("--lookup-plugins=true", func(t *testing.T) {
```
rhuss: Isn't it that we now have `--lookup-plugins` and `--no-lookup-plugins` for boolean params ?

PR #484, opened 2019-11-06
Shorten the --log-http Authorization mask value

### LINKED ISSUE #391: shorten the --log-http Authorization mask value
```
Authorization: *************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************
```
Seems but excessive when something like:
```
Authorization: ****
```
will do - plus it's kind of ugly to see it in the logs. Minor thing but it's the first thing I noticed and laughed at when I saw the logs

daisy-ycguo:
Fixes #391

## Proposed Changes

* display `--log-http` Authorization as `********`

### DIFF
--- a/pkg/util/logging_http_transport.go
+++ b/pkg/util/logging_http_transport.go
@@ -20,7 +20,6 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"os"
-	"strings"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 )
@@ -56,11 +55,7 @@ func (t *LoggingHttpTransport) RoundTrip(r *http.Request) (*http.Response, error
 	for k, v := range r.Header {
 		if sensitiveRequestHeaders.Has(k) {
 			redacted[k] = v
-			l := 0
-			for _, h := range v {
-				l += len(h)
-			}
-			r.Header.Set(k, strings.Repeat("*", l))
+			r.Header.Set(k, "********")
 		}
 	}
 	reqBytes, err := httputil.DumpRequestOut(r, true)
--- a/pkg/util/logging_http_transport_test.go
+++ b/pkg/util/logging_http_transport_test.go
@@ -87,6 +87,8 @@ func TestElideAuthorizationHeader(t *testing.T) {
 	assert.NilError(t, e)
 	s := out.String()
 	assert.Assert(t, strings.Contains(s, "REQUEST"))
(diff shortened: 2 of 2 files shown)


daisy-ycguo:
@maximilien Thanks for review. Test added.

pkg/util/logging_http_transport.go
```diff
@@ -56,11 +55,7 @@ func (t *LoggingHttpTransport) RoundTrip(r *http.Request) (*http.Response, error
 	for k, v := range r.Header {
 		if sensitiveRequestHeaders.Has(k) {
 			redacted[k] = v
-			l := 0
-			for _, h := range v {
-				l += len(h)
-			}
-			r.Header.Set(k, strings.Repeat("*", l))
+			r.Header.Set(k, "(REDACTED FOR PRIVACY)")
```
rhuss: Do we really have to be that verbose? I would be happy to just show 8 asterisks ("**********"). This seems to be a common practice when it comes to erasing passwords. 

Why for 'privacy' ? And not for 'security' ? Is it because the password reveals something personal ?. And "redacted" means to be changed based on the value, but it's just "overwritten" without condition, not "redacted".

Nit-picking on high-level, sorry ;-) Just saying, that I really would prefer a simple commonly used, non-sense value like "**********", as, because it's non-sense, nobody questions its sense ;-)

If my comment goes over the top, feel free to merge nevertheless.

daisy-ycguo: I think we just need to agree on the verbose to show instead of the real authorization string.

rhuss: Let's do a quick vote.

rhuss: Option 1:`(REDACTED FOR PRIVACY)`

rhuss: Option 2: `********`

sixolet: In my first version of this PR it was "ELIDED", and people were like "what does ELIDED even mean?" so I think asterisks are fine.

sixolet: WG meeting approves the asterisks.

daisy-ycguo: Will update today. Thank you all.

daisy-ycguo:
@rhuss @maximilien I updated based on comments. It's ready for review. Thank you.

PR #491, opened 2019-11-09
In waiting for ready, watch falls back to polling.

### LINKED ISSUE #325: Polling watcher option
Right now kn watches resources as they're getting `Ready` using a watch request. Watch is only SHOULD in the Knative API, so we should support a polling normal GET too, as an option. It should be settable both at the command line and in the kn config file.

sixolet:
Fixes #325

Add a new way of using a polling function to generate a watch.Interface(), and fall back to that if the server doesn't support watch. Not every knative server supports watch.


sixolet:
The test issue is some `go.sum` checksum mismatch I'm not sure how to resolve.

rhuss:
@sixolet please rebase on master, I've fixed that.

pkg/wait/poll_watcher.go
```diff
@@ -0,0 +1,189 @@
...
+					w.result <- watch.Event{
+						Type: watch.Error,
+					}
+				} else if old == nil && new != nil {
+					// Added
+					w.result <- watch.Event{
+						Type:   watch.Added,
+						Object: new,
+					}
+				} else if !(ok1 && ok2) {
```
navidshaikh: Should we separate the if check for `err`, `ok1` and `ok2` ? (OR may be a nested switch)

pkg/wait/poll_watcher.go
```diff
@@ -0,0 +1,189 @@
...
+
+	api_errors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/rest"
+)
+
+// PollInterval determins when you should poll.  Useful to mock out, or for
```
navidshaikh: nit: *determines

pkg/wait/wait_for_ready.go
```diff
@@ -110,20 +104,9 @@ func (w *waitForReadyConfig) Wait(name string, timeout time.Duration, msgCallbac
 	}
 }
 
-func addWatchTimeout(opts *v1.ListOptions, timeout time.Duration) {
```
navidshaikh: should be accessible in `poll_watcher.go` as well (as we moved this in `poll_watcher.go`)

Issue #317, opened 2019-07-30
Consider supporting multiple NAMEs on some commands


duglin:
## In what area(s)?

Classifications:
/kind feature
/kind proposal

## Describe the feature:

It would be nice if some commands supported multiple resource names to make life easier for people. For example, I was cleaning up my env and I tried this:
```
$ kn service delete echo helloworld foo
requires the service name.
```
and it died because it wasn't expecting more than one service name. It would be nice if we could support multiple NAMEs on certain commands where it makes sense.

It's also kind of an odd error message right now. I would have expected something more like `only one service name is allowed`, not a message implying I didn't specify any at all.

navidshaikh:
This is needed, I have to clean my dev environment, it needed multiple `kn service delete` commands.

I think we should implement this for `kn service delete` as well as `kn revision delete`
along with say `--all` option to delete all the services or revisions in current namespace.

> It's also kind of an odd error message right now. I would have expected something more like only one service name is allowed, not a message implying I didn't specify any at all.

This is bug.

wslyln:
@daisy-ycguo  just wanted to double check you're not working on multiple names on revision delete

I'd be up to work on it revision delete and trigger delete.

I checked the rest of the commands in kn and it doesn't seem like there are any others that multiple arguments really make sense. (I guess there might be a case of updating multiple services at a time, though it'd be pretty tricky)

duglin:
/remove-lifecycle stale
@daisy-ycguo  @wslyln were one of you going to work on this?

Kaustubh-pande:
I'd like to work on this issue.
/assign

duglin:
#1217 doesn't actually address this (at least I don't think it does)

duglin:
I missed that this was done before #1217 at least for delete
/close

PR #492, opened 2019-11-11
Support multiple NAMEs on service delete

daisy-ycguo:
Partially address #317

## Proposed Changes

* Support multiple NAMEs on `kn service delete`

Release Note:

In the following cases, add a short description of PR to the unreleased section in CHANGELOG.adoc:

- 🧽 Support multiple NAMEs on `kn service delete`

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -12,6 +12,17 @@
 |===
 ////
 
+## v0.11.0 (unreleased)
+
+[cols="1,10,3", options="header", width="100%"]
+|===
+| | Description | PR
+
+| 🧽
+| Support multiple NAMEs on kn service delete
+| https://github.com/knative/client/pull/492[#492]
+|===
+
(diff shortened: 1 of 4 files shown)


rhuss:
@daisy-ycguo please rebase on master, this should fix the test issues.

daisy-ycguo:
Thank you for review. Your comment is addressed. An e2e test case is added.

> We definitely need an e2e test for this.
> 
> What happens if the list has bogus services in it. Should the whole command fail or continue. Likely fail as you have implemented but a message to user that the remaining services did not get deleted is likely needed.
> 
> Let's be clear about the failure scenarios and add appropriate tests.
> 
> Thanks for the contribution.

daisy-ycguo:
> looks good to me, please also add a line to CHANGELOG.adoc for this change. thanks !

Thank you for review. I've add a line to CHANGELOG.adoc.

PR #503, opened 2019-11-13
Update vendor/knative.dev/test-infra

adrcunha:
Major changes:
* display E2E cluster setup during tests
* fix parsing error for autogenerated code in flaky test reporter
* fix automatic dot releases for 0.10+ branches

Part of https://github.com/knative/test-infra/issues/1542

/cc @grantr

PR #504, opened 2019-11-13
Update vendor/knative.dev/test-infra

adrcunha:
Major changes:

* display E2E cluster setup during tests
* fix parsing error for autogenerated code in flaky test reporter
* fix automatic dot releases for 0.10+ branches

Part of knative/test-infra#1542

This is a copy of #503, but running `hack/build.sh -c` in the same commit since it looks like `verify-codegen.sh` didn't like when I ran it in a subsequent commit (dunno why).

/cc @grantr
/cc @sixolet


adrcunha:
This is useless, I got the same error...

PR #514, opened 2019-11-27
Add common codes for sources client

daisy-ycguo:
## Proposed Changes

* Add common codes for sources client, for easy collaboration


pkg/kn/commands/types.go
```diff
@@ -56,6 +59,9 @@ func (params *KnParams) Initialize() {
 	if params.NewClient == nil {
 		params.NewClient = params.newClient
 	}
+	if params.NewSourcesClient == nil {
+		params.NewSourcesClient = params.newSourcesClient
```
navidshaikh: :+1:

PR #522, opened 2019-11-28
Remove explicit flag parsing so that it works with go1.13

chaodaiG:
as said

part of: https://github.com/knative/test-infra/pull/1558


chaodaiG:
> emitMetrics

just to answer the question, it's not set by default in test-infra: https://github.com/knative/test-infra/blob/c65d7bf63c2dde11173923127b91d87965d61533/scripts/e2e-tests.sh#L458

Not sure why this was defined initially here, we should be able to remove this file safely?

navidshaikh:
> we should be able to remove this file safely?

Not set/required by client tests, but reading https://github.com/knative/test-infra/pull/1558#issuecomment-558405923 from @chizhg looks like its required by CI jobs.

PR #540, opened 2019-12-09
Remove unused emit-metrics flag for e2e tests

chizhg:
Fixes https://github.com/knative/test-infra/issues/1568

## Proposed Changes
The `--emit-metrics` can be safely removed now as the relevant job has been deprecated.

/cc @adrcunha 
/cc @chaodaiG


adrcunha:
This doesn't close 1568.

PR #536, opened 2019-12-06
kn source list-types (builtin types)

navidshaikh:
Part of #511 

/lint

## Proposed Changes
 - Only lists the builtin source types
 - Uses client-go dynamic client for listing CRDs
 - Adds DyanmicClient interface to KnParams
 - Adds printing options

<details><summary>Example:</summary>
<p>

Command:
```sh
kn source list-types
TYPE              NAME                                            DESCRIPTION
ApiServerSource   apiserversources.sources.eventing.knative.dev   Kubernetes API Server events source
ContainerSource   containersources.sources.eventing.knative.dev   Container events source
CronJobSource     cronjobsources.sources.eventing.knative.dev     CronJob events source
```

Print flags:
```
./kn source list-types --help    
List available source types

Usage:
  kn source list-types [flags]

Flags:
      --allow-missing-template-keys   If true, ignore any errors in templates when a field or map key is missing in the template. Only applies to golang and jsonpath output formats.
                                      (default true)
  -h, --help                          help for list-types
  -n, --namespace string              Specify the namespace to operate in.
      --no-headers                    When using the default output format, don't print headers (default: print headers).
  -o, --output string                 Output format. One of: json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-file.
      --template string               Template string or path to template file to use when -o=go-template, -o=go-template-file. The template format is golang templates
                                      [http://golang.org/pkg/text/template/#pkg-overview].

Global Flags:
      --config string       kn config file (default is $HOME/.kn/config.yaml)
      --kubeconfig string   kubectl config file (default is $HOME/.kube/config)
      --log-http            log http traffic
```
</p>
</details>

Issue #499, opened 2019-11-13
kn source apiserver - manage ApiServerSources 


daisy-ycguo:
/kind feature

## Describe the feature:

It's a separate item of Eventing MVP ["kn source" create & update typed builtin](https://docs.google.com/document/d/1B8VK-5nuxt3D-i1Up3onmUbzcztQXE0GRgf1lfx49P8/edit#bookmark=id.d382ob8kwx0f), for easily management.

Add commands to create, delete and list ApiServerSource. See below as an example:

```
kn source apiserver create k8sevents --resources Event:v1:true --service-account myaccountname --sink svc:mysvc
kn source apiserver delete k8sevents
kn source apiserver list
```

------------------------

- [x] `kn source apiserver create` - PR #415
- [x] `kn source apiserver update`
- [x] `kn source apiserver describe`
- [x] `kn source apiserver delete` - PR #415

Issue #375, opened 2019-08-19
Use it  as SDK?


magicLian:
If this repo is the Knative Go SDK?   And docs?

sixolet:
It's our intent that you should eventually be able to use it as an SDK. Use caution, right now, as we haven't gone through and marked the APIs we think are stable enough for external use.

PR #542, opened 2019-12-09
feature(source cronjobs): Implementation of CronJobSource management

rhuss:
Contains create/delete/update/describe but not tests yet.

Still todo:

 * Think how to provide a namespace for the sink
 * Support for more sinks
 * Synchronous mode for create & update
 * Add list (or implement "source list")


rhuss:
Still todo: Adding tests.

This PR is based on #499 which should be merged first.

pkg/kn/commands/source/cronjob/client.go
```diff
@@ -0,0 +1,92 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cronjob
+
+import (
+	"fmt"
+
+	apis_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
navidshaikh:
```suggestion

```

pkg/kn/commands/source/cronjob/client.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	if sink != nil {
+		source.Spec.Sink = sink
+	}
+	_, err = c.client.CronJobSources(c.namespace).Update(source)
+	return err
+}
+
+func (c *CronSourceClient) DeleteCronJobSource(name string) error {
+	return c.client.CronJobSources(c.namespace).Delete(name, &apis_v1.DeleteOptions{})
```
navidshaikh:
```suggestion
	return c.client.CronJobSources(c.namespace).Delete(name, &meta_v1.DeleteOptions{})

```

pkg/kn/commands/source/cronjob/cronjob.go
```diff
@@ -0,0 +1,46 @@
...
+	return cronImporterCmd
+}
+
+func newCronJobSourceClient(p *commands.KnParams, cmd *cobra.Command) (*CronSourceClient, error) {
+	namespace, err := p.GetNamespace(cmd)
+	if err != nil {
+		return nil, err
+	}
+
+	config, err := p.GetClientConfig()
```
navidshaikh: adding a note here for related change in another PR, have added `prepareRestConfig` [here](https://github.com/knative/client/pull/536/files#diff-52c760be60a1c01977e8e9bcac3167c3R108) , avoids further processing of client config and can directly be passed for creating clientset, we can refine later for aligning this.

pkg/kn/commands/source/cronjob/create.go
```diff
@@ -0,0 +1,73 @@
...
+	cmd := &cobra.Command{
+		Use:   "create NAME --schedule SCHEDULE --sink SINK --data DATA",
+		Short: "Create a Crontab scheduler source.",
+		Example: `
+  # Create a crontab scheduler 'my-cron-trigger' which fires every minute and sends 'ping' to service 'mysvc' as a cloudevent
+  kn source cronjob create my-cron-trigger --schedule "* * * * */1" --data "ping" --sink svc:mysvc`,
+
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			if len(args) != 1 {
+				return errors.New("'name of the source to create required")
```
navidshaikh:
```suggestion
                              return errors.New("requires the name of the crobjob source to create as single argument")

```

pkg/kn/commands/source/cronjob/delete.go
```diff
@@ -0,0 +1,55 @@
...
+func NewCronJobDeleteCommand(p *commands.KnParams) *cobra.Command {
+	ApiServerDeleteCommand := &cobra.Command{
+		Use:   "delete NAME",
+		Short: "Delete a CronJob source.",
+		Example: `
+  # Delete a CronJob source 'my-cron-trigger'
+  kn source cronjob delete my-cron-trigger`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
+				return errors.New("'kn source cronjob delete' requires the name of the crobjob source to delete as single argument")
```
navidshaikh:
```suggestion
				return errors.New("'requires the name of the crobjob source to delete as single argument")
```

pkg/kn/commands/source/cronjob/describe.go
```diff
@@ -0,0 +1,109 @@
...
+
+	cronJobDescribe := &cobra.Command{
+		Use:   "describe NAME",
+		Short: "Describe a CronJob source.",
+		Example: `
+  # Describe a cronjob source with name 'my-cron-trigger'
+  kn source cronjob describe my-cron-trigger`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
+				return errors.New("source name required as argument")
```
navidshaikh: lets align to other similar messages

docs/cmd/kn_source_cronjob.md
```diff
@@ -0,0 +1,34 @@
...
+```
+      --config string       kn config file (default is $HOME/.kn/config.yaml)
+      --kubeconfig string   kubectl config file (default is $HOME/.kube/config)
+      --log-http            log http traffic
+```
+
+### SEE ALSO
+
+* [kn source](kn_source.md)	 - Event Source command group
+* [kn source cronjob create](kn_source_cronjob_create.md)	 - Create a Crontab scheduler source.
```
daisy-ycguo: Some use `Crontab scheduler source` and some use `CronJob source`. I think it's better to use a consistent word.

pkg/kn/commands/source/source.go
```diff
@@ -19,14 +19,16 @@ import (
...
+	"knative.dev/client/pkg/kn/commands/source/cronjob"
 )
 
 func NewSourceCommand(p *commands.KnParams) *cobra.Command {
 	sourceCmd := &cobra.Command{
 		Use:   "source",
 		Short: "Event Source command group",
 	}
-	//sourceCmd.AddCommand(cron.NewCronCommand(p))
+	//sourceCmd.AddCommand(cron.NewCronJobCommand(p))
```
daisy-ycguo: please remove this line of comment

docs/cmd/kn_source_cronjob.md
rhuss: > Just as knative serving API invocations go to `pkg/serving`, I think all eventing server side invocation methods shall go to `pkg/eventing`. Because some day, we may want to expose these server side invocations as a client-go library. Forgive me if I'm wrong. I think it's one of team decisions.

I might have missed that discussion, do you have some pointers for me ? 

The real design decision is whether we want to make the sources support self-contained and all under one package (which make it easier to refactor e.g. when they will be moved away from eventing proper), or whether we want expose them as API and maintaining it (in that case we should build up the same structure as for serving).

I personally tend to the 'self-contained' approach for sources, but would also be happy to go over eventing pkg.

rhuss: But yeah, probably better to be consistent. Let me change this later. thanks

daisy-ycguo:
@rhuss Here is a short discussion about sdk: https://github.com/knative/client/issues/375. I would strongly suggest that we put all server (serving+eventing) side functions together in a separated folder of command lines.

For end users, it will be very hard for them to use Knative API directly. Knative needs a sdk-go for end users to manage the server side objects, using simple type parameters, not complicated K8s type of objects. In the future, when community decide to have a knative sdk-go, Knative Client could easily separate the server side methods and the command lines, and easily create the basic version of knative sdk-go.

rhuss:
> For end users, it will be very hard for them to use Knative API directly because of the complex K8s CRD types. Knative needs a sdk-go for end-users to manage the server-side objects, using simple type parameters, not complicated K8s types. In the future, when community decide to have a knative sdk-go, Knative Client could easily separate the server side methods and the command lines, and easily create the basic version of knative sdk-go.

Even our API are using "complex CRD types", actually the same as the Knative client SDKs. The benefit of a client-side SDK is not so much hiding the Knative types (there are already a simplification over the standard K8s CRs), but to provide extra, client-related features like:

* CRD version alignment to also work with older clusters (still to implement, currently we send v1alpha1 objects to server but we could adapt this depending on the cluster's version).
* Some syntactic sugar, like creating the client for a specific namespace or some easy way to provide list options.

So not really much (except maybe the version negotiation to come which is probably quite helpful for supporting multiple cluster versions), but I'm not sure if this is what you have in mind. How would a `kn` based Go library much simpler to use than the generated sdk libraries from knative-serving and knative-eventing that kn is using internally? Do you have an example in mind? If you mean to translate all CRDs field to method arguments, I would be strongly against this. Nobody wants methods with 20+ arguments. Maybe we could help with a kind of builder patterns, but at the end, complexity just does not go simply away.

pkg/kn/commands/source/cronjob/cronjob.go
rhuss: +1 I updated to move `prepareRestConfig` to `RestConfig` so that I can access it from here. The issue why I'm not using `param.NewSourcesClient()` and `KnTestCommand` here is, that ...

* ... it does not really deal well when two fake clients are needed. That ApiServerSource create tests work is by accident, as the serving FakeService is not prepared and just returns an empty service when the create command looks up the sink.
* ... the approach doesn't work well with the mock tests (there are preparing the commands on their own to inject the Mockclient instead of a fake service)
* ... its IMO better decoupled if every command can look up its client on its own (which makes later refactoring easier and one does not always have to touch top-level components like KnParam to introduce new features)


I suggest to merge this PR nevertheless now, and I will immediately jump on to refactor the multi client management afterwards.

pkg/kn/commands/source/cronjob/describe.go
rhuss: 👍 for now, but we should go over the whole wording anyway. Not sure if repeating the command just called is helpful. But totally agree that it should be consistent, so I updated it here.

pkg/kn/commands/source/cronjob/cronjob.go
navidshaikh: Alright! :+1:

pkg/eventing/sources/v1alpha1/apiserver_client.go
```diff
@@ -0,0 +1,72 @@
...
+	client_v1alpha1 "knative.dev/eventing/pkg/client/clientset/versioned/typed/sources/v1alpha1"
+
+	kn_errors "knative.dev/client/pkg/errors"
+)
+
+// Interface for working with ApiServer sources
+type KnApiServerSourcesClient interface {
+
+	// Get an ApiServerSource by object
+	CreateApiServerSource(apisvrsrc *v1alpha1.ApiServerSource) (*v1alpha1.ApiServerSource, error)
```
navidshaikh: FYI - In the apiserver update PR I am preparing, I've updated this function definition to only return the `error` as the created object is not used (only used in the AddReactor styles tests and object is used to check the name and namespace).

pkg/kn/commands/types.go
```diff
@@ -104,8 +104,8 @@ func (params *KnParams) newDynamicClient(namespace string) (dynamic_kn.KnDynamic
 	return dynamic_kn.NewKnDynamicClient(client, namespace), nil
 }
 
-// prepareRestConfig returns REST config, which can be to use to create specific clientset
-func (params *KnParams) prepareRestConfig() (*rest.Config, error) {
+// RestConfig returns REST config, which can be to use to create specific clientset
+func (params *KnParams) RestConfig() (*rest.Config, error) {
 	var err error
```
navidshaikh: :+1:

PR #548, opened 2019-12-11
chore: Refactor mock testing

rhuss:
This makes it easier for other components (like eventing) to introduce an own
mock for their tests.

### DIFF
--- a/pkg/kn/commands/service/list_mock_test.go
+++ b/pkg/kn/commands/service/list_mock_test.go
@@ -17,60 +17,21 @@ package service
 import (
 	"strings"
 	"testing"
-	"time"
 
 	"gotest.tools/assert"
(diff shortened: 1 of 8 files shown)


pkg/kn/commands/service/list_mock_test.go
```diff
@@ -17,60 +17,21 @@ package service
...
 )
 
 func TestServiceListAllNamespaceMock(t *testing.T) {
-	client := knclient.NewMockKnClient(t)
-
+	client := knclient.NewMockKnServiceClient(t, "default")
 	r := client.Recorder()
+	setupListExpectations(r)
 
-	r.GetService("svc1", nil, errors.NewNotFound(v1alpha1.Resource("service"), "svc1"))
```
rhuss: removed that parts as they are not about the list test and already covered in the create tests.

PR #562, opened 2019-12-17
Add trigger update command

daisy-ycguo:
## Proposed Changes

* Add trigger update command

### DIFF
--- a/pkg/eventing/v1alpha1/client.go
+++ b/pkg/eventing/v1alpha1/client.go
@@ -23,6 +23,7 @@ import (
 	"knative.dev/eventing/pkg/apis/eventing/v1alpha1"
 	"knative.dev/eventing/pkg/client/clientset/versioned/scheme"
 	client_v1alpha1 "knative.dev/eventing/pkg/client/clientset/versioned/typed/eventing/v1alpha1"
+	duckv1 "knative.dev/pkg/apis/duck/v1"
 )
 
 const (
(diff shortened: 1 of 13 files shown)


pkg/eventing/v1alpha1/client.go
```diff
@@ -118,3 +131,70 @@ func (c *knEventingClient) Namespace() string {
...
+		triggerFilterAttributes := v1alpha1.TriggerFilterAttributes(filters)
+		b.trigger.Spec.Filter = &v1alpha1.TriggerFilter{
+			Attributes: &triggerFilterAttributes,
+		}
+	}
+	return b
+}
+
+// UpdateFilters to update the filters of trigger object
+func (b *TriggerBuilder) UpdateFilters(toUpdate map[string]string, toRemove []string) *TriggerBuilder {
```
rhuss: I wouldn't do any removal or change in the builder. Its about creating an immutable object, so would prefer to have just a `Filter()` method (as in my PR ;-) if you don't mind I would update that later.

pkg/eventing/v1alpha1/client.go
```diff
@@ -118,3 +131,70 @@ func (c *knEventingClient) Namespace() string {
...
+	}
+	for _, key := range toRemove {
+		delete(existing, key)
+	}
+	b.Filters(existing)
+	return b
+}
+
+// Sink to set the subscriber of trigger object
+func (b *TriggerBuilder) Sink(sink *duckv1.Destination) *TriggerBuilder {
```
rhuss: Let's call it Subscriber like in the spec.

pkg/eventing/v1alpha1/client.go
daisy-ycguo: Find. I don't mind. Please update it in a separate PR.

Issue #563, opened 2019-12-17
Improve sink/subscriber support


rhuss:
## In what area(s)?

Eventing

## Describe the feature:

The following features are needed for proper sink/subscriber handling:

- [x] * Add more sink types like define in the spec
- [x] * Allow an URL to be provided
- [ ] * Allow specifying the namespace for the sink (currently: Always default)

navidshaikh:
Broker, knative service and URL can be specified, more types can be configured via sink prefixes in the config.

Currently sink is looked up in the namespace kn is operating in, updated issue description with unchecked item for specifying namespace of the sink.

lionelvillard:
Since sources natively support sending events to plain k8s services, it'd be nice to provide a shortcut, e.g. `svc:my-svc`.

rhuss:
> Since sources natively support sending events to plain k8s services, it'd be nice to provide a shortcut, e.g. `svc:my-svc`.

We discussed this, but we felt that referring to vanilla Kubernetes objects is outside the 'Knative contract'. (e.g. will it also work on CloudRun ?). 

See https://github.com/knative/client/issues/898 for a more in deep discussion on this topic.

rhuss:
Although this issue has been closed, I think we should still think about how to allow the specification of a different namespace for a cross-namespace eventing mesh. 

I create a new issue for this: https://github.com/knative/client/issues/1143

PR #574, opened 2019-12-18
Allow for http and broker sinks

### LINKED ISSUE #570: Allow source creation with brokers
Allow `broker:default` and other brokers as sinks in source creation.

Possibly allow the default broker as a default sink for a source.

sixolet:
Refactors sink code to use a dynamic client to validate the sink, so we don't keep having to write the same parallel code. Paves the way for arbitrary user-configured sinks.

Fixes #570 

Also fixes most of #563 : missing arbitrary ns for sinks, but now has the sink as being in the same ns as the object being created.

### DIFF
--- a/pkg/dynamic/client.go
+++ b/pkg/dynamic/client.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/fake"
 
+	eventing_v1alpha1 "knative.dev/eventing/pkg/apis/eventing/v1alpha1"
 	serving_v1alpha1 "knative.dev/serving/pkg/apis/serving/v1alpha1"
 )
 
@@ -102,6 +103,7 @@ func (c knDynamicClient) RawClient() dynamic.Interface {
 func CreateFakeKnDynamicClient(testNamespace string, objects ...runtime.Object) KnDynamicClient {
 	scheme := runtime.NewScheme()
 	scheme.AddKnownTypeWithName(schema.GroupVersionKind{Group: "serving.knative.dev", Version: "v1alpha1", Kind: "Service"}, &serving_v1alpha1.Service{})
+	scheme.AddKnownTypeWithName(schema.GroupVersionKind{Group: "eventing.knative.dev", Version: "v1alpha1", Kind: "Broker"}, &eventing_v1alpha1.Broker{})
 	client := fake.NewSimpleDynamicClient(scheme, objects...)
 	return NewKnDynamicClient(client, testNamespace)
 }
--- a/pkg/dynamic/client.go
+++ b/pkg/dynamic/client.go
@@ -18,13 +18,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
-	"k8s.io/client-go/dynamic/fake"
-
-	eventing_v1alpha1 "knative.dev/eventing/pkg/apis/eventing/v1alpha1"
-	serving_v1alpha1 "knative.dev/serving/pkg/apis/serving/v1alpha1"
 )
 
 const (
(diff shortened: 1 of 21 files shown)


pkg/kn/commands/flags/sink_test.go
```diff
@@ -0,0 +1,76 @@
...
+		ObjectMeta: metav1.ObjectMeta{Name: "mysvc", Namespace: "default"},
+	}
+	defaultBroker := &eventing_v1alpha1.Broker{
+		TypeMeta:   metav1.TypeMeta{Kind: "Broker", APIVersion: "eventing.knative.dev/v1alpha1"},
+		ObjectMeta: metav1.ObjectMeta{Name: "default", Namespace: "default"},
+	}
+
+	assert.NilError(t, err)
+	cases := []resolveCase{
+		{"svc:mysvc", &duckv1beta1.Destination{
```
rhuss: maybe also a test with "service:mysvc" ?

pkg/kn/commands/source/apiserver/create.go
```diff
@@ -51,13 +51,11 @@ func NewAPIServerCreateCommand(p *commands.KnParams) *cobra.Command {
...
 
-			// create Serving client for resolving service sink
-			servingClient, err := p.NewServingClient(namespace)
+			dynamicClient, err := p.NewDynamicClient(namespace)
 			if err != nil {
 				return err
 			}
-
-			objectRef, err := sinkFlags.ResolveSink(servingClient)
+			objectRef, err := sinkFlags.ResolveSink(dynamicClient.RawClient(), namespace)
```
rhuss: as mentioned elsewhere, better use `dynamicClient` directly as argurment.

sixolet: Done

pkg/dynamic/client.go
```diff
@@ -87,3 +95,16 @@ func (c *knDynamicClient) ListSourcesTypes() (*unstructured.UnstructuredList, er
 	options.LabelSelector = sourcesLabels.String()
 	return c.ListCRDs(options)
 }
+
+func (c knDynamicClient) RawClient() dynamic.Interface {
+	return c.client
+}
+
+// CreateFakeKnDynamicClient gives you a dynamic client for testing contianing the given objects.
+func CreateFakeKnDynamicClient(testNamespace string, objects ...runtime.Object) KnDynamicClient {
```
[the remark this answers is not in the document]
sixolet: I have put the Fake helper method in a different package, even, because you're right it's nice to have test code decoupled from mainline code. On the other hand, I really like the mocking scheme here. Using it was refreshing and much nicer — I would much rather specify the objects in my "fake server" than specify _every_ call to it.. I'll keep this mocking scheme. Might even look to use it in more places.

pkg/kn/commands/source/cronjob/update.go
```diff
@@ -48,7 +48,11 @@ func NewCronJobUpdateCommand(p *commands.KnParams) *cobra.Command {
 				return err
 			}
 
-			servingClient, err := p.NewServingClient(cronSourceClient.Namespace())
+			namespace, err := p.GetNamespace(cmd)
+			if err != nil {
+				return err
+			}
+			dynamicClient, err := p.NewDynamicClient(namespace)
```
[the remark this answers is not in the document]
sixolet: I'll look at sink namespace in another PR

pkg/kn/commands/flags/sink_test.go
sixolet: Done

pkg/kn/commands/trigger/create.go
```diff
@@ -50,17 +50,17 @@ func NewTriggerCreateCommand(p *commands.KnParams) *cobra.Command {
...
 			}
 
-			servingClient, err := p.NewServingClient(namespace)
+			eventingClient, err := p.NewEventingClient(namespace)
 			if err != nil {
 				return err
 			}
 
-			objectRef, err := sinkFlags.ResolveSink(servingClient)
+			objectRef, err := sinkFlags.ResolveSink(dynamicClient.RawClient(), namespace)
```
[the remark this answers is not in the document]
sixolet: done

pkg/dynamic/client_test.go
```diff
@@ -99,3 +99,13 @@ func TestListSourceTypes(t *testing.T) {
 		assert.Equal(t, uList.Items[1].GetName(), "bar")
 	})
 }
+
+// createFakeKnDynamicClient gives you a dynamic client for testing contianing the given objects.
```
navidshaikh:
```suggestion
// createFakeKnDynamicClient gives you a dynamic client for testing containing the given objects.
```

PR #575, opened 2019-12-19
Retrurn error if exist duplicate key in filter option

### LINKED ISSUE #569: Error when multiple filter with the same name are given for kn trigger
In `kn trigger --filter type=bla --filter type=blub`, "blub" will silently win. As this is very likely not the user's intent, an error should be thrown instead.

Same btw for "kn service create --env". The check could be part of a unified list-option handling.

MIBc:
Fixes #569

## Proposed Changes

*Return error if user set duplicate key in `--filter` option. For example, `kn trigger create --filter type=bla --filter type=blub test`.


rhuss:
or then maybe in a next PR ;-)

rhuss:
(was too slow ...)

MIBc:
Sure, I will do.

rhuss:
@MIBc If you like, you could also jump on #577 which I just have created. I think this would be the most comprehensive solution.

MIBc:
Yes. I will try. Thanks!

PR #581, opened 2019-12-20
feature(#580): adds traffic and tags information to revision list

### LINKED ISSUE #580: kn revision list does not show traffic and tagging details
## In what area(s)?

`kn revision`

Classifications:

> /kind feature

## Describe the feature:

During my usage of `kn` for demo (see: github.com/maximilien/knfun/README.md/#traffic-splitting) I realize that the current `kn revision list` is not very useful if it does not show the traffic and tags since these are important information that users need when looking at revision.

maximilien:
Fixes #580

## Proposed Changes

* adds traffic information of revision as an annotation in revision list
* adds tags information of revision as an annotation in revision list
* show traffic and tags when listing revisions

With this change, invocation of `kn revision list` will look like:

```bash
➜  client git:(issue580) ✗ ./kn revision list
NAME                  SERVICE       GENERATION   AGE   CONDITIONS   READY   TRAFFIC   TAGS    REASON
twitter-fn-ycybl-1    twitter-fn    26           28d   3 OK / 4     True    100%
summary-fn-hwkll-31   summary-fn    20           23d   3 OK / 4     True    50%       async
summary-fn-qwfdb-1    summary-fn    19           23d   3 OK / 4     True    50%       sync
watson-fn-wxpyf-15    watson-fn     15           28d   3 OK / 4     True    100%
hello-world-tndhb-2   hello-world   2            28d   3 OK / 4     True    100%

➜  client git:(issue580) ✗ ./kn revision list -s summary-fn
NAME                  SERVICE      GENERATION   AGE   CONDITIONS   READY   TRAFFIC   TAGS    REASON
summary-fn-hwkll-31   summary-fn   20           23d   3 OK / 4     True    50%       async
summary-fn-qwfdb-1    summary-fn   19           23d   3 OK / 4     True    50%       sync
```

<!--
Release Note:

- 🎁 adds revision traffic and tags information when listing revisions

### DIFF
--- a/pkg/kn/commands/revision/list.go
+++ b/pkg/kn/commands/revision/list.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"sort"
 	"strconv"
+	"strings"
 
 	"knative.dev/serving/pkg/apis/serving"
 
@@ -85,6 +86,23 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
 				}
 			}
 
+			// Add revision/service info as annotations: traffic and tags
+			var service *v1alpha1.Service
+			var serviceName string
+			for _, revision := range revisionList.Items {
+				if serviceName == "" || revision.Labels[serving.ServiceLabelKey] != serviceName {
+					serviceName = revision.Labels[serving.ServiceLabelKey]
+					service, err = client.GetService(serviceName)
+					if err != nil {
+						return err
+					}
+				}
+
+				traffic, tags := trafficForRevision(revision.Name, service)
+				revision.Annotations[RevisionTrafficAnnotation] = fmt.Sprintf("%d%%", traffic)
+				revision.Annotations[RevisionTagsAnnotation] = strings.Join(tags, ",")
+			}
+
 			// sort revisionList by configuration generation key
 			sort.SliceStable(revisionList.Items, func(i, j int) bool {
 				a := revisionList.Items[i]
@@ -110,6 +128,7 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
 			if err != nil {
 				return err
 			}
+
 			err = printer.PrintObj(revisionList, cmd.OutOrStdout())
 			if err != nil {
 				return err
--- a/pkg/kn/commands/revision/human_readable_flags.go
+++ b/pkg/kn/commands/revision/human_readable_flags.go
@@ -23,11 +23,18 @@ import (
 	servingv1alpha1 "knative.dev/serving/pkg/apis/serving/v1alpha1"
 )
 
+const (
+	RevisionTrafficAnnotation = "client.knative.dev/traffic"
+	RevisionTagsAnnotation    = "client.knative.dev/tags"
+)
+
 // RevisionListHandlers adds print handlers for revision list command
 func RevisionListHandlers(h hprinters.PrintHandler) {
 	RevisionColumnDefinitions := []metav1beta1.TableColumnDefinition{
 		{Name: "Name", Type: "string", Description: "Name of the revision.", Priority: 1},
 		{Name: "Service", Type: "string", Description: "Name of the Knative service.", Priority: 1},
+		{Name: "Traffic", Type: "string", Description: "Percentage of traffic assigned to this revision.", Priority: 1},
+		{Name: "Tags", Type: "string", Description: "Set of tags assigned to this revision.", Priority: 1},
 		{Name: "Generation", Type: "string", Description: "Generation of the revision", Priority: 1},
 		{Name: "Age", Type: "string", Description: "Age of the revision.", Priority: 1},
 		{Name: "Conditions", Type: "string", Description: "Conditions describing statuses of the revision.", Priority: 1},
@@ -57,6 +64,8 @@ func printRevisionList(revisionList *servingv1alpha1.RevisionList, options hprin
 func printRevision(revision *servingv1alpha1.Revision, options hprinters.PrintOptions) ([]metav1beta1.TableRow, error) {
 	name := revision.Name
 	service := revision.Labels[serving.ServiceLabelKey]
+	traffic := revision.Annotations[RevisionTrafficAnnotation]
+	tags := revision.Annotations[RevisionTagsAnnotation]
 	generation := revision.Labels[serving.ConfigurationGenerationLabelKey]
 	age := commands.TranslateTimestampSince(revision.CreationTimestamp)
 	conditions := commands.ConditionsValue(revision.Status.Conditions)
@@ -68,6 +77,8 @@ func printRevision(revision *servingv1alpha1.Revision, options hprinters.PrintOp
 	row.Cells = append(row.Cells,
 		name,
 		service,
+		traffic,
+		tags,
 		generation,
 		age,
 		conditions,
--- a/pkg/kn/commands/revision/list_flags.go
+++ b/pkg/kn/commands/revision/list_flags.go
@@ -26,7 +26,7 @@ import (
 type RevisionListFlags struct {
 	GenericPrintFlags  *genericclioptions.PrintFlags
 	HumanReadableFlags *commands.HumanPrintFlags
-	ServiceRefFlags    ServiceReferenceFlags
+	ServiceRefFlags    *ServiceReferenceFlags
 }
 
 // AllowedFormats is the list of formats in which data can be displayed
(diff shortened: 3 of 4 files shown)


pkg/kn/commands/revision/list.go
```diff
@@ -85,6 +86,19 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
 				}
 			}
 
+			// Add revision service info as annocations: traffic and tags
+			for _, revision := range revisionList.Items {
+				serviceName := revision.Labels[serving.ServiceLabelKey]
+				service, err := client.GetService(serviceName)
```
rhuss: Can we please cache the service and don't refetch afresh for every revision (belonging to the same service) ?

pkg/kn/commands/revision/list.go
```diff
@@ -85,6 +86,19 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
...
+				if err != nil {
+					return err
+				}
+
+				traffic, tags := trafficForRevision(revision.Name, service)
+				revision.Annotations[RevisionTrafficAnnotation] = fmt.Sprintf("%d%%", traffic)
+				revision.Annotations[RevisionTagsAnnotation] = strings.Join(tags, ",")
```
rhuss: Can we avoid modifying the revision here by e.g. using a  map keyed by revision name ? Imagine if someone later would introduce such annotations for real (so that you would override it here). Also 'misusing' an object by writing to it in a pure read-only use case is not architectural sound.

maximilien: Thought of that, however, the revision is not modified permanently. Only temporarily to hold values that frankly belongs to it. These modified revision objects are in memory and are never seen by 'anyone' else other than the user that invoked the command.

pkg/kn/commands/revision/list.go
maximilien: Fair. Will do.

pkg/kn/commands/revision/list.go
navidshaikh: > Can we avoid modifying the revision here by

+1

We can use `trafficForRevision` in `printRevision` to fetch the `traffic` and `tags`.

maximilien: Read my comment @navidshaikh it does not modify the revision except in memory which makes the rest of the code super clean

maximilien:
> We should re-group the traffic and tags column next to `SERVICE` for eg:
> 
> ```
> ➜  client git:(df045ee0) ✗ ./kn revision list
> NAME           SERVICE   TRAFFIC   TAGS    GENERATION   AGE   CONDITIONS   READY   REASON
> test-wrvjj-1   test      100%      v1,v2   1            31m   3 OK / 4     True    
> ```

Done

pkg/kn/commands/revision/list.go
navidshaikh: Sure! IMO, this information doesn't belong to Revision and we shouldn't piggyback this data onto Revision.

maximilien:
@navidshaikh this is ready to merge IMO.

pkg/kn/commands/revision/list.go
rhuss: I'm still not sold to modify domain objects, even when only "in memory". You never know what happens with that objects. BTW, if you would do such thing with Hibernate (Java ORM mapper), it would be persisted at the end of a transaction, even when it's 'only in memory'. Also, I don't see why an object should be modified in a pure read-only situation (imagine that at some point in the future the Client API decides to make objects returned by read-only operations immutable by throwing an error on setting fields. Your code would break without a real reason.

It's just not a good style to misuse domain objects for temporary information. Not because something could be done easily should be done. (Imagine someone would print out the full object somewhere in between, like in a `kn revision list --verbose` or something. You will see artifical annotations here).

This can be achieved much more elegantly.

maximilien: I disagree since annotations are meant to do things like that. However, I am happy to hear suggestions on how else to do it?

maximilien: > Sure! I think this information doesn't belong to Revision and we shouldn't piggyback this data onto Revision.

It does at least when using the revision list. See why I opened this issue in the first place. It comes from the need to see this information and not having to have to show the details of the service and parsing the out to find the data. This is my feedback from real usage.

maximilien: > (Imagine someone would print out the full object somewhere in between, like in a `kn revision list --verbose` or something. You will see artifical annotations here).

`kn revision list --verbose` call does not exist. If we decide to do it we can easily filter out the annotations or better flag them as such. Which I'd be happy to also do here.

rhuss: > I disagree since annotations are meant to do things like that. However, I am happy to hear suggestions on how else to do it?

That's where I disagree, i.e. that annotations are used for volatile information which never gets persisted ;-). But maybe we could at least agree that listing revisions is a read-only operation and could be implemented with immutable objects that don't allow mutating operations? Why should we break that without a need ?

> `kn revision list --verbose` call does not exist. If we decide to do it we can easily filter out the annotations or better flag them as such. Which I'd be happy to also do here.

That's not really the point, whether this feature exists or not as its just an illustrative example. When you say one can easily add a filter, that is a bad smell, too, to use another hack to fix the previous hack. 

Not sure whether I could convince you that keeping certain architectural constraints even when they are not enforced today is a good thing. So let's merge them at let me make an alternative suggestion in the next PR.

rhuss:
Seems like a flaky error

```
customresourcedefinition.apiextensions.k8s.io/services.serving.knative.dev created
error: unable to recognize "https://storage.googleapis.com/knative-nightly/serving/latest/serving.yaml": no matches for kind "Image" in version "caching.internal.knative.dev/v1alpha1"
Installing the rest of serving components from https://storage.googleapis.com/knative-nightly/serving/latest/serving.yaml
customresourcedefinition.apiextensions.k8s.io/images.caching.internal.knative.dev unchanged
```

/retest

navidshaikh:
/retest

```
timeout: service 'hello' not ready after 600 seconds
```

pkg/kn/commands/revision/list.go
```diff
@@ -85,6 +86,23 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
 				}
 			}
 
+			// Add revision/service info as annotations: traffic and tags
+			var service *v1alpha1.Service
+			var serviceName string
+			for _, revision := range revisionList.Items {
+				if serviceName == "" || revision.Labels[serving.ServiceLabelKey] != serviceName {
```
rhuss: (just working on an followup PR, but wanna leave some extra comments): It is by no means guaranteed that revisions from the list are sorted according to service. Better store in a map for latter referral.

Please don't fix I'm on it.

pkg/kn/commands/revision/list.go
maximilien: > Not sure whether I could convince you that keeping certain architectural constraints even when they are not enforced today is a good thing. 

That's not the part that I disagree with. 

> That's where I disagree, i.e. that annotations are used for volatile information which never gets persisted ;-). 

Yup, that's the conundrum. It's the use of annotations. For me, I feel that this is a good use since the attributes I am adding `traffic` and `tags` are not really "features" of the `Service` nor the `Revision` but of both. There are synthesized from and belong to the association between `Service` and `Revision`, in my view.

> So let's merge them at let me make an alternative suggestion in the next PR.

OK cool. Look forward to it. Hopefully it's not a significant change since that was where I was headed until this approach. But maybe I am missing some easy trick... Happy to learn.

rhuss: > OK cool. Look forward to it. Hopefully it's not a significant change since that was where I was headed until this approach. But maybe I am missing some easy trick... Happy to learn.

I have to apologize a bit :) Ideally I would use classical "view object" used for printing. But the printer framework we are using insist on `runtime.Object`. To implement those is even more hackish than adding to the annotations (I already was there before I returned back to your approach, which definitely is the better in this particular case). So I suggest indeed to stick to this technique for tunneling purposes (if we would have the printing of a list in our hands, its trivial to change with a  `revisionInfo` object containing the revision + tags,traffic from the service). 

However, my point is illustrated quite nicely what happens at the moment, when you are doing a `kn revision list -o yaml`. This includes, in YAML, the temporary annotations. This is incorrect as these annotations are *not part of the revision*. We can fix that, but in this initial PR it was overlooked. Something similar can happen in the future, too. Hopefully not, but anybody touching this code needs to be aware of this particular usage of volatile information attached to an otherwise persisted object.

So, there is no easy trick. Life is complicated ;-)

PR #583, opened 2019-12-20
Support a flag to set ImagePullPolicy

### LINKED ISSUE #578: Support a flag to set ImagePullPolicy
## In what area(s)?

Classifications:
/kind good-first-issue
/kind feature
/kind proposal
/kind doc

## Describe the feature:

As I'm making changes to my app and redeploying it (over and over), I usually do ` kn service delete...` followed by `kn service create...`. However, as I rebuild my container image each time there is no guarantee that the new image will be picked up if it's already in the cache. As a result I find that I need to drop down to yaml to do the `create` so that I can specify `imagePullPolicy: Always`. It would be nice if there was some `kn` flag I could specify to give me the same net result.

MIBc:
Fixes #578

## Proposed Changes

* Add imagePullPolicy flag for service

### DIFF
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -30,11 +30,12 @@ import (
 
 type ConfigurationEditFlags struct {
 	// Direct field manipulation
-	Image   string
-	Env     []string
-	EnvFrom []string
-	Mount   []string
-	Volume  []string
+	Image           string
+	ImagePullPolicy string
+	Env             []string
+	EnvFrom         []string
+	Mount           []string
+	Volume          []string
 
 	RequestsFlags, LimitsFlags ResourceFlags
 	MinScale                   int
@@ -72,6 +73,9 @@ func (p *ConfigurationEditFlags) markFlagMakesRevision(f string) {
 func (p *ConfigurationEditFlags) addSharedFlags(command *cobra.Command) {
 	command.Flags().StringVar(&p.Image, "image", "", "Image to run.")
 	p.markFlagMakesRevision("image")
+	command.Flags().StringVar(&p.ImagePullPolicy, "pull-policy", "",
+		"Image pull policy with the possible values 'Always', 'Never' or 'IfNotPresent'")
+	p.markFlagMakesRevision("pull-policy")
 	command.Flags().StringArrayVarP(&p.Env, "env", "e", []string{},
 		"Environment variable to set. NAME=value; you may provide this flag "+
 			"any number of times to set multiple environment variables. "+
@@ -259,6 +263,17 @@ func (p *ConfigurationEditFlags) Apply(
 		}
 	} else if !p.LockToDigest {
 		servinglib.UnsetUserImageAnnot(template)
+	}
+
+	if cmd.Flags().Changed("pull-policy") {
+		if p.ImagePullPolicy == string(corev1.PullAlways) || p.ImagePullPolicy == string(corev1.PullNever) || p.ImagePullPolicy == string(corev1.PullIfNotPresent) {
+			err = servinglib.UpdateImagePullPolicy(template, p.ImagePullPolicy)
+			if err != nil {
+				return err
+			}
+		} else {
+			return fmt.Errorf("Invalid --pull-policy %q. Only support Always/Never/IfNotPresent", p.ImagePullPolicy)
+		}
 	}
 
 	limitsResources, err := p.computeResources(p.LimitsFlags)
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -12,6 +12,17 @@
 |===
 ////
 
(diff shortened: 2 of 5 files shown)


pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -261,6 +265,17 @@ func (p *ConfigurationEditFlags) Apply(
...
 	}
 
+	if cmd.Flags().Changed("imagePullPolicy") {
+		if p.ImagePullPolicy == string(corev1.PullAlways) || p.ImagePullPolicy == string(corev1.PullNever) || p.ImagePullPolicy == string(corev1.PullIfNotPresent) {
+			err = servinglib.UpdateImagePullPolicy(template, p.ImagePullPolicy)
+			if err != nil {
+				return err
+			}
+		} else {
+			return fmt.Errorf("Invalid --imagePullPolicy %s. Only support Always/Never/IfNotPresent", p.ImagePullPolicy)
```
duglin: s/%s/%q/ so it's quoted

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -72,6 +73,9 @@ func (p *ConfigurationEditFlags) markFlagMakesRevision(f string) {
 func (p *ConfigurationEditFlags) addSharedFlags(command *cobra.Command) {
 	command.Flags().StringVar(&p.Image, "image", "", "Image to run.")
 	p.markFlagMakesRevision("image")
+	command.Flags().StringVar(&p.ImagePullPolicy, "imagePullPolicy", "",
```
duglin: I wonder if we should just call it `pullPolicy` instead to save people typing? While I like consistency with the yaml, we've already deviated on other things, like `--concurrency-limit` instead of `--container-concurrency`, and I think dropping `image` is less controversial.   (yes, I'd love to switch to just `--concurrency` since the use of the word `limit` is less intuitive to me, but that's a different topic)

duglin: also - should it be `--pull-policy` since we seem to use `-`s instead of camelCase ?

rhuss: Yes, option should be dash separated. Also ok with use pull policy (but wouldn't mind image pull policy, too).

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -72,6 +73,9 @@ func (p *ConfigurationEditFlags) markFlagMakesRevision(f string) {
...
+		"Image pull policy. Support Always, Never or IfNotPresent")
```
rhuss:
```suggestion
		"Image pull policy with the possible values 'Always', 'Never' or 'IfNotPresent'")
```

sixolet:
Did image pull policy get added to the knative API when I wasn't looking? Last I checked, it's explicitly something serving abstracts over because revisions set their digest on creating

duglin:
> Did image pull policy get added to the knative API when I wasn't looking? Last I checked, it's explicitly something serving abstracts over because revisions set their digest on creating

I've been using it my my Knative yaml for a while

mattmoor:
> I've been using it my my Knative yaml for a while

To do what?  The only setting that should be relevant is "Never", unless you've disabled digest resolution.

duglin:
To pull the latest. I don't think I'm setting anything special but I've had cases where if I don't use yaml + pullPolicy=Always but instead used `kn service create ...` the image will not have the latest code.

sixolet:
That seems like a serving bug.

sixolet:
That seems like a serving bug.

pkg/kn/commands/service/configuration_edit_flags.go
MIBc: done

pkg/kn/commands/service/configuration_edit_flags.go
MIBc: done

pkg/kn/commands/service/configuration_edit_flags.go
MIBc: done

MIBc:
> Need a test for this. Ideally also an e2e test.
> 

Add unit test and e2e test. Please review.

MIBc:
> Looks good to me in general, please add also to CHANGELOG.adoc

Thanks for your suggestion. Add changelog. Please review.

mattmoor:
> That seems like a serving bug.

I'd agree, but would need to see a repro.

sixolet:
As discussed above: 
* If the revision doesn't get the latest version of the image at the tag on _creation_, it's a serving bug. 
* If it does get the latest version of the image at the tag any time _after_ creation, it's a serving bug. Please file those bugs with the serving component.

PullPolicy is designed to be nullified by Knative serving. As such, we shouldn't include a flag to manipulate it.

PR #589, opened 2019-12-20
Fix misleading error message when verb is bad

### LINKED ISSUE #557: Misleading error message when verb is bad
```
$ kn service udpate test --min-scale=0
unknown flag: --min-scale
```

Notice that I misspelled "update". I've also done this by using "edit" instead of "update" by mistake. So, yes there was a user error, however, the error message really should have said something like "unknown action `udpate`" or "unknown action `edit`". Instead it complains about `--min-scale`, which is technically correct the command I **thought** I was trying to do. So a user could spend quite a long time scratching their heads wondering why we're complaining about an option when the real error is that they asked for an invalid action.

Net: we should check the validity of the "verb" before we complain about the validity of the options.

itsmurugappan:
Fixes #557 

## Proposed Changes

* Validate sub command verbs
* Add tests


pkg/kn/core/root_test.go
```diff
@@ -40,6 +41,51 @@ func TestNewDefaultKnCommand(t *testing.T) {
...
+	rootCmd := NewDefaultKnCommandWithArgs(NewKnCommand(), pluginHandler, args, os.Stdin, os.Stdout, os.Stderr)
+	checkRootCmd(t, rootCmd)
+}
+
+func TestNewDefKnCmdWithArgsforPrimaryCmdValidityErr(t *testing.T) {
+
+	if os.Getenv("EXIT") == "1" {
+		pluginHandler := plugin.NewDefaultPluginHandler(plugin.ValidPluginFilenamePrefixes,
+			"/random/plugins", false)
+		args := []string{"kn", "srevice", "update", "--tag=0.13"}
```
maximilien: Did you mean to have a typo in “service” here as a bad verb?

pkg/kn/core/root_test.go
```diff
@@ -40,6 +41,51 @@ func TestNewDefaultKnCommand(t *testing.T) {
...
+	pluginHandler := plugin.NewDefaultPluginHandler(plugin.ValidPluginFilenamePrefixes,
+		"/random/plugins", false)
+	args := []string{"kn", "service", "update", "--tag=0.13"}
+	rootCmd := NewDefaultKnCommandWithArgs(NewKnCommand(), pluginHandler, args, os.Stdin, os.Stdout, os.Stderr)
+	checkRootCmd(t, rootCmd)
+}
+
+func TestNewDefKnCmdWithArgsforPrimaryCmdValidityErr(t *testing.T) {
+
+	if os.Getenv("EXIT") == "1" {
```
maximilien: I am not sure I understand the need for this `os.Getenv(“EXIT”)`?

Can we test without it?

test/README.md
```diff
@@ -48,10 +48,10 @@ E2E_TAGS="eventing" test/e2e-tests-local.sh
 
 ### Running a single test case
 
-To run one e2e test case, e.g. TestBasicWorkflow
+To run one e2e test case, e.g. TestTrafficSplit
```
maximilien: Since this seems to be a different topic I suggest adding to a different PR on how to run specific e2e tests. Which is valuable in itself.

pkg/kn/core/root_test.go
itsmurugappan: This is to test if primary command "service"  is mispelt

pkg/kn/core/root_test.go
itsmurugappan: This test runs a sub process. So initially we set the env variable and when the subprocess runs it check the variable to execute the block which terminates with os.Exit.

navidshaikh:
For cases like below, the error message is not correct

```
➜  client git:(5e8a5474) ./kn source apiserver noverb
[..]
unknown sub-command "apiserver" for "kn source"%                                                                                                                                              
```

We'll need utility to hook into the root kn command visiting the sub-commands and verb, and reporting accordingly.

itsmurugappan:
> For cases like below, the error message is not correct
> 
> ```
> ➜  client git:(5e8a5474) ./kn source apiserver noverb
> [..]
> unknown sub-command "apiserver" for "kn source"%                                                                                                                                              
> ```
> 
> We'll need utility to hook into the root kn command visiting the sub-commands and verb, and reporting accordingly.

@navidshaikh added an utility, to get the commands visiting the verb before reporting the error. Please review.

navidshaikh:
Sorry for coming back late on this.
Few nits:
 - Extra `%` at the end of error message. 
 - The error message gets hard to track if printed with help text, so we could remove `rootCmd.Help()` and just print the error message.

I checked knctl which adds cobra utilities for this use-case, it nicely provides the list of available commands if incorrect verb is given, would like to implement similar utility here.
/cc: @cppforlife

itsmurugappan:
> Sorry for coming back late on this.
> Few nits:
> 
> * Extra `%` at the end of error message.
> * The error message gets hard to track if printed with help text, so we could remove `rootCmd.Help()` and just print the error message.
> 
> I checked knctl which adds cobra utilities for this use-case, it nicely provides the list of available commands if incorrect verb is given, would like to implement similar utility here.
> /cc: @cppforlife

@navidshaikh Checked out knctl, i have made the below changes similar to it.
* the list of available commands will be shown when the sub command is wrong
* when the primary command is wrong , rootCmd.Help() followed "Invalid command - see available commands/subcommands above"

pkg/kn/core/root_test.go
```diff
@@ -40,6 +41,69 @@ func TestNewDefaultKnCommand(t *testing.T) {
...
+func TestNewDefKnCmdWithArgsforWrongVerb(t *testing.T) {
+
+	if os.Getenv("EXIT") == "1" {
+		pluginHandler := plugin.NewDefaultPluginHandler(plugin.ValidPluginFilenamePrefixes,
+			"/random/plugins", false)
+		args := []string{"kn", "revision", "update", "--tag=0.13"}
+		NewDefaultKnCommandWithArgs(NewKnCommand(), pluginHandler, args, os.Stdin, os.Stdout, os.Stderr)
+		return
+	}
+	cmd := exec.Command(os.Args[0], "-test.run=TestNewDefKnCmdWithArgsforWrongVerb")
```
rhuss: Nice trick to call the same test again from the outside, but I wonder whether we couldn't make it  a real unit test. This currently looks more like an integration test which should be triggered from the outsied (e.g. including it into our e2e test suite which execs to kn).

So I would suggest to call `NewDefaultKnCommndWithArgs` directly within the test, with the arguments prepared  accordingly. This is then would be a real self-container minmal unit -test (which does not have any deps on the external context, e.g. where the test is run). This is faster and more robust. Let's leave integration tests to teh e2e test suite (you could add such a test there, too)

pkg/kn/core/root.go
```diff
@@ -72,18 +72,23 @@ func NewDefaultKnCommandWithArgs(rootCmd *cobra.Command,
...
 			err := plugin.HandlePluginCommand(pluginHandler, cmdPathPieces)
 			if err != nil {
 				rootCmd.Help()
-				fmt.Fprintf(rootCmd.OutOrStderr(), "Unknown command or plugin '%s'.\n", args[1])
+				fmt.Fprintf(rootCmd.OutOrStderr(), "Invalid command '%s' - see available commands/subcommands above .\n", args[1])
+				os.Exit(1)
+			}
+		} else if foundCmd.HasSubCommands() {
+			if _, _, err := rootCmd.Find(innerArgs); err != nil {
+				fmt.Fprintf(rootCmd.OutOrStderr(), showSubcommands(foundCmd))
```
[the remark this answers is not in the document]
rhuss: Also be sure to add not only the parent command  in the error message, but the whole chain up to root (i.e. we have commands with more than one subcommand, like `kn source cronjob create ...`. The error message here should say '... unknown subcommand "bla" for "source cronjob"' (not only "... unknown subcommand "bla" for "cronjob")

pkg/kn/core/root_test.go
itsmurugappan: Thank you @rhuss .  The reason I had to do this was the `NewDefaultKnCommndWithArgs` func exits with os.Exit(1) when there is an error. So for failure cases the test exits without continuing. Is there something I could do for this scenaro ?

pkg/kn/core/root.go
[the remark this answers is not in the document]
itsmurugappan: @rhuss have made these changes. Please validate.

pkg/kn/core/root_test.go
itsmurugappan: @rhuss moved the tests to e2e.

pkg/kn/core/root.go
```diff
@@ -288,3 +292,22 @@ func width() (int, error) {
 	width, _, err := terminal.GetSize(int(os.Stdout.Fd()))
 	return width, err
 }
+
+func getCommands(args []string, innerArg string) string {
+	var commands []string
```
navidshaikh:
```suggestion
	commands := []string{"kn"}
```

pkg/kn/core/root.go
```diff
@@ -288,3 +292,22 @@ func width() (int, error) {
...
+	}
+	return ""
+}
+
+func showSubcommands(cmd *cobra.Command, args []string, innerArg string) string {
+	var strs []string
+	for _, subcmd := range cmd.Commands() {
+		strs = append(strs, subcmd.Name())
+	}
+	return fmt.Sprintf("Error: unknown subcommand '%s' for %s. Available subcommands: %s\nRun 'kn --help' for usage.\n", innerArg, getCommands(args, innerArg), strings.Join(strs, ", "))
```
navidshaikh:
```suggestion
	return fmt.Sprintf("Error: unknown subcommand '%s' for '%s'. Available subcommands: %s\nRun 'kn --help' for usage.\n", innerArg, getCommands(args, innerArg), strings.Join(strs, ", "))
```

navidshaikh: With above two suggested changes:
```
kn source apiserver noverb
Error: unknown subcommand 'noverb' for 'kn source apiserver'. Available subcommands: create, delete, describe, list, update
Run 'kn --help' for usage.
```

pkg/kn/core/root.go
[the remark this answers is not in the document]
rhuss: 👍 Thanks a lot for your contribution. Looks good (and @navidshaikh already confirmed that :).

PR #590, opened 2019-12-23
Improve 'source apiserver update --resource'

### LINKED ISSUE #565: source apiserver update --resource should follow the common update semantics for list like structures
I.e. the same as for `kn service update --env` or `kn trigger update --filter`. I.e. add new resources, overwrite existing, and remove when suffixed with `-`.

daisy-ycguo:
Fixes #565

## Proposed Changes

* Improve `source apiserver update --resource` to follow the common update semantics
* Support remove resource when suffixed with `-`


pkg/kn/commands/source/apiserver/flags.go
```diff
@@ -41,27 +41,44 @@ type APIServerSourceUpdateFlags struct {
...
+func (f *APIServerSourceUpdateFlags) getUpdateAPIServerResourceArray() ([]resourceSpec, []resourceSpec, error) {
+	var added []resourceSpec
+	var removed []resourceSpec
+	for _, r := range f.Resources {
+		if strings.HasSuffix(r, "-") {
+			resourceSpec, err := getValidResource(r[:len(r)-1])
+			if err != nil {
+				return nil, nil, err
+			}
+			removed = append(removed, *resourceSpec)
```
navidshaikh: Lets get https://github.com/knative/client/pull/592 in first and update this PR accordingly.

navidshaikh: @daisy-ycguo : #592 is merged, please update the PR to use the shared utility to parse minus suffix.

daisy-ycguo: Thank you @navidshaikh 
resources at `apiserver update` are not key=value format. So it's not convenient to use ParseMinusSuffix here. But I added a common utility in parsing_helper.go called `AddListAndRemovalListFromArray` to parse minus suffix in an array.

pkg/kn/commands/source/apiserver/flags.go
```diff
@@ -41,27 +42,45 @@ type APIServerSourceUpdateFlags struct {
...
+			return nil, nil, err
 		}
-		resourceList = append(resourceList, resourceRef)
+		removed = append(removed, *resourceSpec)
 	}
-	return &resourceList, nil
+	return added, removed, nil
 }
 
 func getValidResource(resource string) (*resourceSpec, error) {
```
maximilien: you should add comment for all code. I would rather see for public methods and functions but since you added above then should be consistent...

pkg/kn/commands/source/apiserver/flags.go
```diff
@@ -95,9 +114,9 @@ func (f *APIServerSourceUpdateFlags) Add(cmd *cobra.Command) {
 		`The mode the receive adapter controller runs under:,
 "Ref" sends only the reference to the resource,
 "Resource" send the full resource.`)
-	cmd.Flags().StringSliceVar(&f.Resources,
+	cmd.Flags().StringArrayVar(&f.Resources,
 		"resource",
```
maximilien: Why does the name `resource` not match the variable `f.Resources`? Should this not be `"resources"`?

pkg/kn/commands/source/apiserver/flags_test.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+func TestGetAPIServerResourceArray(t *testing.T) {
+	t.Run("get single apiserver resource", func(t *testing.T) {
+		createFlag := APIServerSourceUpdateFlags{
+			ServiceAccountName: "test-sa",
+			Mode:               "Ref",
+			Resources:          []string{"Service:serving.knative.dev/v1alpha1:true"},
+		}
+		created, _ := createFlag.getAPIServerResourceArray()
```
maximilien: why are you ignoring errors? Just assert that they are nil

pkg/kn/commands/source/apiserver/flags_test.go
```diff
@@ -0,0 +1,148 @@
...
+package apiserver
+
+import (
+	"testing"
+
+	"gotest.tools/assert"
+)
+
+func TestGetAPIServerResourceArray(t *testing.T) {
+	t.Run("get single apiserver resource", func(t *testing.T) {
```
maximilien: I think since there are many combination you are testing you should offer a small comment summary of the table and what combination you are testing. This would even help you if you need to add or remove tests. Or seperate them into their own `t.Run` calls

pkg/util/parsing_helper_test.go
```diff
@@ -92,3 +92,9 @@ func TestParseMinusSuffix(t *testing.T) {
 	assert.DeepEqual(t, expectedMap, inputMap)
 	assert.DeepEqual(t, expectedStringToRemove, stringToRemove)
 }
+
+func TestAddListAndRemovalListFromArray(t *testing.T) {
```
maximilien: I would add tests for corner cases. For instance, 
1. add one item and remove none, 
2. add no items, remove one, 
3. add one item and remove one, 
4. add none and remove none.

docs/cmd/kn_source_apiserver_create.md
```diff
@@ -26,7 +26,7 @@ kn source apiserver create NAME --resource RESOURCE --service-account ACCOUNTNAM
                                  "Ref" sends only the reference to the resource,
                                  "Resource" send the full resource. (default "Ref")
   -n, --namespace string         Specify the namespace to operate in.
-      --resource strings         Specification for which events to listen, in the format Kind:APIVersion:isController, e.g. Deployment:apps/v1:true.
+      --resource stringArray     Specification for which events to listen, in the format Kind:APIVersion:isController, e.g. Deployment:apps/v1:true.
```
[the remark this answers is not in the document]
daisy-ycguo: `Deployment:apps/v1:true` is just the format of `Kind:APIVersion:isController`. APIVersion usually looks like:
- `serving.knative.dev/v1` or
- simply `v1`, only means kubernetes API V1.

I could change it to `v1` if it looks confused.

docs/cmd/kn_source_apiserver_update.md
```diff
@@ -26,7 +26,7 @@ kn source apiserver update NAME --resource RESOURCE --service-account ACCOUNTNAM
                                  "Ref" sends only the reference to the resource,
                                  "Resource" send the full resource. (default "Ref")
   -n, --namespace string         Specify the namespace to operate in.
-      --resource strings         Specification for which events to listen, in the format Kind:APIVersion:isController, e.g. Deployment:apps/v1:true.
+      --resource stringArray     Specification for which events to listen, in the format Kind:APIVersion:isController, e.g. Deployment:apps/v1:true.
```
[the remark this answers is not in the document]
daisy-ycguo: see my previous comments

pkg/eventing/sources/v1alpha1/apiserver_client.go
```diff
@@ -156,6 +159,43 @@ func (b *APIServerSourceBuilder) Resources(resources []v1alpha1.ApiServerResourc
...
+		resources = []v1alpha1.ApiServerResource{}
+		b.apiServerSource.Spec.Resources = resources
+	}
+	resourceRef := v1alpha1.ApiServerResource{
+		APIVersion: version,
+		Kind:       kind,
+		Controller: isController,
+	}
+	b.apiServerSource.Spec.Resources = append(resources, resourceRef)
+	return b
```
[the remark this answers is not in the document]
daisy-ycguo: I follow the same style in other eventing clients, e.g. [cronjob_client](https://github.com/knative/client/blob/master/pkg/eventing/sources/v1alpha1/cronjob_client.go#L153) and [apiserver_client](https://github.com/knative/client/blob/master/pkg/eventing/sources/v1alpha1/apiserver_client.go#L174). If you prefer to change the return, I will suggest to create a separate issue to track and we can change them all together.

pkg/eventing/sources/v1alpha1/apiserver_client.go
```diff
@@ -156,6 +159,43 @@ func (b *APIServerSourceBuilder) Resources(resources []v1alpha1.ApiServerResourc
...
+		Kind:       kind,
+		Controller: isController,
+	}
+	for i, k := range resources {
+		if reflect.DeepEqual(k, resourceRef) {
+			resources = append(resources[:i], resources[i+1:]...)
+			return b, nil
+		}
+	}
+	return b, fmt.Errorf("cannot find resource %s:%s:%t to remove", version, kind, isController)
```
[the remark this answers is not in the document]
daisy-ycguo: If the resource could be found, the function will return at the line #193 without error.
If the resource could not be found after traversing the whole structure, the function will return at line #196 with an error.
Could you help to explain more of your comment ?

pkg/eventing/sources/v1alpha1/apiserver_client.go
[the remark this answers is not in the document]
daisy-ycguo: b is short for builder. Again, I will suggest to create a separate issue to change all client codes, including  cronjob_client, eventing_client and apiserver_client.

pkg/kn/commands/source/apiserver/flags.go
daisy-ycguo: `resource` is a flag. `Resources` is an array of `resource`.
User uses `--resource abc --resource def` to set multiple resources, which will be saved in `f.Resources`.

pkg/kn/commands/source/apiserver/flags_test.go
daisy-ycguo: will add some tests to test errors. Thank you for pointing out.

pkg/kn/commands/source/apiserver/flags_test.go
daisy-ycguo: simple comments are added before test cases. e.g. line 38 and line 52.
But I think maybe separate them into a single t.Run is more readable.

pkg/util/parsing_helper_test.go
daisy-ycguo: will do.

pkg/kn/commands/source/apiserver/flags.go
```diff
@@ -41,27 +42,45 @@ type APIServerSourceUpdateFlags struct {
...
-		resourceRef := v1alpha1.ApiServerResource{
-			APIVersion: resourceSpec.apiVersion,
-			Kind:       resourceSpec.kind,
-			Controller: resourceSpec.isController,
+		resourceList = append(resourceList, *resourceSpec)
+	}
+	return resourceList, nil
+}
+
+// getAPIServerResourceArray is to return an array of ApiServerResource from a string. A sample is Event:v1:true
```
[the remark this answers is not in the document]
daisy-ycguo: will do

pkg/eventing/sources/v1alpha1/apiserver_client.go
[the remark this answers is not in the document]
rhuss: This is classical pattern used in many languages which allows for chained calls for elegant, DSL-like APIs. Do you remember the review comment where there was a check for a nil-received in a method (in some dependency) ? That is exactly for that use case of chained calls, where an intermediate call could return nil. So its also used in Go quite a bit (and btw is very popular in Java as well).

Some references illustrating that chaining is quite popular in Go, too (and they all return the object itself):

* https://github.com/mauricioklein/go-chainable
* https://vikash1976.wordpress.com/2017/02/03/method-chaining-in-go-lang-through-interface/
* https://riptutorial.com/go/example/13465/chaining-methods
* https://stackoverflow.com/questions/27297702/go-method-chaining-and-error-handling
* ....

BTW, you can still ignore the returned value if you like.

pkg/kn/commands/source/apiserver/flags.go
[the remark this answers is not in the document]
daisy-ycguo: done.

pkg/kn/commands/source/apiserver/flags.go
daisy-ycguo: Added a comment.

maximilien:
Is this ready? Also, looks like you have some conflicts to resolve. Thanks.

pkg/util/parsing_helper_test.go
daisy-ycguo: done.

docs/cmd/kn_source_apiserver_create.md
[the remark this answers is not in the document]
rhuss: I would keep `apps/v1` as `v1` without an apigroup is more the exception than the rule. Giving a full example makes more sense.

pkg/kn/commands/source/apiserver/flags_test.go
daisy-ycguo: done.

pkg/eventing/sources/v1alpha1/apiserver_client.go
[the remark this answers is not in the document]
rhuss: @maximilien  is right, chaining on works on methods which return a single return value, so returning two values makes no sense in this pattern.

But as we discussed in the other PR having mutating methods on a builder is an anti-pattern (as then the builder turns more into a regular domain objects). Maybe we can solve this use case differently, too ?

If not, we could agree that removing a non-existing resource is just a no-op and gets ignored, so that you can return directly the builder again.

pkg/kn/commands/source/apiserver/update.go
```diff
@@ -74,11 +74,19 @@ func NewAPIServerUpdateCommand(p *commands.KnParams) *cobra.Command {
...
-				b.Resources(*resources)
+				for _, k := range removed {
+					_, err = b.RemoveResource(k.ApiVersion, k.Kind, k.IsController)
+					if err != nil {
+						return err
+					}
+				}
+				for _, k := range added {
+					b.AddResource(k.ApiVersion, k.Kind, k.IsController)
+				}
```
rhuss: I would follow the same scheme here as we had for trigger filters in #603 , I think it covers the same use case.

daisy-ycguo: ok. will do.

daisy-ycguo: Resource is not a string map, but an array. I cannot leverage stringmap here.
I have to think about how to handle.

pkg/kn/commands/source/apiserver/flags.go
```diff
@@ -40,47 +42,99 @@ type APIServerSourceUpdateFlags struct {
...
-			APIVersion: resourceSpec.apiVersion,
-			Kind:       resourceSpec.kind,
-			Controller: resourceSpec.isController,
+		resourceList = append(resourceList, *resourceSpec)
+	}
+	return resourceList, nil
+}
+
+// updateExistingAPIServerResourceArray is to update an array of resources.
+func (f *APIServerSourceUpdateFlags) updateExistingAPIServerResourceArray(existing []v1alpha1.ApiServerResource) ([]v1alpha1.ApiServerResource, error) {
```
daisy-ycguo: I would like to move this method to parsing_helper.go to provide a common method to update existing array.
My question is which type (or key word) I should use to describe a general array here ?

I tried to use []interface{} but I get an error:
```
cannot use existing (type []"knative.dev/eventing/pkg/apis/sources/v1alpha1".ApiServerResource) as type []interface {} in argument to util.UpdateArray
```

rhuss: That's not easily possible, see https://golang.org/doc/faq#convert_slice_of_interface

pkg/kn/commands/source/apiserver/update.go
rhuss: Ah, right. Unfortunately, there are no generics in golang, that would make things much easier. Its also not easy to simulate with interfaces ;(

pkg/kn/commands/source/apiserver/flags.go
rhuss: What you could do is to introduce a simple utility method which converts an `[]ApiServerResource` to a `[]interface{}` and feed that into the generic method like in 

```golang
fromISlice(UpdateArray(toISlice(existing), toISliceP(added), toISliceP(removed)))
```

and do the conversion between ApiServerSource and interfaces slices before and after calling the function. But I don't think its worth the hassle.

daisy-ycguo: oh, ok. Then I keep as it is. 
Please review the latest version. 
Thank you.

daisy-ycguo:
> Is this ready? Also, looks like you have some conflicts to resolve. Thanks.

Yes, it's ready for a review.

pkg/kn/commands/source/apiserver/flags.go
```diff
@@ -40,47 +42,99 @@ type APIServerSourceUpdateFlags struct {
...
+
+	addedArray, removedArray := util.AddedAndRemovalListsFromArray(f.Resources)
+	for _, r := range addedArray {
+		resourceSpec, err := getValidResource(r)
+		if err != nil {
+			return nil, nil, err
+		}
+		added = append(added, *resourceSpec)
+	}
+	for _, r := range removedArray {
```
rhuss: you could put the loop into a seperate function and call it twice, for adds and removals. That would reduce duplicated code a bit.

pkg/kn/commands/source/apiserver/flags.go
```diff
@@ -40,47 +42,99 @@ type APIServerSourceUpdateFlags struct {
...
+//getValidResource is to parse resource spec from a string
+func getValidResource(resource string) (*v1alpha1.ApiServerResource, error) {
+	var isController bool
+	var version string
 	var err error
 
-	parts := strings.Split(resource, apiVersionSplitChar)
+	parts := strings.SplitN(resource, apiVersionSplitChar, 3)
+	if len(parts[0]) == 0 {
+		return nil, fmt.Errorf("cannot find KIND in resource specification %s", resource)
```
rhuss:
```suggestion
		return nil, fmt.Errorf("cannot find 'Kind' part in resource specification %s (expected: <Kind:ApiVersion[:controllerFlag]>", resource)
```

pkg/kn/commands/source/apiserver/flags.go
```diff
@@ -40,47 +42,99 @@ type APIServerSourceUpdateFlags struct {
...
+	}
 	kind := parts[0]
-	if len(parts) < 2 {
-		return nil, fmt.Errorf("no APIVersion given for resource %s", resource)
+
+	if len(parts) >= 2 && len(parts[1]) > 0 {
+		version = parts[1]
+	} else {
+		version = "v1"
```
rhuss: Do we really want a default to "v1" ? I would prefer to be always explicit with the ApiVersion and make it mandatory (as you have to do it for any k8s resource, too, when deploying to the server)

pkg/kn/commands/source/apiserver/flags.go
```diff
@@ -40,47 +42,99 @@ type APIServerSourceUpdateFlags struct {
...
 	}
-	version := parts[1]
-	if len(parts) >= 3 {
+	if len(parts) >= 3 && len(parts[2]) > 0 {
 		isController, err = strconv.ParseBool(parts[2])
 		if err != nil {
 			return nil, fmt.Errorf("cannot parse controller flage in resource specification %s", resource)
```
rhuss:
```suggestion
			return nil, fmt.Errorf("controller flag is not a boolean in resource specification %s (expected: <Kind:ApiVersion[:controllerFlag]>)", resource)
```

pkg/kn/commands/source/apiserver/flags.go
daisy-ycguo: done

pkg/kn/commands/source/apiserver/flags.go
daisy-ycguo: done

pkg/kn/commands/source/apiserver/flags.go
daisy-ycguo: done

pkg/kn/commands/source/apiserver/flags.go
daisy-ycguo: I just want to make a simple format for ApiServerSource, e.g. Event or Pod or Deployment could only be like `--resource Event --resource Pod --resource Deployment`, other than `--resource Event:v1 --resource Pod:v1 --resource Deployment:v1` which doesn't look artistic.

So if no ApiVersion provides, I use v1 instead. If you don't agree with this kind of contraction rule, I could delete this function.

rhuss: My point is that this is a bit surprising for Kubernetes users, as the apiVersion is a mandatory field in a resource descriptors. Also, imagine now that the Event or Pod spec evolves to a v2 version. Would you then also update the default to get to this newest version ? Will a user still using v1 and, more important, is she aware of that not the latest event version is used ? Maybe even v1 is getting removed later on then, making the default useless.

I understand that you want to make it as easy and concise as possible. However, its one of  key principle of Kubernetes (and hence Knative, too) that to identify a resource you alway need apiVersion + kind. 

So I really would be explicit here to (a) be on par with the Kubernetes spec and (b) to avoid future surprises (and (c) to avoid 'hacks' like '::' to get to the third option).

rhuss: btw, `Deployment` is in `apps/v1`, not `v1`. There are only a handful of resources that are really in v1.

daisy-ycguo: oh. Got your point. Let me remove the default value.

daisy-ycguo: Done. The latest version is submitted and ready for review. @rhuss

daisy-ycguo:
@maximilien @rhuss Are we ready to merge this ? I'm on vacation. I could still address any comments if you have. Thank you.

PR #591, opened 2019-12-23
fixes(#308): adds e2e tests for plugins

### LINKED ISSUE #308: Kn plugins integration test

maximilien:
Fixes #308

## Proposed Changes

* adds `test/e2e/plugins_test.go`
* adds a new option to `e2eTest` to prevent namespace creation (since some test might not need ns)

Release Note:

- 🧽 backfill e2e tests for plugins

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -47,13 +47,16 @@ const (
 var m sync.Mutex
 
 type e2eTest struct {
-	env env
-	kn  kn
+	env                    env
+	kn                     kn
+	createNamespaceOnSetup bool
+	namespaceCreated       bool
 }
 
 func NewE2eTest(t *testing.T) *e2eTest {
 	return &e2eTest{
-		env: buildEnv(t),
+		env:                    buildEnv(t),
+		createNamespaceOnSetup: true,
 	}
 }
 
@@ -61,8 +64,10 @@ func NewE2eTest(t *testing.T) *e2eTest {
 func (test *e2eTest) Setup(t *testing.T) {
 	test.env.Namespace = fmt.Sprintf("%s%d", test.env.Namespace, getNamespaceCountAndIncrement())
 	test.kn = kn{t, test.env.Namespace, Logger{}}
-	test.CreateTestNamespace(t, test.env.Namespace)
-	time.Sleep(20 * time.Second)
+	if test.createNamespaceOnSetup {
+		test.CreateTestNamespace(t, test.env.Namespace)
+		time.Sleep(20 * time.Second)
+	}
 }
 
 func getNamespaceCountAndIncrement() int {
@@ -83,7 +88,9 @@ func getServiceNameAndIncrement(base string) string {
 
 // Teardown clean up
 func (test *e2eTest) Teardown(t *testing.T) {
-	test.DeleteTestNamespace(t, test.env.Namespace)
+	if test.namespaceCreated {
+		test.DeleteTestNamespace(t, test.env.Namespace)
+	}
 }
 
 // CreateTestNamespace creates and tests a namesspace creation invoking kubectl
@@ -99,6 +106,7 @@ func (test *e2eTest) CreateTestNamespace(t *testing.T, namespace string) {
 	if !matchRegexp(t, expectedOutputRegexp, out) {
 		t.Fatalf("Expected output incorrect, expecting to include:\n%s\n Instead found:\n%s\n", expectedOutputRegexp, out)
 	}
+	test.namespaceCreated = true
 }
 
 // CreateTestNamespace deletes and tests a namesspace deletion invoking kubectl
@@ -113,6 +121,7 @@ func (test *e2eTest) DeleteTestNamespace(t *testing.T, namespace string) {
 	if !matchRegexp(t, expectedOutputRegexp, out) {
 		t.Fatalf("Expected output incorrect, expecting to include:\n%s\n Instead found:\n%s\n", expectedOutputRegexp, out)
 	}
+	test.namespaceCreated = false
 }
 
 // WaitForNamespaceDeleted wait until namespace is deleted
--- a/test/e2e/plugins_test.go
+++ b/test/e2e/plugins_test.go
@@ -1,0 +1,210 @@
+// Copyright 2019 The Knative Authors
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+
+//     http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or im
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build e2e
+
+package e2e
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"gotest.tools/assert"
+	"knative.dev/client/pkg/util"
+)
+
+const (
+	TestPluginCode string = `#!/bin/bash
+
+echo "Hello Knative, I'm a Kn plugin"
+echo "  My plugin file is $0"
+echo "  I recieved arguments: $1 $2 $3 $4"`
+
+	KnConfigDefault string = `plugins-dir: %s
+lookup-plugins: %s`
+
+	FileModeReadable   = 0644
+	FileModeReadWrite  = 0666
+	FileModeExecutable = 0777
+)
+
+func TestPluginWorkflow(t *testing.T) {
+	var (
+		knConfigDir, knPluginsDir, knPluginsDir2  string
+		knConfigPath, knPluginPath, knPluginPath2 string
+		lookupPlugins                             bool
+		err                                       error
+	)
+
+	t.Parallel()
+	test := NewE2eTest(t)
+	setup := func(t *testing.T) {
+		test.createNamespaceOnSetup = false
+		test.Setup(t)
+		defer test.Teardown(t)
+
+		knConfigDir, err = ioutil.TempDir("", "kn-config")
+		assert.Assert(t, err == nil)
+
+		knPluginsDir = filepath.Join(knConfigDir, "plugins")
+		err = os.MkdirAll(knPluginsDir, FileModeExecutable)
+		assert.Assert(t, err == nil)
+
+		knPluginsDir2 = filepath.Join(knConfigDir, "plugins2")
+		err = os.MkdirAll(knPluginsDir2, FileModeExecutable)
+		assert.Assert(t, err == nil)
+
+		knConfigPath, err = test.createConfig(t, "config.yaml", "", knConfigDir)
+		assert.Assert(t, err == nil)
+		assert.Assert(t, knConfigPath != "")
+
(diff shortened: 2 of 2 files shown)


test/e2e/plugins_test.go
```diff
@@ -0,0 +1,210 @@
...
+		t.Run("when using --plugin-dir", func(t *testing.T) {
+			t.Run("when --plugins-dir has a plugin", func(t *testing.T) {
+				t.Run("when --lookup-plugins is false", func(t *testing.T) {
+					lookupPlugins = false
+
+					setup(t)
+					defer teardown(t)
+
+					knFlags := []string{fmt.Sprintf("--plugins-dir=%s", knPluginsDir), "--lookup-plugins=false"}
+					assert.Assert(t, lookupPlugins == false)
```
navidshaikh:
```suggestion

```

`lookupPlugins` not being operated after its set.

test/e2e/plugins_test.go
```diff
@@ -0,0 +1,210 @@
...
+	err := ioutil.WriteFile(pluginFilename, []byte(pluginCode), FileModeExecutable)
+	if err != nil {
+		return "", err
+	}
+	return pluginFilename, nil
+}
+
+func (test *e2eTest) listPlugin(t *testing.T, knFlags []string, expectedPlugins []string, unexpectedPlugins []string) {
+	knArgs := addArgs([]string{}, knFlags...)
+	knArgs = addArgs(knArgs, []string{"plugin", "list"}...)
```
navidshaikh:
```suggestion
	knArgs := append(knFlags, []string{"plugin", "list"}...)
```

test/e2e/plugins_test.go
```diff
@@ -0,0 +1,210 @@
...
+		assert.Check(t, util.ContainsAll(out, output))
+	}
+}
+
+func addArgs(cmdArgs []string, args ...string) []string {
+	for _, arg := range args {
+		cmdArgs = append(cmdArgs, arg)
+	}
+	return cmdArgs
+}
```
navidshaikh: Remove.

test/e2e/plugins_test.go
```diff
@@ -0,0 +1,210 @@
...
+	knArgs := addArgs([]string{}, knFlags...)
+	knArgs = addArgs(knArgs, []string{"plugin", "list"}...)
+
+	out, err := test.kn.RunWithOpts(knArgs, runOpts{NoNamespace: true})
+	assert.Assert(t, err == nil)
+	assert.Check(t, util.ContainsAll(out, expectedPlugins...))
+	assert.Check(t, util.ContainsNone(out, unexpectedPlugins...))
+}
+
+func (test *e2eTest) runPlugin(t *testing.T, knFlags []string, pluginName string, args []string, expectedOutput []string) {
```
navidshaikh: `expectedOutput` can just be string here?

test/e2e/plugins_test.go
```diff
@@ -0,0 +1,210 @@
...
+	"gotest.tools/assert"
+	"knative.dev/client/pkg/util"
+)
+
+const (
+	TestPluginCode string = `#!/bin/bash
+
+echo "Hello Knative, I'm a Kn plugin"
+echo "  My plugin file is $0"
+echo "  I recieved arguments: $1 $2 $3 $4"`
```
navidshaikh: *received

test/e2e/plugins_test.go
```diff
@@ -0,0 +1,210 @@
...
+				})
+
+				t.Run("with --lookup-plugins is true", func(t *testing.T) {
+					lookupPlugins = true
+
+					setup(t)
+					defer teardown(t)
+
+					knFlags := []string{fmt.Sprintf("--plugins-dir=%s", knPluginsDir), "--lookup-plugins=true"}
+					assert.Assert(t, lookupPlugins == true)
```
navidshaikh:
```suggestion

```

unnecessary check

test/e2e/plugins_test.go
```diff
@@ -0,0 +1,210 @@
...
+					})
+				})
+			})
+		})
+	})
+}
+
+// Private
+
+func (test *e2eTest) createConfig(t *testing.T, configName, configCode, configPath string) (string, error) {
```
navidshaikh:
```suggestion
func (test *e2eTest) createConfig(t *testing.T, configFilename, configData, configDir string) (string, error) {
```
(symbol renames)

actually should be ordered as Dir, Filename, Data


```suggestion
func (test *e2eTest) createConfig(t *testing.T, configDir, configFilename, configData string) (string, error) {
```

test/e2e/plugins_test.go
```diff
@@ -0,0 +1,210 @@
...
+
+	t.Parallel()
+	test := NewE2eTest(t)
+	setup := func(t *testing.T) {
+		test.createNamespaceOnSetup = false
+		test.Setup(t)
+		defer test.Teardown(t)
+
+		knConfigDir, err = ioutil.TempDir("", "kn-config")
+		assert.Assert(t, err == nil)
```
navidshaikh:
```suggestion
		assert.NilError(t, err)
```

navidshaikh: also for other similar checks

test/e2e/plugins_test.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// Private
+
+func (test *e2eTest) createConfig(t *testing.T, configName, configCode, configPath string) (string, error) {
+	configFilename := filepath.Join(configPath, configName)
+	err := ioutil.WriteFile(configFilename, []byte(configCode), FileModeReadWrite)
+	if err != nil {
+		return "", err
+	}
```
navidshaikh:
```
	assert.NilError(t, err)
```

test/e2e/plugins_test.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+func (test *e2eTest) createPlugin(t *testing.T, pluginName, pluginCode, pluginPath string) (string, error) {
+	pluginFilename := filepath.Join(pluginPath, pluginName)
+	err := ioutil.WriteFile(pluginFilename, []byte(pluginCode), FileModeExecutable)
+	if err != nil {
+		return "", err
+	}
+	return pluginFilename, nil
+}
```
navidshaikh: Actually, `createConfig` and `createPlugin` should be merged into `createFile` taking `Dir, Filename, data, FilePermission` args.

test/e2e/plugins_test.go
```diff
@@ -0,0 +1,210 @@
...
+		assert.Assert(t, knConfigPath != "")
+
+		knPluginPath, err = test.createPlugin(t, "kn-helloe2e", TestPluginCode, knPluginsDir)
+		assert.Assert(t, err == nil)
+		assert.Assert(t, knPluginPath != "")
+
+		knPluginPath2, err = test.createPlugin(t, "kn-hello2e2e", TestPluginCode, knPluginsDir2)
+		assert.Assert(t, err == nil)
+		assert.Assert(t, knPluginPath2 != "")
+	}
```
navidshaikh: IMO, we should take off this code from setup and move to individual test runs. We should then also be able to test cases  like, if the plugin name overrides the built in commands/verbs.

test/e2e/common.go
```diff
@@ -47,22 +47,27 @@ const (
...
 
 // Setup set up an environment for kn integration test returns the Teardown cleanup function
 func (test *e2eTest) Setup(t *testing.T) {
 	test.env.Namespace = fmt.Sprintf("%s%d", test.env.Namespace, getNamespaceCountAndIncrement())
 	test.kn = kn{t, test.env.Namespace, Logger{}}
-	test.CreateTestNamespace(t, test.env.Namespace)
-	time.Sleep(20 * time.Second)
+	if test.createNamespaceOnSetup {
+		test.CreateTestNamespace(t, test.env.Namespace)
+		time.Sleep(20 * time.Second)
```
navidshaikh: why `time.Sleep`?

test/e2e/plugins_test.go
maximilien: OK. Good suggestion. Thanks.

test/e2e/plugins_test.go
maximilien: 👍

test/e2e/plugins_test.go
maximilien: 👍

test/e2e/plugins_test.go
maximilien: Thx

test/e2e/plugins_test.go
maximilien: OK

test/e2e/common.go
maximilien: We always had that sleep since namespaces take time to create. See previous code.

We can experiment with smaller values but that's a separate PR.

I had to make namespace creation optional since this plugin e2e would have created a namespace but never used it.

test/e2e/plugins_test.go
maximilien: OK

test/e2e/plugins_test.go
maximilien: Well because the `lookupPlugins` is a variable, the `setup(t)` could have changed it and since this block is about `lookupPlugins` either `true` or `false` wanted to ensure that.

test/e2e/plugins_test.go
maximilien: Want to check for multiple lines

test/e2e/plugins_test.go
maximilien: I already have tests for "plugin name overrides the built in commands/verbs." in the `pkg/commands/plugin` package and you won't see the overrides until the plugin is ran.

test/e2e/common.go
maximilien: Was in previous code so kept it. I think was there to give time for namespace to be created.

I think we should probably reduce the time since NS creation might be faster. However, that's a separate PR.

I'll open an issue to try to get e2e tests faster and point to that line as one place.

maximilien: I created issue #598

maximilien:
@navidshaikh I need to work on this a bit more. Will submit updates sometimes tomorrow.

test/e2e/plugins_test.go
maximilien: OK. Good point. Done.

maximilien:
> @navidshaikh I need to work on this a bit more. Will submit updates sometimes tomorrow.

Actually, got a bit of breather with other things and completed it before leaving for home. See update. Thanks.

maximilien:
@navidshaikh I am not sure why the IT tests failed. Let me try it again.

test/e2e/plugins_test.go
navidshaikh: The `err` here is already asserted for NilError, `if err != nil` check is not needed. 
Also err is being checked here, so `createFile` method doesn't need to return the `err` in this case.

maximilien: OK fair. Pushed new updates to address. Thx.

PR #593, opened 2019-12-26
"Faas" scenario — create a trigger alongside its source.

### LINKED ISSUE #498: kn trigger create - create a source implicitely (based on #479)
As described in the [MVP doc](https://docs.google.com/document/d/1B8VK-5nuxt3D-i1Up3onmUbzcztQXE0GRgf1lfx49P8/edit#heading=h.rr500f66txqz), we want to not only create a trigger, but also create an associated source CR, whose results only go to that trigger.

The feature described in this issue add the possibility to specify a source when creating a trigger. It is based on #479 and has the trigger creation without source as a prerequisite. 

Copy of the spec related to implicit source creation:

-----------------
....
In addition to the trigger, an additional Source can be created automatically like with kn source create. The difference to kn source create is, that the source is connected exclusively to this trigger (and you automatically use the same broker). The source type is provided with a --source option.

Except that the Source's sink can't be specified, all other, source specific commands are available. These can be given after a --(two dashes) separator on the command line which indicates the end of the trigger specific options. The following example will create the same trigger as above, but also instantiate a generic source for the storage CRD:

```
# Create a trigger + a storage generic source   
$ kn trigger create ochoa \
    --sink event-display \
    --filter "type=dev.knative.foo" \
    --source generic \
    -- \
    --source-name storage \
    --event-types finalize \
    --parameters "bucket=my-bucket" \
    --secret "gcsSecret=google-cloud-key:key.json"
```

For the options possible after the delimiter (--) see kn source <type> create. Whether the delimiter is required depends on the likelihood that a typed source might also use the --filter or --broker options.

Only sources which support the ceOverride mechanism as described in [Knative Eventing Source Spec](https://github.com/knative/eventing/pull/1856) can be used together with a Trigger in this way as this override is used to connect the Source exclusively to this trigger by creating a unique value, which is added to every event and a Trigger filter which filters exactly on this unique value. So it's also a task of this command to verify whether the given source supports this feature. For generic Sources, the openAPIV3 schema as given by the CRD needs to be examined, for typed Sources, all built-in sources should support this and plugins will add this information (supported: yes/no) in its manifest command.

When making a Source too, the CLI will create the Trigger and the Source simultaneous and waits by default until the Source is ready (the same behaviour like for kn service create, which supports also an --async flag). This should be idempotent, such that if the Trigger creates succeeds, then the Source create fails, the user should be able to rerun the same command to try again. Deleting the Trigger will delete the Source (via Kubernetes garbage collection).

The example above would create the following CO:

```yaml
apiVersion: eventing.knative.dev/v1alpha1
kind: Trigger
metadata:
  name: ochoa
  namespace: default
    annotations:
      knative.dev/dependency: '{"kind":"Storage","name":"ochoa","apiVersion":"events.cloud.run/v1alpha1"}'
spec:
  broker: default
  filter:
    attributes:
      # This attribute is injected by the CLI into both the 
      # Trigger and the Source to link them.
      knsourcetrigger: storages.events.cloud.run/namespaces/default/names/ochoa
      type: dev.knative.foo
  subscriber:
    ref:
      kind: Service
      name: event-display
---
apiVersion: events.cloud.run/v1alpha1
kind: Storage
metadata:
  name: ochoa
  namespace: default
  ownerReferences:
    - apiVersion: eventing.knative.dev/v1alpha1
      controller: true
      kind: Trigger
      name: ochoa
      uid: ce9350a1-c8fb-11e9-9d25-42010a8000a0
spec:
  bucket: my-bucket
  ceOverrides:
    extensions:
      knsourcetrigger: storages.events.cloud.run/namespaces/default/names/ochoa  
  eventTypes:
   - finalize
  gcsSecret:
    name: google-cloud-key
    key: key.json
  sink:
   apiVersion: eventing.knative.dev/v1alpha1
    kind: Broker
    name: default
```

sixolet:
This only so far handles the non-plugin source creation, but it's a start.

NB: based on #574 — MERGE THAT FIRST then rebase -i.

To do in this PR before merging:

- [ ] Add `ceOverrides.knsourcetrigger` / `filter.knsourcetrigger` matching attributes to both source and trigger.
- [ ] Create trigger before source.
- [ ] Add dependency attribute to trigger
- [ ] Add owner reference to source
- [ ] Add `--async` flag, and do an appropriate wait for readiness (ignore unready due to dependency not yet present)
- [ ] Tests
- [ ] Changelog

To do outside this PR:
- [ ] Merge PR #574 for the broker allowed as a sink targer
- [ ] Make flag for `ceOverrides` setting in sources
- [ ] Make *hidden* flag for setting owner reference in sources.

Fixes #498

### DIFF
--- a/pkg/dynamic/client.go
+++ b/pkg/dynamic/client.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/fake"
 
+	eventing_v1alpha1 "knative.dev/eventing/pkg/apis/eventing/v1alpha1"
 	serving_v1alpha1 "knative.dev/serving/pkg/apis/serving/v1alpha1"
 )
 
@@ -102,6 +103,7 @@ func (c knDynamicClient) RawClient() dynamic.Interface {
 func CreateFakeKnDynamicClient(testNamespace string, objects ...runtime.Object) KnDynamicClient {
 	scheme := runtime.NewScheme()
 	scheme.AddKnownTypeWithName(schema.GroupVersionKind{Group: "serving.knative.dev", Version: "v1alpha1", Kind: "Service"}, &serving_v1alpha1.Service{})
+	scheme.AddKnownTypeWithName(schema.GroupVersionKind{Group: "eventing.knative.dev", Version: "v1alpha1", Kind: "Broker"}, &eventing_v1alpha1.Broker{})
 	client := fake.NewSimpleDynamicClient(scheme, objects...)
 	return NewKnDynamicClient(client, testNamespace)
 }
--- a/pkg/dynamic/client.go
+++ b/pkg/dynamic/client.go
@@ -18,13 +18,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
-	"k8s.io/client-go/dynamic/fake"
-
-	eventing_v1alpha1 "knative.dev/eventing/pkg/apis/eventing/v1alpha1"
-	serving_v1alpha1 "knative.dev/serving/pkg/apis/serving/v1alpha1"
 )
 
 const (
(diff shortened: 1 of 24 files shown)


rhuss:
@sixolet Is this PR ready for a review or would make it sense to review it later when more tasks from the list above are implemented ?

sixolet:
It's still WIP

maximilien:
You also need to rebase. Best.

rhuss:
@sixolet where do we stand with this PR ? Do you want to continue to work on this "FaaS" direction now replaced the direction the Functions WG is heading to ?

rhuss:
This PR needs probably a lot of love to be aligned with the main branch, so it's probably better to start afresh.

If you don't mind and if you don't have any objections until next Tuesday, I would close the PR then.

rhuss:
Closing, for now, feel free to reopen in case.

PR #595, opened 2020-01-06
Delete k8s resource if error happen at setup step

### LINKED ISSUE #576: e2e test fails if there are resouces created by the previous test
## Describe the feature:

The e2e test may fail in sometime, for example failing download docker image.  When start test again, it will fail for some existed resource. In this situation, I have to clear resource manually.
Is there a command or shell to clear all resource the test has created?

MIBc:
Fixes #576

## Proposed Changes

* Delete k8s cluster resource: `clusterrole`, `clusterrolebinding`.
* Adjust `defer` function position at Test case. The `defer` function can run, if `t.Fatalf` called in `Setup` function. Otherwise, the `defer` func will not run when  `t.Fatalf` called.

### DIFF
--- a/test/e2e/basic_workflow_test.go
+++ b/test/e2e/basic_workflow_test.go
@@ -28,8 +28,8 @@ import (
 func TestBasicWorkflow(t *testing.T) {
 	t.Parallel()
 	test := NewE2eTest(t)
-	test.Setup(t)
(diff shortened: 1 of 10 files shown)


MIBc:
@googlebot I fixed it.

PR #603, opened 2020-01-09
Enable trigger filter to be optional

daisy-ycguo:
## Proposed Changes

* Make trigger filter to be optional
* Handle filter flag in a same way as other string array flags
* use Builder to create a trigger in the creation command

### DIFF
--- a/pkg/eventing/v1alpha1/client.go
+++ b/pkg/eventing/v1alpha1/client.go
@@ -187,21 +187,6 @@ func (b *TriggerBuilder) Filters(filters map[string]string) *TriggerBuilder {
 	return b
 }
 
-func (b *TriggerBuilder) RemoveFilters(keys []string) *TriggerBuilder {
-	filter := b.trigger.Spec.Filter
-	if filter == nil {
-		return b
-	}
-	attributes := filter.Attributes
-	if attributes == nil {
-		return b
-	}
-	for _, k := range keys {
-		delete(*attributes, k)
-	}
-	return b
-}
-
 // Build to return an instance of trigger object
 func (b *TriggerBuilder) Build() *v1alpha1.Trigger {
 	return b.trigger
--- a/pkg/eventing/v1alpha1/client.go
+++ b/pkg/eventing/v1alpha1/client.go
@@ -170,7 +170,7 @@ func (b *TriggerBuilder) Broker(broker string) *TriggerBuilder {
 	return b
 }
 
-func (b *TriggerBuilder) AddFilter(key, value string) *TriggerBuilder {
+func (b *TriggerBuilder) Filters(filters map[string]string) *TriggerBuilder {
 	filter := b.trigger.Spec.Filter
 	if filter == nil {
 		filter = &v1alpha1.TriggerFilter{}
@@ -181,11 +181,13 @@ func (b *TriggerBuilder) AddFilter(key, value string) *TriggerBuilder {
 		attributes = &v1alpha1.TriggerFilterAttributes{}
 		filter.Attributes = attributes
 	}
-	(*attributes)[key] = value
+	for k, v := range filters {
+		(*attributes)[k] = v
+	}
 	return b
 }
 
-func (b *TriggerBuilder) RemoveFilter(key string) *TriggerBuilder {
+func (b *TriggerBuilder) RemoveFilters(keys []string) *TriggerBuilder {
 	filter := b.trigger.Spec.Filter
 	if filter == nil {
 		return b
@@ -194,7 +196,9 @@ func (b *TriggerBuilder) RemoveFilter(key string) *TriggerBuilder {
 	if attributes == nil {
 		return b
 	}
-	delete(*attributes, key)
+	for _, k := range keys {
+		delete(*attributes, k)
+	}
 	return b
 }
 
--- a/pkg/eventing/v1alpha1/client_test.go
+++ b/pkg/eventing/v1alpha1/client_test.go
@@ -128,12 +128,12 @@ func TestListTrigger(t *testing.T) {
 
 func TestTriggerBuilder(t *testing.T) {
 	a := NewTriggerBuilder("testtrigger")
-	a.AddFilter("type", "foo").AddFilter("source", "bar")
+	a.Filters(map[string]string{"type": "foo", "source": "bar"})
 
 	t.Run("update filter", func(t *testing.T) {
 		b := NewTriggerBuilderFromExisting(a.Build())
 		assert.DeepEqual(t, b.Build(), a.Build())
-		b.AddFilter("type", "new").RemoveFilter("source")
+		b.Filters(map[string]string{"type": "new"}).RemoveFilters([]string{"source"})
 		expected := &v1alpha1.TriggerFilter{
 			Attributes: &v1alpha1.TriggerFilterAttributes{
 				"type": "new",
@@ -145,7 +145,7 @@ func TestTriggerBuilder(t *testing.T) {
 	t.Run("update filter with only deletions", func(t *testing.T) {
 		b := NewTriggerBuilderFromExisting(a.Build())
 		assert.DeepEqual(t, b.Build(), a.Build())
-		b.RemoveFilter("source")
+		b.RemoveFilters([]string{"source"})
 		expected := &v1alpha1.TriggerFilter{
 			Attributes: &v1alpha1.TriggerFilterAttributes{
 				"type": "foo",
@@ -157,7 +157,7 @@ func TestTriggerBuilder(t *testing.T) {
 	t.Run("update filter with only updates", func(t *testing.T) {
 		b := NewTriggerBuilderFromExisting(a.Build())
 		assert.DeepEqual(t, b.Build(), a.Build())
-		b.AddFilter("type", "new")
+		b.Filters(map[string]string{"type": "new"})
 		expected := &v1alpha1.TriggerFilter{
 			Attributes: &v1alpha1.TriggerFilterAttributes{
 				"type":   "new",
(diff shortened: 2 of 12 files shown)


pkg/kn/commands/trigger/create.go
```diff
@@ -73,19 +72,19 @@ func NewTriggerCreateCommand(p *commands.KnParams) *cobra.Command {
...
 			}
-			if filters == nil {
-				return fmt.Errorf(
-					"cannot create trigger '%s' "+
-						"because filters are required", name)
-			}
 
-			trigger := constructTrigger(name, namespace, triggerUpdateFlags.Broker, filters)
-			trigger.Spec.Subscriber = &duckv1.Destination{
+			trigger := client_v1alpha1.NewTriggerBuilder(name)
```
rhuss: better call it `triggerBuilder`

pkg/kn/commands/trigger/create.go
```diff
@@ -73,19 +72,19 @@ func NewTriggerCreateCommand(p *commands.KnParams) *cobra.Command {
...
+			trigger.Namespace(namespace)
```
rhuss: you can chain the calls directly as in 

```golang
triggerBuilder := NewTriggerBuilder(name).Namespace(...).Broker(....).Subscriber(....)
```

pkg/kn/commands/trigger/create.go
```diff
@@ -73,19 +72,19 @@ func NewTriggerCreateCommand(p *commands.KnParams) *cobra.Command {
...
+			trigger.Broker(triggerUpdateFlags.Broker)
+			for k, v := range filters {
```
rhuss: We could conside to add a `Filters(map[string]string)` builder method so you could easily do the full chaining (and providers a nicer API for the builder)

pkg/kn/commands/trigger/create.go
daisy-ycguo: fixed.

pkg/kn/commands/trigger/create.go
daisy-ycguo: fixed.

pkg/kn/commands/trigger/create.go
daisy-ycguo: fixed.

pkg/eventing/v1alpha1/client.go
```diff
@@ -170,7 +170,7 @@ func (b *TriggerBuilder) Broker(broker string) *TriggerBuilder {
 	return b
 }
 
-func (b *TriggerBuilder) AddFilter(key, value string) *TriggerBuilder {
+func (b *TriggerBuilder) AddSingleFilter(key, value string) *TriggerBuilder {
```
rhuss: I'd suggest to keep `AddFilter` and jut add a `Filters` method for setting the full field (like for the other methods)

pkg/eventing/v1alpha1/client.go
```diff
@@ -198,6 +205,13 @@ func (b *TriggerBuilder) RemoveFilter(key string) *TriggerBuilder {
 	return b
 }
 
+func (b *TriggerBuilder) RemoveMultipleFilters(keys []string) *TriggerBuilder {
```
rhuss: Do wee need this ? (and also RemoveFilter). At the end its a builder with a very short lifetime. If you don't want to have something added, just don't add it. IMO, we should only have methods which 'build up' something.

daisy-ycguo: The remove filter will remove existing filters setting by `--filter name-`.
If I need to remove some existing filters, what should I do ? Get the existing map of filters, then remove some keys outside builder, and then set back to builder ?

pkg/eventing/v1alpha1/client.go
daisy-ycguo: In the latest version, I keep only Filters to add multiple filter.

pkg/eventing/v1alpha1/client.go
daisy-ycguo: I keep only RemoveFilters in the latest version to remove those filters setting by `--filter name-`

pkg/kn/commands/trigger/update.go
```diff
@@ -82,12 +82,7 @@ func NewTriggerUpdateCommand(p *commands.KnParams) *cobra.Command {
 					return fmt.Errorf(
 						"cannot update trigger '%s' because %s", name, err)
 				}
-				for k, v := range updated {
-					b.AddFilter(k, v)
-				}
-				for _, k := range removed {
-					b.RemoveFilter(k)
-				}
+				b.Filters(updated).RemoveFilters(removed)
```
rhuss: Sorry, my last nit-pick (promised :). What do you think about:

```golang
b.Filters(maputil.RemoveKeysFromMap(updated, removed))
```

and a `maputil.RemoveFrom()` (but maybe in an already existing util package):

```golang
func RemoveKeysFromMap(original map[string]string, toRemove []string) map[string]string {
     ....
}
```

That way this method could be re-used for other use cases and we don't have to add a `Remove` method to a builder.

wdyt ?

daisy-ycguo: never mind.
The problem of your suggestion is that I need to get the original value of filters map to remove. I think we want to hide the detail of trigger `Spec` by Builder. Except I add a method to Builder `GetFilters()`, I have to manage the trigger `Spec` structure to get the filter. The code may like below:
```
var attributes *map[string]string
attributes = nil
if b.Build().Spec.Filter != nil {
    attributes = b.Build().Spec.Filter.Attributes
}
b.Filter(maputil.AddAndRemoveFromMap(attributes, updated,removed))
```

and
```
func AddAndRemoveFromMap(original *map[string]string, toAdd map[string]string, toRemove []string) map[string]string {
......
}
```

If I add a `GetFilters()` to Builder, it may look like
```
b.Filter(maputil.AddAndRemoveFromMap(b.GetFilters(), updated,removed))
```
But it violates the rule that builder is to build something.

Do I make sense ?

rhuss: My suggestion:

```golang
// Extract filters directly from original trigger (not builder). Return nil if none.
existing := extractFilters(trigger)

// Use chained util method, so that they can bey reused in different contexts:
// * type StringMap map[string]string
// * func StringMap(map [string]string) StringMap : Just cast or copy to the new type (then call it "NewStringMap")
// * func (m SringMap) Merge(toMerge map[string]string) StringMap 
//    Keys from the toMerge map override those from the wrapped map
// * func (m map[string]string) RemovKeys(toRemove string[]) for removing the keys of a map. If this no
b.Filters(maputil.StringMap(existing).Merge(updated).Remove(removed))
```

This would have the benefit:

* Builder is still immutable and write-only
* Reusable string map manipulation functions which can be easily used in different contexts
* Easy to use API (maybe a bit harder to implement but obey one of the principles of good API design: "It should be as easy to use as possible, even when its might harder to implement that parts that are hidden from the user". 

Instead of introducing a "StringMap" you could equally cast to the type "StringMap" via `existing.(StringMap).Merge(updated).Remove(toRemove)`

Does this make sense ?

daisy-ycguo: Thanks. I made a few update to your suggestion.

I used builder as parameter in extractFilters, just to make sure no dependencies to `knative.dev/eventing/pkg/apis/eventing/v1alpha1` in the update command. Here is the definition of `extractFilters`
```
func extractFilters(builder *client_v1alpha1.TriggerBuilder) util.StringMap
```

rhuss: > I used builder as parameter in `extractFilters`, not trigger. I don't like to introduce a dependency to `knative.dev/eventing/pkg/apis/eventing/v1alpha1` in the update command. If I use trigger as parameter, I have to add this dependency. Here is the definition of `extractFilters`

Don't we already have the trigger available in this method ? with `extractFilters()` I mean just a function within `update.go` itself (just holding the loop that you've mentioned). I don't think we should put that into the builder.

I refer to this line: https://github.com/knative/client/blob/5c2ac9682343b141af049aee8dd084104fd14918/pkg/kn/commands/trigger/update.go#L69

rhuss: Ah, you mean a goimport. Good point, but we have the implicit dependency anyway (e.g. via the constructor arg of the builder or the return value). Alternatively you could add the function to client_v1alpha1 package, like `client_v1alpha1.ExtractFilters(trigger)`


It's better than diving into a builder. E.g. in generally when you call `Build()` on a Builder this is considered often to be a final action, so you can't call Build() a second time on the same builder (because it performs some more persisten action). So I would prefer to avoid that in favor of the utility in the client package.

daisy-ycguo: OK. I added a line to goimport in `update.go`, and used trigger as input parameter in the latest version. Please review. Thank you.

pkg/kn/commands/trigger/update.go
```diff
@@ -112,3 +110,16 @@ func NewTriggerUpdateCommand(p *commands.KnParams) *cobra.Command {
 
 	return cmd
 }
+
+func extractFilters(trigger *v1alpha1.Trigger) util.StringMap {
+	if trigger.Spec.Filter == nil {
+		return nil
```
rhuss: You should return an empty map here, otherwise you can a panic when you call `existing.Merge` and `existing` is nil. Its also inconsistent to return different things when `Filter == nil` or `Filter.Attributes == nil`

pkg/eventing/v1alpha1/client_test.go
```diff
@@ -142,22 +142,10 @@ func TestTriggerBuilder(t *testing.T) {
 		assert.DeepEqual(t, expected, b.Build().Spec.Filter)
 	})
 
-	t.Run("update filter with only deletions", func(t *testing.T) {
+	t.Run("update filter with add and update", func(t *testing.T) {
```
maximilien: this description reads odd... any way to make this clearer?

daisy-ycguo: fixed. Thank you.

pkg/kn/commands/trigger/update.go
daisy-ycguo: Changed it in the latest version. Thank you.

PR #605, opened 2020-01-10
refactor,fix(revision list): Refactoring and fixing issues

rhuss:
This PR fixes the following issues of `kn revision list`:

* `kn revision list -A` didn't work as the internal service-get requires a namespace (that from the revision)
* `kn revison list -s myservice myrevision` was not supported
* Service caching assumed that the revision list returned by the server is grouped by servicename. This is not guaranteed.
* `kn revision list -o yaml` returned yaml with the artificial and bogus annotations used for printing out in columns.
* No "0%" traffic is printed anymore

The following refactorings were added:

* The main run function was simplified and split up in multiple methods
* List-filter handling was made straight forward (and hopefully easier to understand)

I abandoned the initial intention to remove the temporary in-memory set annotations for tags and traffic as it does not fit well with the resource printer framework which only works with original K8s resource runtime. Object (and e.g. use serialization directly). I still don't like the hackish approach to misuse annotations as data container for temporary data ~~(please note, how I need to write a wrapped printer to get rid of the annotations for fixing the `-o yaml` use case, which otherwise prints out those annotations, which are not existant on the server-side)~~ (this has been replaced by a more direct approach to check for machine output), but for the sake of simplicity let's keep it (not dogmatic here).


rhuss:
cc: @maximilien

rhuss:
just removed the printer wrapper by checking directly whether a generic print output is requested. So should be simpler now.

maximilien:
You probably need to add some tests to bump the coverage of `pkg/kn/commands/revision/revision.go`

maximilien:
> chore: Simplified

Will review by next WG. However, can we agree to start that "Simplified" is a misnomer :)

rhuss:
> Will review by next WG. However, can we agree to start that "Simplified" is a misnomer :)

Thanks !  Really don't like the main function compared to the original monolithic (and faulty) run function ?

> * The main run function was simplified and split up in multiple methods

```golang
func(cmd *cobra.Command, args []string) error {
  namespace, err := p.GetNamespace(cmd)
  if err != nil {
    return err
  }
  client, err := p.NewServingClient(namespace)
  if err != nil {
    return err
  }

  // Create list filters
  var params []serving_v1alpha.ListConfig
  params, err = appendServiceFilter(params, client, cmd)
  if err != nil {
    return err
  }
  params, err = appendRevisionNameFilter(params, client, args)
  if err != nil {
    return err
  }

  // Query for list with filters
  revisionList, err := client.ListRevisions(params...)
  if err != nil {
    return err
  }

  // Stop if nothing found
  if len(revisionList.Items) == 0 {
    fmt.Fprintf(cmd.OutOrStdout(), "No revisions found.\n")
    return nil
  }

  // Only add temporary annotations if human readable output is requested
  if !revisionListFlags.GenericPrintFlags.OutputFlagSpecified() {
    err = enrichRevisionAnnotationsWithServiceData(p.NewServingClient, revisionList)
    if err != nil {
      return err
    }
  }

  // Sort by revision
  sortRevisionInfosByGeneration(revisionList)

  // Print out infos via printer
  printer, err := revisionListFlags.ToPrinter()
  if err != nil {
    return err
  }

  return printer.PrintObj(revisionList, cmd.OutOrStdout())
},
```

rhuss:
> You probably need to add some tests to bump the coverage of `pkg/kn/commands/revision/revision.go`

right, will do, if we can find some consense whether this PR is good enough to include.

pkg/kn/commands/revision/list_flags.go
```diff
@@ -55,7 +53,7 @@ func (f *RevisionListFlags) ToPrinter() (hprinters.ResourcePrinter, error) {
 	return p, nil
 }
 
-// AddFlags receives a *cobra.Command reference and binds
+// Flags receives a *cobra.Command reference and binds
```
navidshaikh: revert this change?

navidshaikh:
The output for `-A` flag needs fix
```
./kn revision list -A
NAME           SERVICE   TRAFFIC   TAGS   GENERATION   AGE    CONDITIONS   READY   REASON
foo1-jnmmc-1   foo1                       1            5d1h   0 OK / 4     False   ExitCode1 : Container failed with: Manage your Knative building blocks:

* Serving: Manage your services and release new software to them.
* Eventing: Manage event subscriptions and channels. Connect up event sources.

Usage:
  kn [command]

Available Commands:
  completion  Output shell completion code
  plugin      Plugin command group
  source      Event source command group
  trigger     Trigger command group

Flags:
      --config string        kn config file (default is $HOME/.kn/config.yaml)
  -h, --help                 help for kn
      --kubeconfig string    kubectl config file (default is $HOME/.kube/config)
      --log-http             log http traffic
      --lookup-plugins       look for kn plugins in $PATH
      --plugins-dir string   kn plugins directory (default "~/.kn/plugins")

Use "kn [command] --help" for more information about a command.
subcommand is required

hello-pcryk-1   hello   100%         1     4d    3 OK / 4   True   
svc-rngcc-1     svc     100%         1     11d   3 OK / 4   True   
```

pkg/kn/commands/revision/revision.go
```diff
@@ -29,3 +31,26 @@ func NewRevisionCommand(p *commands.KnParams) *cobra.Command {
...
+	}
+	var percent int64
+	var tags []string
+	for _, target := range service.Status.Traffic {
+		if target.RevisionName == revision {
+			if target.Percent != nil {
+				percent += *target.Percent
+			}
+			if target.Tag != "" {
+				tags = append(tags, target.Tag)
```
navidshaikh: For a revision tagged twice in targets: revision list will show the sum of traffic this revision is receiving and comma separated list of tags assigned. Should we list comma separated traffic portions as we're doing here for tags ?

rhuss: Well, I think this is an information that is not so important for a list overview entry. Important is the total amount of traffic targeted to this list. I'm fine with the current view.

pkg/kn/commands/revision/list_flags.go
rhuss: oops, not sure how this came in. Will fix.

rhuss:
Thanks for the feedback, will fix the -A view and add some more tests.

pkg/kn/commands/revision/revision.go
navidshaikh: > that is not so important for a list overview entry.

+1

maximilien:
> > * The main run function was simplified and split up in multiple methods

That I like but could be done in previous as well :) My "simplified" comment was w.r.t. other aspects of the code... which seems to have grown.

Anyhow, will try to do a thorough review after. Notice you are planning to make more changes from @navidshaikh's feedback.

rhuss:
> The output for `-A` flag needs fix

I can't reproduce it. The output looks really strange, too, with the help message inbetween list columns. Could you please tell me a bit more about your setup ?

rhuss:
> That I like but could be done in previous as well :) My "simplified" comment was w.r.t. other aspects of the code... which seems to have grown.

Cool. Let's try to get it simpler, looking forward to your feedback.

btw, one issue with the current approach of a namespaced `KnServingClient` is, when the namespace is nil (for -A), then only list operations work on that client instance, no get (you can't get a single object without a namespace).
So when you call `kn revision list -A` and we have to enrich service infos (tags, traffic), we have to create individual `KnServingClient` for each revision's namespace to get the referenced service. 

That makes the code a bit more complicated (as I'm also caching the clients) in `kn revision list` but as this is a corner case (other list operations don't require access to related objects), I don't think we have to revise our design decision to use a namespaced `KnServingClient`.

navidshaikh:
```
./kn revision list -A
NAMESPACE   NAME            SERVICE   TRAFFIC   TAGS   GENERATION   AGE     CONDITIONS   READY   REASON
default     svc-rngcc-1     svc       100%             1            12d     3 OK / 4     True    
tkn-kn      foo1-jnmmc-1    foo1                       1            5d23h   0 OK / 4     False   ExitCode1 : Container failed with: Manage your ...
tkn-kn0     hello-pcryk-1   hello     100%             1            4d23h   3 OK / 4     True    
```

The help text in between revision list was coming from container error message - Truncating the text brings the list columns in shape!
lgtm

rhuss:
> Nicer than first PR before I reviewed but lots more verbose than using the annotations ;)

Maybe you missed it, but this PR still uses the annotations that you have introduced. It's in fact the easier approach and most suitable in this situation because otherwise you would have to implement/fake a `runtime.Object` to be able to work with the printers package. That would the bigger evil (I tried it and it looks very ugly, bloated and confusing :).

If one takes care (e.g. don't add the annotations for machine output via `-o yaml`) the annotation piggy back is not so bad (apologies for being that negative at the beginning of the discussion). Still some risk to break later, but thats imo.

> Finally, IMO the traffic and tags info is so useful that it should be in `revision list` by default. But I am OK to wait for more feedback from real world usage.

wdym ? This is included in this PR and they are printed by default. Sorry, I don't get that point ... This PR was not about reverting your contribution, it was about to fix certain issues (see PR description) and refactor a bit. Part of the fix, unfortunately, included to make it a bit more complicated (i.e. when using --all-namespaces you need to create a fresh KnServingClient for each revision to use the revision's namespace for fetching its service).

pkg/kn/commands/revision/list.go
```diff
@@ -58,101 +62,192 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
...
+			if len(revisionList.Items) == 0 {
+				fmt.Fprintf(cmd.OutOrStdout(), "No revisions found.\n")
+				return nil
 			}
 
-			// sort revisionList by configuration generation key
-			sort.SliceStable(revisionList.Items, func(i, j int) bool {
-				a := revisionList.Items[i]
-				b := revisionList.Items[j]
+			// Add namespace column if no namespace is given ("-A")
```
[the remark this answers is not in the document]
navidshaikh: if user sets `-A`  (i.e. `namespace=""`), `WithNamespace` is set to `true`, thereby adding a Namespace column in the list.

pkg/kn/commands/revision/list.go
```diff
@@ -58,101 +62,192 @@ func NewRevisionListCommand(p *commands.KnParams) *cobra.Command {
...
 	default:
-		return nil, fmt.Errorf("'kn revision list' accepts maximum 1 argument")
+		return nil, fmt.Errorf("'kn revision list' accepts maximum 1 argument, not %d arguments as given", len(args))
+	}
+}
+
+// sortRevisions sorts revisions by namespace, service and generation (in this order)
+func sortRevisions(revisionList *v1alpha1.RevisionList) {
+	// sort revisionList by configuration generation key
+	sort.SliceStable(revisionList.Items, func(i, j int) bool {
```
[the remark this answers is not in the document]
rhuss: Do you mean to move this inner function to a named function ? That makes sense, will do.

pkg/kn/commands/revision/list.go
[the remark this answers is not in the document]
rhuss: yep `-A` is a shortoption for `--all-namespaces`

navidshaikh:
/retest

smoke tests: timeout: service 'hello' not ready after 600 seconds

rhuss:
> smoke tests: timeout: service 'hello' not ready after 600 seconds

At some point in time we should get rid of that flake :)

/retest

navidshaikh:
/retest

ERROR: timeout waiting for pods to come up
ERROR: Knative setup failed

PR #611, opened 2020-01-20
fixes(#598): checks for namespace creation instead of waiting

### LINKED ISSUE #598: Could e2e tests run faster?
## In what area(s)?

e2e

Classifications:
> /kind good-first-issue
> /kind tests

## Ask your question here:

Could the e2e tests be faster. There are places for possible easy optimizations.

For instance @navidshaikh has identified a `time.Sleep(20 * time.Second)` in line of [commons.go](https://github.com/knative/client/blob/e0877da3d66e0c93a51eac0a3c39577bbe3c31d3/test/e2e/common.go#L65) which perhaps could be reduced or even removed all together?

maximilien:
Fixes #598

## Proposed Changes

* checks for namespace creation instead of waiting

CHANGELOG.adoc:

- 🧽 Update or clean up current behaviour


test/e2e/common.go
```diff
@@ -127,14 +127,23 @@ func (test *e2eTest) DeleteTestNamespace(t *testing.T, namespace string) {
...
 		t.Fatalf("Error deleting namespace %s, timed out", namespace)
 	}
 }
 
+// WaitForNamespaceCreated wait until namespace is created
+func (test *e2eTest) WaitForNamespaceCreated(t *testing.T, namespace string) {
+	logger := Logger{}
+	created := checkNamespace(t, namespace, true, MaxRetries, logger)
+	if !created {
+		t.Fatalf("Error creating namespace %s, timed out", namespace)
```
duglin: %q so it's quoted. I think it would be better to return an error and print that instead of assuming it's due to a timeout. Then see next comment below....

test/e2e/common.go
```diff
@@ -127,14 +127,23 @@ func (test *e2eTest) DeleteTestNamespace(t *testing.T, namespace string) {
...
+	}
+}
+
 // Private functions
-func checkNamespaceDeleted(t *testing.T, namespace string, maxRetries int, logger Logger) bool {
+func checkNamespace(t *testing.T, namespace string, created bool, maxRetries int, logger Logger) bool {
```
duglin: I think it would be good to return an error (even in the "timeout" case) so that if something goes wrong we can echo it back to the user and they can try to fix it.

rhuss:
> CHANGELOG.adoc:
> 
>     * 🧽 Update or clean up current behaviour

Thanks for giving the CHANGELOG some love here, but it would be easier for me (or whoever makes the next release) if you would add this directory to CHANGELOG.adoc and add it as part of your PR.

test/e2e/common.go
```diff
@@ -143,7 +152,14 @@ func checkNamespaceDeleted(t *testing.T, namespace string, maxRetries int, logge
...
 		output, _ := kubectlGetNamespace()
-		if !strings.Contains(output, namespace) {
+
+		// check for namespace deleted
+		if !created && !strings.Contains(output, namespace) {
+			return true
+		}
+
+		// check for namespace created
+		if created && strings.Contains(output, namespace) {
```
rhuss: Can we add at least a minimal sleep (1 or 2s) at least ? Just to avoid to spam the API server. Ideally one would start a watch here for looking for ADD and DELETE events for the namespace, but if you do polling we should be a more patient.

test/e2e/common.go
maximilien: But the `t.Fatalf(..)` call will fail the test and the user will see the error...

test/e2e/common.go
maximilien:
```golang
t.Fatalf("Error creating namespace %s, timed out", namespace)
```
Fails the test and shows message to user.

test/e2e/common.go
duglin: Ignore my comments - I didn't realize this is just for testing - I thought this was "real" code :-)

maximilien:
> > CHANGELOG.adoc:
> > ```
> > * 🧽 Update or clean up current behaviour
> > ```
> 
> Thanks for giving the CHANGELOG some love here, but it would be easier for me (or whoever makes the next release) if you would add this directory to CHANGELOG.adoc and add it as part of your PR.

Not sure I understand this request... I essentially selected that line when submitting the PR. This comes from the template. Am I missing something?

test/e2e/common.go
maximilien: Sure, that's reasonable. Will update now.

maximilien: Actually there is a wait between retries in line 169: `time.Sleep(RetrySleepDuration)`

None needed in the `if` statement since that is the positive path :)

maximilien:
@navidshaikh and @rhuss can I get a LGTM here. This is done and ready and increases the speed of e2e in appreciable manner. Thx.

test/e2e/common.go
rhuss: sorry, that was not clear in the diff.

rhuss: @maximilien however, the retry interval is 30s (which is even longer than the 20s sleep). So when the namespace is not gone after the first request it takes at least 30s before the next try.

I suggest to reduce the retry timeout to 5s if you really aim for a considerable speedup.

rhuss:
> Not sure I understand this request... I essentially selected that line when submitting the PR. This comes from the template. Am I missing something?

Sorry, didn't state that clearly. Actually I would like to change the template to mention that this line should be directly added into the CHANGELOG.adoc file instead (and add it to the PR files) of just a comment in the PR. Because otherwise still someone need to pick that up and copy it over, which actually is not of much help as one has to go over every PR. At the moment we still have to do it, but for the future it would be really helpful.

No worries here, but I will update the template to add a note.

rhuss:
@maximilien well, I just saw that this is already mentioned in the template:

```
<!--
Release Note:

In the following cases, add a short description of PR to the unreleased section in CHANGELOG.adoc:

- 🎁 Add new feature
- 🐛 Fix bug
- 🧽 Update or clean up current behaviour
- 🗑️ Remove feature or internal logic

See other entries in CHANGELOG.adoc as an example.

PLEASE DON'T ADD THAT LINE HERE IN THE PULL-REQUEST DESCRIPTION BUT DIRECTLY IN CHANGELOG.ADOC AND ADD CHANGELOG.ADOC AS PART OF YOUR PULL-REQUEST.
-->
```

Do you have a suggestion how to state it more clearly ?

navidshaikh:
I tested the PR locally and works fine.

There are few improvements we can bring to e2e tests setup/teardown in a subsequent PR,
for example: parsing the error while creating the test namespace, if it `AlreadyExists`, 
we can delete and recreate it.

maximilien:
> @maximilien well, I just saw that this is already mentioned in the template:
> 
> ```
> 
> ```
> 
> Do you have a suggestion how to state it more clearly ?

Ah, OK. Make sense :) Thanks and will update and reduce the inter wait timeout. Cheers 🍻

test/e2e/common.go
rhuss: I refer to this line: https://github.com/knative/client/blob/1406f7de3db975ce2f0c0aae17a60feb1ac46e4e/test/e2e/common.go#L44

maximilien: OK, thx. Pushing update now.

PR #614, opened 2020-01-21
feature(serving): Add --autoscale-window

### LINKED ISSUE #613: Add --autoscale-window for managing the autoscale window
This should add an annotation and allow changing the scale down window on a per service basis.

rhuss:
Fixes #613

### DIFF
--- a/docs/cmd/kn_service_create.md
+++ b/docs/cmd/kn_service_create.md
@@ -44,7 +44,7 @@ kn service create NAME --image IMAGE [flags]
 ```
       --annotation stringArray    Service annotation to set. name=value; you may provide this flag any number of times to set multiple annotations. To unset, specify the annotation name followed by a "-" (e.g., name-).
       --async                     Create service and don't wait for it to become ready.
-      --autoscale-window string   Autoscale window duration for when to scale down. (eg: 10s)
+      --autoscale-window string   Duration to look back for making auto-scaling decisions. The service is scaled to zero if no request was received in during that time. (eg: 10s)
       --concurrency-limit int     Hard Limit of concurrent requests to be processed by a single replica.
       --concurrency-target int    Recommendation for when to scale up based on the concurrent number of incoming request. Defaults to --concurrency-limit when given.
   -e, --env stringArray           Environment variable to set. NAME=value; you may provide this flag any number of times to set multiple environment variables. To unset, specify the environment variable name followed by a "-" (e.g., NAME-).
--- a/docs/cmd/kn_service_create.md
+++ b/docs/cmd/kn_service_create.md
@@ -42,31 +42,32 @@ kn service create NAME --image IMAGE [flags]
 ### Options
 
 ```
-      --annotation stringArray   Service annotation to set. name=value; you may provide this flag any number of times to set multiple annotations. To unset, specify the annotation name followed by a "-" (e.g., name-).
-      --async                    Create service and don't wait for it to become ready.
-      --concurrency-limit int    Hard Limit of concurrent requests to be processed by a single replica.
-      --concurrency-target int   Recommendation for when to scale up based on the concurrent number of incoming request. Defaults to --concurrency-limit when given.
-  -e, --env stringArray          Environment variable to set. NAME=value; you may provide this flag any number of times to set multiple environment variables. To unset, specify the environment variable name followed by a "-" (e.g., NAME-).
-      --env-from stringArray     Add environment variables from a ConfigMap (prefix cm: or config-map:) or a Secret (prefix secret:). Example: --env-from cm:myconfigmap or --env-from secret:mysecret. You can use this flag multiple times. To unset a ConfigMap/Secret reference, append "-" to the name, e.g. --env-from cm:myconfigmap-.
-      --force                    Create service forcefully, replaces existing service if any.
-  -h, --help                     help for create
-      --image string             Image to run.
-  -l, --label stringArray        Service label to set. name=value; you may provide this flag any number of times to set multiple labels. To unset, specify the label name followed by a "-" (e.g., name-).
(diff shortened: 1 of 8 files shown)


rhuss:
Not 100% sure about the option name:

* `--autoscale-window`
* `--scale-window`
* `--window`

'guess autoscale is the best choice here ...

rhuss:
No idea about the IT test error:

```
   --- FAIL: TestTrafficSplit/update_tag_from_testing_to_staging_for_@latest_revision (6.58s)
        traffic_split_test.go:372: assertion failed: 
            --- expectedTargets
            +++ formattedActualTargets
              []e2e.TargetFields{
              	{
            - 		Tag:      "staging",
            + 		Tag:      "testing",
              		Revision: "echo5-rev-1",
              		Percent:  100,
              		Latest:   true,
              	},
              }
            
    --- PASS: TestTrafficSplit/update_tag_from_testing_to_staging_for_a_revision_(non_@latest) (12.73s)
```

Let's check whether this is a flake.

/retest

duglin:
I believe this PR uses this annotation: https://github.com/knative/serving/blob/master/pkg/apis/autoscaling/register.go#L67  right?

Note the comment in there says:
```
// WindowAnnotationKey is the annotation to specify the time
// interval over which to calculate the average metric
```

If so, does `Autoscale window duration for when to scale down` accurately describe it? When I read the cmd line option text it sounds like the value used will be the number of seconds before an idle pod will be killed. But is that true? I'm not fully up to speed on the usage of the 'window' but from the text in the golang code it looks more like it's used as part of the averaging logic, and not necessarily an upper-bound on the idle timeout of a pod. If I'm correct, then I think the option text needs to be changed to more accurately reflect what this value is used for.... or if we're looking to modify the idle timeout, we may need to tweak a different value in the yaml.

rhuss:
> If so, does `Autoscale window duration for when to scale down` accurately describe it?

In fact its not only for scaling down but for autoscaling in general (let me update that). If I understand correctly (but please correct me @markusthoemmes ), this value specifies the window after which the traffic average is caculated and the:

* If its 0 then scaled to 0
* If its larger than the concurrency target scaled up (based on much the current average concurrency is over the target value)

So the shortest possible description is probably "Time window after which to determine whether to autoscale" Does this sounds better ?

markusthoemmes:
It's the window **over** which statistics are averaged. We actually recalculate every two seconds today.

The sideffect is that we only scale down to 0 **after** we receive straight 0s for the entire window. That's because of math 😂 . The average will only be 0 if all values are 0.

duglin:
> Time window after which to determine whether to autoscale

That's better and if we can't come up with anything else then that's ok for now.
But, I do wonder if it's still a little misleading and your use of the word "granularity" in our slack chat got me wondering if we should make it clear that this flag doesn't necessarily have a direct relation to when things scale, but instead it controls how often we check "something" in our auto-scaling logic.

So, perhaps something more like:
`Time window over which the autoscaling metrics are averaged.`
A bit geeky, but sadly this flag is geeky :-)

markusthoemmes:
> but instead it controls how often we check "something" in our auto-scaling logic.

That's not true.

It in fact directly affects when things are scaled to 0 because of the maths behind it. The distinction between *checking after* and *calculating over* is super important here. The window simply specifies the amount of time we look in the past for historic data, maybe that's more useful?

```
The amount of historic data that is taken into account for autoscaling decisions.
```

rhuss:
> It in fact directly affects when things are scaled to 0 because of the maths behind it. The distinction between _checking after_ and _calculating over_ is super important here. The window simply specifies the amount of time we look in the past for historic data, maybe that's more useful?

To harden my understanding, for the *first* scaling event to happen, you need at least wait for this time window to pass, correct ? . 

I mean, if the window is 30 minutes, you have to wait 30 minutes until you scale up because of increased load. After that, the scaling decision is made every 2 seconds (with the data from the last 30 minutes). Is this correct? (asking because that was what I actually observer, i.e. that no scale-up happens when I set this window really large and I'm doing a load test before that window has been reached for the first time).

rhuss:
> The amount of historic data that is taken into account for autoscaling decisions.

I like that, maybe including _window_ as its part of the option name:

`Duration to look back for making autoscaling decisions. The service is scaled to zero if no request was received in during that time.`

duglin:
> I mean, if the window is 30 minutes, you have to wait 30 minutes until you scale up because of increased load. 

I don't think this is true because I can see things scale up immediately when I set CC=1 and all pods are busy. This is why I said this window is "part" of the logic w.r.t. scaling up.  But I'm sure @markusthoemmes will correct me if I'm wrong ;-)

markusthoemmes:
We don't wait for the window to fill first, no.

rhuss:
> We don't wait for the window to fill first, no.

ok, then I will have to investigate further why upscale didn't work for me when setting the window to 30 minutes, with a concurrency-target of 10 I get not more than 2 containers for 100 concurrent clients.

markusthoemmes:
Depending on which version you're running, the behavior might be slightly different.

As of our latest code, we assume that the entire window is "there". For a fresh revision we're therefore considering for example 29 minutes of straight 0s in your case. That'll heavily bias the scale towards the lower end.

rhuss:
> As of our latest code, we assume that the entire window is "there". For a fresh revision we're therefore considering for example 29 minutes of straight 0s in your case. That'll heavily bias the scale towards the lower end.

Ah, I see. But wouldn't it be better that you just take what you have during the first 30 minutes to calculate the average (e.g. average only over 10minutes if your service is running for 10minutes, not filling up 20 minutes with 0), maybe with some portion of leeway (eg soften by multiplying with some weighting function or adding a shorter duration of zeros, like 5minutes) ? But that's just a autoscaling greenhorn speaking here :)

PR #617, opened 2020-01-21
Add --pull-secret flag

### LINKED ISSUE #616: Support image pull secrets
Classifications:
> /kind feature

## Describe the feature:
Support `ImagePullSecrets` field as enabled in Serving v0.11.1 release.
Support setting and clearing image pull secrets for service to pull images behind registry auth.

navidshaikh:
Fixes #616

 - Add --pull-secret flag for service create/update operations
 - Setting empty string to flag clears the pull secrets
 - List `Image Pull Secret` in `service describe` default output

/lint

### DIFF
--- a/docs/cmd/kn_service_create.md
+++ b/docs/cmd/kn_service_create.md
@@ -62,6 +62,7 @@ kn service create NAME --image IMAGE [flags]
   -n, --namespace string          Specify the namespace to operate in.
       --no-lock-to-digest         do not keep the running image for the service constant when not explicitly specifying the image. (--no-lock-to-digest pulls the image tag afresh with each new revision)
   -p, --port int32                The port where application listens on.
+      --pull-secrets string       Image pull secrets to set. Empty image pull secrets will result to clear the pull secrets.
       --requests-cpu string       The requested CPU (e.g., 250m).
       --requests-memory string    The requested memory (e.g., 64Mi).
       --revision-name string      The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
--- a/docs/cmd/kn_service_create.md
+++ b/docs/cmd/kn_service_create.md
@@ -62,11 +62,11 @@ kn service create NAME --image IMAGE [flags]
   -n, --namespace string          Specify the namespace to operate in.
       --no-lock-to-digest         do not keep the running image for the service constant when not explicitly specifying the image. (--no-lock-to-digest pulls the image tag afresh with each new revision)
   -p, --port int32                The port where application listens on.
-      --pull-secrets string       Image pull secrets to set. Empty image pull secrets will result to clear the pull secrets.
+      --pull-secret string        Image pull secret to set. An empty argument ("") clears the pull secret. The referenced secret must exist in the service's namespace.
       --requests-cpu string       The requested CPU (e.g., 250m).
       --requests-memory string    The requested memory (e.g., 64Mi).
       --revision-name string      The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
-      --service-account string    Service account name to set. Empty service account name will result to clear the service account.
+      --service-account string    Service account name to set. An empty argument ("") clears the service account. The referenced service account must exist in the service's namespace.
       --volume stringArray        Add a volume from a ConfigMap (prefix cm: or config-map:) or a Secret (prefix secret: or sc:). Example: --volume myvolume=cm:myconfigmap or --volume myvolume=secret:mysecret. You can use this flag multiple times. To unset a ConfigMap/Secret reference, append "-" to the name, e.g. --volume myvolume-.
       --wait-timeout int          Seconds to wait before giving up on waiting for service to be ready. (default 600)
 ```
--- a/pkg/kn/commands/service/describe.go
+++ b/pkg/kn/commands/service/describe.go
@@ -171,6 +171,9 @@ func writeService(dw printers.PrefixWriter, service *v1alpha1.Service) {
 	if (service.Spec.Template != nil) && (service.Spec.Template.Spec.ServiceAccountName != "") {
 		dw.WriteAttribute("ServiceAccount", service.Spec.Template.Spec.ServiceAccountName)
 	}
+	if service.Spec.Template != nil && service.Spec.Template.Spec.ImagePullSecrets != nil {
+		dw.WriteAttribute("ImagePullSecrets", service.Spec.Template.Spec.ImagePullSecrets[0].Name)
+	}
 }
 
 // Write out revisions associated with this service. By default only active
--- a/pkg/kn/commands/service/describe.go
+++ b/pkg/kn/commands/service/describe.go
@@ -169,10 +169,10 @@ func writeService(dw printers.PrefixWriter, service *v1alpha1.Service) {
 		}
 	}
 	if (service.Spec.Template != nil) && (service.Spec.Template.Spec.ServiceAccountName != "") {
-		dw.WriteAttribute("ServiceAccount", service.Spec.Template.Spec.ServiceAccountName)
+		dw.WriteAttribute("Service Account", service.Spec.Template.Spec.ServiceAccountName)
 	}
 	if service.Spec.Template != nil && service.Spec.Template.Spec.ImagePullSecrets != nil {
-		dw.WriteAttribute("ImagePullSecrets", service.Spec.Template.Spec.ImagePullSecrets[0].Name)
+		dw.WriteAttribute("Image Pull Secret", service.Spec.Template.Spec.ImagePullSecrets[0].Name)
 	}
 }
 
--- a/pkg/serving/config_changes.go
+++ b/pkg/serving/config_changes.go
@@ -407,6 +407,18 @@ func UpdateServiceAccountName(template *servingv1alpha1.RevisionTemplateSpec, se
 	return nil
 }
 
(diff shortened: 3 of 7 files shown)


docs/cmd/kn_service_create.md
```diff
@@ -61,6 +61,7 @@ kn service create NAME --image IMAGE [flags]
   -n, --namespace string         Specify the namespace to operate in.
       --no-lock-to-digest        do not keep the running image for the service constant when not explicitly specifying the image. (--no-lock-to-digest pulls the image tag afresh with each new revision)
   -p, --port int32               The port where application listens on.
+      --pull-secrets string      Image pull secrets to set. Empty image pull secrets will result to clear the pull secrets.
```
maximilien: This "Empty image pull secrets will result to clear the pull secrets" reads odd...  Perhaps: "Empty image pull secrets will clear the pull secrets field"

pkg/serving/config_changes_test.go
```diff
@@ -551,6 +551,17 @@ func TestUpdateServiceAccountName(t *testing.T) {
...
 
+func TestUpdateImagePullSecrets(t *testing.T) {
+	template, _ := getV1alpha1RevisionTemplateWithOldFields()
+	template.Spec.ImagePullSecrets = nil
+
+	UpdateImagePullSecrets(template, "quay")
+	assert.Equal(t, template.Spec.ImagePullSecrets[0].Name, "quay")
+
+	UpdateImagePullSecrets(template, " ")
+	assert.Check(t, template.Spec.ImagePullSecrets == nil)
```
maximilien: Hmm, is that true. An empty space value will result in `nil`. From code above 

```golang
if pullsecrets == "" {
		template.Spec.ImagePullSecrets = nil
	} else {
		template.Spec.ImagePullSecrets = []corev1.LocalObjectReference{{
			Name: pullsecrets,
		}}
	}
```

seems like it would be an `ImagePullSecrets` with a `LocalObjectReference` of name " ". No?

test/presubmit-integration-tests-latest-release.sh
```diff
@@ -19,7 +19,7 @@
 # We currently take 0.10.0 for Serving and 0.10.2 for Eventing as the
 # latest release version.
 
-export KNATIVE_SERVING_VERSION="0.11.0"
+export KNATIVE_SERVING_VERSION="0.12.0"
```
maximilien: Does this belong here? Seems like #618 is more appropriate.

docs/cmd/kn_service_create.md
rhuss: Maybe: "No pull secret is used when no secret is provided as an argument"

btw, this is another example why we might to separate better the create and update help messages as for a `kn service create` an empty `--pull-secrets` doesn't make sense at all (and would be a no-op)

test/presubmit-integration-tests-latest-release.sh
rhuss: True, but as we are going to merge #618 soon anyway, I don't mind much.

pkg/kn/commands/service/describe.go
```diff
@@ -171,6 +171,9 @@ func writeService(dw printers.PrefixWriter, service *v1alpha1.Service) {
 	if (service.Spec.Template != nil) && (service.Spec.Template.Spec.ServiceAccountName != "") {
 		dw.WriteAttribute("ServiceAccount", service.Spec.Template.Spec.ServiceAccountName)
 	}
+	if service.Spec.Template != nil && service.Spec.Template.Spec.ImagePullSecrets != nil {
+		dw.WriteAttribute("ImagePullSecrets", service.Spec.Template.Spec.ImagePullSecrets[0].Name)
```
rhuss: I wonder whether we should use the label with spaces (like in "Image Pull Secrets", maybe also for "Service Account") as we use this for other parts in describe, too, for an easier consumption by a human reader.

But I just saw, that even for a `kubectl describe` this is not consistent (i.e. mixed usage of camel-case and space separated labels). So probably not a big thing :)

pkg/serving/config_changes.go
```diff
@@ -397,6 +397,18 @@ func UpdateServiceAccountName(template *servingv1alpha1.RevisionTemplateSpec, se
 	return nil
 }
 
+// UpdateImagePullSecrets updates the image pull secrets used for the corresponding knative service
+func UpdateImagePullSecrets(template *servingv1alpha1.RevisionTemplateSpec, pullsecrets string) {
+	pullsecrets = strings.TrimSpace(pullsecrets)
```
rhuss: @maximilien this is where leading and trailing spaces are removed, therefor a test with " " works (and also tests that trim)

docs/cmd/kn_service_create.md
navidshaikh: >  this is another example why we might to separate better the create and update help messages as for a kn service create an empty --pull-secrets doesn't make sense at all (and would be a no-op)

+1, good point!

I kept the message in format as we've for ServiceAccount
```
      --service-account string   Service account name to set. Empty service account name will result to clear the service account.
```
The pull secrets are set per revision of the service, you can create a service specifying the pull secrets and a private image, then update service with another revision with a public image and clearing the pull secrets.
Also note, created a service with pull-secrets and private image, any subsequent updates to service without image and empty pull secrets (`""`) works fine too.

pkg/serving/config_changes_test.go
navidshaikh: No. The zero value for ImagePullSecrets (type slice) is nil here.

test/presubmit-integration-tests-latest-release.sh
navidshaikh: I'd created this PR before 618 with version `0.11.1`. :-)
Let me take this off completely.

pkg/kn/commands/service/describe.go
navidshaikh: It occurred to me as well, though I went along with other fields as mentioned for ServiceAccount. Will change it to space separated.

docs/cmd/kn_service_create.md
navidshaikh: I've updated the help text as below for `--pull-secrets` and `--service-account` flag.
```
Image pull secrets to set. Empty image pull secrets will the clear the pull secrets.

```
```
Service account name to set. Empty service account name will clear the service account.
```

pkg/kn/commands/service/describe.go
navidshaikh: Changed to `Image Pull Secrets` and `Service Account`.

docs/cmd/kn_service_create.md
```diff
@@ -62,10 +62,11 @@ kn service create NAME --image IMAGE [flags]
   -n, --namespace string          Specify the namespace to operate in.
       --no-lock-to-digest         do not keep the running image for the service constant when not explicitly specifying the image. (--no-lock-to-digest pulls the image tag afresh with each new revision)
   -p, --port int32                The port where application listens on.
+      --pull-secrets string       Image pull secrets to set. Empty image pull secrets will the clear the pull secrets.
```
rhuss:
```suggestion
      --pull-secrets string       Image pull secrets to set. An empty argument will clear the pull secrets.
```

pkg/kn/commands/service/describe.go
```diff
@@ -169,7 +169,10 @@ func writeService(dw printers.PrefixWriter, service *v1alpha1.Service) {
 		}
 	}
 	if (service.Spec.Template != nil) && (service.Spec.Template.Spec.ServiceAccountName != "") {
-		dw.WriteAttribute("ServiceAccount", service.Spec.Template.Spec.ServiceAccountName)
+		dw.WriteAttribute("Service Account", service.Spec.Template.Spec.ServiceAccountName)
+	}
+	if service.Spec.Template != nil && service.Spec.Template.Spec.ImagePullSecrets != nil {
+		dw.WriteAttribute("Image Pull Secrets", service.Spec.Template.Spec.ImagePullSecrets[0].Name)
```
rhuss: So you are using only the first element, right ? IMO it would be more consequent to call the option `--image-pull-secret` (singular) than `--image-pull-secrets`

docs/cmd/kn_service_create.md
```diff
@@ -62,10 +62,11 @@ kn service create NAME --image IMAGE [flags]
...
       --requests-cpu string       The requested CPU (e.g., 250m).
       --requests-memory string    The requested memory (e.g., 64Mi).
       --revision-name string      The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
-      --service-account string    Service account name to set. Empty service account name will result to clear the service account.
+      --service-account string    Service account name to set. Empty service account name will clear the service account.
```
rhuss:
```suggestion
      --service-account string    Service account name to set. An empty argument clears the service account. The referenced service account must exist in the service's namespace.
```

rhuss: well, actually I suggest the change on in the edit_flags.go file. But youi know whay I mean ;-)

pkg/kn/commands/service/describe.go
navidshaikh: I removed `image` from it. Do you mean to add `image` in the flag as well apart from singular comment?

Yes, we're using only the first element.
IIUC only one image is allowed per Knative Service, but should we keep the flag name closer to the actual field name in spec?

navidshaikh:
/retest

```
Failed to successfully execute 'kn service create hello --image gcr.io/knative-samples/helloworld-go -n kne2etests6': Execution error: stderr: 'RevisionMissing: Revision "hello-mzmck-1" referenced in traffic not found.
```

navidshaikh:
New Changes: 
- Renamed `--pull-secrets` to singular `--pull-secret`
- describe output now lists keys as `Service Account` and `Image Pull Secret`
- Help messages as follow:
```
--pull-secret string        Image pull secret to set. An empty argument ("") clears the pull secret. The referenced secret must exist in the service's namespace.
```
and
```
--service-account string    Service account name to set. An empty argument ("") clears the service account. The referenced service account must exist in the service's namespace.
```

pkg/kn/commands/service/describe.go
rhuss: nope, `--pull-secret` is ok as the context is clear. My bad.

PR #625, opened 2020-01-27
feature(source binding): Support for Sink Binding as source

### LINKED ISSUE #624: Add `kn source binding ...` command
`kn source binding` set of commands should be added, perferably still for v0.12.0

The tricky part will be how to specify the `subject`. That probably needs to be split over multiple options like in `--subject apiVersion:kind:name` or `--subject apiVersion:kind --match labelkey1:labelvalue1 --match labelkey2:labelvalue2`

rhuss:
Sink bindings are managed like any other source. Sinks are specified as usual (with prefix and name),
'subjects' (the other end of the binding) is managed via a shortcut notation:

* with name: `<kind>:<apiVersion>:<name>`
* with label selector: `<kind>:<apiVersion>:key1=value1,key2=value2`

~With `--subject-namespace` and additional namespace can be provided (shoudl be possible for a sink, too but is not yet)~ As discussed on Slack, neither for a sink nor a subject the namespace is setable by the user. So I removed that option.

The implementation already uses the new sink binding from the `sources.knative.dev` group
and hence is a bit inconsistent to the still old usage kf `sources.eventing.knative.dev` for apiserver source and cronjob.

However as we will move over to `sources.knative.dev` very soon (right after v0.12.0) release,
this is was more appropriates.

Still WIP, but eventually fixes #624

Task list:

- [X] create
- [X] update
- [X] delete
- [X] describe
- [X] list


rhuss:
One question to the community: I would like to include this feature into 0.12.0 which is due tomorrow. However, because of time constraints and outstanding review, I would suggest to postpone the release to Wednesday (and I would some help with some quick reviewers ;-)

wdyt ? We can discuss this tomorrown on the WG call, however as I'm on a train at that time, not sure how it will work out for me.

rhuss:
ready for review

pkg/eventing/legacysources/v1alpha1/client.go
```diff
@@ -18,7 +18,7 @@ import (
 	client_v1alpha1 "knative.dev/eventing/pkg/legacyclient/clientset/versioned/typed/legacysources/v1alpha1"
 )
 
-// KnSourcesClient to Eventing Sources. All methods are relative to the
+// KnSinkBindingClient to Eventing Sources. All methods are relative to the
```
navidshaikh:
```suggestion
// KnSourcesClient to Eventing Sources. All methods are relative to the
```

pkg/kn/commands/source/binding/binding.go
```diff
@@ -0,0 +1,121 @@
...
+		Ref: destination.Ref,
+		URI: destination.URI,
+	}
+}
+
+func toReference(subjectArg string) (*tracker.Reference, error) {
+	parts := strings.SplitN(subjectArg, ":", 3)
+	if len(parts) < 3 {
+		return nil, fmt.Errorf("invalid subject argument '%s': not in format kind:api/version:nameOrSelector", subjectArg)
+	}
```
navidshaikh: lets use `strings.Split` instead.
Given an input like `a:b:c:d:e`, `SplitN` of 3 will always return 3 elements in slice (with last member of slice containing `:`)  and the `if` check here won't catch that.

Also we could use `strings.Trim` before doing split to remove any `:` given as prefix or suffix.

pkg/kn/commands/source/binding/binding.go
```diff
@@ -0,0 +1,121 @@
...
+	kind := parts[0]
```
navidshaikh: About the order in which kind, apiVersion, name/labelSelector are specified:
I think this differs from the order mentioned in the issue https://github.com/knative/client/issues/624#issue-555436582
`--subject apiVersion:kind:name`.  

I think we should parse `apiVersion` as the first part here followed by `kind`?

pkg/kn/commands/source/binding/binding.go
```diff
@@ -0,0 +1,121 @@
...
+	return reference, nil
+}
+
+func parseSelector(labelSelector string) (map[string]string, error) {
+	selector := map[string]string{}
+	for _, p := range strings.Split(labelSelector, ",") {
+		keyValue := strings.SplitN(p, "=", 2)
+		if len(keyValue) != 2 {
+			return nil, fmt.Errorf("invalid subject label selector '%s'. format: key1=value,key2=value", labelSelector)
+		}
```
navidshaikh: same comment as above, lets use `Split` here instead of `SplitN`

pkg/kn/commands/source/binding/binding.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	return reference, nil
+}
+
+func parseSelector(labelSelector string) (map[string]string, error) {
+	selector := map[string]string{}
+	for _, p := range strings.Split(labelSelector, ",") {
+		keyValue := strings.SplitN(p, "=", 2)
+		if len(keyValue) != 2 {
+			return nil, fmt.Errorf("invalid subject label selector '%s'. format: key1=value,key2=value", labelSelector)
```
navidshaikh:
```suggestion
			return nil, fmt.Errorf("invalid subject label selector '%s', expected format: key1=value,key2=value", labelSelector)
```

pkg/kn/commands/source/binding/create.go
```diff
@@ -0,0 +1,100 @@
...
+func NewBindingCreateCommand(p *commands.KnParams) *cobra.Command {
+	var bindingFlags bindingUpdateFlags
+	var sinkFlags flags.SinkFlags
+
+	cmd := &cobra.Command{
+		Use:   "create NAME --subject SCHEDULE --sink SINK --ce-override KEY=VALUE",
+		Short: "Create a sink binding source.",
+		Example: `
+  # Create a sink binding source, which connects a deployment 'myapp' with a Knative service 'mysvc'
+  kn source binding create my-binding --subject "" --sink svc:mysvc`,
```
navidshaikh: need value for `--subject`

docs/cmd/kn_source_binding.md
```diff
@@ -0,0 +1,35 @@
...
+```
+      --config string       kn config file (default is $HOME/.kn/config.yaml)
+      --kubeconfig string   kubectl config file (default is $HOME/.kube/config)
+      --log-http            log http traffic
+```
+
+### SEE ALSO
+
+* [kn source](kn_source.md)	 - Event source command group
+* [kn source binding create](kn_source_binding_create.md)	 - Create a sink binding source.
```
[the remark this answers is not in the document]
rhuss: good point. c&p error probably, I settled on "sink binding" (all lower case)

pkg/kn/commands/source/binding/create.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/spf13/cobra"
+
+	"knative.dev/client/pkg/kn/commands"
+	"knative.dev/client/pkg/kn/commands/flags"
+	v1alpha12 "knative.dev/client/pkg/sources/v1alpha1"
+	"knative.dev/client/pkg/util"
+)
+
+// NewCronJobCreateCommand is for creating CronJob source COs
```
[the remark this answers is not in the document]
rhuss: yeah, quite a bit has been taken over from other sources as blueprint. One of the first things after 0.12.0 is out is to combine duplicated code.

I also will run a grep through the binding files to find any left-overs like this.

pkg/kn/commands/source/binding/describe.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func writeSink(dw printers.PrefixWriter, namespace string, sink *duckv1.Destination) {
+	subWriter := dw.WriteAttribute("Sink", "")
+	if sink.Ref.Namespace != "" && sink.Ref.Namespace != namespace {
+		subWriter.WriteAttribute("Namespace", sink.Ref.Namespace)
+	}
+	subWriter.WriteAttribute("Name", sink.Ref.Name)
+	ref := sink.Ref
+	if ref != nil {
```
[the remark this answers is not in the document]
rhuss: Really ? Don't think so. If `sink.Ref` is nil then `ref` just is set to `nil`, too. Its only when you would have used `sink.Ref.something` that you would get a NPE.

pkg/kn/commands/source/binding/flags.go
```diff
@@ -0,0 +1,122 @@
...
+		r, err := printSinkBinding(&binding, options)
+		if err != nil {
+			return nil, err
+		}
+		rows = append(rows, r...)
+	}
+	return rows, nil
+}
+
+// subjectToString converts a reference to a string representation
```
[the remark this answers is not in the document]
rhuss: yes, it's probably to have in `binding.go` as the reference, for now, is really binding specific (not yet useful of that package). Will move it.

When we see that it makes sense to use outside of the binding package, we can make it public and move it up.

pkg/kn/commands/source/binding/list.go
```diff
@@ -0,0 +1,74 @@
...
+
+	listCommand := &cobra.Command{
+		Use:   "list",
+		Short: "List sink bindings.",
+		Example: `
+  # List all sink binding in YAML format
+  kn source binding list -o yaml`,
+
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			// TODO: filter list by given source name
```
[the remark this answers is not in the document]
rhuss: Agree that long-living TOODs ;-P are a bad smell. I definitely plan some to spent some cycles for a cleanup early next week (including to get rid of all TODOs.

The filter here is a bit ambivalent, as it still needs to discuss how to implement such a filter: Prefix matching, exact matching, regexp matching ? Should then be implemented for all list like functionality the same way and at once (therefor you find similar todos on other list command)

So, let's keep it for now. But you should beat me again if this stays longer here than the next release :)

pkg/kn/commands/source/binding/binding.go
rhuss: I deliberately to split on max 3, so that the final part could contain ':' in label values. If its less than 3 parts, you still only get 1 or 2 so that check still works. 

So what would you do when having 4, 5, ... parts  ? You could ignore them or join them again with ":". So  SplitN is better here, too.

Convinced ;-) ?

pkg/kn/commands/source/binding/binding.go
rhuss: Ah good point about the documentation. I also found that `list` use the other order.

I've chosen `kind:apiVersion` because `trigger` does it the same way. Not saying this is the correct way (however I tend to like to see the kind as most important information which sould come first), but we should be consistent. As we cant change the triggers right now, letst adopt to the trigger's convention.

pkg/kn/commands/source/binding/binding.go
rhuss: Again my argument that with SplitN you can allow labels holding `=` as values. Not sure how realistic this is, but definitely possible. If you use `Split` you have to check the case when its _more than 2_ parts. With SplitN you can be sure that its _at most 2_ parts (but can be less)

pkg/kn/commands/source/binding/binding.go
navidshaikh: Yes ;-)

pkg/kn/commands/source/binding/binding.go
navidshaikh: :+1:

pkg/kn/commands/source/binding/binding.go
navidshaikh: sounds good, thanks!

pkg/kn/commands/source/binding/delete.go
```diff
@@ -0,0 +1,55 @@
...
+func NewBindingDeleteCommand(p *commands.KnParams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "delete NAME",
+		Short: "Delete a sink binding.",
+		Example: `
+  # Delete a sink binding with name 'my-binding'
+  kn source binding delete my-binding`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
+				return errors.New("'requires the name of the sink bindinbg to delete as single argument")
```
navidshaikh:
```suggestion
				return errors.New("requires the name of the sink binding to delete as single argument")
```

pkg/kn/commands/source/binding/describe.go
```diff
@@ -0,0 +1,146 @@
...
+func NewBindingDescribeCommand(p *commands.KnParams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "describe NAME",
+		Short: "Describe a sink binding.",
+		Example: `
+  # Describe a sink binding with name 'mysinkbinding'
+  kn source binding describe mysinkbinding`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 1 {
+				return errors.New("'kn source binding describe' requires the name of the source as argument")
```
navidshaikh:
```suggestion
				return errors.New("'kn source binding describe' requires the name of the source as single argument")
```

pkg/kn/commands/source/binding/update.go
```diff
@@ -0,0 +1,96 @@
...
+// NewBindingUpdateCommand prepares the command for a sink binding update
+func NewBindingUpdateCommand(p *commands.KnParams) *cobra.Command {
+	var bindingFlags bindingUpdateFlags
+	var sinkFlags flags.SinkFlags
+
+	cmd := &cobra.Command{
+		Use:   "update NAME --subject SCHEDULE --sink SINK --ce-override OVERRIDE",
+		Short: "Update a sink binding.",
+		Example: `
+  # Update the the subject of a sink binding 'my-binding' to a new cronjob with label selector 'app=ping'
```
navidshaikh:
```suggestion
  # Update the subject of a sink binding 'my-binding' to a new cronjob with label selector 'app=ping'  
```

pkg/kn/commands/source/binding/update.go
```diff
@@ -0,0 +1,96 @@
...
+	cmd := &cobra.Command{
+		Use:   "update NAME --subject SCHEDULE --sink SINK --ce-override OVERRIDE",
+		Short: "Update a sink binding.",
+		Example: `
+  # Update the the subject of a sink binding 'my-binding' to a new cronjob with label selector 'app=ping'  
+  kn source binding update my-binding --subject cronjob:batch/v1beta1:app=ping"`,
+
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			if len(args) != 1 {
+				return errors.New("name of sink binding required")
```
navidshaikh:
```suggestion
				return errors.New("requires the name of the sink binding to update as single argument")
```

pkg/kn/commands/source/binding/update.go
```diff
@@ -0,0 +1,96 @@
...
+				if err != nil {
+					return err
+				}
+				b.Subject(reference)
+			}
+			binding, err := b.Build()
+			if err != nil {
+				return err
+			}
+			err = sinkBindingClient.UpdateSinkBinding(binding)
```
navidshaikh: dont see the `ce-override` flags processing for update?

rhuss: Good point, will add it.

rhuss: good catch. Just added it.

pkg/sources/v1alpha1/binding_client.go
```diff
@@ -0,0 +1,263 @@
+// Copyright © 2019 The Knative Authors
```
navidshaikh: `s/trigger/sinkBinding/gI`

pkg/kn/commands/source/binding/update.go
```diff
@@ -0,0 +1,101 @@
...
+				b.Sink(toDuckV1(destination))
+			}
+			if cmd.Flags().Changed("subject") {
+				reference, err := toReference(bindingFlags.subject, namespace)
+				if err != nil {
+					return err
+				}
+				b.Subject(reference)
+			}
+			err = updateCeOverrides(bindingFlags, b)
```
navidshaikh: will need a check here like above flags if it's changed `if cmd.Flags().Changed("ce-overrides")`, otherwise update operation without this flag would nil any key value pair set in create.

rhuss: ah, right. Next round ;-)

rhuss: The check is in the update function itself. We could move it upwards, but lets do it next time.

navidshaikh: my bad,
the check is covered in `updateCeOverrides` function as
`if bindingFlags.ceOverrides != nil {`

PR #638, opened 2020-02-05
Fix capitalization of generated flags

### LINKED ISSUE #636: generated negative flag `--no-xxx` description are not well capitalized
## In what area(s)?

utilities

Classifications:
> /kind good-first-issue
> /kind cleanup

## Describe the feature:

Found during PR #629. It's a _nitpick_ issue but one that should eventually be addressed nonetheless.

When autogenerating the negative of a `bool` flag, e.g., `--cluster-local` and `--no-cluster-local`, the description of the positive flag is used to create the description of the negative by adding: `do not ...` so in this case:

```text
do not Specify that the service be private. (--no-cluster-local will make the service publicly available (default true)
```

which is not correct... Same for other `--no-xxx` flags like `--no-lock-to-digest` whose positive description starts with lower case. Suggestion is to make the generated description smartly generate:

```text
Do not specify that the service be private. (--no-cluster-local will make the service publicly available (default true)
```

So add `Do not` and correctly capitalize the start of the positive description.

cc @navidshaikh

dsimansk:
Fixes #636 

## Proposed Changes

* When generating negative flag, change first character of original usage desc to lower case


rhuss:
```
Failed to successfully execute 'kn service create hello --image gcr.io/knative-samples/helloworld-go -n kne2etests6': Execution error: stderr: 'RevisionMissing: Revision "hello-pvkvk-1" referenced in traffic not found.
' error: 'exit status 1'FAIL	knative.dev/client/test/e2e	31.400s
FAIL
```

Same error as in #640 , we should fix that flake somehow.

rhuss:
Again:

```
Failed to successfully execute 'kn service create hello --image gcr.io/knative-samples/helloworld-go -n kne2etests4': Execution error: stderr: 'RevisionMissing: Revision "hello-hrwhj-1" referenced in traffic not found.
' error: 'exit status 1'FAIL	knative.dev/client/test/e2e	38.095s
FAIL
```

I think has todo with BYO revision name feature (as the name here is clearly one that we generate on the client side.) From my feeling it looks like that the route is created before the revision has been established.

rhuss:
Flake should be fixed now with #644 

/retest

PR #639, opened 2020-02-06
Rename --async to --no-wait and deprecate it

### LINKED ISSUE #438: Rename --async to --no-wait
For operations for that we support sync/async, we default to wait, and offer an option `--async` for background operations.

I suggest renaming this to `--no-wait` as it seems that `--wait` is the common way to specify this behaviour when it is not the default. We can also introduce a `--wait` that would be a no-op now but help users to express their intent better and save them from changing of default values.

Other projects where I see `--wait` as options are:

* Riff
* Helm
* TriggerMesh CLI

and probably some more.

dsimansk:
Fixes #438

## Proposed Changes

* Add `--no-wait` flag with same behaviour as `--async`
* Mark `--async` as deprecated in documentation and display warning message when used

Example of deprecation warning:
```bash
➜  client git:(issue-438) kn service create svc-test --image gcr.io/knative-samples/helloworld-go --async  

DEPRECATED WARNING: flag --async is going to be removed in future release, please use --no-wait instead.

Service 'svc-test' created in namespace 'default'.
➜  client git:(issue-438) kn service update svc-test --image gcr.io/knative-samples/helloworld-go --async

DEPRECATED WARNING: flag --async is going to be removed in future release, please use --no-wait instead.

Service 'svc-test' updated in namespace 'default'.
➜  client git:(issue-438) kn service update svc-test --image gcr.io/knative-samples/helloworld-go --no-wait
Service 'svc-test' updated in namespace 'default'.

```

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -28,6 +28,10 @@
 | Add `--cmd` and `--arg` for customization of container entrypoint
 | https://github.com/knative/client/pull/635[#635]
 
+| 🧽
+| Add `no-wait` instead of `--async` and deprecate it
+https://github.com/knative/client/pull/639[#639]
+
 ## v0.12.0 (2020-01-29)
 
 [cols="1,10,3", options="header", width="100%"]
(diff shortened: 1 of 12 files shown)


rhuss:
> Wonder if we should have a time limit on DEPRECATED flags and commands? Perhaps mentioning in the docs that commands/flags marked as DEPRECATED will be removed in future after `n` number of releases? I'd vote for `n == 2`. Thoughts?

Yes, I think we definitely should establish a rule. For API versioning we have [Knative Release Principles](https://docs.google.com/document/d/1GXDe6163lO8ohtLUMPCplbjm3OTpYo_4U8f9K1fgZPo/edit) and there is a rule to support `n ... n-3` (so that in this case, we would remove the option in kn 0.16.0). 

On the other hand, there is also the table about deprecation, which states which feature can be removed after which time, based on the maturity level.

So if we would say `kn` itself is "alpha", then we could remove immediately. If we say "beta", then only after 9 months. However, we have not decided how we label ourselves (i would say, still alpha, but we could also say, the serving part is beta, eventing alpha. It's difficult. :)

PR #647, opened 2020-02-11
Fix --image flag to only allow single occurence

### LINKED ISSUE #533: Does not reject multiple `--image` parameters

> Paste output of 'kn version'

```
$ kn version
Version:      v0.10.0
Build Date:   2019-11-07 15:02:21
Git Revision: 22c65d6
Supported APIs:
- serving.knative.dev/v1alpha1 (knative-serving v0.10.0)
```

0.10

## Expected Behavior

When I type `kn update service --image example.com/first --image example.com/second`, I expect to be told that I cannot provide two images.

## Actual Behavior

`kn` accepts the command line without error or warning. It applies the last-provided argument. In the example above this means that the updated service has `example.com/second` as its image.

1. Create a service.
1. Update the service, using two different `--image` parameters.
1. Observe that no error occurs.
1. Use describe to observe that second parameter is applied.

dsimansk:
Fixes #533 

## Proposed Changes

* Introduce new type `singletonString` for flag values that only allow single value assignment, otherwise return error
* Fix `--image` flag with the new type.

Example:
```bash
➜  client git:(issue-533) kn service create svc-test --image gcr.io/knative-samples/helloworld-go --image gcr.io/foo/bar:barz --no-wait 
invalid argument "gcr.io/foo/bar:barz" for "--image" flag: value is already set
```

### DIFF
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -30,7 +30,7 @@ import (
 
 type ConfigurationEditFlags struct {
 	// Direct field manipulation
-	Image   singletonString
+	Image   uniqueStringArg
 	Env     []string
 	EnvFrom []string
 	Mount   []string
@@ -67,24 +67,24 @@ type ResourceFlags struct {
 	Memory string
 }
 
-// -- singletonString Value
+// -- uniqueStringArg Value
 // Custom implementation of flag.Value interface to prevent multiple value assignment.
-// Useful to enforce single use of flag, e.g. --image.
-type singletonString string
+// Useful to enforce unique use of flags, e.g. --image.
+type uniqueStringArg string
 
-func (s *singletonString) Set(val string) error {
+func (s *uniqueStringArg) Set(val string) error {
 	if len(*s) > 0 {
-		return errors.New("value is already set")
+		return errors.New("can be provided only once")
 	}
-	*s = singletonString(val)
+	*s = uniqueStringArg(val)
 	return nil
 }
 
-func (s *singletonString) Type() string {
+func (s *uniqueStringArg) Type() string {
 	return "string"
 }
 
-func (s *singletonString) String() string { return string(*s) }
+func (s *uniqueStringArg) String() string { return string(*s) }
 
 // markFlagMakesRevision indicates that a flag will create a new revision if you
 // set it.
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -30,7 +30,7 @@ import (
 
 type ConfigurationEditFlags struct {
 	// Direct field manipulation
-	Image   string
+	Image   singletonString
 	Env     []string
 	EnvFrom []string
 	Mount   []string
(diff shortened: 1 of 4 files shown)


pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -67,6 +67,25 @@ type ResourceFlags struct {
...
 }
 
+// -- singletonString Value
+// Custom implementation of flag.Value interface to prevent multiple value assignment.
+// Useful to enforce single use of flag, e.g. --image.
+type singletonString string
+
+func (s *singletonString) Set(val string) error {
+	if len(*s) > 0 {
+		return errors.New("value is already set")
```
rhuss: I'm not sure how this "feels" like on the CLI. If this is the only error printed, then it might be confusing (especially when you have a large list of options). It would be better to actually print *which* argument was duplicated (i.e. `--image` here). I think you could implement that by using a type like

```golang
type uniqueStringArg struct {  
   // which kind of argument it is
    kind string 

   // Argument value
   value string
}

func newUniqueStringArg(kind string) uniqueStringArg {
     return uniqueStringArg{kind: kind}
}
```

and then partially pre-initialize the ConfiguarionEditFlags like in 

```golang
ceF = ConfigurationEditFlags{
    Image: newUniqueStringArg("image")
}
```

and then finally adapt the error message to include the `kind`. 

Does this make sense ?

dsimansk: Actually, the error already contains the flag.
```
```

dsimansk: Maybe a bit more descriptive error message?

rhuss: Ah, if this is the case, then I'm fine with having the current solution.

However for the error message I woudl use something like "flag can be provided only once" (so from the POV of the user, not from the flag's POV)

dsimansk: OK, anyway I do like your suggested name `uniqueStringArg` a bit more. I'll change the name, error message and finally add some unit tests.

dsimansk: Reflected review feedback in https://github.com/knative/client/pull/647/commits/c68fe1dbf790fab2b341d56de8f543ad9c80378d and added unit tests in https://github.com/knative/client/pull/647/commits/c9902886ceed29d5fe57068f603dedb8500636d2.

pkg/kn/commands/service/create_test.go
```diff
@@ -134,6 +134,12 @@ func TestServiceCreateImage(t *testing.T) {
 	}
 }
 
+func TestServiceCreateWithMultipleImages(t *testing.T) {
+	_, _, _, err := fakeServiceCreate([]string{
+		"service", "create", "foo", "--image", "gcr.io/foo/bar:baz", "--image", "gcr.io/bar/foo:baz", "--no-wait"}, false, false)
+	assert.Error(t, err, "invalid argument \"gcr.io/bar/foo:baz\" for \"--image\" flag: can be provided only once")
```
rhuss: What we also do typically in unit test to assert on the content of the error message: not the exact wording but that key context information is included. This helps us to ensure that the error message is meaningful, but the test is also as robust as possible so that the exact wording could be changed without breaking the tests.

We have an utility `util.ContainsAll()` which helps here. Check for other tests that use that method.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -67,6 +67,25 @@ type ResourceFlags struct {
 	Memory string
 }
 
+// -- uniqueStringArg Value
+// Custom implementation of flag.Value interface to prevent multiple value assignment.
```
maximilien: Clever :)

PR #658, opened 2020-02-12
Harmonize list headers for trigger and sources

### LINKED ISSUE #567: Harmonize list headers for "trigger list", "cronjob list" and "apiserver list"
Actually the columns wrt to Sink / Subscription are different:

* CronJon and Apiserver source display the sink shortcut ('svc:myservice')
* Trigger shows the resolved URL ("SUBSCRIBER_URL")

Also, the condition and reason columns should be the same as for services. 
Should we add AGE, too ?

daisy-ycguo:
Fixes #567

## Proposed Changes

* Change `SUBSCRIBER_URI` to `SINK` in trigger list
* Add `CONDITIONS` to trigger list
* Add `AGE` to trigger and source ( cronjob, apiserver, binding ) list
* Move `SinkToString` to package `flags` and reuse in trigger list

### DIFF
--- a/pkg/kn/commands/flags/sink.go
+++ b/pkg/kn/commands/flags/sink.go
@@ -105,3 +105,18 @@ func parseSink(sink string) (string, string) {
 		return parts[0], parts[1]
 	}
 }
+
+// SinkToString prepares a sink for list output
+func SinkToString(sink duckv1.Destination) string {
+	if sink.Ref != nil {
+		if sink.Ref.Kind == "Service" {
+			return fmt.Sprintf("svc:%s", sink.Ref.Name)
+		} else {
+			return fmt.Sprintf("%s:%s", strings.ToLower(sink.Ref.Kind), sink.Ref.Name)
+		}
+	}
+	if sink.URI != nil {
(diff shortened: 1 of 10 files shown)


pkg/kn/commands/flags/sink.go
```diff
@@ -105,3 +105,18 @@ func parseSink(sink string) (string, string) {
...
 	}
 }
+
+// SinkToString prepares a sink for list output
+func SinkToString(sink duckv1.Destination) string {
+	if sink.Ref != nil {
+		if sink.Ref.Kind == "Service" {
+			return fmt.Sprintf("svc:%s", sink.Ref.Name)
+		} else {
+			return fmt.Sprintf("%s:%s", sink.Ref.Kind, sink.Ref.Name)
```
rhuss: can we make the kind lower case ? That's how we use it normally.

daisy-ycguo: Done. covert to lower case in the latest version.

rhuss:
thanks !

/lgtm

PR #660, opened 2020-02-12
feature(service update): Add generic mechanism to include retries on conflicts

rhuss:
Move the retry logic for retrying in case of a resource conflict into the KnClient.

This is wip (e.g. unit tests are not working yet), but demonstrate how to move the retry logic into the client. If we decided that its worth the effort, I will update the tests, too.

/hold

### DIFF
--- a/pkg/kn/commands/service/update.go
+++ b/pkg/kn/commands/service/update.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 
 	"github.com/spf13/cobra"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
 	"knative.dev/client/pkg/kn/commands/flags"
 	"knative.dev/client/pkg/kn/traffic"
@@ -74,15 +73,12 @@ func NewServiceUpdateCommand(p *commands.KnParams) *cobra.Command {
 				return err
 			}
 
-			var retries = 0
-			for {
-				name := args[0]
-				service, err := client.GetService(name)
-				if err != nil {
-					return err
-				}
-				service = service.DeepCopy()
-				latestRevisionBeforeUpdate := service.Status.LatestReadyRevisionName
+			// Use to store the latest revision name
+			var latestRevisionBeforeUpdate string
+			name := args[0]
+
+			updateFunc := func(service *servingv1.Service) (*servingv1.Service, error) {
+				latestRevisionBeforeUpdate = service.Status.LatestReadyRevisionName
 				var baseRevision *servingv1.Revision
 				if !cmd.Flags().Changed("image") && editFlags.LockToDigest {
 					baseRevision, err = client.GetBaseRevision(service)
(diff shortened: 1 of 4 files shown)


dsimansk:
@rhuss just a little side note, you should rebase on master to get the latest unit tests change. (I was bit by it already :)).

navidshaikh:
+1

rhuss:
/unhold

think it would be good to review now (and merge maybe)

pkg/kn/commands/service/update.go
```diff
@@ -92,48 +88,46 @@ func NewServiceUpdateCommand(p *commands.KnParams) *cobra.Command {
...
-				} else {
-					if waitFlags.Async {
-						fmt.Fprintf(out, "\nWARNING: flag --async is deprecated and going to be removed in future release, please use --no-wait instead.\n\n")
-					}
-					fmt.Fprintf(out, "Service '%s' updated in namespace '%s'.\n", args[0], namespace)
+				fmt.Fprintln(out, "")
+				return showUrl(client, name, latestRevisionBeforeUpdate, "updated", out)
+			} else {
+				if waitFlags.Async {
+					fmt.Fprintf(out, "\nWARNING: flag --async is deprecated and going to be removed in future release, please use --no-wait instead.\n\n")
```
maximilien: Isn't this renamed to --no-wait with #438

rhuss: Yes, but we still accept `--async` even it is deprecated.

The deprecation process is supposed to be:
* Support both `--no-wait` and `--async` for the next n releases (n == 4 for knative/serving, but I think we could be faster). 
* Print a warning if the user uses a deprecated, with a hint to the replacement.

rhuss:
Agreed, we should move the constant upwards, and reuse it. I'm also thinking about to make this configurable and allow the user to chose, whether she wants the automatic retry or not (as this can be also dangerous and the optimistic locking feature is here for some reason). I will open a new issue for that.

I would suggest, that the second implementation of that pattern moves the constant upwards. (And leave it for now here)

rhuss:
@daisy-ycguo : Here's the issue for an enhancement request for those retry feature: #662

pkg/serving/v1/client.go
```diff
@@ -207,6 +218,37 @@ func (cl *knServingClient) UpdateService(service *servingv1.Service) error {
 	return updateServingGvk(service)
 }
 
+// Update the given service with a retry in case of a conflict
+func (cl *knServingClient) UpdateServiceWithRetry(name string, updateFunc serviceUpdateFunc, nrRetries int) error {
+	return updateServiceWithRetry(cl, name, updateFunc, nrRetries)
+}
+
+// Extracted to be usable with the Mocking client
+func updateServiceWithRetry(cl KnServingClient, name string, updateFunc serviceUpdateFunc, nrRetries int) error {
```
navidshaikh: should there be more cases to retry other than `apierrors.IsConflict(err)`; we can make that configurable as well, need to check.

navidshaikh:
/retest

```
failed: Post https://webhook.knative-serving.svc:443/?timeout=30s: no endpoints available for service "webhook"
' error: 'exit status 1'FAIL	knative.dev/client/test/e2e	3.039s
```

PR #659, opened 2020-02-12
e2e: Dump stdout and describe ksvc

navidshaikh:
/hold

PR #661, opened 2020-02-12
fix(service wait): Wrong check for error window

rhuss:
The error window introduced with #644 had a wrong conditional. Fixed that and added a test which would have detected this.

Also, this should fix some issues which we tried to detect with #659.


rhuss:
@navidshaikh fix for a stupid error. btw, I'm not sure anymore whether the original flake really was a race in reconciliation, it looks more like an optimistic lock error. See https://github.com/knative/serving/issues/6784 for tracking this.

PR #667, opened 2020-02-14
chore(e2e): Refactor to better track errors

rhuss:
* Collect results from various `kn` calls to provide the context
* Changed sub-Run() commands to t.Log() to reduce the log noise and get file + line information in the log output
* Removed any Fatal() calls to allows the test  to properly finish
* Only output via t.Log() to get the output at the proper places and allow Prow to better pick up the test results.
* Most important: EMOJIS !! 🦆🦆🦆🦆🦆


rhuss:
@navidshaikh I think this is good to merge and makes troubleshooting e2e issues much easier.

PR #668, opened 2020-02-14
feat(XDG home): use XDG_CONFIG_HOME as default config location

### LINKED ISSUE #429: Use XDG_CONFIG_HOME for location of the `kn` config
## In what area(s)?

/kind feature

## Describe the feature:
Rather than contribute to the config pollution of the `$HOME` dir on Unix based systems can we following the [XDG Base Directory Specification](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html)?

dsimansk:
Before proceeding further, I'd like to ask for feedback for the following change, especially to the deprecated fallback part and plugin dir handling. 

@rhuss @navidshaikh 
Does it seem like a right direction to go?

TODO:
* Update docs about config location
* Update help/usage messages with new default location
* Add tests

Fixes #429 

## Proposed Changes

* Change default config path to: 
  * `~/.config/kn` for unix-like OS 
  * `%APPDATA%\kn` for MS Win
* Respect env variable `XDG_CONFIG_HOME`
* Check for pre-existing configuration dir on deprecated location and fallback to it. Otherwise use XDG recommended default.

### DIFF
--- a/pkg/kn/core/root.go
+++ b/pkg/kn/core/root.go
@@ -140,8 +140,9 @@ func NewKnCommand(params ...commands.KnParams) *cobra.Command {
 	}
 
 	// Persistent flags
-	rootCmd.PersistentFlags().StringVar(&commands.CfgFile, "config", "", "kn config file (default is $HOME/.kn/config.yaml)")
-	rootCmd.PersistentFlags().StringVar(&p.KubeCfgPath, "kubeconfig", "", "kubectl config file (default is $HOME/.kube/config)")
+	rootCmd.PersistentFlags().StringVar(&commands.CfgFile, "config", "", "kn config file (default is "+
+		filepath.Join(commands.Cfg.DefaultConfigDir, "config.yaml")+")")
+	rootCmd.PersistentFlags().StringVar(&p.KubeCfgPath, "kubeconfig", "", "kubectl config file (default is ~/.kube/config)")
 	flags.AddBothBoolFlags(rootCmd.PersistentFlags(), &p.LogHTTP, "log-http", "", false, "log http traffic")
 
 	plugin.AddPluginFlags(rootCmd)
--- a/pkg/kn/core/root.go
+++ b/pkg/kn/core/root.go
@@ -19,7 +19,8 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path"
+	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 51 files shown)


pkg/kn/core/root.go
```diff
@@ -225,6 +224,42 @@ func initConfig() {
...
+	// Respect %APPDATA% on MS Windows
+	// C:\Documents and Settings\username\Application Data
+	if runtime.GOOS == "windows" {
+		return filepath.Join(os.Getenv("APPDATA"), "kn"), nil
+	}
+	// Respect XDG_CONFIG_HOME if set
+	if xdgHome := os.Getenv("XDG_CONFIG_HOME"); xdgHome != "" {
+		return filepath.Join(xdgHome, "kn"), nil
+	}
+	// Fallback to XDF default for both Linux and macOS
```
maschmid: s/XDF/XDG ?

dsimansk: Typo fixed, thanks!

dsimansk:
> I am not convinced that `XDG_CONFIG_HOME` is the right thing to do. It deviates from `kubectl` and the rest of CNCF community AFAIK.

As mentioned in the issue, one example is Helm, that moved to [XDG spec](https://github.com/helm/helm/pull/5443/files).

> 
> If `kuebctl` and others are also planning to change to this new fashion then I happy to stand corrected but would love to see the evidence (link to issue or PR).

I haven't found any reference for `kubectl`. Honestly, I'd say there'd be overwhelming push against it due to usage of cli itselft and handful of other tools that rely on current config location.

> 
> Otherwise, I think it's ill-advised now to make this change. Especially since we have not reached version 1.0 and this will only confuse people.

Well, I'd say that discussion and consensus should happen prior to reaching 1.0. It would be significantly more difficult to change location and provided meaningful backward-compatibility/migration after 1.0. Currently config is limited to pointing to plugin's dir + default plugin's dir and lookup strategy. 
However, there's either WIP or feature-requests on third party plugin contribs, plugins provided by server etc. I'd say before 1.0 `kn` will produce a significantly more data on user's filesystem. 
There's nothing wrong with current location, other than contributing to home polution that XDG is trying to address.

dsimansk:
To be more precise, there's [a request](https://github.com/kubernetes/kubernetes/issues/56402) for `kubectl` support, but without too much attention from SIG over 2 years and counting.

rhuss:
I also think that if we want to settle on that convention, we should do it now as we only 2 configuration values that are plugin related but we don't have any plugins contribution so I'm not sure if these options are really used.

Also, I think we should follow the XDG convention as it doesn't hurt us and for some people this is really important. I'm pretty sure, `kubectl` would settle on this, too, if there wasn't a huge backwards compatibility issue (that we don't have).

I suggest to bring this discussion forward to the WG call and doing a poll here.

rhuss:
Quick Poll: Should we switch to `~/.config/kn/` as default config location to play nicely with the XDG spec and help to keep the home directory clean ? (similar like Helm 3 does)

rhuss:
**Yes, it's the right time to do it**

rhuss:
**No, lets keep in on the `~/.kn/`**

rhuss:
Ok let's move in the XDG direction.

PR #669, opened 2020-02-17
add "kn service export" (#653)

### LINKED ISSUE #653: Add "kn service export"
## Describe the feature:

A `kn service export` should export a specific Knative service in yaml/json so that it can be easily recreated. It's different to `kn service describe -o yaml` as it strips all runtime specific elements like the status and metadata like the namespace. The idea is that the exported description can be easily committed to a CM system and re-created with `kn service import` (or `kubectl apply -f` if direct Kubernetes access is possible).

An extension could be to allow also an export of all attached revisions (like with `--history`), but that would imly also to export owner references and not sure whether revisions can be created manually.

itsmurugappan:
Fixes #653

## Proposed Changes

* Add new command kn export
* Add unit test


itsmurugappan:
@rhuss I checked out in the serving slack and have made the changes accordingly. I have added 2 flows 
1. ```kn service export service-name -o yaml``` - exports latest service
2. ```kn service export service-name --history -o yaml``` - for getting the service list with current active revisions.

Please review.

pkg/kn/commands/service/export.go
```diff
@@ -0,0 +1,240 @@
...
+	machineReadablePrintFlags := genericclioptions.NewPrintFlags("")
+
+	command := &cobra.Command{
+		Use:   "export NAME",
+		Short: "export a service",
+		Example: `
+  # Export a service in yaml format
+  kn service export foo -n bar -o yaml
+  # Export a service in json format
+  kn service export foo -n bar -o yaml`,
```
dsimansk:
```suggestion
  kn service export foo -n bar -o json`,
```

pkg/kn/commands/service/export.go
```diff
@@ -0,0 +1,240 @@
...
+	"sort"
+	"strconv"
+
+	"github.com/spf13/cobra"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+
+	"knative.dev/client/pkg/kn/commands"
+	clientservingv1alpha "knative.dev/client/pkg/serving/v1"
```
dsimansk: @rhuss has gone through renaming imports exercise a while ago, it should be the following:
```suggestion
	clientservingv1 "knative.dev/client/pkg/serving/v1"
```

pkg/kn/commands/service/export.go
itsmurugappan: Changed. Thank you!

pkg/kn/commands/service/export.go
itsmurugappan: Changed. Thank you!

rhuss:
thanks a lot ! I will give it a review tomorrow ...

rhuss:
The test is failing because of

```
    🦆 kn service update hello --env TARGET=kn --port 8888 --namespace kne2etests0
    ┃ Updating Service 'hello' in namespace 'kne2etests0':
    ┃ 
    ┃   0.017s Ready to serve.
    ┃ 
    ┃ Service 'hello' with latest revision 'hello-rsmrk-1' (unchanged) is available at URL:
    ┃ http://hello.kne2etests0.example.com
    ┃ 
```

not detecting a new revision which is created in the background (and so doesn't wait for that new revision to become ready)

itsmurugappan:
> The test is failing because of
> 
> ```
>   ┣━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
>     🦆 kn service update hello --env TARGET=kn --port 8888 --namespace kne2etests0
>     ┃ Updating Service 'hello' in namespace 'kne2etests0':
>     ┃ 
>     ┃   0.017s Ready to serve.
>     ┃ 
>     ┃ Service 'hello' with latest revision 'hello-rsmrk-1' (unchanged) is available at URL:
>     ┃ http://hello.kne2etests0.example.com
>     ┃ 
> ```
> 
> not detecting a new revision which is created in the background (and so doesn't wait for that new revision to become ready)

Thanks. I reran and it was successfull.

rhuss:
> Thanks. I reran and it was successfull.

yes, its a race. We already have an idea how to fix this.

Sorry, I'm still late with the review. I really hope to make it before the weekend because I will be on PTO next week. But I will hand this over if I won't make it.

pkg/kn/commands/service/export.go
```diff
@@ -0,0 +1,233 @@
...
+
+	command := &cobra.Command{
+		Use:   "export NAME",
+		Short: "export a service",
+		Example: `
+  # Export a service in yaml format
+  kn service export foo -n bar -o yaml
+  # Export a service in json format
+  kn service export foo -n bar -o json`,
+		RunE: func(cmd *cobra.Command, args []string) error {
```
maximilien: Can we split this long (100+ lines) `RunE` function into smaller more manageable private functions with clear names? That would make this much easier to read, review, debug, and test. Thx.

pkg/kn/commands/service/export.go
```diff
@@ -0,0 +1,233 @@
...
+			if err != nil {
+				return err
+			}
+
+			history, err := cmd.Flags().GetBool("history")
+			if err != nil {
+				return err
+			}
+
+			// Print out machine readable output if requested
```
maximilien: From here to below could be easy refactor to a `emitService(...)` or similar.

pkg/kn/commands/service/export.go
```diff
@@ -0,0 +1,233 @@
...
+			if len(args) != 1 {
+				return errors.New("'kn service export' requires name of the service as single argument")
+			}
+			if !machineReadablePrintFlags.OutputFlagSpecified() {
+				return errors.New("'kn service export' requires output format")
+			}
+			serviceName := args[0]
+
+			namespace, err := p.GetNamespace(cmd)
+			if err != nil {
```
maximilien: Does it make sense to log error or some output to help end user?

pkg/kn/commands/service/export.go
```diff
@@ -0,0 +1,233 @@
...
+	var params []clientservingv1.ListConfig
+	params = append(params, clientservingv1.WithService(latestSvc.ObjectMeta.Name))
+
+	// Query for list with filters
+	revisionList, err := client.ListRevisions(params...)
+	if err != nil {
+		return nil, err
+	}
+
+	sortRevisions(revisionList)
```
maximilien: Wondering if an option on sort order would be useful? Perhaps with current as the default?

pkg/kn/commands/service/export_test.go
```diff
@@ -0,0 +1,322 @@
...
+	servinglib "knative.dev/client/pkg/serving"
+	knclient "knative.dev/client/pkg/serving/v1"
+	"knative.dev/client/pkg/util/mock"
+	"knative.dev/pkg/ptr"
+	apiserving "knative.dev/serving/pkg/apis/serving"
+	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
+)
+
+func TestServiceExport(t *testing.T) {
+
```
maximilien: No new line needed

pkg/kn/commands/service/export_test.go
```diff
@@ -0,0 +1,322 @@
...
+	json.Unmarshal([]byte(output), &actSvc)
+
+	assert.DeepEqual(t, expectedService, &actSvc)
+
+	// Validate that all recorded API methods have been called
+	r.Validate()
+}
+
+func TestServiceExportwithMultipleRevisions(t *testing.T) {
+
```
maximilien: Remove new line

pkg/kn/commands/service/export_test.go
```diff
@@ -0,0 +1,322 @@
...
+	client := knclient.NewMockKnServiceClient(t)
+
+	// Recording:
+	r := client.Recorder()
+
+	r.GetService(expectedService.ObjectMeta.Name, expectedService, nil)
+
+	r.ListRevisions(mock.Any(), revs, nil)
+
+	output, err := executeServiceCommand(client, "export", expectedService.ObjectMeta.Name, "-r", "-o", "json")
```
maximilien: you should change this to generate YAML and verify. The one above does JSON as well. Seems redundant to test that output format twice.

maximilien:
Also (obviously) tests are failing. Please address as well. Thx

pkg/kn/commands/service/export.go
itsmurugappan: Thanks for your comments @maximilien . Have tried my best to keep the ```RunE``` func short (50 lines :-)) . Since the flows (export service with and without history) return differnt objects I had to call that from RunE func.

pkg/kn/commands/service/export.go
itsmurugappan: Sort is done to orders the revisions in the order it was created.

pkg/kn/commands/service/export.go
itsmurugappan: Returning the error should print it on the users screen.

pkg/kn/commands/service/export_test.go
itsmurugappan: have added another test for yaml

rhuss:
@itsmurugappan I saved some time to work on a review this week. Please stay tuned ;-)

pkg/kn/commands/service/export.go
```diff
@@ -0,0 +1,230 @@
...
+
+// NewServiceExportCommand returns a new command for exporting a service.
+func NewServiceExportCommand(p *commands.KnParams) *cobra.Command {
+
+	// For machine readable output
+	machineReadablePrintFlags := genericclioptions.NewPrintFlags("")
+
+	command := &cobra.Command{
+		Use:   "export NAME",
+		Short: "export a service",
```
dsimansk:
```suggestion
		Short: "Export a service.",
```
Could you please change the short description to be consistent with the rest of commands. It's used in generated `.md` as a sentence.

itsmurugappan:
> I have a objections to current format of tests, the common approach for the rest of cmds is to split unit and mock tests. However, it isn't impacting functionality added and can be addressed later.

@dsimansk Thanks for reviewing. I will make the changes. Regarding unit tests , all the unit tests were supposed to move to mock based on this issue https://github.com/knative/client/issues/358 .
Please let me know if my understanding is wrong.

dsimansk:
@itsmurugappan sorry for the confusion. My perspective was a bit different based on parts that I worked on recently (the old way) and I didn't know about #358 the new way. So your approach is actually right. :)

itsmurugappan:
> LGTM pending e2e test. Just to make sure that this gets executed on real services. Thx.

will add the e2e

itsmurugappan:
> Sounds good ?

Sure.

pkg/kn/commands/service/export.go
```diff
@@ -0,0 +1,230 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   latestSvc.ObjectMeta.Name,
+			Labels: latestSvc.ObjectMeta.Labels,
+		},
+		TypeMeta: latestSvc.TypeMeta,
+	}
+
+	exportedSvc.Spec.Template = servingv1.RevisionTemplateSpec{
+		Spec:       revision.Spec,
+		ObjectMeta: latestSvc.Spec.ConfigurationSpec.Template.ObjectMeta,
```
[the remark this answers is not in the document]
itsmurugappan: From revision they were lot other values.

From ksvc is see only the below

```
    metadata:
      creationTimestamp: null
      name: hello-v3
```

PR #629, opened 2020-02-01
fixes(#606): adds --cluster-local / --no-cluster-local flags

maximilien:
When specified on 'service create' the `--cluster-local` flag will
make the created service 'private' by setting its config visibility
to `cluster-local`. This is done with label:

`serving.knative.dev/visibility: cluster-local`

The `--no-cluster-local` will remove the label.

Fixes #606 

## Proposed Changes

* add `--cluster-local` and `--no-cluster-local`
* when `--cluster-local` is specified then create service and add label: `serving.knative.dev/visibility: cluster-local`
* add unit test for creating service with --cluster-local

PR #672, opened 2020-02-17
Prevent service-specific labels from being put on the revision template

duglin:
See: https://github.com/knative/serving/pull/6865

/lint

### DIFF
--- a/pkg/serving/config_changes.go
+++ b/pkg/serving/config_changes.go
@@ -362,6 +362,12 @@ func UpdateResources(template *servingv1.RevisionTemplateSpec, requestsResourceL
 	return nil
 }
 
+// ServiceOnlyLabels should only appear on the Service and NOT on the
+// Revision template
+var ServiceOnlyLabels = map[string]struct{}{
+	serving.GroupName + "/visibility": {},
+}
+
 // UpdateLabels updates the labels identically on a service and template.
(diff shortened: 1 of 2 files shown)


rhuss:
Thanks for the PR and this fix is a quick fix for this issue, but does not solve the underlying issue that we already discussed while implementing labels: How can a user specify to which object to apply the label ? The service or the template ? For now its both, but there are situation (like this) where we definitely want a more fine grained resolution.

So I'm +1 for this if we fix the other issue (like using `--label` for both , `--label-service` for only the service and `--label-pod` for the pod template. same for annotations.), too and then remove that workaround (because then the use can specify this on her own).

For this very specific label we will have direct support, too, so I don't think people will provided that as label anyway.

How does this sound ?

duglin:
Do we know when https://github.com/knative/client/pull/629 will be merged? It seems to be taking a while and this is kind of a blocker for us. I'm ok with removing this one #629 is merged.

rhuss:
@navid, that's a question to you --^ :)

duglin:
Let me ask this... if we did the split label feature you mentioned... would we still want code to skip "known" labels that don't belong on Revisions if the user used `--label` ?  Might be good, dunno

duglin:
thanks @rhuss

PR #674, opened 2020-02-17
docs: Update kn examples with real image

### LINKED ISSUE #673: Example help section should use a real image name
## Describe the feature:

Currently we are using fake names in the help sections of e.g. `kn service create --help`

```
Examples:

  # Create a service 'mysvc' using image at dev.local/ns/image:latest
  kn service create mysvc --image dev.local/ns/image:latest

  # Create a service with multiple environment variables
  kn service create mysvc --env KEY1=VALUE1 --env KEY2=VALUE2 --image dev.local/ns/image:latest

  # Create or replace a service 's1' with image dev.local/ns/image:v2 using --force flag
  # if service 's1' doesn't exist, it's just a normal create operation
  kn service create --force s1 --image dev.local/ns/image:v2

  # Create or replace environment variables of service 's1' using --force flag

  # Create service 'mysvc' with port 80
  kn service create mysvc --port 80 --image dev.local/ns/image:latest

  # Create or replace default resources of a service 's1' using --force flag
  # (earlier configured resource requests and limits will be replaced with default)
  # (earlier configured environment variables will be cleared too if any)
  kn service create --force s1 --image dev.local/ns/image:v1

  # Create a service with annotation
  kn service create s1 --image dev.local/ns/image:v3 --annotation sidecar.istio.io/inject=false
```

We should replace `dev.local/ns/image` with a *real* image. I suggest gcr.io/knative-samples/helloworld-go but would prefer one with a shorter name (like one from docker.io so we can use it without registry prefix).

This should be done for all commands that uses image names.

navidshaikh:
Fixes #673 

## Proposed Changes

* Now uses `gcr.io/knative-samples/helloworld-go:latest`

### DIFF
--- a/docs/cmd/kn_service_create.md
+++ b/docs/cmd/kn_service_create.md
@@ -14,29 +14,29 @@ kn service create NAME --image IMAGE [flags]
 
 ```
 
-  # Create a service 'mysvc' using image at dev.local/ns/image:latest
-  kn service create mysvc --image dev.local/ns/image:latest
+  # Create a service 'mysvc' using image at gcr.io/knative-samples/helloworld-go:latest
+  kn service create mysvc --image gcr.io/knative-samples/helloworld-go:latest
 
   # Create a service with multiple environment variables
-  kn service create mysvc --env KEY1=VALUE1 --env KEY2=VALUE2 --image dev.local/ns/image:latest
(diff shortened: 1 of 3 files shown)


navidshaikh:
/close

following up discussion on issue/slack.

PR #679, opened 2020-02-19
Add --user flag for service create and update

### LINKED ISSUE #678: Add a flag to set RunAsUser
## Describe the feature:

RunAsUser is a must for clusters which has pod security policies to check for non root users. Currently kn cli doesn't have an option to set the user running the container. 
It will be a handy feature for clusters which have psp , afaik most organizations have it.

```
kn service create --image=foo --run-as-user 1001
```

itsmurugappan:
Fixes #678 

## Proposed Changes

* Add run as user flag
* add tests for create and update


pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -189,6 +190,8 @@ func (p *ConfigurationEditFlags) addSharedFlags(command *cobra.Command) {
 		"",
 		"Image pull secret to set. An empty argument (\"\") clears the pull secret. The referenced secret must exist in the service's namespace.")
 	p.markFlagMakesRevision("pull-secret")
+	command.Flags().Int64VarP(&p.RunAsUser, "run-as-user", "", 0, "The user to run the container (e.g., 1001).")
```
daisy-ycguo: better to clarify it's the UID, for example, `The user ID to run the container`

itsmurugappan: Thank you. Have made the change.

navidshaikh:
How about just `--user` instead of `--run-as-user` ?

itsmurugappan:
> How about just `--user` instead of `--run-as-user` ?

gave the same name used by k8 container spec.

itsmurugappan:
> How about just `--user` instead of `--run-as-user` ?

@daisy-ycguo @rhuss your thoughts on this ?

navidshaikh:
> gave the same name used by k8 container spec. Can change to user if its more appropriate

There are other parameters as well in `SecurityContext` which we might add in future (like group, privileged, etc). IMO `--user` flag conveys the message here that one wants to run this container with given user. 
Are you also planning to add related property for `--group`? (may be in a next PR)

itsmurugappan:
> > gave the same name used by k8 container spec. Can change to user if its more appropriate
> 
> There are other parameters as well in `SecurityContext` which we might add in future (like group, privileged, etc). IMO `--user` flag conveys the message here that one wants to run this container with given user.
> Are you also planning to add related property for `--group`? (may be in a next PR)

Makes sense. I will change accordingly.

itsmurugappan:
> > gave the same name used by k8 container spec. Can change to user if its more appropriate
> 
> There are other parameters as well in `SecurityContext` which we might add in future (like group, privileged, etc). IMO `--user` flag conveys the message here that one wants to run this container with given user.
> Are you also planning to add related property for `--group`? (may be in a next PR)

Makes sense. I will change accordingly.

PR #682, opened 2020-02-19
feat(wait): add wait for delete operation

### LINKED ISSUE #671: Add `--wait` and `--no-wait` for all delete operation
## Describe the feature:

In order to avoid race conditions when on delete an object and immediately recreate it, by default we should also wait for the end of a delete operation, like we do for `service create` and `service update`.

Therefor a `--wait` and a `--no-wait` option should be added to all "delete" operation. This can reuse our general wait handling util libs.

dsimansk:
Fixes #671 

## Proposed Changes

* Add `waitForEvent` logic to wait package
* Refactor `client.DeleteService()`
  * `timeout == 0` asyn delete without wait
  * otherwise wait a given time for delete op to complete
* Change `service delete` to wait by default
* Add `--no-wait` flag to `delete` op

Generated doc for `--no-wait` flag seems a bit off, as it was meant for ready use-case primarly. I'll take a closer after actual waiting logic in reviewed.


docs/cmd/kn_service_delete.md
```diff
@@ -24,8 +24,11 @@ kn service delete NAME [flags]
 ### Options
 
 ```
+      --async              DEPRECATED: please use --no-wait instead. Delete service and don't wait for it to become ready.
```
rhuss: don't add that, not deeded to add a deprecated option.

dsimansk: It's still part of `AddConditionWaitFlags()` and I wish I could create overloaded method...:) But of course, I agress it shouldn't be created with deprecated flag.

rhuss: yep, I've seen. its ok.

pkg/wait/wait_for_ready.go
```diff
@@ -178,6 +196,30 @@ func (w *waitForReadyConfig) waitForReadyCondition(start time.Time, name string,
...
+	if err != nil {
+		return err, 0
+	}
+	defer watcher.Stop()
+	start := time.Now()
+	// channel used to transport the error
+	errChan := make(chan error)
+	for {
+		select {
+		case <-time.After(timeout):
```
rhuss: Not sure if this is correct (i know, it the same as before, but still ;-), as it looks like that the timer starts afresh for every event that comes in. Shouldn't it be the timeout for the overall operation ? (so we would need to run the timer in a separate gofunc and communicate via channel, a bit like in the delaye case we have in waitForReady)

pkg/wait/wait_for_ready.go
```diff
@@ -178,6 +196,30 @@ func (w *waitForReadyConfig) waitForReadyCondition(start time.Time, name string,
...
+			if w.eventDone(&event) {
+				return nil, time.Since(start)
+			}
+		}
+	}
+}
+
 // Going over Unstructured to keep that function generally applicable.
 // Alternative implemenentation: Add a func-field to waitForReadyConfig which has to be
 // provided for every resource (like the conditions extractor)
```
rhuss: you can remove that comments

pkg/serving/v1/client.go
```diff
@@ -69,12 +69,14 @@ type KnServingClient interface {
...
 	// Delete a service by name
-	DeleteService(name string) error
+	DeleteService(name string, timeout time.Duration) error
 
 	// Wait for a service to become ready, but not longer than provided timeout.
 	// Return error and how long has been waited
 	WaitForService(name string, timeout time.Duration, msgCallback wait.MessageCallback) (error, time.Duration)
 
+	// WaitForEvent(kind, name string, timeout time.Duration, done wait.EventDone) error
+
```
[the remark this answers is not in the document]
dsimansk: No, it's a leftover from my refactors. I'll remove it.

pkg/wait/wait_for_ready.go
dsimansk: I think, you're right. I had a feeling the timeout isn't working actually. Anyway I should add more test cases to unit tests.

dsimansk:
@rhuss pls check if the following change is acceptable https://github.com/knative/client/pull/682/commits/b97e2ca8ffadd478bf1f194cce2379f137370476 as part of this PR. :)

rhuss:
@dsimansk starring at this latest integration error at https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_client/682/pull-knative-client-integration-tests-latest-release/1230890397631254528 I think it's not only good enough to wait for the service to be deleted, but also on all revisions attached to this service. As you can see, in the log, the service was deleted sync (per this PR), but the *next* create still references a revision from this old service in its traffic split. This indicates to me, that although the service has been removed, that it's not the case for its owned revision (therefore somehow the subsequent `service create` could still select this old revision  which is just about to delete (and that doesn't how up anymore in the dumps which happens some time later).

rhuss:
Do you think its hard to implement to wait for the service _and all object which have this service as object reference_ ?

rhuss:
btw, yesterday I stumbled upon: https://github.com/k14s/kapp/blob/master/docs/apply-waiting.md which is more of a generic "wait" on resources. Not sure about the implementation, but we could have a look, too.

navidshaikh:
Nice observation on the logs! :+1: 

A service is created with BYO revision name and deleted.
Another service with same "name" in the same namespace is created with BYO revision name, why client should generate the same revision name (as generated in earlier service create) ?

dsimansk:
@navidshaikh I've reverted sync revision delete commit https://github.com/knative/client/pull/682/commits/9bb60ac25e2f2a30a637a233d851c19d73729628 due to several side-effects that popped up in E2E tests.
I'd say this PR is good for a review and I'll open a follow-up with sync revision delete, once it's a bit more stable.

PR #683, opened 2020-02-19
Allow configuring output for --log-http

### LINKED ISSUE #390: --log-http shouldn't write to stderr
When using `--log-http` it would be nice if it didn't write the log to stderr, but appended to a file (either well-defined name, or value specified by the user). Having that output appear mixed in with the normal output makes it harder to parse/extract. Often I might turn on this option when running a set of scripts so I can decide later if I want to debug it by looking at the http flows. Having all of this output mixed in with the normal output makes it much harder to process.

akerekes:
Change log-http from boolean to string, if no argument is given write log to stderr, otherwise treat argument as file to write to.

/lint

Fixes #390 

## Proposed Changes

* Change behavior of --log-http: if set without argument, write the http log to stderr, if an argument is set, treat it as a file path to where the logs are written

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -21,6 +21,10 @@
 | | Description | PR
 
 | 🧽
+| Allow configure destination for --log-http
+| https://github.com/knative/client/pull/683[#683]
+
+| 🧽
 | Support multiple revisions on `kn revision delete`
 | https://github.com/knative/client/pull/657[#657]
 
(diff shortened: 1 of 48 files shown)


rhuss:
@akerekes Any input to the comments/questions above ? 

A lot has changed in the meantime with a lot of conflicts, so I would suggest to close this PR and start afresh on the current main branch.

If you don't have any objections until next Tuesday, I would close this PR for now (but feel free to reopen it any time).

rhuss:
closing for now, feel free to reopen in case.

PR #703, opened 2020-03-05
Add option for adding labels only to service and/or template

### LINKED ISSUE #675: Add option for adding labels only to service and/or template
## Describe the feature:

At the moment, when using `kn service create --label` then the labe is put into two places: The service's metadata but also on the revisiontemplate. This is find in many cases, but sometimes one needs more fine grained control to where to put labels.

I suggest to use the following options:

* `--label` for setting a label on both, service & template
* `--label-service` for only setting on the service
* `--label-pod` for setting a label only on the revision spec (so could be also `--label-revision`)

I suggest to use `--label-service` over `--service-label` so that the options appear together when sorting alphabetically (like in `kn service create --help`). 

Same should be implemented for `update`

Also for annotations we have the same issue.

An alternative approach would be to use a prefix to the label key to indicate the location of the label

* `--label mylabel=value` to put on both spots
* `--label svc:mylabel=value` to put only on services (`:` is a char not allowed in a label key)
* `--label pod:mylabel=value` to put only on the temlate

This solution would have the advantage to not add to the ever growing option-space. As I expect that 90% of the cases use the pure `--label` form to set on both entitites, I tend to prefer this variant over the one with separate options.

The separator char (':') in this case needs to be checked, whether it can be also used for annotations (so would need to be an illegal char for an annotation key, too). Both, labels and annotations should use the same separator character.

This issue would also make the fix in #672 obsolete so part of this feature would be to cleanup #672, too.

shashwathi:
## Description

Added command line  flags `label-service` and `label-revision` to add labels more granularly for service and revision template objects.

Fixes #675 

/lint

cc @andrew-su 
/assign @rhuss


rhuss:
@shashwathi thanks a lot for the PR ! The test is failing because the test cluster is already on the latest Knative eventing with some breaking changes. We need the latest update to be merged in to get this PR in.

Please stay tuned.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -362,14 +376,33 @@ func (p *ConfigurationEditFlags) Apply(
...
+		revisionMap := make(util.StringMap)
+
+		labelsAllMap, err := util.MapFromArrayAllowingSingles(p.Labels, "=")
 		if err != nil {
 			return errors.Wrap(err, "Invalid --label")
 		}
+		serviceMap.Merge(labelsAllMap)
+		revisionMap.Merge(labelsAllMap)
+
+		labelServiceMap, err := util.MapFromArrayAllowingSingles(p.LabelService, "=")
```
rhuss: we could put the duplicate code into a single method called twice. This time we can _guarantee_ that the specific labels are treated the same.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -47,6 +47,8 @@ type ConfigurationEditFlags struct {
 	AutoscaleWindow            string
 	Port                       int32
 	Labels                     []string
+	LabelService               []string
```
rhuss: would calls it "LabelsService" (plural, like for `Labels`)

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -47,6 +47,8 @@ type ConfigurationEditFlags struct {
...
+	LabelRevision              []string
```
rhuss: dito

pkg/serving/config_changes.go
```diff
@@ -374,31 +374,34 @@ func UpdateResources(template *servingv1.RevisionTemplateSpec, requestsResourceL
 	return nil
 }
 
-// ServiceOnlyLabels should only appear on the Service and NOT on the
-// Revision template
-var ServiceOnlyLabels = map[string]struct{}{
-	serving.GroupName + "/visibility": {},
-}
+// // ServiceOnlyLabels should only appear on the Service and NOT on the
```
rhuss: Please remove it not needed anymore.

pkg/serving/config_changes.go
```diff
@@ -374,31 +374,34 @@ func UpdateResources(template *servingv1.RevisionTemplateSpec, requestsResourceL
...
+// // Revision template
+// var ServiceOnlyLabels = map[string]struct{}{
+// 	serving.GroupName + "/visibility": {},
+// }
 
 // UpdateLabels updates the labels identically on a service and template.
 // Does not overwrite the entire Labels field, only makes the requested updates
-func UpdateLabels(service *servingv1.Service, template *servingv1.RevisionTemplateSpec, toUpdate map[string]string, toRemove []string) error {
+func UpdateLabels(service *servingv1.Service, template *servingv1.RevisionTemplateSpec,
```
rhuss: Having functions with more than 4 or 5 arguments is a code smell, especially when the have the same types. This is not good API design and confuses users and code readers. Instead, its better to either split up a 'god' function (i.e. in a simpler method which might then be called twice) or introduce an _option struct_ (which allows for "named" parameters (field of the structure) in favor of positional arguments). If possible the formere solution is easier to implement and more lightweight. The latter (option struct) should be used only if it can't be avoid. 

This seems kind of a nit-pick comment for now, but maintaining god methods that do too much is really hard. Also using smaller method can help in guaranteeing consistency.

rhuss: E.g. one could also use two methods `UpdateServiceLabels` or `UpdateRevisionTemplateLabels` (or maybe also `UpdatePodLabels` as this is where the lable will end up and might be easier to understand).

I think this is the better option.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -160,10 +162,22 @@ func (p *ConfigurationEditFlags) addSharedFlags(command *cobra.Command) {
...
 			"any number of times to set multiple labels. "+
 			"To unset, specify the label name followed by a \"-\" (e.g., name-).")
 	p.markFlagMakesRevision("label")
+	command.Flags().StringArrayVarP(&p.LabelService, "label-service", "", []string{},
+		"Service label to set. name=value; you may provide this flag "+
+			"any number of times to set multiple labels. "+
+			"To unset, specify the label name followed by a \"-\" (e.g., name-). This flag takes "+
+			"precedence over \"label\" flag.")
+	p.markFlagMakesRevision("label-service")
+	command.Flags().StringArrayVarP(&p.LabelRevision, "label-revision", "", []string{},
```
rhuss: Shouldn't it be really `--label-pod` instead of `--label-revision` as it is about adding the label to the pod spec ?

andrew-su:
@googlebot I consent.

shashwathi:
@rhuss : Yes we will follow up with another Pr for unit and integration test cases. Thanks for the feedback on the PR. Its ready for another review 👍 
cc @andrew-su

rhuss:
thanks looks good ! Let's merge this know and follow up with more tests after the release.

I added issue https://github.com/knative/client/issues/718  to not forget.

/lgtm

PR #704, opened 2020-03-06
Setup `./hack/update-deps.sh` so that we can auto-update CLI.

mattmoor:
## Desciption

This script is the hook that the bots key off of to float our dependencies forward via a PR each day that lets us determine whether things have broken via Prow.  In our other repos, this generally lets us know about breaking changes across repos within 1 business day, and has proven valuable.

## Changes

Add a script to automatically float select dependencies forward via `./hack/update-deps.sh --upgrade`.

/lint

### DIFF
--- a/hack/update-deps.sh
+++ b/hack/update-deps.sh
@@ -1,0 +1,57 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The Knative Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


rhuss:
@mattmoor thanks ! Could you adapt the PR to only include the script ? I'm  asking because in #705 I'm doing the update already manually (because a lot of things has changed wrt/ eventing) and that would definitley conflict with this PR.

mattmoor:
Done!

rhuss:
/retest

no worries the nightly test is expected to fail because of eventing breaking changes (removing legacyclient). Fix is underway.

hack/update-deps.sh
```diff
@@ -0,0 +1,57 @@
...
+VERSION="master"
+
+# The list of dependencies that we track at HEAD and periodically
+# float forward in this repository.
+FLOATING_DEPS=(
+  "knative.dev/test-infra"
+  "knative.dev/pkg@${VERSION}"
+  "knative.dev/serving@${VERSION}"
+  "knative.dev/eventing@${VERSION}"
+  "knative.dev/eventing-contrib@${VERSION}"
```
rhuss: we don't yet have a dependency on eventing-contrib so I would remove it here for now.

PR #710, opened 2020-03-09
Format markdown

mattmoor:
Produced via:
  `prettier --write --prose-wrap=always $(find -name '*.md' | grep -v vendor | grep -v .github)`
/assign maximilien rhuss
/cc maximilien rhuss


navidshaikh:
fix for links https://github.com/knative/client/pull/714

mattmoor:
So the only rich-text diffs github reports appear benign (trailing whitespace?)

mattmoor:
Another notable thing: once this bulk cleanup is done, I can enable the automated code review that leaves these as suggested edits on code reviews (along with all the other automated bot codereviews that client seems to be missing out on 🤔  ).

navidshaikh:
all the changes in this PR to actual mardown files will be reverted next time we run `./hack/build.sh` as the cobra's auto doc generator formats the markdown differently. :-/

mattmoor:
Perhaps we could have that script run `prettier` as well?

mattmoor:
Also fwiw @navidshaikh, in other repos we have a script `verify-codegen.sh` that tries to catch this sort of problem (modification of generated files)

navidshaikh:
Yes, though it would add `prettier` as then dev dependency, as the docs generated part of `./hack/build.sh`. Meanwhile, we can avoid the `docs/cmd` dir for the command used for this PR?
```
prettier --write  $(find -name '*.md' | grep -v vendor | grep -v .github | grep -v docs/cmd/)
```

rhuss:
> Also fwiw @navidshaikh, in other repos we have a script `verify-codegen.sh` that tries to catch this sort of problem (modification of generated files)

We do this, too, but it verifies that you run via build.sh. Not sure why it haven't caught this issue.

navidshaikh:
For docs only PR (no code change) like this PR; verify code gen step didnt run. Verified it by https://github.com/knative/client/pull/717 and filed https://github.com/knative/test-infra/issues/1799.

PR #715, opened 2020-03-09
fix(docs): Revert markdown changes for cobra's auto-docs

navidshaikh:
Revert sub-set of changes from https://github.com/knative/client/pull/710 which
 are auto-generated by cobra for cmd docs. (run `./hack/build.sh` on current HEAD)

/cc @mattmoor if we can update the format markdown command for client meanwhile
```
prettier --write --prose-wrap=always $(find -name '*.md' | grep -v vendor | grep -v .github | grep -v docs/cmd/)
```

### DIFF
--- a/docs/cmd/kn.md
+++ b/docs/cmd/kn.md
@@ -6,8 +6,8 @@ Knative client
 
 Manage your Knative building blocks:
 
-- Serving: Manage your services and release new software to them.
-- Eventing: Manage event subscriptions and channels. Connect up event sources.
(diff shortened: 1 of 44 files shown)


mattmoor:
/lgtm

As I said, we should probably add a presubmit check like verify-codegen to block changes to these not through the generator.

We should probably also mark these as generated in `.gitattributes`.

mattmoor:
I added ` | grep -v docs/cmd/` for now.

PR #717, opened 2020-03-09
DNM: prettier change

navidshaikh:
CI should catch this as blocking change
/hold


navidshaikh:
@rhuss : CI didnt catch that. ;-)
Last time I remember similar issue with return status of the command in test-infra script. Will check further tomorrow..

PR #726, opened 2020-03-10
feat(trigger): add --inject-broker flag to trigger create command

dsimansk:
## Description

Allow user to add default broker when creating trigger.

## Changes

* Add flag `--inject-broker` to trigger create commnad

### DIFF
--- a/pkg/eventing/v1alpha1/client.go
+++ b/pkg/eventing/v1alpha1/client.go
@@ -161,8 +161,12 @@ func (b *TriggerBuilder) Subscriber(subscriber *duckv1.Destination) *TriggerBuil
 }
 
 // Broker to set the broker of trigger object
-func (b *TriggerBuilder) Broker(broker string) *TriggerBuilder {
+func (b *TriggerBuilder) Broker(broker string, inject bool) *TriggerBuilder {
 	b.trigger.Spec.Broker = broker
+	if inject {
+		meta_v1.SetMetaDataAnnotation(&b.trigger.ObjectMeta, "knative-eventing-injection", "enabled")
+	}
+
 	return b
 }
 
--- a/pkg/eventing/v1alpha1/client.go
+++ b/pkg/eventing/v1alpha1/client.go
@@ -161,11 +161,15 @@ func (b *TriggerBuilder) Subscriber(subscriber *duckv1.Destination) *TriggerBuil
 }
 
 // Broker to set the broker of trigger object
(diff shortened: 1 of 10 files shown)


pkg/eventing/v1alpha1/client.go
```diff
@@ -161,8 +161,12 @@ func (b *TriggerBuilder) Subscriber(subscriber *duckv1.Destination) *TriggerBuil
 }
 
 // Broker to set the broker of trigger object
-func (b *TriggerBuilder) Broker(broker string) *TriggerBuilder {
+func (b *TriggerBuilder) Broker(broker string, inject bool) *TriggerBuilder {
 	b.trigger.Spec.Broker = broker
+	if inject {
+		meta_v1.SetMetaDataAnnotation(&b.trigger.ObjectMeta, "knative-eventing-injection", "enabled")
```
rhuss: There is supposed to be a constant somewhere in eventing holding this value.

pkg/eventing/v1alpha1/client.go
```diff
@@ -161,8 +161,12 @@ func (b *TriggerBuilder) Subscriber(subscriber *duckv1.Destination) *TriggerBuil
 }
 
 // Broker to set the broker of trigger object
-func (b *TriggerBuilder) Broker(broker string) *TriggerBuilder {
+func (b *TriggerBuilder) Broker(broker string, inject bool) *TriggerBuilder {
```
rhuss: I would add a separate builder method like `"InjectBroker()"` instead of mangling it with the broker. Also, when injection is enabled the broker has to be set to "default" anyway, so there are only limited combinations of this argument possible.

pkg/eventing/v1alpha1/client.go
```diff
@@ -163,6 +163,14 @@ func (b *TriggerBuilder) Subscriber(subscriber *duckv1.Destination) *TriggerBuil
 // Broker to set the broker of trigger object
 func (b *TriggerBuilder) Broker(broker string) *TriggerBuilder {
 	b.trigger.Spec.Broker = broker
+
+	return b
+}
+
+// InjectBroker to add annotation to setup default broker
+func (b *TriggerBuilder) InjectBroker() *TriggerBuilder {
```
rhuss: final request: Could we make it taking a boolean arg and enable it when its true, otherwise delete it ? This is useful if you hand over builders or init them from an existing trigger.

dsimansk: Sure, sounds useful.

PR #735, opened 2020-03-10
missing paren

duglin:
Just typo


dsimansk:
@duglin pls run `./hack/build.sh` to regenerate docs. It should fix the failing build.

PR #766, opened 2020-03-30
Add e2e test for different service and revision label

### LINKED ISSUE #718: Add e2e tests and command line tests for "--labels" options
See https://github.com/knative/client/pull/703#pullrequestreview-371018971 for more details.

/assign shashwathi

shashwathi:
## Description

Add e2e test for different service and revision label

## Reference

Fixes #718 

/assign @rhuss

### DIFF
--- a/test/e2e/service_options_test.go
+++ b/test/e2e/service_options_test.go
@@ -111,6 +111,10 @@ func TestServiceOptions(t *testing.T) {
 	validateUserID(r, "svc6", uid)
 	serviceUpdate(r, "svc6", "--user", strconv.FormatInt(uid+1, 10))
 	validateUserID(r, "svc6", uid+1)
+
+	t.Log("create and validate service and revision labels")
+	serviceCreateWithOptions(r, "svc7", "--label-service", "svc=helloworld-svc", "--label-revision", "rev=helloworld-rev")
+	validateLabels(r, "svc7", map[string]string{"svc": "helloworld-svc"}, map[string]string{"rev": "helloworld-rev"})
 }
 
 func serviceCreateWithOptions(r *test.KnRunResultCollector, serviceName string, options ...string) {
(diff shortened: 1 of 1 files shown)


test/e2e/service_options_test.go
```diff
@@ -167,6 +172,22 @@ func (test *e2eTest) validateServiceAnnotations(t *testing.T, r *KnRunResultColl
 	}
 }
 
+func (test *e2eTest) validateLabels(t *testing.T, r *KnRunResultCollector, serviceName string, serviceLabels, revisionLabels map[string]string) {
+	metadataLabelsJsonpathFormat := "jsonpath={.metadata.labels.%s}"
+	templateLabelsJsonpathFormat := "jsonpath={.spec.template.metadata.labels.%s}"
+
+	for k, v := range serviceLabels {
+		out := test.kn.Run("service", "describe", serviceName, "-o", fmt.Sprintf(metadataLabelsJsonpathFormat, k))
```
maximilien: Could you not run the command once and verify output with all labels after?

shashwathi: Thanks for the feedback. I have updated the PR to do the same.

test/e2e/service_options_test.go
```diff
@@ -167,6 +171,23 @@ func (test *e2eTest) validateServiceAnnotations(t *testing.T, r *KnRunResultColl
...
+	assert.NilError(t, err)
+
+	gotRevisionLabels := service.Spec.Template.ObjectMeta.GetLabels()
+	for k, v := range revisionLabels {
+		assert.Equal(t, gotRevisionLabels[k], v)
+	}
+	gotServiceLabels := service.ObjectMeta.GetLabels()
+	for k, v := range serviceLabels {
+		assert.Equal(t, gotServiceLabels[k], v)
+	}
```
rhuss, line 188: For a full check, you would also need to check that every label in `gotServiceLabels` has been verified (i.e. that the labels returned by the service are really all stored in `gotServiceLabels`).

Imagine that `GetLabels` returns an empty map (so no label has been created, which would be a bug). But your validation would still succeed.

Actually you should _remove_ the entry in `gotServiceLabels` after each _assert_ and then check after the loop that `gotServiceLabels` is indeed empty.

rhuss: Same btw for the loop above, too.

shashwathi: > Imagine that GetLabels returns an empty map (so no label has been created, which would be a bug). But your validation would still succeed.

If GetLabels returns an empty map and `serviceLabels` has any label to be validated then validation logic will fail.

rhuss: Sorry, you are right with `GetLabel` returning an empty map would lead to an error. But would happen if `GetLabel` returns _more_ labels than expected ? Is this supposed to be an error is this something that is also expected ? 

Of course this is a question for what to check. If you have full control over the service, I would expect that there are no extra labels being added to revisions and annotations (so would still suggest to check that all of the labels you get with `GetLabel` are really expected). In other words, I would check that both maps are equal not that the expected map is subset of the actual map.

shashwathi: > But would happen if GetLabel returns more labels than expected ?

This is not suppose to be an error IMO. Function responsibility is to verify expectedLabels are present in received service labels, not to check both are an exact match. 

>  If you have full control over the service, I would expect that there are no extra labels being added to revisions and annotations

Thanks for explaining the pretext here. I see your argument now about checking for exact match. 👍

shashwathi: I have updated the PR. Please review it at your convenience

PR #770, opened 2020-04-01
Remove the delete propagation flag

duglin:
In it's current state it now takes me about 25 seconds for the `kn delete`
to complete. Before https://github.com/knative/client/pull/682 it used to be
almost immediate. This is because we now pass in the
`DeletePropagationBackground` flag. I believe this is a mistake, not only
because of the 20+ seconds of additional time to delete things, but IMO
the CLI should talk to the server in the same way regardless of the --wait
flag. That flag should just be a CLI thing to indicate if the user wants the CLI
to wait for the server to complete but not HOW the server should do the delete.

### DIFF
--- a/pkg/kn/commands/wait_flags.go
+++ b/pkg/kn/commands/wait_flags.go
@@ -42,8 +42,15 @@ type WaitFlags struct {
 func (p *WaitFlags) AddConditionWaitFlags(command *cobra.Command, waitTimeoutDefault int, action, what, until string) {
 	waitUsage := fmt.Sprintf("%s %s and don't wait for it to be %s.", action, what, until)
 	//TODO: deprecated flag should be removed in next release
+
+	// Special-case 'delete' command so it comes back to the user immediately
+	noWaitDefault := false
+	if action == "Delete" {
+		noWaitDefault = true
+	}
+
 	command.Flags().BoolVar(&p.Async, "async", false, "DEPRECATED: please use --no-wait instead. "+waitUsage)
-	command.Flags().BoolVar(&p.NoWait, "no-wait", false, waitUsage)
+	command.Flags().BoolVar(&p.NoWait, "no-wait", noWaitDefault, waitUsage)
 
 	timeoutUsage := fmt.Sprintf("Seconds to wait before giving up on waiting for %s to be %s.", what, until)
 	command.Flags().IntVar(&p.TimeoutInSeconds, "wait-timeout", waitTimeoutDefault, timeoutUsage)
--- a/docs/cmd/kn_revision_delete.md
+++ b/docs/cmd/kn_revision_delete.md
@@ -24,7 +24,7 @@ kn revision delete NAME [flags]
       --async              DEPRECATED: please use --no-wait instead. Delete revision and don't wait for it to be deleted.
   -h, --help               help for delete
   -n, --namespace string   Specify the namespace to operate in.
-      --no-wait            Delete revision and don't wait for it to be deleted.
+      --no-wait            Delete revision and don't wait for it to be deleted. (default true)
       --wait-timeout int   Seconds to wait before giving up on waiting for revision to be deleted. (default 600)
 ```
 
--- a/docs/cmd/kn_service_delete.md
+++ b/docs/cmd/kn_service_delete.md
@@ -27,7 +27,7 @@ kn service delete NAME [flags]
       --async              DEPRECATED: please use --no-wait instead. Delete service and don't wait for it to be deleted.
   -h, --help               help for delete
   -n, --namespace string   Specify the namespace to operate in.
-      --no-wait            Delete service and don't wait for it to be deleted.
+      --no-wait            Delete service and don't wait for it to be deleted. (default true)
       --wait-timeout int   Seconds to wait before giving up on waiting for service to be deleted. (default 600)
 ```
 
--- a/pkg/kn/commands/wait_flags_test.go
+++ b/pkg/kn/commands/wait_flags_test.go
@@ -115,3 +115,12 @@ func TestAddWaitUsageMessage(t *testing.T) {
 		t.Error("wrong until message")
 	}
 }
+
+func TestAddWaitUsageDelete(t *testing.T) {
+	flags := &WaitFlags{}
+	cmd := cobra.Command{}
+	flags.AddConditionWaitFlags(&cmd, 60, "Delete", "blub", "deleted")
+	if !strings.Contains(cmd.UsageString(), "deleted. (default true)") {
+		t.Error("Delete has wrong default value for --no-wait")
+	}
+}
--- a/pkg/serving/v1/client.go
+++ b/pkg/serving/v1/client.go
@@ -262,7 +262,7 @@ func updateServiceWithRetry(cl KnServingClient, name string, updateFunc serviceU
 // For `timeout == 0` delete is performed async without any wait.
 func (cl *knServingClient) DeleteService(serviceName string, timeout time.Duration) error {
 	if timeout == 0 {
-		return cl.deleteService(serviceName, v1.DeletePropagationBackground)
+		return cl.deleteService(serviceName)
 	}
 	waitC := make(chan error)
 	go func() {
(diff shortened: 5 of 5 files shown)


dsimansk:
It was introduced as a part of synchronous delete operation to ensure all dependant resources are deleted for the service and we wait in-sync for the service resource to be gone.
- `DeletePropagationBackground` is added for `--no-wait` and shouldn't cause any blocking. Afaik, it should be cluster default behaviour, but I do agree we could pass empty value to ensure server-side is honoured. 
- `DeletePropagationForeground` is added is current default and configured byt `--wait` flag. It's part of synchronous delete operation, otherwise there should be other mechanism to ensure removal of everything related to service.

From my local test `delete --no-wait` is immediate and `delete --wait` takes some time as expected from synchronous operation.

duglin:
I think the biggest issue is that `--no-wait` is not the default for `kn` so the new 20+second delay I was seeing was very noticeable/unexpected and gave me a bad UX. Perhaps just changing the default would be sufficient

However, I still wonder if `--no-wait` should change what we send to the server. I suspect many folks will think of it the way I did... it's just CLI thingy that only changes how long it takes for my cmd prompt to come back to me - it doesn't change what we tell the server to do - which I believe is true for other `--no-wait` flags on our other commands. If we want the user to be able to influence the DeletePropagation flag then I think we should introduce a flag to specifically allow the user to control that knob w/o overloading the `--no-wait` flag.

But I actually think that's a bit too low level for our CLI. I think the biggest concern for people is "when can I create a new ksvc with the same name w/o getting an error about that name being used?". And the default/DeletePropagationForeground does that. How long it takes to delete other resources that are hidden from the user, and would not influence any subsequent cmd they execute, isn't really that interesting except to a pretty advanced user.

dsimansk:
> I think the biggest issue is that --no-wait is not the default for kn so the new 20+second delay I was seeing was very noticeable/unexpected and gave me a bad UX. Perhaps just changing the default would be sufficient

The motivation to have `--wait` as default is alignment to `create` or `update`. Personally, I don't have a strong preference here.

>However, I still wonder if --no-wait should change what we send to the server. I suspect many folks will think of it the way I did... it's just CLI thingy that only changes how long it takes for my cmd prompt to come back to me - it doesn't change what we tell the server to do - which I believe is true for other --no-wait flags on our other commands. If we want the user to be able to influence the DeletePropagation flag then I think we should introduce a flag to specifically allow the user to control that knob w/o overloading the --no-wait flag.

>But I actually think that's a bit too low level for our CLI. I think the biggest concern for people is "when can I create a new ksvc with the same name w/o getting an error about that name being used?". And the default/DeletePropagationForeground does that. How long it takes to delete other resources that are hidden from the user, and would not influence any subsequent cmd they execute, isn't really that interesting except to a pretty advanced user.

The overall goal to have a synchronous delete was to ensure there aren't unexpected race condition between calling CRUD commands in rapid succession, e.g. scripts, E2E tests etc. The case of "service name collision" is on point here as well. I understand `Service` resource as an umbrella that should ensure removal of everything related to prevent unwanted dangling resources. 

Wrt the delete policy flag, I lean strongly toward "too low level" approach. The policy should be opinionated default of synchronous delete. In this case it's part of delete everything strategy.

duglin:
> The motivation to have `--wait` as default is alignment to create or update. Personally, I don't have a strong preference here.

Totally agree, but in this case I think we only need to wait for the ksvc to vanish, not the hidden resources.

duglin:
Net: 20+ seconds for an out-of-the-box `kn delete` is not good :-)

rhuss:
Kubectl uses cascading delete by default, and we should do that, too. Because if you don't do that, who is going to delete the revisions of a service afterwards ? (afaik, if you don't do a cascading delete those revisions, even when they have an ownerReference will just stay around forever. tbv)

However, I think for `delete` we should really switching to async by default. Because when you delete something you more often don't care anymore about the thing, which is different than when you create something which you might want to use immediate.

So my suggestion is:

* Do cascading delete all the time
* Switch to async delete by default, which can be switched on with `--wait` .We probably will have to adapt the E2E tests, too to add `--wait`

rhuss:
My fear is if we don't do a cascade delete we are left with orphaned objects.

duglin:
to be clear... I don't think cascading or not is the issue, either way we'll delete stuff. The question is how we ask the server to do the delete and whether we wait on the CLI side or not.

Switching to `--no-wait` by default is fine, as long as it will at least wait until the ksvc is gone - and that's always been quick for me and what `kn` used to do. I think the issue here is whether kn's `--wait` flag is just a CLI thing or not because we're mixing two concepts into one with the current code base. 

I'll repeat what I said above:
>However, I still wonder if --no-wait should change what we send to the server. I suspect many folks will think of it the way I did... it's just CLI thingy that only changes how long it takes for my cmd prompt to come back to me - it doesn't change what we tell the server to do - which I believe is true for other --no-wait flags on our other commands. If we want the user to be able to influence the DeletePropagation flag then I think we should introduce a flag to specifically allow the user to control that knob w/o overloading the --no-wait flag.

I'd recommend that we keep `--wait` controlling how long the CLI waits for the server to complete it's task - strictly a CLI-side flag.  We can then discuss a second flag to indicate what that "task" is - meaning is it what it used to do (background delete propagation), or is it a foreground delete.

maximilien:
Also @duglin consider adding tests :)

rhuss:
For me whether we wait for all objects to be deleted (foreground) or to only the main object to be deleted (background) is also a client concern, there is no different background operation. Serverside the eventual result will be the same, whether its foreground or background deletion (with foreground delete policy relying on the [kubernetes garbage collector](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#controlling-how-the-garbage-collector-deletes-dependents (), thanks @dsimansk for the link).

So the end result will be the same: The object and all its dependent objects are deleted (the option here is actually not related whether to cascade or not, I misunderstood that).

Its just the way _how_ the client instructs the deletion, so I'm totally fine that this way differs for `--wait` vs `--no-wait`, to make `--no-wait` as fast as possible and `--wait` as safe as possible (important for automated use cases like our e2e tests):

* Use the option that @dismansk has selected for `--wait` (foreground) and `--no-wait` (background), so keep the original code
* However: Make `--no-wait` the default for the reasons mentioned in https://github.com/knative/client/pull/770#issuecomment-607913796 to keep a good interactive UX.

rhuss:
> Wouldn't it make more sense to have a `--cascade-delete` or similar flag that does not default to true? Seems like removing that feature completely is not a step forward...

I would do *always* a cascade delete as I don't see the use case for non-cascade delete on the abstraction level of `kn`. The question is whether we do the cascade delete explicit (foreground) or rely on the k8s garbage collector (background).

duglin:
I'm not thrilled with the idea that `--wait` has different semantics based on the CLI cmd. If it means "CLI returns immediately but the server action is the same" for everything except delete then I think that inconsistency is a problem and not an ideal UX.

If you believe that the client might care about how the delete happens then linking it with whether the CLI returns immediately or not would be bad, because this means that someone can't choose 2 of the 4 possible combination of semantics.

In the end, I strongly believe that `kn service delete foo` needs to return as quickly as possible and it means that I can reuse that Ksvc name immediately, like it used to. Forcing a flag to get that semantics would be bad IMO. 

So I'd like to suggest that we deal with that first and then have a discussion around whether, and how, to change the semantics of the delete on the server. I still like a new flag to control that because I do believe linking it with `--wait` is mixing topics/concerns.

duglin:
Updated PR to just set the default to be `--no-wait` per our slack chat

pkg/kn/commands/wait_flags.go
```diff
@@ -42,8 +42,15 @@ type WaitFlags struct {
 func (p *WaitFlags) AddConditionWaitFlags(command *cobra.Command, waitTimeoutDefault int, action, what, until string) {
 	waitUsage := fmt.Sprintf("%s %s and don't wait for it to be %s.", action, what, until)
 	//TODO: deprecated flag should be removed in next release
+
+	// Special-case 'delete' command so it comes back to the user immediately
+	noWaitDefault := false
+	if action == "Delete" {
```
dsimansk, line 48: I'd rather see default value provided through function's argument. The current approach will change default for all delete operations, e.g. also Revisions. What about a new config struct that for now can have 2 fields, default value and timeout.

duglin: oh I didn't realize you could be more specific... I'll fix....

rhuss: btw, I would be happy to switch to no-wait as default for all delete operations (this would then be also easy to reason about).

duglin: I do like consistency :-)  We just need to figure out which way it spans.  Making all deletes the same (as long as it doesn't have negative side-effects) seems ok to me.  @dsimansk ?

duglin:
Weird  - the CLA bot was happy yesterday

pkg/kn/commands/wait_flags.go
dsimansk: I wasn't sure about all the usage. but wait flags are used in Service (delete) and Revision (delete), no objections to go with `--no-wait` for all delete then.

duglin: ok - if we do that, then I think the PR is ready for review

duglin:
@rgregg any ideas on the CLA issue?

duglin:
@googlebot rescan

duglin:
OK CLA check is now happy - ready for review

PR #771, opened 2020-04-01
fix(769): various issues with godoc documents

maximilien:
## Description

Fixes various godoc comments that are not conforming as well as some /lint issues that came up

## Changes

* correct signatures for some functions with with non-conforming names
* various corrections to Godoc comments for exported code in new lib/test
* fix CHANGELOG.doc for new location for common test code

## Reference

Fixes #769

/lint

PR #773, opened 2020-04-01
fix(#772) refactored / simplified parameters to KnResultCollector

### LINKED ISSUE #772: refactor: simplify parameters in e2e tests
### Refactor

From PR #765 

```
looking at result of refactoring, every lib util now need to take 3 testing related params, before it actually takes any operation related param, e.g:

func serviceCreate(t *testing.T, it *test.KnTest, r *test.KnRunResultCollector, serviceName string) {
There is repetition,
see:

type KnRunResultCollector struct {
  results    []KnRunResult
  extraDumps []string
  t          *testing.T
}
We can access testing.T from KnRunResultCollector in every test util function, this can be fixed
separately from this PR. @rhuss: WDYT?
```
> /kind good-first-issue
> /kind refactor

maximilien:
## Description

Simplify and consolidated parameters to `KnResultCollector`

## Changes

* pass `testing.T` to `KnResultCollector`
* pass `test.KnTest` to `KnResultCollector`
* change functions calls to only using new `KnResultCollector` parameters

## Reference

Fixes #772

/lint


maximilien:
Please ignore the `golint` Godoc issues as they are solved in #771 which I can rebase into this when its merged.

cc @navidshaikh

test/e2e/version_test.go
```diff
@@ -28,10 +28,13 @@ import (
...
 
-	r := test.NewKnRunResultCollector(t)
+	it, err := test.NewKnTest()
+	assert.NilError(t, err)
+
+	r := test.NewKnRunResultCollector(t, it)
 	defer r.DumpIfFailed()
 
-	out := test.Kn{}.Run("version")
+	out := it.Kn().Run("version")
```
navidshaikh, line 37: from [logs](https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/knative_client/773/pull-knative-client-integration-tests/1245495159261499396#1:build-log.txt%3A694):
```
--- FAIL: TestVersion (1.20s)
    result_collector.go:75: ERROR: unknown flag: --namespace
    result_collector.go:101: === FAIL: =======================[[ERROR]]========================
        🦆 kn version --namespace kne2etests6
        ┃ 
        🔥 unknown flag: --namespace
        🔥 
        =================================================================
```

Here `test.NewKnTest()` would create a new namespace for this test and `Kn().Run()` would check if namespace is non empty and add `--namespace` flag in the command.

We can replace this test as
```
func TestVersion(t *testing.T) {
  t.Parallel()
  r := test.NewKnRunResultCollector(t, nil)
  defer r.DumpIfFailed()
  out := test.RunKn("", []string{"version"})
  r.AssertNoError(out)
  assert.Check(t, util.ContainsAll(out.Stdout, "Version"))
}
```
run
```
➜  client git:(1de3841e) ✗ go test ./test/e2e/ -count=1 -test.v --tags e2e -run TestVersion
=== RUN   TestVersion
=== PAUSE TestVersion
=== CONT  TestVersion
--- PASS: TestVersion (0.04s)
PASS
ok  	knative.dev/client/test/e2e	0.052s
```

maximilien: Slight change to be cleaner but generally should address

maximilien:
Ok to merge @navidshaikh

test/e2e/version_test.go
```diff
@@ -28,10 +28,13 @@ import (
 func TestVersion(t *testing.T) {
 	t.Parallel()
 
-	r := test.NewKnRunResultCollector(t)
+	it, err := test.NewKnTest()
```
navidshaikh, line 31: `NewKnTest()` will create a namespace for this test.
1. We don't need namespace creation for version test.
2. We dont have `tearDown` in this test, which leaving behind the created namespace.
Ran this locally from PR:
```
➜  client git:(c8c54c52) go test ./test/e2e/ -count=1 -test.v --tags e2e -run TestVersion                                                                        
=== RUN   TestVersion
=== PAUSE TestVersion
=== CONT  TestVersion
--- PASS: TestVersion (5.54s)
PASS
ok  	knative.dev/client/test/e2e	5.544s

➜  client git:(c8c54c52) kubectl get ns|grep kne2etests
kne2etests0                                             Active   56s
```
If `kn` doesn't require a namespace to execute, lets not create one? Call to `NewKnTest()` will create one.

I'd still suggest the earlier review comment fix for version test, no need to define a new method for only version test case.

```
func TestVersion(t *testing.T) {
  t.Parallel()
  r := test.NewKnRunResultCollector(t, nil)
  defer r.DumpIfFailed()
  out := test.RunKn("", []string{"version"})
  r.AssertNoError(out)
  assert.Check(t, util.ContainsAll(out.Stdout, "Version"))
}
```

navidshaikh: If we go without this, we'll have a namespace left behind from e2e test run which will break running e2e locally against same cluster in next run.

maximilien: I think having a way to run w/o namespace is important since remember that this code is now library that can be used outside. I added code to tear down the namespace when the test is done:

```go
+	defer func() {
+		assert.NilError(t, it.Teardown())
+	}()
```

maximilien: Note also that the new code:

```golang
// RunNoNamespace the 'kn' CLI with args but no namespace
func (k Kn) RunNoNamespace(args ...string) KnRunResult {
	return RunKn("", args)
}
```

Is pretty much just a wrapper to make clear we are running `kn` on default namespace which is a useful thing in general. Keeps the `VersionTest` consistent with the other tests. We still have various `e2e` tests that need to be converged to this model so I think it's best to keep the simple `VersionTest` in line with the others since people will use that as an exemplar.

navidshaikh: Well, as `test.RunKn` is an exported method, this could actually be an example for how to run `kn` without creating a namespace in tests. Say a plugin does not require a namespace to perform its task, its e2e can refer version tests.

PR #778, opened 2020-04-03
fix(issue #762): correct error message when updating service

### LINKED ISSUE #762: Misleading error message with multiple names
### Bug report

```
$ kn service update echo echo1 --image duglin/echo
requires the service name
```

### Expected behavior

I would have expected either something like:
```
$ kn service create echo echo1 --image duglin/echo
'service create' requires the service name given as single argument
```
or for it to do the update on both services (I kind of like this option).

See above

### kn version

master/head as of 2020/03/26

maximilien:
## Description

Correct error message when calling `service update` with more than one arg

## Changes

* fix issue with TestServiceUpdateWithMultipleImages running create vs update
* correct message when updating service and passing many names
* added TestServiceDescribeWithMultipleNames and fix to describe command
* added TestServiceCreateWithMultipleNames
* fix error message for service delete since many names can be passed

## Reference

Fixes #762

/lint


pkg/kn/commands/service/delete.go
```diff
@@ -40,7 +40,7 @@ func NewServiceDeleteCommand(p *commands.KnParams) *cobra.Command {
 
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if len(args) < 1 {
-				return errors.New("requires the service name")
+				return errors.New("requires the service name(s)")
```
navidshaikh, line 43:
```suggestion
				return errors.New("'service delete' requires the service name(s)")
```

CHANGELOG.adoc
```diff
@@ -26,6 +26,10 @@
 | Fix plugin lookup with file ext on Windows
 | https://github.com/knative/client/pull/774[#774]
 
+| 🐛
+| correct error message when updating service
```
navidshaikh, line 30:
```suggestion
| Correct error message when updating service
```

maximilien:
@navidshaikh done! Thanks.

PR #782, opened 2020-04-06
Add -a flag as alias for --annotation

duglin:
I find that I'm using annotations more and more and since `-a`
wasn't being used I decided we should save people from typing too much

## Changes

* add support for `-a` as an alias for `--annotation`

/lint

### DIFF
--- a/docs/cmd/kn_service_create.md
+++ b/docs/cmd/kn_service_create.md
@@ -45,7 +45,7 @@ kn service create NAME --image IMAGE [flags]
 ### Options
 
 ```
-      --annotation stringArray        Service annotation to set. name=value; you may provide this flag any number of times to set multiple annotations. To unset, specify the annotation name followed by a "-" (e.g., name-).
+  -a, --annotation stringArray        Service annotation to set. name=value; you may provide this flag any number of times to set multiple annotations. To unset, specify the annotation name followed by a "-" (e.g., name-).
       --arg stringArray               Add argument to the container command. Example: --arg myArg1 --arg --myArg2 --arg myArg3=3. You can use this flag multiple times.
       --async                         DEPRECATED: please use --no-wait instead. Create service and don't wait for it to be ready.
       --autoscale-window string       Duration to look back for making auto-scaling decisions. The service is scaled to zero if no request was received in during that time. (eg: 10s)
--- a/test/e2e/service_options_test.go
+++ b/test/e2e/service_options_test.go
@@ -87,6 +87,13 @@ func TestServiceOptions(t *testing.T) {
 	validateServiceAnnotations(r, "svc3", map[string]string{"alpha": "direwolf", "brave": ""})
 	serviceDelete(r, "svc3")
 
+	t.Log("create, update and validate service with annotations but -a")
+	serviceCreateWithOptions(r, "svc3a", "-a", "alpha=wolf", "-a", "brave=horse")
+	validateServiceAnnotations(r, "svc3a", map[string]string{"alpha": "wolf", "brave": "horse"})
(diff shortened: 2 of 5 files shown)


navidshaikh:
We also have other flags starting with `a`, `--async` (deprecated but still there), `--all-namespaces` with `-A` alias,  `--args` and `--autoscale-window`. IMO, if there are multiple flags starting with same letter, it becomes confusing to map aliases (`--args` flag is closest ).
And, this depends on how popular the flag is to make a short hand alias for (as done for `--label` flag). 

@duglin : should auto-completion help here?

duglin:
Yep - it's a bit of a guessing game, and popularity contest.

I don't think it's confusing though since people learn quickly what works. `-l` seems to be ok and we have lots of `--l...` long named flags.

But, as of now, of all the `--a...` flags, I do think `--annotation` would be the most popular, so I'm just trying to make life a bit easier.

docs/cmd/kn_service_create.md
```diff
@@ -45,7 +45,7 @@ kn service create NAME --image IMAGE [flags]
 ### Options
 
 ```
-      --annotation stringArray       Service annotation to set. name=value; you may provide this flag any number of times to set multiple annotations. To unset, specify the annotation name followed by a "-" (e.g., name-).
+  -a, --annotation stringArray       Service annotation to set. name=value; you may provide this flag any number of times to set multiple annotations. To unset, specify the annotation name followed by a "-" (e.g., name-).
```
rhuss, line 48: Im fine with that (except maybe that it can be easily confused with `-A` which is "for all namespaces")

rhuss: on the other hand, why not `-a` for `--arg` ? 

But as we already have `-e` and `-l`, for consistencies sake I'm fine with `-a` (but in general we should take care to give away single letter options not easily)

duglin: I think --annotation would be more than --arg, plus "arg" is shorter than "annotation" - so between the two I don't think "arg" needs as much UX help :-)

As for -A and -a .... not sure what to say, other than these things are case sensitive for a reason.

rhuss:
After no objections from the latest WG call, lets merge this PR.

Thanks, @duglin !

/lgtm
/approve

test/e2e/service_options_test.go
```diff
@@ -85,6 +85,13 @@ func TestServiceOptions(t *testing.T) {
 	validateServiceAnnotations(t, it, r, "svc3", map[string]string{"alpha": "direwolf", "brave": ""})
 	serviceDelete(t, it, r, "svc3")
 
+	t.Log("create, update and validate service with annotations but -a")
+	serviceCreateWithOptions(t, it, r, "svc3", "-a", "alpha=wolf", "-a", "brave=horse")
+	validateServiceAnnotations(t, it, r, "svc3", map[string]string{"alpha": "wolf", "brave": "horse"})
+	serviceUpdate(t, it, r, "svc3", "-a", "alpha=direwolf", "-a", "brave-")
+	validateServiceAnnotations(t, it, r, "svc3", map[string]string{"alpha": "direwolf", "brave": ""})
+	serviceDelete(t, it, r, "svc3")
+
```
rhuss, line 94: You probably have to adapt this part to fit the refactoring landed recently (that's why the build tests are failing)

duglin:
@rhuss fixed testing issues

duglin:
Added an entry to the change log too - I think it's all ready now

rhuss:
/retest

Still looks like a flake

duglin:
/hold

I'm doing some testing to see why the tests are failing

Issue #789, opened 2020-04-08
`service describe` show repetitive revisions if service has multi revisions


MIBc:
### Bug report

If a service has more than three revisons. The 'kn service describe -v' will show repetitive revisions.
Here is a screenshot：
![image](https://user-images.githubusercontent.com/4477478/78783192-c9de3780-79d5-11ea-8a3c-c78a358a2e7f.png)

I have read the [code](https://github.com/knative/client/blob/master/pkg/kn/commands/service/describe.go#L324).
The reason is that the revison would be covered by next one.
We should use deep copy.

### Expected behavior

Show every revision in knative service.

### Steps to reproduce the problem

Creating more than three revisions for a knative service.
### kn version
v0.13

/kind bug

dsimansk:
Fixed in PR #790.

/close

PR #790, opened 2020-04-08
Show all revisions when run `service describe -v`

MIBc:
Fixes: #https://github.com/knative/client/issues/789
* The `kn service describe -v` command shows repetitive revisions, because
  the revision would be covered by next one.

### DIFF
--- a/pkg/kn/commands/service/describe.go
+++ b/pkg/kn/commands/service/describe.go
@@ -267,7 +267,7 @@ func getRevisionDescriptions(client clientservingv1.KnServingClient, service *se
 			return nil, fmt.Errorf("cannot extract revision from service %s: %v", service.Name, err)
 		}
 		revisionsSeen.Insert(revision.Name)
-		desc, err := newRevisionDesc(revision, &target, service)
+		desc, err := newRevisionDesc(*revision, &target, service)
 		if err != nil {
 			return nil, err
 		}
@@ -302,7 +302,7 @@ func completeWithLatestRevisions(client clientservingv1.KnServingClient, service
 		if err != nil {
 			return nil, err
 		}
-		newDesc, err := newRevisionDesc(rev, nil, service)
+		newDesc, err := newRevisionDesc(*rev, nil, service)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


pkg/kn/commands/service/describe.go
```diff
@@ -321,7 +321,7 @@ func completeWithUntargetedRevisions(client clientservingv1.KnServingClient, ser
 			continue
 		}
 		revisionsSeen.Insert(revision.Name)
-		newDesc, err := newRevisionDesc(&revision, nil, service)
+		newDesc, err := newRevisionDesc(revision.DeepCopy(), nil, service)
```
rhuss, line 324: Couldn't we just past the revision by calue, not by reference ?  I think this would be even cleaner. wdyt ?

pkg/kn/commands/service/describe_test.go
```diff
@@ -432,28 +432,31 @@ func TestServiceDescribeVerbose(t *testing.T) {
 	r := client.Recorder()
 
 	// Prepare service
-	expectedService := createTestService("foo", []string{"rev1", "rev2"}, goodConditions())
+	expectedService := createTestService("foo", []string{"rev1", "rev2", "rev3"}, goodConditions())
```
maximilien, line 435: Do we have a test for the base case with one revision? I think there is but worth double checking. Thx

pkg/kn/commands/service/describe.go
MIBc: That will modify the parameter type in `newRevisionDesc` function. It brings some changes which need to test.

pkg/kn/commands/service/describe_test.go
MIBc: We have the base case with one revision.

MIBc:
> I wonder however if we could just change the signature to pass by-value not by-reference, so that a deep copy would not be necessary. Tbh, I don't yet understand why (and if) the called method modifies the revision

I found this [code](https://github.com/knative/client/blob/master/pkg/kn/commands/service/describe.go#L319)will cover the revision. But not create a new `revision` variable.

rhuss:
> I found this [code](https://github.com/knative/client/blob/master/pkg/kn/commands/service/describe.go#L319) would cover the revision. But not create a new `revision` variable.

I think I understand (excellent find by the way!), but if we don't move the pointer but the value to the function for extracting the revision description, this should be similar to a copy, or ? (not a deep copy, but I think a shallow copy should be good enough, too).

So changing that function signature from taking a point to taking a value should be good enough, i think.

MIBc:
The reason we don't use deep copy is that it cost more time. Right?

rhuss:
> The reason we don't use deep copy is that it cost more time. Right?

I'm not really worried about costs, it's just that I think its cleaner and simpler, but I'm not dogmatic.  If you think a deep copy has other advantages or is more idiomatic, I'm happy, too ;)

PR #794, opened 2020-04-10
e2e: Foreground delete for revisions and services in e2e

navidshaikh:
## Description
Since the default behavior is background deletion, add `--no-wait=false` to service/revision delete operations in e2e to avoid any race conditions and flakes

## Changes
- Add `--no-wait=false` for service/revision delete operations in e2e.

/lint

### DIFF
--- a/test/e2e/revision_test.go
+++ b/test/e2e/revision_test.go
@@ -98,7 +98,7 @@ func revisionListWithService(r *test.KnRunResultCollector, serviceNames ...strin
 }
 
 func revisionDelete(r *test.KnRunResultCollector, revName string) {
-	out := r.KnTest().Kn().Run("revision", "delete", "--no-wait=false", revName)
+	out := r.KnTest().Kn().Run("revision", "delete", "--wait", revName)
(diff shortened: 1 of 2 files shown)


test/e2e/revision_test.go
```diff
@@ -91,7 +91,7 @@ func revisionListWithService(r *test.KnRunResultCollector, serviceNames ...strin
 }
 
 func revisionDelete(r *test.KnRunResultCollector, revName string) {
-	out := r.KnTest().Kn().Run("revision", "delete", revName)
+	out := r.KnTest().Kn().Run("revision", "delete", "--no-wait=false", revName)
```
rhuss, line 94: Wasn't there a `--wait` option, too ?

rhuss: At least I thouhgt that was what we have decided to use for boolean flags (`--bla` and `--no-bla` ...)

navidshaikh: No. 
The flag isnt using `flags.AddBothBoolFlagsUnhidden` which auto defines the other flag.
I think we'd need to fix this to use ^^ util for this bool flag definition.

navidshaikh: Using `--wait` flag now.

PR #802, opened 2020-04-14
Make wait, no-wait and async flags per bool var CLI convention

### LINKED ISSUE #800: wait, no-wait and async flag should follow bool var CLI conventions
We're using cobra util to define `no-wait` bool flag. It gets confusing to pass true/false to flags starting with `no`. We've `AddBothBoolFlagsUnhidden ` utility to take care of bool flags according to convention.
This is to track updating the flag definition to use mentioned utility.

See https://github.com/knative/client/pull/794#discussion_r407985406

navidshaikh:
Fixes #800

## Description
 - Deprecated bool vars can be supported for CLI convention
 - Bind --async flag value to --no-wait

```
➜  kn service create foo --image gcr.io/foo/bar --async --no-wait
only one of (DEPRECATED) --async, --wait and --no-wait may be specified

➜  kn service create foo --image gcr.io/foo/bar --async --wait   
only one of (DEPRECATED) --async, --wait and --no-wait may be specified

➜  kn service create foo --image gcr.io/foo/bar --async --no-wait --wait
only one of (DEPRECATED) --async, --wait and --no-wait may be specified

➜  kn service create foo --image gcr.io/foo/bar --async=false           
use --wait instead of providing "false" to --async
```
/lint

### DIFF
--- a/docs/cmd/kn_service_delete.md
+++ b/docs/cmd/kn_service_delete.md
@@ -24,10 +24,11 @@ kn service delete NAME [flags]
 ### Options
 
 ```
-      --async              DEPRECATED: please use --no-wait instead. Delete service and don't wait for it to be deleted.
+      --async              DEPRECATED: please use --no-wait instead. Do not wait for 'service Delete' operation to be completed.
   -h, --help               help for delete
   -n, --namespace string   Specify the namespace to operate in.
-      --no-wait            Delete service and don't wait for it to be deleted. (default true)
+      --no-wait            Do not wait for 'service Delete' operation to be completed.
+      --wait               Wait for 'service Delete' operation to be completed. (default true)
       --wait-timeout int   Seconds to wait before giving up on waiting for service to be deleted. (default 600)
 ```
 
--- a/docs/cmd/kn_service_delete.md
+++ b/docs/cmd/kn_service_delete.md
@@ -24,11 +24,11 @@ kn service delete NAME [flags]
 ### Options
 
 ```
-      --async              DEPRECATED: please use --no-wait instead. Do not wait for 'service Delete' operation to be completed.
+      --async              DEPRECATED: please use --no-wait instead. Do not wait for 'service delete' operation to be completed. (default true)
   -h, --help               help for delete
   -n, --namespace string   Specify the namespace to operate in.
-      --no-wait            Do not wait for 'service Delete' operation to be completed.
-      --wait               Wait for 'service Delete' operation to be completed. (default true)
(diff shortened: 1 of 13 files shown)


pkg/kn/flags/bool.go
```diff
@@ -65,6 +68,25 @@ func ReconcileBoolFlags(f *pflag.FlagSet) error {
 		if err != nil {
 			return
 		}
+
+		// handle DEPRECATD flags
```
rhuss, line 72:
```suggestion
		// handle DEPRECATED flags
```

pkg/kn/flags/bool.go
```diff
@@ -65,6 +68,25 @@ func ReconcileBoolFlags(f *pflag.FlagSet) error {
...
+		if strings.HasPrefix(flag.Usage, deprecatedPrefix) {
```
rhuss, line 73: Not sure if this is needed as you below check for "async" only.  I would wait for a second deprecated option before starting to abstract this.

pkg/kn/flags/bool.go
```diff
@@ -65,6 +68,25 @@ func ReconcileBoolFlags(f *pflag.FlagSet) error {
...
+			if flag.Changed {
+				// handle async flag
+				if flag.Name == "async" {
```
rhuss, line 76: Can be combined with the previous `if` to a single conditional with `&&`

pkg/kn/flags/bool.go
```diff
@@ -65,6 +68,25 @@ func ReconcileBoolFlags(f *pflag.FlagSet) error {
...
+					if f.Lookup("wait").Changed || f.Lookup("no-wait").Changed {
+						err = fmt.Errorf("only one of (DEPRECATED) --%s, --wait and --no-wait may be specified", flag.Name)
```
rhuss, line 78: You already know that the flag name is "async", no need to add it via parameter here.

docs/cmd/kn_service_delete.md
```diff
@@ -24,10 +24,11 @@ kn service delete NAME [flags]
 ### Options
 
 ```
-      --async              DEPRECATED: please use --no-wait instead. Delete service and don't wait for it to be deleted.
+      --async              DEPRECATED: please use --no-wait instead. Do not wait for 'service Delete' operation to be completed.
```
navidshaikh, line 27:
```suggestion
      --async              DEPRECATED: please use --no-wait instead. Do not wait for 'service delete' operation to be completed.
```

PR #819, opened 2020-04-22
refactor kn service export to export revisions

### LINKED ISSUE #728: "kn service export --with-revisions" should create revision COs directly
Currently, `kn service export --with-revisions` exports only services which are update one after each other. However, as it is allowed to create revisions directly it makes more sense to export those directly, with the proper owner-references.

An `kn service import` then could then also adapt the revision names in case of collisions.

itsmurugappan:
## Description
Changes made to export revisions directly with kn service export.
https://github.com/knative/client/issues/728#issuecomment-603928462 - This was the approach followed

## Changes

 
* --with-revisions option will export service and revision list
* --with-revisions --kubernetes-resources will export a service list (kubectl friendly)
* add annotations to export

## Reference

Fixes #728

### DIFF
--- a/docs/cmd/kn_service_export.md
+++ b/docs/cmd/kn_service_export.md
@@ -18,10 +18,10 @@ kn service export NAME [flags]
   kn service export foo -n bar -o yaml
   # Export a service in JSON format
   kn service export foo -n bar -o json
-  # Export a service with revisions in JSON format
-  kn service export foo --with-revisions -n bar -o json
-  # Export a service with revisions in kubectl friendly format
-  kn service export foo --with-revisions --kubernetes-resources -n bar -o json
+  # Export a service with revisions
+  kn service export foo --with-revisions --mode=resources -n bar -o json
+  # Export services in kubectl friendly format, as a list kind, one service item for each revision
+  kn service export foo --with-revisions --mode=kubernetes -n bar -o json
 ```
 
 ### Options
@@ -29,7 +29,7 @@ kn service export NAME [flags]
 ```
       --allow-missing-template-keys   If true, ignore any errors in templates when a field or map key is missing in the template. Only applies to golang and jsonpath output formats. (default true)
   -h, --help                          help for export
-      --kubernetes-resources          Export all routed revisions in kubectl friendly format(experimental)
+      --mode string                   Format for exporting all routed revisions. One of kubernetes|resources (experimental)
   -n, --namespace string              Specify the namespace to operate in.
   -o, --output string                 Output format. One of: json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-file.
       --template string               Template string or path to template file to use when -o=go-template, -o=go-template-file. The template format is golang templates [http://golang.org/pkg/text/template/#pkg-overview].
--- a/docs/cmd/kn_service_export.md
+++ b/docs/cmd/kn_service_export.md
@@ -18,6 +18,10 @@ kn service export NAME [flags]
   kn service export foo -n bar -o yaml
   # Export a service in json format
   kn service export foo -n bar -o json
+  # Export a service with revisions in json format
+  kn service export foo --with-revisions -n bar -o json
+  # Export a service with revisions in kubectl friendly format
+  kn service export foo --with-revisions --kubernetes-resources -n bar -o json
 ```
 
 ### Options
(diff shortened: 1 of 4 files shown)


docs/cmd/kn_service_export.md
```diff
@@ -18,13 +18,18 @@ kn service export NAME [flags]
   kn service export foo -n bar -o yaml
   # Export a service in json format
   kn service export foo -n bar -o json
+  # Export a service with revisions in json format
```
maximilien, line 21: JSON

pkg/kn/commands/service/export_test.go
```diff
@@ -35,23 +33,37 @@ import (
 
 type expectedServiceOption func(*servingv1.Service)
 type expectedRevisionOption func(*servingv1.Revision)
+type expectedServiceListOption func(*servingv1.ServiceList)
+type expectedRevisionListOption func(*servingv1.RevisionList)
+type podSpecOption func(*v1.PodSpec)
+
+func TestServiceExportError(t *testing.T) {
+	// New mock client
```
maximilien, line 41: Not clear the value of this comment since code is clear :)

itsmurugappan:
> Any way to have the expected YAML in the tests? That way it’s clear and just compares output from test to YAML itself?

like a set of expected files in a folder and compare against them ? I think for e2e we can do them.

rhuss:
Thanks a lot ! I was a bit busy the last week, but will work on the review soon

pkg/kn/commands/service/export.go
```diff
@@ -230,3 +302,13 @@ func revisionListSortFunc(revisionList *servingv1.RevisionList) func(i int, j in
 		return a.Name > b.Name
 	}
 }
+
+func stripGeneratedFieldsfromService(svc *servingv1.Service) {
+	delete(svc.ObjectMeta.Annotations, "serving.knative.dev/creator")
```
rhuss, line 307: Maybe we can the annotations that should be removed into an array and put it top-level in this file with a prominent comment what this array is about? This probably makes maintaining this list easier (as I expect it will require some maintenance and/or fine-tuning)

pkg/kn/commands/service/export.go
```diff
@@ -230,3 +302,13 @@ func revisionListSortFunc(revisionList *servingv1.RevisionList) func(i int, j in
 		return a.Name > b.Name
 	}
 }
+
+func stripGeneratedFieldsfromService(svc *servingv1.Service) {
```
rhuss, line 306: Can we rename it to `stripIgnoredAnnotationsFromService` and use an array `IGNORED_SERVICE_ANNOTATIONS` (see comment below) ?

pkg/kn/commands/service/export.go
```diff
@@ -230,3 +302,13 @@ func revisionListSortFunc(revisionList *servingv1.RevisionList) func(i int, j in
...
+	delete(svc.ObjectMeta.Annotations, "serving.knative.dev/creator")
+	delete(svc.ObjectMeta.Annotations, "serving.knative.dev/lastModifier")
+	delete(svc.ObjectMeta.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
+}
+
+func stripGeneratedFieldsfromRevision(revision *servingv1.Revision) {
```
rhuss, line 312: Same renaming suggestion as above

pkg/kn/commands/service/export.go
```diff
@@ -181,20 +224,49 @@ func exportServiceWithActiveRevisions(latestSvc *servingv1.Service, client clien
...
 }
 
-func setTrafficSplit(latestSvc *servingv1.Service, exportedSvc servingv1.Service) servingv1.Service {
+func exportActiveRevisionList(latestSvc *servingv1.Service, client clientservingv1.KnServingClient) (*servingv1.Service, *servingv1.RevisionList, error) {
+	var exportedRevItems []servingv1.Revision
 
-	exportedSvc.Spec.RouteSpec = latestSvc.Spec.RouteSpec
-	return exportedSvc
+	//get revisions to export from traffic
+	revsMap := getRevisionstoExport(latestSvc)
```
rhuss, line 231: Please rename to `getRevisionsToExport` (capitalized `To`)

rhuss: Same for `constructServicefromRevision` to `constructServiceFromRevision`

pkg/kn/commands/service/export.go
```diff
@@ -181,20 +224,49 @@ func exportServiceWithActiveRevisions(latestSvc *servingv1.Service, client clien
...
+
+	// Query for list with filters
+	revisionList, err := client.ListRevisions(clientservingv1.WithService(latestSvc.ObjectMeta.Name))
```
rhuss, line 234: This code (up to `sortRevisions`) is shared with the other method so it could be extracted into an own function.

itsmurugappan:
> wdyt ? We can also do it in two steps: First merge this PR and the think about that `KnExport` (or `KnativeExport` type).

@rhuss  I also think its an excellent idea to introduce a new type. It will be a much cleaner approach.

PR #825, opened 2020-05-05
Skip LatestReadyRevisionName if Revision is Pending or Unknown

### LINKED ISSUE #823: Confusing Error Message from kn service describe When Revision Not Available
### Bug report 

When using `kn service describe` on a Service where the Revision is not available, the following error message is returned:

```
resource name may not be empty
```

It would be helpful to display a more informative error message or display that the Revision is not ready as part of the describe command. 

Debugging this locally, this error comes from [`completeWithLatestRevisions`](https://github.com/knative/client/blob/c41e9fd923299a735865572d79e8770aae9655d9/pkg/kn/commands/service/describe.go#L303) and is returned here:

https://github.com/knative/client/blob/c41e9fd923299a735865572d79e8770aae9655d9/pkg/kn/commands/service/describe.go#L119

image

It appears an empty string is being passed to `GetRevision` causing the error. 

### Expected behavior

`kn service describe` should describe the Service without issue. Also, regardless of the root of this issue, there should be more information about why this error message is occurring. 

Find a way to create a `Service` with a `Revision` stuck in a pending state:

```
$ kn service list
NAME            URL                                                                                     LATEST
  AGE     CONDITIONS   READY     REASON
helloworld-go   http://helloworld-go.deploy-your-first-application-using-knative-w01-s001.example.com
  6m49s   0 OK / 3     Unknown   RevisionMissing : Configuration "helloworld-go" is waiting for a Revision to b
ecome ready.
```

Run `kn service describe` on the service. 

### kn version

```
Version:      v0.14.0
Build Date:   2020-04-22 08:56:32
Git Revision: c41e9fd
Supported APIs:
* Serving
  - serving.knative.dev/v1 (knative-serving v0.14.0)
* Eventing
  - sources.knative.dev/v1alpha2 (knative-eventing v0.14.1)
  - eventing.knative.dev/v1alpha1 (knative-eventing v0.14.1)
```

Serving: v0.14.0

danielhelfand:
## Description

This pull request adds a check for if `LatestReadyRevisionName` is `""` with `kn service describe`, indicating the Revision is in a Pending or Unknown state. This will prevent `kn service describe` from returning an error message of `resource name may not be empty` due to `LatestReadyRevisionName` being an empty string when passed to the `GetRevision` function in ` completeWithLatestRevisions`.

## Changes

Add a check for if `revisionName == ""` and continue to the next loop iteration if true. Also added a unit test for this case and edited helper functions. 

Open to any different interpretation of how to treat this case as far as whether the revisionName should be skipped or other details should be returned.

## Reference

Fixes #823 

/lint

### DIFF
--- a/pkg/kn/commands/service/describe.go
+++ b/pkg/kn/commands/service/describe.go
@@ -294,7 +294,7 @@ func orderByConfigurationGeneration(descs []*revisionDesc) []*revisionDesc {
 
 func completeWithLatestRevisions(client clientservingv1.KnServingClient, service *servingv1.Service, revisionsSeen sets.String, descs []*revisionDesc) ([]*revisionDesc, error) {
 	for _, revisionName := range []string{service.Status.LatestCreatedRevisionName, service.Status.LatestReadyRevisionName} {
-		if revisionsSeen.Has(revisionName) {
+		if revisionName == "" || revisionsSeen.Has(revisionName) {
 			continue
 		}
 		revisionsSeen.Insert(revisionName)
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -12,6 +12,14 @@
 | https://github.com/knative/client/pull/[#]
 ////
 
+[cols="1,10,3", options="header", width="100%"]
+|===
+| | Description | PR
+
+| 🐛
+| Skip LatestReadyRevisionName if Revision is Pending or Unknown
+| https://github.com/knative/client/pull/825[#825]
+
 ## v0.14.0 (2020-04-21)
 
 [cols="1,10,3", options="header", width="100%"]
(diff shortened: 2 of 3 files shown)


pkg/kn/commands/service/describe.go
```diff
@@ -294,7 +294,7 @@ func orderByConfigurationGeneration(descs []*revisionDesc) []*revisionDesc {
 
 func completeWithLatestRevisions(client clientservingv1.KnServingClient, service *servingv1.Service, revisionsSeen sets.String, descs []*revisionDesc) ([]*revisionDesc, error) {
 	for _, revisionName := range []string{service.Status.LatestCreatedRevisionName, service.Status.LatestReadyRevisionName} {
-		if revisionsSeen.Has(revisionName) {
+		if revisionsSeen.Has(revisionName) || revisionName == "" {
```
rhuss, line 297: I would move the check for `revisionName == ""` to the first position as it doesn't make sense to check for an empty name in the has anyway.

danielhelfand: Good catch! Addressed.

rhuss:
Thanks !

/lgtm

PR #830, opened 2020-05-07
DNM: Run serving specific e2e tests using tags

navidshaikh:
/hold

### DIFF
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -50,9 +50,9 @@ run() {
 }
 
 integration_test() {
-  header "Running tests for Knative Serving $KNATIVE_SERVING_VERSION and Eventing $KNATIVE_EVENTING_VERSION"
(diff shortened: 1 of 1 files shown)


navidshaikh:
> I like this and would even split the eventing-e2e into their own so that we could do something like: ./hack/build.sh --eventing-e2e or similar.

@maximilien : We already have this using go tags. You can set `eventing` or `serving` to `E2E_TAGS` and run as:

```
E2E_TAGS=eventing ./test/local-e2e-tests.sh
```
https://github.com/knative/client/pull/832 should fix any unscoped method being referenced in scoped tags.

navidshaikh:
@coryrc : I've opened this PR to re-produce issue mentioned in https://github.com/knative/test-infra/issues/2064

navidshaikh:
/close

Closing the PR, https://github.com/knative/test-infra/issues/2064 can be re-produced as mentioned in this PR.
cc: @chizhg @chaodaiG

PR #833, opened 2020-05-11
[master] Format markdown

mattmoor:
Produced via:
  `prettier --write --prose-wrap=always $(find -name '*.md' | grep -v vendor | grep -v .github | grep -v docs/cmd/)`
/assign maximilien rhuss
/cc maximilien rhuss

### DIFF
--- a/conventions/cli.md
+++ b/conventions/cli.md
@@ -1,11 +1,13 @@
 ## `kn` User Interface Principles
 
-This document describes the conventions that are used for all `kn` commands and options.
-It is normative in the sense that any new feature that introduces or changes user-facing commands or options needs to adhere to the given principles.
+This document describes the conventions that are used for all `kn` commands and
+options. It is normative in the sense that any new feature that introduces or
+changes user-facing commands or options needs to adhere to the given principles.
 
-Also, the given rules apply to plugins as well.
-Especially any plugin that wants to be promoted to the [knative/client-contrib](https://github.com/knative/client-contrib) plugin repository has to adhere to these rules.
-
+Also, the given rules apply to plugins as well. Especially any plugin that wants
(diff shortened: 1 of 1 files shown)


rhuss:
Can we please switch off the automatic markdown formatting ? I understand why this make sense for formatting code, but it does not make sense for authoring documents. Markdown is not a programming language and each other author has different preferences. E.g. I highly prefer to write one sentence per line ("semantic linefeed"). This allows to easily rearrange parts and keeps you focussed on the current sentence. See e.g. https://rhodesmill.org/brandon/2012/one-sentence-per-line/ or https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line for the rational behind this. It makes it also much easier to look at markdown diffs.

The prettifier completely mess this up. @mattmoor isn't there really a way to opt out from the markdown reformatter ?

Or asked otherwise: What is the benefit for formatting markdown (i'm not aware of a generally enforced standard) ?

// cc: @maximilien @navidshaikh

mattmoor:
Hey @rhuss, sorry for the late response, this was sitting under a large pile of github notifications in my inbox.  I've disabled the periodic job that runs `prettier` for client, ultimately these jobs are intended to make folks lives easier and improve repo hygiene, if it's stressing anyone it's having the opposite effect, so my apologies 😉 

Ultimately the benefit I've seen is consistency in documents where the stylistic differences of multiple authors makes it particularly awkward to read as the style shifts across sections or even sentences.

But like I said, it's off now, and apologies again for any stress is may have caused.

rhuss:
@mattmoor 🙏 thanks, and no worries. The bots are really helpful (just not so for markdown imo :-P

Issue #579, opened 2019-12-19
Split eventing/serving commands in help message to increase discoverability


ahmetb:
v0.11 added eventing commands:
```
Available Commands:
  completion  Output shell completion code
  help        Help about any command
  plugin      Plugin command group
  revision    Revision command group
  route       Route command group
  service     Service command group
  source      Event source command group
  trigger     Trigger command group
  version     Prints the client version
```

This might be confusing to users going through `kn --help` to figure out how to use it. 

Simply, `source` says  "Event", but other things don't say "Serving".

So one must know which CRDs are Event vs Serving.

Consider doing what kubectl does:

```
Basic Commands (Beginner):
  create         Create a resource from a file or from stdin.
  expose         Take a replication controller, service, deployment or pod and expose it as a new Kubernetes Service
  run            Run a particular image on the cluster
  set            Set specific features on objects

Basic Commands (Intermediate):
  explain        Documentation of resources
  get            Display one or many resources
  edit           Edit a resource on the server
  delete         Delete resources by filenames, stdin, resources and names, or by resources and label selector

Deploy Commands:
  rollout        Manage the rollout of a resource
  scale          Set a new size for a Deployment, ReplicaSet or Replication Controller
  autoscale      Auto-scale a Deployment, ReplicaSet, or ReplicationController

Cluster Management Commands:
  certificate    Modify certificate resources.
  cluster-info   Display cluster info
  top            Display Resource (CPU/Memory/Storage) usage.
  cordon         Mark node as unschedulable
  uncordon       Mark node as schedulable
  drain          Drain node in preparation for maintenance
  taint          Update the taints on one or more nodes

Troubleshooting and Debugging Commands:
  describe       Show details of a specific resource or group of resources
  logs           Print the logs for a container in a pod
  attach         Attach to a running container
  exec           Execute a command in a container
  port-forward   Forward one or more local ports to a pod
  proxy          Run a proxy to the Kubernetes API server
  cp             Copy files and directories to and from containers.
  auth           Inspect authorization

Advanced Commands:
  diff           Diff live version against would-be applied version
  apply          Apply a configuration to a resource by filename or stdin
  patch          Update field(s) of a resource using strategic merge patch
  replace        Replace a resource by filename or stdin
  wait           Experimental: Wait for a specific condition on one or many resources.
  convert        Convert config files between different API versions
  kustomize      Build a kustomization target from a directory or a remote url.

Settings Commands:
  label          Update the labels on a resource
[...]

PR #834, opened 2020-05-12
fixes(#827): allow plugins to extend all command groups

### LINKED ISSUE #827: Supporting extending all command groups?
### Feature request

PR #818 allows extensions to the `source` command group. This is needed since `kn` has core commands under `source` to support some eventing sources, e.g., `ping`. However, since the number of variety of eventing sources is large and growing, we envisioned additional eventing source support to be added as plugins. See for instance the `source kafka` and `source github` and other plugins being developed in `client-contrib` repository.

So the question then comes up should we generalize support for extending and overriding of core commands. This was raised by @navidshaikh and @rhuss. And proposed by Navid in [comments](https://github.com/knative/client/pull/818#discussion_r419379336) to PR #818.

While both sides of the issues are useful. Before implementing we will use this issue to discuss and eventually vote a direction. So watch for the comment with pole for voting below once we have agreement on the feature request.

### Use case

There are two principle use cases:

1. As a `kn` user and developer, I would like to add a new command to an existing command group via a plugin. For instance, a `debug` command to the core `service` command group.

2. As a `kn` user and developer, I would like to override an existing command with my own implementation via a plugin. So imagine you have a `fancy-create` command for `service` that allows some extra decorations to services and other features when user calls `create`. You would like to have user call `kn service create` vs `kn service fancy-create`. 

Please note that use case 2 extends on use case 1. So use case 2 can only exist if we allow use case 1.

### UI Example

```
# assume a plugin called kn-service-debug in ~/.config/kn/plugins directory
# also assume user has a new value in ~/.config/kn/config.yml set, 
# e.g., allow-core-extensions: true
kn service debug my-service # executes the kn-service-debug plugin

# similarly for use case 2 above, if a plugin named ~/.config/kn/plugins/kn-service-create is 
# present it will be executed for all `service create` calls but other commands in `service` 
# will be left as is 
kn service create my-service ... # executes ~/.config/kn/plugins/kn-service-create
kn service update my-service ... # executes core `update` command
```

> /kind proposal

maximilien:
## Description

Allow `kn` plugins to extend existing command groups. So a plugin named: `kn-service-blah` is now allowed. It extends the `service` command group with command `blah`. These extensions are for any existing command groups. This is per issue #827 

## Changes

* allow plugins to extends all command groups, e.g., kn-service-blah is OK
* prevent plugins that shadow commands of existing groups, e.g., kn-service-create is NOT OK
* list plugins that extend existing groups
* warn plugins that shadow command of existing groups
* add UTs for above

## Reference

Fixes #827

/lint


maximilien:
To reviewers. Some local tests I have done that might be useful to verify this yourself.

```bash
➜  client git:(issue827) ./kn plugin list
/Users/maximilien/.config/kn/plugins/kn-curl
/Users/maximilien/.config/kn/plugins/kn-service-blah
/Users/maximilien/.config/kn/plugins/kn-service-create
/Users/maximilien/.config/kn/plugins/kn-source-create
/Users/maximilien/.config/kn/plugins/kn-source_github
ERROR:
  - kn-service-create overwrites existing built-in command: kn service create
kn-service-create overwrites existing built-in command: kn service create
➜  client git:(issue827) cat /Users/maximilien/.config/kn/plugins/kn-service-blah
#!/bin/bash
echo "Hello kn: $1 $2"
➜  client git:(issue827) ./kn service blah max
Hello kn: max
➜  client git:(issue827) ./kn service create
required flag(s) "image" not set
➜  client git:(issue827) ./kn service list
NAME          URL                                                                                                            LATEST                AGE   CONDITIONS   READY   REASON
hello-world   http://hello-world-default.kn-plugins-a4511d9cf0901749cf183d1384ce2bca-0000.sjc04.containers.appdomain.cloud   hello-world-kzyhx-1   55d   3 OK / 3     True
```

maximilien:
Ok looks like I need to add more tests to `root.go` so that the coverage gets above 50%.

Will work on that tomorrow.

/hold

maximilien:
Looks like I am still not over 50% coverage for `root.go` :(

pkg/kn/commands/plugin/plugin.go
```diff
@@ -53,16 +53,13 @@ func BindPluginsFlagToViper(cmd *cobra.Command) {
 	viper.SetDefault("lookup-plugins", false)
 }
 
-// AllowedExtensibleCommandGroups the list of command groups that can be
-// extended with plugins, e.g., a plugin named `kn-source-kafka` for Kafka
-// event sources is allowed. This is defined as a fixed [...]string since
-// cannot defined Golang []string constants
-var AllowedExtensibleCommandGroups = [...]string{"source"}
+// CoreCommandNames names of all core `kn` commands
+var CoreCommandNames = []string{}
```
rhuss, line 57: I don't see the point why to track the list of all commands as you can get the list anyway when checking for the plugin. Can't we just remove `CoreCommanNames` and `InAllowedExtensibleCommandGroups()` as the latter will always return true when you iterate over all commands ?

rhuss: Also it looks to me that "InAllowedExtensibleCommandGroups()" is not really used by any code ?

pkg/kn/commands/plugin/plugin_test.go
```diff
@@ -74,11 +74,15 @@ func TestNewPluginCommand(t *testing.T) {
 }
 
 func TestInAllowedExtensibleCommandGroups(t *testing.T) {
```
rhuss, line 76: These test then could be also removed

pkg/kn/core/root.go
```diff
@@ -78,18 +78,36 @@ func NewDefaultKnCommandWithArgs(rootCmd *cobra.Command,
...
-				fmt.Fprintf(rootCmd.OutOrStderr(), showSubcommands(foundCmd, cmdPathPieces, innerArgs[0]))
+		}
+
+		// look for case of plugins with command that shadows existing command
+		if foundCmd.HasSubCommands() && len(innerArgs) > 0 {
+			cmdName := innerArgs[0]
+			for _, subcommand := range foundCmd.Commands() {
+				if subcommand.Name() == cmdName {
+					fmt.Fprintf(rootCmd.OutOrStderr(), fmt.Sprintf("Error: subcommand '%s' for '%s' already exists.\nRun 'kn --help' for usage.\n", cmdName, foundCmd.Name()))
+					os.Exit(1)
```
rhuss, line 103: I know this is not part of this PR, but doing an `os.Exit()` deep within the call hierarchy is a bad smell as its hard to detect and troubleshoot (imagine some upper layer would have to do some cleanup like removing files or so). Its much better to move the error upwards and have only `os.Exit()` on the top level. But just saying (as I have it noticed here for the first time), maybe we can address this is in a later PR.

pkg/kn/core/root.go
```diff
@@ -340,7 +363,9 @@ func removeKnPluginFlags(args []string) []string {
 		if arg == "--plugins-dir" ||
 			strings.HasPrefix(arg, "--plugins-dir=") ||
 			arg == "--lookup-plugins" ||
-			strings.HasPrefix(arg, "--lookup-plugins=") {
+			strings.HasPrefix(arg, "--lookup-plugins=") ||
+			// remove -test.* args which are added when running go test
+			strings.HasPrefix(arg, "-test.") {
```
rhuss, line 368: Could you elaborate a bit on this ? I don't understand where this can happen.

pkg/kn/core/root_test.go
```diff
@@ -89,12 +90,56 @@ func TestNewDefaultKnCommandWithArgs(t *testing.T) {
...
+				beforeEach(t)
+				args = []string{pluginPath, pluginName}
+				setup(t)
+				defer afterEach(t)
+
+				checkRootCmd(t, rootCmd)
+			})
+
+			t.Run("when plugin extends and shadows an existing command group it fails", func(t *testing.T) {
+				pluginName = "service-create"
```
rhuss, line 137: Can we also please have a test for deeper nested plugins like `kn-source-ping-log`  (which should be possible as `kn source ping` is not a leaf-command)

pkg/kn/commands/plugin/plugin.go
maximilien: Yes, nice catch. I had `InAllowedExtensibleCommandGroups()`before fore previous version for `source` extension. Not needed anymore. Will remove.

pkg/kn/core/root.go
maximilien: If you don't do this then running tests add these `-test.*` args and they fail the code during tests.

pkg/kn/core/root.go
maximilien: Yeah, let's move this to another PR and also hunt for similar smells other places. I'll open an issue.

pkg/kn/commands/plugin/plugin.go
maximilien: Actually, I am using it in the `verifier.go`:

```go
	// Check both, commands with underscore and with dash because plugins can be called with both
	overwrittenCommands := make(map[string]bool)
	for _, c := range [][]string{cmds, convertUnderscoresToDashes(cmds)} {
		cmd, _, err := v.root.Find(c)
		if err == nil {
			if !InAllowedExtensibleCommandGroups(cmd.Name()) {
				overwrittenCommands[cmd.CommandPath()] = true
			}
		}
	}
```

I'll move it there since it's a freeform function and make it private. Initially it was used in multiple places (root.go and verifier.go) and I needed to have it in a place that made sense for both.

pkg/kn/commands/plugin/plugin_test.go
maximilien: See above.

pkg/kn/core/root.go
maximilien: Issue open. Will deal with it after.

pkg/kn/core/root_test.go
maximilien: OK adding that now.

maximilien:
/retest

```bash
go: github.com/google/licenseclassifier upgrade => v0.0.0-20200402202327-879cb1424de0
go: downloading github.com/sergi/go-diff v1.0.0
go: github.com/sergi/go-diff upgrade => v1.1.0
go: downloading github.com/sergi/go-diff v1.1.0
2020/05/22 21:54:37 Error creating license classifier: cannot register licenses: EOF
Step failed: default_build_test_runner
============================
==== BUILD TESTS FAILED ====
============================
==== Fri May 22 14:54:37 PDT 2020
============================
Step failed: run_build_tests
+ EXIT_VALUE=1
+ set +o xtrace
```

maximilien:
Getting a weird error....

```bash
go: github.com/google/licenseclassifier upgrade => v0.0.0-20200402202327-879cb1424de0
go: downloading github.com/sergi/go-diff v1.0.0
go: github.com/sergi/go-diff upgrade => v1.1.0
go: downloading github.com/sergi/go-diff v1.1.0
2020/05/22 21:54:37 Error creating license classifier: cannot register licenses: EOF
Step failed: default_build_test_runner
============================
==== BUILD TESTS FAILED ====
============================
==== Fri May 22 14:54:37 PDT 2020
============================
Step failed: run_build_tests
+ EXIT_VALUE=1
+ set +o xtrace
```

maximilien:
Same consistent CI issue:

```bash
2020/05/22 21:54:37 Error creating license classifier: cannot register licenses: EOF
Step failed: default_build_test_runner
``` 

I'll have to take a look at this next Tuesday --- long weekend here in US due to Memorial Day.

rhuss:
Looks like an incompatible change in the test-infra scripts. @coryrc do you have an idea what has changed ? I think we have the same kind of error over there for client-contrib tests, too.

rhuss:
@maximilien the IT tests should be running again, but there's a conflict now which needs to be resolved.

CHANGELOG.adoc
```diff
@@ -56,6 +56,10 @@
 | Add `-a` flag as an alias for `--annotation`
 | https://github.com/knative/client/pull/782[#782]
 
+| 🎁
+| Allow plugins to extends all command groups
```
navidshaikh, line 60:
```suggestion
| Allow plugins to extend all command groups
```

maximilien:
OK thanks @rhuss and @navidshaikh. Just merged. Let's see what other tests I need to add or not.

maximilien:
@rhuss and @navidshaikh this passes now and is ready. However, I want to make some refactoring to simplify the code in `root.go` a bit. Mostly to make it more readable. Will push an update in next hour and we should be good to go.

rhuss:
Cool, let's get that merged now. 

As I'm starting to work now to include grouping for help messages (#579) and also to weave in plugin help message is needed. This will probably lead to some refactoring also in this part, so please don't touch those in the next week for now.

/lgtm
/approve

PR #840, opened 2020-05-15
Fix test infra

coryrc:
## Description

This allows us to remove go.mod circular dependency for test-infra.

We aren't pulling any Go files from test-infra, so we never should
have been using dep or go mod to pull the scripts in.

TODO: a separate commit to remove knative.dev/test-infra from go.mod

PR #841, opened 2020-05-15
Too many things when updating knative.dev/test-infra

coryrc:
This is from running `go get -d knative.dev/test-infra`

Just a demonstration for https://github.com/knative/test-infra/issues/2069


go.mod
```diff
@@ -7,23 +7,23 @@ require (
...
+	github.com/spf13/viper v1.6.2
+	golang.org/x/crypto v0.0.0-20200302210943-78000ba7a073
 	gomodules.xyz/jsonpatch/v2 v2.1.0 // indirect
 	gotest.tools v2.2.0+incompatible
 	k8s.io/api v0.17.4
 	k8s.io/apimachinery v0.17.4
-	k8s.io/cli-runtime v0.17.0
-	k8s.io/client-go v0.17.4
+	k8s.io/cli-runtime v0.17.3
+	k8s.io/client-go v11.0.1-0.20190805182717-6502b5e7b1b5+incompatible
```
rhuss, line 21: I'm not happy with this, as kn's code also use client-go and so the dependency should be controlled by the `kn`'s dependency not on behalf to the test infrastructure. Also, this should be also always a fixed version.

go.mod
```diff
@@ -7,23 +7,23 @@ require (
 	github.com/google/go-containerregistry v0.0.0-20200413145205-82d30a103c0a // indirect
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/openzipkin/zipkin-go v0.2.2 // indirect
-	github.com/pkg/errors v0.8.1
+	github.com/pkg/errors v0.9.1
 	github.com/robfig/cron v1.2.0 // indirect
-	github.com/spf13/cobra v0.0.5
+	github.com/spf13/cobra v0.0.6
```
rhuss, line 12: fyi, this update does not change anything because of the pinning we use for getting in some bug fixes (see end of go.mod)

go.mod
```diff
@@ -7,23 +7,23 @@ require (
...
+	k8s.io/cli-runtime v0.17.3
+	k8s.io/client-go v11.0.1-0.20190805182717-6502b5e7b1b5+incompatible
 	knative.dev/eventing v0.14.1
-	knative.dev/pkg v0.0.0-20200414233146-0eed424fa4ee
+	knative.dev/pkg v0.0.0-20200514052058-c75d324f8b8b
 	knative.dev/serving v0.14.0
-	knative.dev/test-infra v0.0.0-20200413202711-9cf64fb1b912
-	sigs.k8s.io/yaml v1.1.0
+	knative.dev/test-infra v0.0.0-20200514223200-ef4fd3ad398f
+	sigs.k8s.io/yaml v1.2.0
```
rhuss, line 26: as mentioned above, updating direct dependencies that are also used by `kn` because of a test-infra (and hence those dependencies are only transitively coming in via test-infra) should not be applied blindly.

It could be that its ok, it could be not. This needs to be checked, but as a general rule of thumb, transitive dependencies must not overshadow direct dependencies.

rhuss:
Ah, I see. This is just a test according to https://github.com/knative/test-infra/issues/2069 (this was not clear from the pr description)

rhuss:
but I think the comments are still valid and align with the reasoning in https://github.com/knative/test-infra/issues/2069

coryrc:
> but I think the comments are still valid and align with the reasoning in [knative/test-infra#2069](https://github.com/knative/test-infra/issues/2069)

I agree with you, sorry about the distraction! I'll update description.

/uncc cppforlife
/uncc sixolet

rhuss:
As we have now merged the infra-test update PR #840 , I'm closing this PR.

PR #846, opened 2020-05-18
Fix kn source -h

### LINKED ISSUE #845: kn source -h Returns Error: unknown command 'source'
### Bug report

Running `kn source -h` returns `Error: unknown command 'source'`. This appears due to how `kn` validates commands with plugins. 

This occurs because [`HandlePluginCommand`](https://github.com/knative/client/blob/0df54bc500ad960568e5d00285d3f112b9db665e/pkg/kn/commands/plugin/handler.go#L126) searches for a binary on the local path and returns [`Could not find plugin to execute`](https://github.com/knative/client/blob/0df54bc500ad960568e5d00285d3f112b9db665e/pkg/kn/commands/plugin/handler.go#L151). 

This results in the command being reported as unknown:

https://github.com/knative/client/blob/0df54bc500ad960568e5d00285d3f112b9db665e/pkg/kn/core/root.go#L87

This is solved by adding a `-h` check in `HandlePluginCommand`:

```
if cmdArgs[idx] == "-h" {
    return nil
}
```

### Expected behavior

`kn source -h` returns the following:

```
Usage:
  kn source [flags]
  kn source [command]

Available Commands:
  apiserver   Kubernetes API Server Event Source command group
  binding     Sink binding command group
  list        List available sources
  list-types  List available source types
  ping        Ping source command group

Flags:
  -h, --help   help for source

Global Flags:
      --config string       kn config file (default is ~/.config/kn/config.yaml)
      --kubeconfig string   kubectl config file (default is ~/.kube/config)
      --log-http            log http traffic

Use "kn source [command] --help" for more information about a command.

```

Run `kn source -h`

### kn version

What is currently in master.

N/A

danielhelfand:
## Description

This pull request adds a check for if `-h` or `--help` is specified with a plugin command. There should be no need to check for plugin executable availability if `--help` is specified and will avoid an error being returned that a command does not exist as documented in #845.

## Changes

* Add check for if `-h` or `--help` is specified in `HandlePluginCommand`
* Add test

## Reference

Fixes #845 

/lint

### DIFF
--- a/pkg/kn/core/root.go
+++ b/pkg/kn/core/root.go
@@ -403,7 +403,7 @@ func showSubcommands(cmd *cobra.Command, args []string, innerArg string) string 
 	for _, subcmd := range cmd.Commands() {
 		strs = append(strs, subcmd.Name())
 	}
-	return fmt.Sprintf("Error: unknown subcommand '%s' for '%s'. Available subcommands: %s\nRun 'kn --help' for usage.\n", innerArg, getCommands(args, innerArg), strings.Join(strs, ", "))
+	return fmt.Sprintf("Error: unknown subcommand '%s' for '%s'.\nAvailable subcommands: %s\nRun 'kn --help' for usage.\n", innerArg, getCommands(args, innerArg), strings.Join(strs, ", "))
 }
 
 func helpOptionsPresent(args []string) bool {
--- a/test/e2e/basic_workflow_test.go
+++ b/test/e2e/basic_workflow_test.go
@@ -77,12 +77,11 @@ func TestWrongCommand(t *testing.T) {
 	r := test.NewKnRunResultCollector(t, it)
 	defer r.DumpIfFailed()
 
-	out := test.Kn{}.Run("source", "apiserver", "noverb", "--tag=0.13")
-	assert.Check(t, util.ContainsAll(out.Stderr, "Error", "unknown sub-command", "noverb"))
+	out := test.Kn{}.Run("source", "apiserver", "noverb")
+	assert.Check(t, util.ContainsAll(out.Stderr, "unknown sub-command", "noverb"))
 	r.AssertError(out)
 
 	out = test.Kn{}.Run("rev")
-	assert.Check(t, util.ContainsAll(out.Stderr, "unknown command", "rev"))
+	assert.Check(t, util.ContainsAll(out.Stderr, "unknown command", "rev", "revision"))
 	r.AssertError(out)
-
 }
--- a/pkg/kn/commands/plugin/handler.go
+++ b/pkg/kn/commands/plugin/handler.go
@@ -15,7 +15,6 @@
 package plugin
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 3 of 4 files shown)


rhuss:
Thanks for the PR ! Could you a bit elaborate on your use case ? I'm asking because I think that a 

`kn --help admin` should be the same as `kn admin --help` and print out the help message of the plugin itself (i.e. executing `kn-admin --help`). So I think checking for a valid plugin makes sense and *only* if no or an invalid plugin is given then print out the help (+ an error message if an unknown command/plugin is specified).

danielhelfand:
> So I think checking for a valid plugin makes sense and _only_ if no or an invalid plugin is given then print out the help (+ an error message if an unknown command/plugin is specified).

Thanks for the explanation. My apologies as I am not fully familiar with `kn`'s plugin strategy. I pushed up a change that I think addresses this, but let me know if there are use cases I am not considering here. 

Basically, instead of returning an error for `if len(foundBinaryPath) == 0`, move that check to before attempting to execute the plugin. If the plugin is invalid, the execution returns an error.

```go
if len(foundBinaryPath) != 0 {
    err := pluginHandler.Execute(foundBinaryPath, append([]string{foundBinaryPath}, cmdArgs[len(remainingArgs):]...), os.Environ())
    if err != nil {
        return err
    }
}
```

pkg/kn/core/root.go
```diff
@@ -87,7 +87,8 @@ func NewDefaultKnCommandWithArgs(rootCmd *cobra.Command,
 				fmt.Fprintf(rootCmd.OutOrStderr(), "Error: unknown command '%s' \nRun 'kn --help' for usage.\n", args[1])
 				os.Exit(1)
 			}
-		} else if foundCmd.HasSubCommands() {
+		}
+		if foundCmd.HasSubCommands() {
```
danielhelfand, line 91: This can either be kept as an `else if` and checking would need to be done in the `if` block so an error is still returned if the command is invalid, or it can be checked each time as its own block.

rhuss:
Thanks! I will have a look tomorrow but maybe we should wait until #834 is merged as I think it can conflict.

pkg/kn/core/root.go
maximilien: Thanks. I actually also had a fix in my pending PR for allowing extending all subcommands groups. But thanks for submitting this. I’ll merge mine.

rhuss:
@danielhelfand I think there is still a mismatch in the assertion of the e2e test (see test failure). Could you have a look please ?

rhuss:
@danielhelfand would it be possible that we get this PR merged until tomorrow? I'm asking because I'm planning a bigger refactoring in this area (mostly because for implementing #579 and plugin help messages).

If not, it could be that we later have to change this PR considerably because of conflicts which are likely.

Also, tomorrow the next release will be cut, so would be cool if we could get that feature in.

danielhelfand:
@rhuss Should be fixed with some minor tweaks I'll point out in review messages. Let me know if there's any additional changes you would like, and I can address today. Thanks!

test/e2e/basic_workflow_test.go
```diff
@@ -77,12 +77,11 @@ func TestWrongCommand(t *testing.T) {
 	r := test.NewKnRunResultCollector(t, it)
 	defer r.DumpIfFailed()
 
-	out := test.Kn{}.Run("source", "apiserver", "noverb", "--tag=0.13")
```
danielhelfand, line 80: Seems with changes that flag validation is taking place, so error was coming back saying `--tag` is not a valid flag for `kn source apiserver noverb`. I removed `--tag` for now.

PR #853, opened 2020-05-25
Replace pkg/errors with stdlib errors.

markusthoemmes:
## Description

As per title. Since Golang 1.13, the stdlib has vastly improved regarding error wrapping etc. Make use of that.

## Changes

* Replaced `errors.New` with the stdlib.
* Replaced `errors.Wrap(f)` with `fmt.Errorf`

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -17,8 +17,8 @@
 | | Description | PR
 
 | 🐣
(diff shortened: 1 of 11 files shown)


rhuss:
We have an issue with test-infra scripts right now. So I'm waiting to #840 to be merged and the tackle this testing issues.

PR #854, opened 2020-05-25
Update deps, adapting to API changes

rhuss:
Updates to the latest master dependencies in serving, including incompatible API changes (i.e. adding `Deprecate` prefix to API method names.


rhuss:
Lets Markus do it ...

Issue #815, opened 2020-04-21
Honour deletionTimestamp for "kn revision delete", too


rhuss:
As followup to https://github.com/knative/client/pull/805 add the deletion timestamp handling also for `kn revision delete`.

MIBc:
I want to continue working on it.

rhuss:
Yes, sorry for already merging the other PR but we are cutting a release today (0.14.0), so wanted to have your contribution to be included.

MIBc:
Why we need to check deletionTimestamp for revision delete? 
Unlike update command. Deleting revision will not hang until timeout. It may raise `resource not found` error if revision has been deleted.

rhuss:
Not sure if I understand you comment correcly. Wouldn't it make sense to have consistent behaviour for deleting any resource ? I.e. if the the`deletionTimestamp` is set then no deletion is attempt again (regardless of what we delete).

PR #860, opened 2020-05-27
Check deletionTimestamp for "kn revision delete"

MIBc:
Fixes: #815

### DIFF
--- a/pkg/serving/v1/client.go
+++ b/pkg/serving/v1/client.go
@@ -383,7 +383,7 @@ func (cl *knServingClient) DeleteRevision(name string, timeout time.Duration) er
 		return clienterrors.GetError(err)
 	}
 	if revision.GetDeletionTimestamp() != nil {
-		return fmt.Errorf("can't delete revision %s because it has been marked for deletion", name)
+		return fmt.Errorf("can't delete revision '%s' because it has been already marked for deletion", name)
 	}
 	if timeout == 0 {
 		return cl.deleteRevision(name)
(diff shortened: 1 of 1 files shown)


pkg/serving/v1/client.go
```diff
@@ -378,6 +378,13 @@ func getBaseRevision(cl KnServingClient, service *servingv1.Service) (*servingv1
 
 // Delete a revision by name
 func (cl *knServingClient) DeleteRevision(name string, timeout time.Duration) error {
+	revision, err := cl.client.Revisions(cl.namespace).Get(name, v1.GetOptions{})
+	if err != nil {
+		return clienterrors.GetError(err)
+	}
+	if revision.GetDeletionTimestamp() != nil {
+		return fmt.Errorf("can't delete revision %s because it has been marked for deletion", name)
```
navidshaikh, line 386:
```suggestion
		fmt.Printf("\nWARNING: can't delete revision %s because it is already marked for deletion.\n\n", name)
		return nil
```

navidshaikh: Well, looking at https://github.com/knative/client/blob/c6537a8e842782225dfd3ec0a5281262aec50587/pkg/kn/commands/revision/delete.go#L55-L60

We're not exiting with non-zero exit code at all, so returning with `fmt.Errorf` should be okay as well.

navidshaikh: @MIBc : You can discard my comments above and follow suggestion given by @rhuss.

maximilien, line 386: I think we've consistently used `'%s'` in messages... Please check and follow other messages

navidshaikh, line 386:
```suggestion
		return fmt.Errorf("can't delete revision '%s' because it has been already marked for deletion", name)
```

navidshaikh: Combined a couple of suggestions from @rhuss @maximilien

PR #866, opened 2020-05-29
Update go.mod to specify the module is go1.14

dprotaso:
Thus the go commands will default to -mod=vendor

See: https://golang.org/doc/go1.14#go-command


dprotaso:
I dunno what's up with the build test - the scripts have deviated from other repos 

```
---- Fri May 29 15:18:50 PDT 2020
Step failed: default_build_test_runner
============================
==== BUILD TESTS FAILED ====
============================
==== Fri May 29 15:19:20 PDT 2020
============================
Step failed: run_build_tests
```

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.6.2
 	golang.org/x/crypto v0.0.0-20200302210943-78000ba7a073
+	golang.org/x/tools v0.0.0-20200329025819-fd4102a86c65
```
rhuss, line 9: why that ?

hack/tools.go
```diff
@@ -0,0 +1,23 @@
+// +build tools
```
rhuss, line 1: why is this file needed ?

dprotaso:
go1.14 with the presence of a vendor folder will always use `-mod=vendor` for various `go` commands.

Your `verify-codegen.sh` tries to install the package but it's missing from vendor. To vendor it we need to reference the package being installed - the convention used is to have a `tools.go` file.

dprotaso:
Otherwise error looks like

```
# ./hack/verify-codegen.sh
can't load package: package golang.org/x/tools/cmd/goimports: cannot find package "." in:
	/go/client/vendor/golang.org/x/tools/cmd/goimports
```

third_party/VENDOR-LICENSE/github.com/hashicorp/hcl/hcl/scanner/scanner.go
```diff
@@ -0,0 +1,652 @@
...
+	}
+
+	if ch != eof {
+		s.unread()
+	}
+	return token.NUMBER
+}
+
+// scanMantissa scans the mantissa beginning from the rune. It returns the next
+// non decimal rune. It's used to determine wheter it's a fraction or exponent.
```
mattmoor, line 356: Found potential misspelling(s):
```suggestion
// non decimal rune. It's used to determine whether it's a fraction or exponent.
```

third_party/VENDOR-LICENSE/github.com/hashicorp/hcl/json/scanner/scanner.go
```diff
@@ -0,0 +1,451 @@
...
+	// If we have a larger number and this is zero, error
+	if zero && pos != s.srcPos {
+		s.err("numbers cannot start with 0")
+	}
+
+	return token.NUMBER
+}
+
+// scanMantissa scans the mantissa beginning from the rune. It returns the next
+// non decimal rune. It's used to determine wheter it's a fraction or exponent.
```
mattmoor, line 250: Found potential misspelling(s):
```suggestion
// non decimal rune. It's used to determine whether it's a fraction or exponent.
```

third_party/VENDOR-LICENSE/github.com/hashicorp/hcl/hcl/scanner/scanner.go
mattmoor, line 356: Found potential misspelling(s):
```suggestion
// non decimal rune. It's used to determine whether it's a fraction or exponent.
```

third_party/VENDOR-LICENSE/github.com/hashicorp/hcl/json/scanner/scanner.go
mattmoor, line 250: Found potential misspelling(s):
```suggestion
// non decimal rune. It's used to determine whether it's a fraction or exponent.
```

dprotaso:
weird - update licenses seems broken

dprotaso:
Other option if you don't want to pin the tool to a specific `sha` is to use test-infra's `run_go_tool`

dprotaso:
so I think go code is getting pulled into `third_party` if it's unable to determine the license for the file.

started a thread here: https://knative.slack.com/archives/CA1DTGZ2N/p1591322638315800

Probably not worth conflating this PR with that concern.

dprotaso:
This should be good now - if you want me to drop vendoring `goimports` let me know and I can switch to the `go_run_tool` option

go.mod
rhuss: ok, I see. It's for goimports used during build. Does go has the notion of a "developer dependency" (like npm or Maven) ? I'm not big fan to mix in dependencies here which are not part of the application itself.

hack/tools.go
rhuss: I see, its that you can keep the depedency in go.mod on goimport so that it can be vendored and is not accidentally removed.

It still feels very hackish to me and is a clear smell, that dev dependencies should be managed separately (but it looks like its a go limitation). 

I could see two solutions here:

* Don't compile devtools like "goimports" on the fly but use a given binary (like from a container image)
* Use an own kind of vendor by committing dev dependencies in a different vendor dire (vendor_dev) and fiddle with GOPATH before calling `go get goimports`. 

I agree that using goimports as it is now is not the correct way, also because the `go get` call will update `modules.txt` which sometimes leads to a diff to the comitted modules.txt when running in the CI. Up to now fixed the modules.txt manually in such a case, but vendoring seems to be the better solution here.

test/presubmit-tests.sh
```diff
@@ -50,10 +50,8 @@ function post_build_tests() {
   fi
 }
 
-# Run the unit tests with an additional flag '-mod=vendor' to avoid
-# downloading the deps in unit tests CI job
 function unit_tests() {
-  report_go_test -race -mod=vendor ./... || failed=1
+  report_go_test -race ./... || failed=1
```
rhuss, line 54: Can we keep `-mod=vendor` to the commands and add it where missing ? 

It's better to be explicit about the mode we want to run then to rely on the default of the go runtime. If you only see this line you have no idea whether vendoring is used or not, but you have to check which go version is used. 

Adding `-mod=vendor`  or `-mod=mod` everywhere is much better than relying on defaults.

hack/build.sh
```diff
@@ -130,7 +130,7 @@ source_format() {
 
 go_build() {
   echo "🚧 Compile"
-  go build -mod=vendor -ldflags "$(build_flags $(basedir))" -o kn ./cmd/...
+  go build -ldflags "$(build_flags $(basedir))" -o kn ./cmd/...
```
rhuss, line 133: please keep `-mod=vendor` for the reasons mentioned below.

hack/tools.go
rhuss: If this does not work out or is too involved, than we can of course vendor `goimports`.

But we should keep `-mod=vendor` for all go commands everywhere to be explicit and not relying on defaults which can change (which seems to happen quite a bit in go-land as just seen when going from go 1.13 to 1.14)

rhuss:
> This should be good now - if you want me to drop vendoring `goimports` let me know and I can switch to the `go_run_tool` option

@dprotaso If this avoids messing with the client's go.mod this would be then the preferred solution (if this also works locally outside the CI)

dprotaso:
@rhuss take a look

dprotaso:
rebased since there was a merge conflict

rhuss:
Thanks ! Let me check later ...

hack/build.sh
```diff
@@ -16,6 +16,8 @@
 
 set -o pipefail
 
+source $(dirname $0)/../scripts/test-infra/library.sh
```
rhuss, line 19: Unfortunately, that won't work for macOS. See https://github.com/knative/test-infra/issues/2143

`build.sh` is supposed to run on any Unix like OS, also macOS, _without_ any other dependencies. So having Bash 4 dependency here is a no-go, so we can include any test-infra scripts in build.sh.

hack/build.sh
```diff
@@ -112,19 +114,8 @@ go_fmt() {
...
-     # Run go fmt instead
-     go_fmt
-  else
-     echo "🧽 ${X}Format"
-     goimports -w $(echo $source_dirs)
-     find $(echo $source_dirs) -name "*.go" -print0 | xargs -0 gofmt -s -w
-  fi
+  run_go_tool golang.org/x/tools/cmd/goimports goimports -w $(echo $source_dirs)
+  find $(echo $source_dirs) -name "*.go" -print0 | xargs -0 gofmt -s -w
   set -e
```
rhuss, line 119: Sorry for my former comment about using run_go_tool as I haven't known that the test-infra scripts are not running on the stock bash of macOS. 

However, I wonder whether we could inline the `run_go_tool` code here, making it running on macOs ? 

And even is `run_go_tool` happens to run on macOs (by accident), I still wouldn't include `library.sh` as long as there is a strong commitment to stick to bash version 3 syntax only.

dprotaso:
inlined

hack/build.sh
```diff
@@ -16,6 +16,9 @@
 
 set -o pipefail
 
+[[ ! -v REPO_ROOT_DIR ]] && REPO_ROOT_DIR="$(git rev-parse --show-toplevel)"
```
rhuss, line 19: @dprotaso It's that -v which does not work in bash v3

No worries, I'll fix that tomorrow

rhuss:
> /lgtm

@maximilien why ? It hasn't been tested on macOS (as requested) and breaks now the build for every macOs user.

rhuss:
> inlined

@dprotaso I meant "inline the *logic* of that function", not to inline it _literally_, as this is not better than sourcing `library.sh` (as seen in #882 )

dprotaso:
> > inlined
> 
> @dprotaso I meant "inline the _logic_ of that function", not to inline it _literally_, as this is not better than sourcing `library.sh` (as seen in #882 )

I see my mistake - sorry for the trouble

rhuss:
> I see my mistake - sorry for the trouble

no worries, the fix was trivial :). Thanks for working on aligning the Knative repos !

PR #867, opened 2020-06-01
chore: Remove OWNERS files in vendor/ when updating deps

rhuss:


navidshaikh:
The default build tests reporting the diff for removed OWNERS files, either this can added in test-infra build tests (to have this consistent among other repos) or override build tests in client repo. @rhuss WDYT ?

rhuss:
I just added this to build.sh (to remove the ownerfiles too)

Unfortunately we can't source test-infra scripts into build.sh as the rely on bash 4 which is not available ootb on macos. But the build script should really work for developers ootb

We could move the diff check to not use build sh -x but directly update-deps.sh which might be even the better solution

PR #870, opened 2020-06-02
chore: Unpin dependencies to follow master again

rhuss:
Reset version to "master" and update-deps.sh --upgrade after branching off for release-0.15


rhuss:
@navidshaikh I think this is good to merge, to unpin the version on master again.

PR #872, opened 2020-06-03
Update flag names to --request and --limit

navidshaikh:
## Description
 - Use singular flag names and support comma separated or repeated flag values
 - Update tests and docs

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -21,13 +21,6 @@
 | 🐛
 | Update flag names to --request and --limit
 | https://github.com/knative/client/pull/872[#872]
-|===
-
-## v0.15.0 (2020-06-02)
-
-[cols="1,10,3", options="header", width="100%"]
-|===
-| | Description | PR
 
 | 🐛
 | Fix `kn source -h`
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -12,6 +12,17 @@
 | https://github.com/knative/client/pull/[#]
 ////
 
+## v0.15.1 (2020-06-03)
+
+[cols="1,10,3", options="header", width="100%"]
+|===
+| | Description | PR
+
+| 🐛
+| Update flag names to --request and --limit
+| https://github.com/knative/client/pull/872[#872]
+|===
+
 ## v0.15.0 (2020-06-02)
 
 [cols="1,10,3", options="header", width="100%"]
--- a/docs/cmd/kn_service_create.md
+++ b/docs/cmd/kn_service_create.md
@@ -44,7 +44,7 @@ kn service create NAME --image IMAGE [flags]
   # Create a service with 250MB memory, 200m CPU requests and a GPU resource limit
   # [https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/]
   # [https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/]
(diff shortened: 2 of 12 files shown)


CHANGELOG.adoc
```diff
@@ -12,6 +12,17 @@
...
+
+[cols="1,10,3", options="header", width="100%"]
+|===
+| | Description | PR
+
+| 🐛
+| Update flag names to --request and --limit
+| https://github.com/knative/client/pull/872[#872]
+|===
+
```
rhuss, line 25: I think we can remove the v0.15.0 header as this version has never been officially released.

docs/cmd/kn_service_create.md
```diff
@@ -67,9 +67,9 @@ kn service create NAME --image IMAGE [flags]
...
       --label-revision stringArray    Revision label to set. name=value; you may provide this flag any number of times to set multiple labels. To unset, specify the label name followed by a "-" (e.g., name-). This flag takes precedence over "label" flag.
       --label-service stringArray     Service label to set. name=value; you may provide this flag any number of times to set multiple labels. To unset, specify the label name followed by a "-" (e.g., name-). This flag takes precedence over "label" flag.
-      --limits string                 The resource requirement limits for this Service. For example, 'cpu=100m,memory=256Mi'.
-      --limits-cpu string             DEPRECATED: please use --limits instead. The limits on the requested CPU (e.g., 1000m).
-      --limits-memory string          DEPRECATED: please use --limits instead. The limits on the requested memory (e.g., 1024Mi).
+      --limit strings                 The resource requirement limits for this Service. For example, 'cpu=100m,memory=256Mi'.
```
rhuss: why is the type here still `strings` and not `stringArray` as for `label` ?

rhuss:
It's possible that labels use an array whereas limits/requests use a slice. 

I think we should later adapt labels etc. you slices, too. + we should think about how we can share code better to avoid that kind of differences in the future.

docs/cmd/kn_service_create.md
navidshaikh: if we use `StringSliceVar` type flags, we get `strings` (see traffic flags) and
if we use `StringArrayVar` type flags, we get `stringArray`

navidshaikh:
> It's possible that labels use an array whereas limits/requests use a slice.

Yes (see traffic flags)

> I think we should later adapt labels etc. you slices, too. + we should think about how we can share code better to avoid that kind of differences in the future.

Yes, we'd need to fix this for rest of the flags to work as mentioned in the examples of convention at https://github.com/knative/client/blob/master/conventions/cli.md#maps

CHANGELOG.adoc
navidshaikh: yup, fixed.

pkg/kn/flags/resources_test.go
```diff
@@ -37,49 +37,75 @@ func parseQuantity(value string) resource.Quantity {
...
+			corev1.ResourceList{corev1.ResourceName("cpu"): parseQuantity("200m")},
+			false,
+		},
 	}
 	for _, c := range cases {
 		options := &ResourceOptions{}
 		options.Requests = c.requests
 		options.Limits = c.limits
-		err := options.Validate()
+		reqToRemove, limToRemove, err := options.Validate()
```
rhuss, line 91: Maybe don't call it validate as the main purpose is no to extract the values. What's about `options.Extract()` or `options.Parse()` (or `options.ParseAndValidate()`) ?

navidshaikh: Sure, will update the name when we unify the other similar flags.

PR #877, opened 2020-06-08
Refactor main flow and introduce explicit plugin and config handling

rhuss:
This PR is a significant refactoring of the main flow of how commands and plugins are processed. I'm aware such big changes should not be introduced lightly and only for good reasons. While working to implement #579 and to give some guidance for how to improve the plugin error message flow (cc: @dsimansky), I stumbled upon the following issues (and more):

* Calling out to plugins happens as a side-effect in `NewKnDefaultCommand()`. This code is more or less taken over literally from `kubectl`, but it definitely has it has issues in understandability and maintainability. Plugin error handling is tough here (i.e. how to return the error)
* The options `--plugins-dir` and `--lookup-plugins` are parsed manually (without cobra support), and it was buggy (i.e. doesn't honour both `--plugin-dir=/my/dir` and `--plugin-dir /my/dir`).
* Mangling of args is not transparent.
* Flow of execution is not easy to follow, i.e. not easy to see what happens where.
* Handling of the configuration file is not straight-forward; also it is not clear why the config file has to be written back (which completely messes up any user formatting or comments).
* Too much functionality has ended up in the `kn/commands/plugins` package but should be more top-level as it is used there also to verify the plugin
* The [NewDefaultKnCommandWithArgs](https://github.com/knative/client/blob/3c5ceff655ec4543907a681c300912076fa49e53/pkg/kn/core/root.go#L68) constructor is very complex and too long.

To fix this and to generally enhance the main flow, the following significant changes have been introduced in this PR:

* The plugin handling has been moved out of the `KnDefaultCommand` constructor where it was executed as a side-effect. The original code from `kubectl` suffers from the same issue that plugin handling is not a top-level concern but was very likely introduced as an after-thought. Instead, the plugin handling is done now by a `PluginManager` which is explicitly called in `main()`.
* Configuration and bootstrap option handling is centralized in the package `option`. After the bootstrap happened, the content of the configuration file, as well as any other global configuration, can be obtained from methods on  `config.GlobalConfig`. Also, all flag handling is delegated to cobra so that no own parsing is needed.
* Many of the logic in `pkg/kn/commands/plugin` for plugin management has been moved up to `pkg/kn/plugin` as this code is not only relevant for `plugin list` but also for the bootstrap process.

This PR will also help to make the implementation of the following feature requests easier:

* #763 - Extract common `kn` functionality for plugins and specialized Knative clients
* #579 - Split eventing/serving commands in help message to increase discoverability
* #333 - Introduce --debug and --verbose as global options
[...]

PR #880, opened 2020-06-08
Return Error Message from Using --untag with Nonexistent Tag

### LINKED ISSUE #874: traffic tags: Error if requested for removing non-existing tag
### Bug report
If requested to untag a non-existing tag in traffic block, add an error.

### Expected behavior
`service update` command returns error message indicating the tag doesnt exists for service and non-zero exit status.

1. `kn service create hello --image knativesamples/helloworld`
2. `kn service update hello --untag foo --no-wait`

### kn version
latest

v0.15.0

danielhelfand:
## Description

This pull request returns an error message when using `kn service update serviceName --untag tagName --no-wait` on a tag that does not exist.

The error returned by `kn` will be as follows, but I am open to any change with regard to the message:

```
Error: tag tagName does not exist
```

In the event multiple nonexistent tags are specified, `kn` will return multiple errors:

```
kn service update serviceName --untag tagName --untag tagName2 --no-wait

Error: tag tagName does not exist
Error: tag tagName2 does not exist
```

If a tag that does exist is specified with a nonexistent tag, the error will be displayed and no update will be made. 

Please let me know if this should warrant an e2e test as well.

## Reference

Fixes #874

/lint

### DIFF
--- a/pkg/kn/commands/service/update.go
+++ b/pkg/kn/commands/service/update.go
@@ -92,7 +92,7 @@ func NewServiceUpdateCommand(p *commands.KnParams) *cobra.Command {
 				}
 
 				if trafficFlags.Changed(cmd) {
-					traffic, err := traffic.Compute(cmd, service.Spec.Traffic, &trafficFlags)
+					traffic, err := traffic.Compute(cmd, service.Spec.Traffic, &trafficFlags, service.Name)
 					if err != nil {
 						return nil, err
 					}
--- a/pkg/kn/commands/service/update_test.go
+++ b/pkg/kn/commands/service/update_test.go
@@ -743,7 +743,7 @@ func TestServiceUpdateTagDoesNotExist(t *testing.T) {
 	_, _, _, err := fakeServiceUpdate(orig, []string{
 		"service", "update", "foo", "--untag", "foo", "--no-wait"})
 
-	assert.Error(t, err, "Error: tag foo does not exist")
+	assert.Assert(t, util.ContainsAll(err.Error(), "tag(s)", "foo", "not present", "service", "foo"))
 }
 
 func newEmptyService() *servingv1.Service {
--- a/pkg/kn/commands/service/update_test.go
+++ b/pkg/kn/commands/service/update_test.go
@@ -737,6 +737,15 @@ func TestServiceUpdateDeletionTimestampNotNil(t *testing.T) {
 	assert.ErrorContains(t, err, "service")
 }
 
+func TestServiceUpdateTagDoesNotExist(t *testing.T) {
+	orig := newEmptyService()
+
+	_, _, _, err := fakeServiceUpdate(orig, []string{
+		"service", "update", "foo", "--untag", "foo", "--no-wait"})
+
+	assert.Error(t, err, "Error: tag foo does not exist")
+}
+
 func newEmptyService() *servingv1.Service {
 	ret := &servingv1.Service{
 		TypeMeta: metav1.TypeMeta{
--- a/pkg/kn/traffic/compute.go
+++ b/pkg/kn/traffic/compute.go
@@ -83,15 +83,15 @@ func (e ServiceTraffic) isTagPresent(tag string) bool {
 	return false
 }
 
-func (e ServiceTraffic) untagRevision(tag string) error {
+func (e ServiceTraffic) untagRevision(tag string, serviceName string) string {
 	for i, target := range e {
 		if target.Tag == tag {
 			e[i].Tag = ""
-			return nil
+			return ""
 		}
 	}
 
-	return fmt.Errorf("Error: tag %s does not exist", tag)
+	return tag
 }
 
 func (e ServiceTraffic) isRevisionPresent(revision string) bool {
@@ -269,7 +269,8 @@ func verifyInputSanity(trafficFlags *flags.Traffic) error {
 }
 
 // Compute takes service traffic targets and updates per given traffic flags
-func Compute(cmd *cobra.Command, targets []servingv1.TrafficTarget, trafficFlags *flags.Traffic) ([]servingv1.TrafficTarget, error) {
+func Compute(cmd *cobra.Command, targets []servingv1.TrafficTarget,
+	trafficFlags *flags.Traffic, serviceName string) ([]servingv1.TrafficTarget, error) {
 	err := verifyInputSanity(trafficFlags)
 	if err != nil {
 		return nil, err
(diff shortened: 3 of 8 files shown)


pkg/kn/commands/service/update_test.go
```diff
@@ -737,6 +737,15 @@ func TestServiceUpdateDeletionTimestampNotNil(t *testing.T) {
 	assert.ErrorContains(t, err, "service")
 }
 
+func TestServiceUpdateTagDoesNotExist(t *testing.T) {
+	orig := newEmptyService()
+
+	_, _, _, err := fakeServiceUpdate(orig, []string{
+		"service", "update", "foo", "--untag", "foo", "--no-wait"})
+
+	assert.Error(t, err, "Error: tag foo does not exist")
```
rhuss, line 746: I'd suggest to change this to (please add the import, too, if missing0:

```go
assert.Assert(t, util.ContainsAll(err.Error(), "tag", "foo", "not exist"))
```

It's always very fragile to assert on the exact wording of something, which might slight improvements tedious to work on (as then always a test needs to be fixed). Instead, just check that the important _context information_ is contained in the error message. This focus also having good error message which as much context information as possible to help a user to understand and fix the error. E.g. Here you could also add the service name which would help the user to understand which service you wanted to tag. The more context, the better.

Also, please use error message according to the golang standard: Starting lower case, no punctuation at the need. The "Error:" prefix is also redundant here (its an error anyway), and in fact the top-level routine which prints the error will add the error prefix anyway. So better remove it here, too.

We have an issues open (not sure where), where we want to check for all error messages to be consistently following this rule. Might be a good exercise to automate that and make it part of the build process.

pkg/kn/traffic/compute.go
```diff
@@ -276,8 +278,17 @@ func Compute(cmd *cobra.Command, targets []servingv1.TrafficTarget, trafficFlags
...
-		traffic.untagRevision(tag)
+		err = traffic.untagRevision(tag)
+		if err != nil {
+			errStrings = append(errStrings, err.Error())
+		}
+	}
+
+	// Return all errors from untagging revisions
+	if len(errStrings) > 0 {
+		return nil, fmt.Errorf(strings.Join(errStrings, "\n"))
```
rhuss, line 291: Please don't add any formatting like `\n` to the end of an error message. Use a `", "` instead as separator.

pkg/kn/traffic/compute_test.go
```diff
@@ -278,6 +278,18 @@ func TestComputeErrMsg(t *testing.T) {
 			[]string{"--traffic", "echo-v1=40", "--traffic", "echo-v1=60"},
 			"repetition of revision reference echo-v1 is not allowed, use only once with --traffic flag",
 		},
+		{
+			"untag single tag that does not exist",
+			append(newServiceTraffic([]servingv1.TrafficTarget{}), newTarget("latest", "echo-v1", 100, false)),
+			[]string{"--untag", "foo"},
+			"Error: tag foo does not exist",
```
rhuss, line 285: See my first comment: I would never check on the exact wording of an error message as this might change (e.g. like for the suggestions I made, e.g. to avoid the `Error` prefix and add the service information.

pkg/kn/commands/service/update.go
```diff
@@ -92,7 +92,7 @@ func NewServiceUpdateCommand(p *commands.KnParams) *cobra.Command {
 				}
 
 				if trafficFlags.Changed(cmd) {
-					traffic, err := traffic.Compute(cmd, service.Spec.Traffic, &trafficFlags)
+					traffic, err := traffic.Compute(cmd, service.Spec.Traffic, &trafficFlags, service.Name)
```
danielhelfand, line 95: Added `serviceName` parameter to `traffic.Compute` to pass service name for error messaging. Let me know if there is another option for getting the service name in `compute.go` that would get rid of the need to pass data from here.

pkg/kn/traffic/compute.go
```diff
@@ -276,8 +279,17 @@ func Compute(cmd *cobra.Command, targets []servingv1.TrafficTarget, trafficFlags
...
-		traffic.untagRevision(tag)
+		tagName := traffic.untagRevision(tag, serviceName)
+		if tagName != "" {
+			errTagNames = append(errTagNames, tagName)
+		}
+	}
+
+	// Return all errors from untagging revisions
+	if len(errTagNames) > 0 {
+		return nil, fmt.Errorf("tag(s) %s not present for any revisions of service %s", strings.Join(errTagNames, ", "), serviceName)
```
danielhelfand, line 292: Updated formatting of error message to include all invalid tags used with `--untag` for a particular service and including name for more context with the error message.

pkg/kn/traffic/compute_test.go
danielhelfand: Due to the structure of the tests, can we leave this in place for now? They seem to focus on the exact error message via `assert.Error(t, err, testCase.errMsg)`. However, would be happy to adapt the tests here maybe as a separate issue/pr.

pkg/kn/traffic/compute.go
danielhelfand: Believe this is addressed by comment below:

https://github.com/knative/client/pull/880#discussion_r437513614

pkg/kn/commands/service/update_test.go
danielhelfand: Just to clarify here, the `error: ` prefix won't be added unless included in the message it seems:

```
./kn service update hello --untag foo --no-wait

tag(s) foo not present for any revisions of service hello
```

pkg/kn/traffic/compute_test.go
rhuss: I'm fine with that.

pkg/kn/commands/service/update_test.go
rhuss: Totally agree, that a consistent error handling is missing, but for consistency's sake, I think there should be a single point which prepares an error message for print out.

In fact, in the refactoring in #877 there is a single exit point in case of a custom error, that now adds the appropriate prefix --> https://github.com/knative/client/blob/a71cc3fd91b9722e983929709c84ee9fd29939c5/cmd/kn/main.go#L38-L41

We can more error messaging mangling at this point (like uppercasing the first letter), but if we assume that the change above gets merged, then we should already avoid an `Error:` prefix.

rhuss: But don't let that block this PR as we have to go over all error messages later anyway.

pkg/kn/traffic/compute.go
```diff
@@ -276,8 +279,17 @@ func Compute(cmd *cobra.Command, targets []servingv1.TrafficTarget, trafficFlags
 	traffic := newServiceTraffic(targets)
 
 	// First precedence: Untag revisions
+	var errTagNames []string
 	for _, tag := range trafficFlags.UntagRevisions {
-		traffic.untagRevision(tag)
+		tagName := traffic.untagRevision(tag, serviceName)
+		if tagName != "" {
+			errTagNames = append(errTagNames, tagName)
+		}
```
navidshaikh, line 287: We're returning the value from function which is passed as its own arguments, 
I think `untagRevsion` could return a bool value based on status of the operation?

Issue #812, opened 2020-04-18
Support plural resource nouns


mattmoor:
### Feature request

In `kubectl` and `tkn` I can use either the singular or plural for resource names, but with `kn` I get an error when using the plural:

```shell
kn services list
Error: unknown command 'services'
Run 'kn --help' for usage.
```

IMO `services` should simply alias `service`.

rhuss:
+1 especially for commands that operate on multiple resource instances (but I wouldn't also mind so support both variants unconditionally eg. also `kn services create`)

navidshaikh:
should we also consider bringing other aliases with this? 

for example
 - `svc` for services
 - `rev` for revisions
etc

duglin:
svc might be too easily confused with k8s svc

rhuss:
> svc might be too easily confused with k8s svc

don't think this is a strong argument as the same applies for 'service' itself, too. (i.e. that it can be confused with a k8s 'service')

I thin we can consider command shortcuts but I would do it one step after the other. @navidshaikh would you mind to open an extra issue for this ? There we could also collect the possible shorcuts (like also `src` for `source` ...)

mattmoor:
To chime in on shortnames, muscle memory would have me typing `ksvc`, which is how we alias it in kubectl 😇

rhuss:
Fixed with 0.18

PR #887, opened 2020-06-16
Add grouping for help message + streamlined help messages

### LINKED ISSUE #579: Split eventing/serving commands in help message to increase discoverability

rhuss:
The top-level looks like

```
kn is the command line interface for managing Knative Serving and Eventing resources

 Find more information about Knative at: https://knative.dev

Serving Commands:
  service     Manage Knative services
  revision    Manage service revisions
  route       List and show service routes

Eventing Commands:
  source      Manage event sources
  trigger     Manage event triggers

Other Commands:
  plugin      Manage kn plugins
  completion  Output shell completion code
  version     Show the version of this client

Use "kn <command> --help" for more information about a given command.
Use "kn options" for a list of global command-line options (applies to all commands).
```

The following changes have been applied:

* Add CommandGroups for grouping commands together
* Add flexible templating for the help messages
* Moved global options to an own command ('kn options', much like 'kubectl options')
* Aligned wording and typography of help messages

These features has been highly inspired by kubectl grouping & help templating but has been considerably been stripped down to the needs of kn.

Fixes #579

### DIFF
--- a/cmd/kn/main.go
+++ b/cmd/kn/main.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	"regexp"
 	"strings"
 	"time"
 
@@ -35,8 +36,9 @@ func init() {
 
 func main() {
 	err := run(os.Args[1:])
-	if err != nil {
+	if err != nil && len(os.Args) > 1 {
 		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		fmt.Fprintf(os.Stderr, "Run '%s --help' for usage\n", extractCommandPathFromErrorMessage(err.Error(), os.Args[0]))
 		// This is the only point from where to exit when an error occurs
 		os.Exit(1)
 	}
@@ -106,7 +108,13 @@ func stripFlags(args []string) ([]string, error) {
 				*commandsFound = append(*commandsFound, arg)
 			}
 		},
+		SilenceErrors: true,
+		SilenceUsage:  true,
 	}
+
+	// No help and usage functions to prin
+	extractCommand.SetHelpFunc(func(*cobra.Command, []string) {})
+	extractCommand.SetUsageFunc(func(*cobra.Command) error { return nil })
 
 	// Filter out --help and -h options to avoid special treatment which we don't
 	// need here
@@ -172,10 +180,22 @@ func validateRootCommand(cmd *cobra.Command) error {
 	if err == nil && foundCmd.HasSubCommands() && len(innerArgs) > 0 {
 		argsWithoutFlags, err := stripFlags(innerArgs)
 		if len(argsWithoutFlags) > 0 || err != nil {
-			return errors.Errorf("unknown sub-command '%s' for '%s'. Available sub-commands: %s", innerArgs[0], foundCmd.Name(), strings.Join(root.ExtractSubCommandNames(foundCmd.Commands()), ", "))
+			return errors.Errorf("unknown sub-command '%s' for '%s'. Available sub-commands: %s", innerArgs[0], foundCmd.CommandPath(), strings.Join(root.ExtractSubCommandNames(foundCmd.Commands()), ", "))
 		}
 		// If no args where given (only flags), then fall through to execute the command itself, which leads to
 		// a more appropriate error message
 	}
 	return nil
+}
+
+// extractCommandPathFromErrorMessage tries to extract the command name from an error message
+// by checking a pattern like 'kn service' in the error message. If not found, return the
+// base command name.
+func extractCommandPathFromErrorMessage(errorMsg string, arg0 string) string {
+	extractPattern := regexp.MustCompile(fmt.Sprintf("'(%s\\s.+?)'", arg0))
+	command := extractPattern.FindSubmatch([]byte(errorMsg))
+	if command != nil {
+		return string(command[1])
+	}
+	return arg0
 }
--- a/cmd/kn/main.go
+++ b/cmd/kn/main.go
@@ -37,8 +37,7 @@ func init() {
 func main() {
 	err := run(os.Args[1:])
 	if err != nil && len(os.Args) > 1 {
-		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
-		fmt.Fprintf(os.Stderr, "Run '%s --help' for usage\n", extractCommandPathFromErrorMessage(err.Error(), os.Args[0]))
+		printError(err)
 		// This is the only point from where to exit when an error occurs
 		os.Exit(1)
 	}
@@ -188,6 +187,12 @@ func validateRootCommand(cmd *cobra.Command) error {
 	return nil
 }
 
+// printError prints out any given error
+func printError(err error) {
+	fmt.Fprintf(os.Stderr, "Error: %s\n", cleanupErrorMessage(err.Error()))
+	fmt.Fprintf(os.Stderr, "Run '%s --help' for usage\n", extractCommandPathFromErrorMessage(err.Error(), os.Args[0]))
+}
+
 // extractCommandPathFromErrorMessage tries to extract the command name from an error message
 // by checking a pattern like 'kn service' in the error message. If not found, return the
 // base command name.
@@ -199,3 +204,10 @@ func extractCommandPathFromErrorMessage(errorMsg string, arg0 string) string {
 	}
 	return arg0
 }
+
+// cleanupErrorMessage remove any redundance content of an error message
+func cleanupErrorMessage(msg string) string {
+	regexp := regexp.MustCompile("(?i)^error:\\s*")
+	return string(regexp.ReplaceAll([]byte(msg), []byte("")))
+}
+
--- a/cmd/kn/main.go
+++ b/cmd/kn/main.go
@@ -210,4 +210,3 @@ func cleanupErrorMessage(msg string) string {
 	regexp := regexp.MustCompile("(?i)^error:\\s*")
 	return string(regexp.ReplaceAll([]byte(msg), []byte("")))
 }
-
--- a/cmd/kn/main_test.go
+++ b/cmd/kn/main_test.go
@@ -24,7 +24,7 @@ import (
 	"github.com/spf13/cobra"
 	"gotest.tools/assert"
 
-	"knative.dev/client/pkg/kn/commands"
+	"knative.dev/client/lib/test"
 	"knative.dev/client/pkg/kn/root"
 	"knative.dev/client/pkg/util"
 )
@@ -243,7 +243,7 @@ func TestRunWithError(t *testing.T) {
 		},
 	}
 	for _, d := range data {
-		capture := commands.CaptureOutput(t)
+		capture := test.CaptureOutput(t)
 		printError(errors.New(d.given))
 		stdOut, errOut := capture.Close()
 
@@ -261,7 +261,7 @@ func TestRun(t *testing.T) {
 		os.Args = oldArgs
 	})()
 
-	capture := commands.CaptureOutput(t)
+	capture := test.CaptureOutput(t)
 	err := run(os.Args[1:])
 	out, _ := capture.Close()
 
--- a/cmd/kn/main_test.go
+++ b/cmd/kn/main_test.go
@@ -15,8 +15,10 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/spf13/cobra"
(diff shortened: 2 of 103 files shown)


pkg/templates/command_groups.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/spf13/cobra"
+)
+
+// A command group is for grouping together commands
+type CommandGroup struct {
+
+	// Title for command group shown in help/usage messages
+	Header string
+
+	// List of commans for this group
```
mattmoor, line 28: Found potential misspelling(s):
```suggestion
	// List of commands for this group
```

rhuss:
Please note that testing still needs to be added, this PR is however already open for review to gather early feedback.

pkg/templates/command_groups.go
mattmoor, line 28: Found potential misspelling(s):
```suggestion
	// List of commands for this group
```

rhuss:
> Some general comments would be to check for error messages that already include the prefix to avoid double prints, such as [here](https://github.com/knative/client/blob/d725ef594a2e0f4659b42a279eb0d5acfe80f1b1/pkg/printers/tableprinter.go#L66). Maybe that's handled, but in general should be removed and discouraged in the future.

Since there is a single exit point we are free to mangle the error message as we want. But I consider this more like a workaround since adding an "error: " prefix in an error message looks quite redundant (it's clear that it is an error if the prefix is part of an error message, also the formatting of the error message should be left to the layer which deals with the error). However, we don't have much influence on the wording of errors returned from library calls.

I will add some tests to check the error wording.

cmd/kn/main.go
```diff
@@ -172,10 +179,35 @@ func validateRootCommand(cmd *cobra.Command) error {
...
+	}
+	return arg0
+}
+
+// cleanupErrorMessage remove any redundance content of an error message
+func cleanupErrorMessage(msg string) string {
+	regexp := regexp.MustCompile("(?i)^error:\\s*")
+	return string(regexp.ReplaceAll([]byte(msg), []byte("")))
+}
+
```
mattmoor, line 213: Format Go code:
```suggestion

```

cmd/kn/main_test.go
```diff
@@ -226,6 +228,32 @@ func TestStripFlags(t *testing.T) {
 	}
 }
 
+func TestRunWithError(t *testing.T) {
+	data := []struct {
+		given string
```
mattmoor, line 233: Format Go code:
```suggestion
		given    string
```

cmd/kn/main_test.go
```diff
@@ -226,6 +228,32 @@ func TestStripFlags(t *testing.T) {
...
+		expected string
+	} {
```
mattmoor, line 235: Format Go code:
```suggestion
	}{
```

cmd/kn/main_test.go
```diff
@@ -226,6 +228,32 @@ func TestStripFlags(t *testing.T) {
...
+		{
+			"unknown sub-command blub",
+			"Error: unknown sub-command blub",
+		},
+		{
+			"error: unknown type blub",
+			"Error: unknown type blub",
+		},
+
```
mattmoor, line 244: Format Go code:
```suggestion

```

rhuss:
Added the missing testing bits.

pkg/kn/root/root.go
```diff
@@ -77,28 +70,47 @@ func NewRootCommand() (*cobra.Command, error) {
...
+				source.NewSourceCommand(p),
+				trigger.NewTriggerCommand(p),
+			},
+		},
+		{
+			Header: "Other Commands:",
+			Commands: []*cobra.Command{
+				plugin.NewPluginCommand(p),
+				completion.NewCompletionCommand(p),
+				version.NewVersionCommand(p),
```
navidshaikh, line 95: @rhuss : should this include `options` command as well?

pkg/templates/templates.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// sectionTipsGlobalOptions is the help template section that displays the 'options' hint for displaying global flags.
+	sectionTipsGlobalOptions = `Use "{{rootCmdName}} options" for a list of global command-line options (applies to all commands).`
+)
+
+// usageTemplate if the template for 'usage' used by most commands.
+func usageTemplate() string {
+	sections := []string{
+		"\n\n",
+		//		sectionAliases,
```
navidshaikh, line 73: :+1:

pkg/kn/root/root.go
rhuss: Kubectl doesn't include "options" as a regular command, so I tend to do the same. It's state prominently at the bottom how to get to the global options. And "options" is really kind of a no-op, much like "help" which is also not listed. So I think, that's ok to not include it here.

pkg/templates/command_groups.go
```diff
@@ -0,0 +1,66 @@
...
+
+// SetRootUsage sets our own help and usage function messages to the root command
+func (g CommandGroups) SetRootUsage(rootCmd *cobra.Command) {
+	engine := &templateEngine{
+		RootCmd:       rootCmd,
+		CommandGroups: g,
+	}
+	setHelpFlagsToSubCommands(rootCmd)
+	rootCmd.SilenceUsage = true
+	rootCmd.SilenceErrors = true
```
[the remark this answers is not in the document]
rhuss: true, let me change that.

pkg/templates/command_groups.go
```diff
@@ -0,0 +1,66 @@
...
+	rootCmd.SetFlagErrorFunc(func(c *cobra.Command, err error) error {
+		return errors.Errorf("%s for '%s'", err.Error(), c.CommandPath())
+	})
+	rootCmd.SetUsageFunc(engine.usageFunc())
+	rootCmd.SetHelpFunc(engine.helpFunc())
+}
+
+func setHelpFlagsToSubCommands(parent *cobra.Command) {
```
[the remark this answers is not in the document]
rhuss: but it adds "[options]" instead of "[flags]" --> https://github.com/rhuss/knative-client/blob/41338722cdc2124dda2b3a594527749a31d8b735/pkg/templates/template_engine.go#L134

Again, inline with `kubectl` and "options" is a good wording here, as all the rest is just optional. Everything mandatory should be mentioned in the "Use" attribute.

pkg/templates/templates.go
rhuss: Need to fix the alias section but this could be done when we introduce aliases (like "services" or "ksvc")

pkg/kn/commands/source/apiserver/apiserver.go
```diff
@@ -26,8 +26,8 @@ import (
 // NewAPIServerCommand for managing ApiServer source
 func NewAPIServerCommand(p *commands.KnParams) *cobra.Command {
 	apiServerSourceCmd := &cobra.Command{
-		Use:   "apiserver",
-		Short: "Kubernetes API Server Event Source command group",
+		Use:   "apiserver COMMAND",
```
maximilien, line 29: Shouldn't this be `api-server COMMAND`?

pkg/kn/commands/testing_helper.go
```diff
@@ -93,33 +93,44 @@ func CreateDynamicTestKnCommand(cmd *cobra.Command, knParams *KnParams, objects
 
 }
 
-type StdoutCapture struct {
-	r, w *os.File
-	t    *testing.T
+type OutputCapture struct {
```
maximilien, line 96: Perhaps this `testing_helper.go` should move to `lib/test` to make it useable in `client-contrib` plugins?

rhuss: Good idea, going to move it over there.

pkg/kn/commands/source/apiserver/apiserver.go
rhuss: We settled on `apiserver` for the command (like in `kn source apiserver list`), so changing it now would be backwards incompatible. We could do this if its improve the UX considerably but should be aware that it involves more work than just renaming it here. 

What we can do though is to add an alias `api-server` for `apiserver` alongside the suggestions in #812 .

docs/cmd/kn_source_apiserver_list.md
```diff
@@ -1,13 +1,13 @@
 ## kn source apiserver list
```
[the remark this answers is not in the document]
rhuss: see comment below.

docs/cmd/kn_plugin.md
```diff
@@ -1,16 +1,16 @@
 ## kn plugin
 
-Plugin command group
+Manage kn plugins
 
 ### Synopsis
 
-Provides utilities for interacting and managing with kn plugins.
+Manage kn plugins
```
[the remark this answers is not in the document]
rhuss: 👍 for this but maybe we should have a first release for the client-contrib plugins. We can add this info any time with a link to a consolidated list.

docs/cmd/kn_source_binding_delete.md
```diff
@@ -1,13 +1,13 @@
 ## kn source binding delete
 
-Delete a sink binding.
+Delete a sink binding
 
 ### Synopsis
 
-Delete a sink binding.
+Delete a sink binding
```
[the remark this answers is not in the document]
rhuss: I think this is due to the way how the docs are generated and that the `BindingDeleteCommand` doesn't have a `Long` description (so the short description is reused). 

We should add to every command a longer descriptions (if it makes sense, here e.g. one could add a brief explanation what a sink binding is, maybe with a link to the binding docs).

As this PR doesn't change anything with regard to the Long description, I'd suggest to open a new issue for this and work separately on consolidating the long descriptions.

pkg/templates/command_groups.go
```diff
@@ -0,0 +1,66 @@
...
+func (g CommandGroups) SetRootUsage(rootCmd *cobra.Command) {
+	engine := &templateEngine{
+		RootCmd:       rootCmd,
+		CommandGroups: g,
+	}
+	setHelpFlagsToSubCommands(rootCmd)
+	rootCmd.SilenceUsage = true
+	rootCmd.SilenceErrors = true
+	rootCmd.SetFlagErrorFunc(func(c *cobra.Command, err error) error {
+		return errors.Errorf("%s for '%s'", err.Error(), c.CommandPath())
```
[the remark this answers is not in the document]
rhuss: done

pkg/kn/commands/testing_helper.go
rhuss: I just moved the `OutputCapture` to `lib/test`. 

I think the other parts are more for internally testing `KnCommands` that is specific of how `kn` organizes it commands (also there might be some change here, too).

pkg/kn/commands/source/apiserver/apiserver.go
navidshaikh: :+1 for alias

pkg/templates/templates.go
navidshaikh: yea, aliases could go in separate section, +1 for the placeholder..

navidshaikh:
@rhuss : CHANGELOG missing, lets bring it in another PR

rhuss:
Thanks !

PR #893, opened 2020-06-18
chore: add missing CHANGELOG entry for grouped help messages

rhuss:


rhuss:
@navidshaikh here's the missing changelog entry

PR #896, opened 2020-06-19
Revisit sink prefixes: Support only 'ksvc' prefix and drop 'svc', 'service' support for knative service

### LINKED ISSUE #898: Discussion: Sink Mappings revisited
Currently, we have a set of fixed mappings for the `--sink` option of sources and triggers:

* `broker:` for specifying a broker
* `service:` & `svc:` for specifying a Knative service
* `http:` and `https:` for direct URL
* no prefix: Knative service

We can also add new mapping in the configuration file `~/.config/kn/config.yaml`  with 

```yaml
eventing:
  sink-mappings:
  - prefix: k8service
    resource: services
    group: core
    version: v1
```

However, I think this is not optimal:

* Knative service spec occupies too many "slots" (no prefix, `svc:`, `service:`)
* A K8s Service which is a perfectly valid target can not easily be specified, there is no good prefix available anymore

The question is also how many mapping we would like to add to `kn` as builtin mappings.
There are two possible ways:

* **Intersection**: Only restrict to addressable Knative entities (services, broker, channels)
* **Union**: Also allow entities from the underlying platform, like Kubernetes or CloudRun. Prominent example here is a Kubernetes service, but one could imagine also other mappings.

My suggestion is going for the latter approach and changing the built-in mappings to:

* Default is still a Knative service (i.e. can be specified without prefix)
* With prefix, only `ksvc:` is support for a Knative service
* `svc:` is used to map to a Kubernetes service
* More prefix could be added, even when they are targeting objects that are not covered by the Knative spec.

Note that this can't be easily achieved in a backwards-compatible way. Still, I believe we should make adjustments here (also considering that we also stated that the eventing support in kn was experimental).

What are your opinion on this topic ?

navidshaikh:
Fixes #898 

- 'ksvc' prefix --> Knative Service (or no prefix means knative service)
- Update all the examples and tests
- Add examples for broker, service and URI in flag help message
- Drop support for 'svc' and 'service' prefixes for knative service, hint in err msg if these prefixes are given

help message:
```
  -s, --sink string               Addressable sink for events. You can specify broker, knative service or URI. Examples: '--sink broker:nest' for 'nest' broker, '--sink
                                  https://event.receiver.uri' for an URI (identified by 'http' or 'https'), '--sink ksvc:receiver' or '--sink receiver' for 'receiver' knative service.
                                  If prefix is not provided, it is considered as knative service.
```

### DIFF
--- a/pkg/kn/commands/flags/sink.go
+++ b/pkg/kn/commands/flags/sink.go
@@ -38,11 +38,11 @@ func (i *SinkFlags) Add(cmd *cobra.Command) {
 		"s",
 		"",
 		"Addressable sink for events. "+
-			"You can specify broker, knative service or URI. "+
-			"Examples: '--sink broker:nest' for 'nest' broker, "+
-			"'--sink https://event.receiver.uri' for an URI (identified by 'http' or 'https'), "+
-			"'--sink ksvc:receiver' or '--sink receiver' for 'receiver' knative service. "+
-			"If prefix is not provided, it is considered as knative service.")
+			"You can specify a broker, Knative service or URI. "+
+			"Examples: '--sink broker:nest' for a broker 'nest', "+
+			"'--sink https://event.receiver.uri' for an URI with an 'http://' or 'https://' schema, "+
+			"'--sink 'ksvc:receiver' or simply '--sink receiver' for a Knative service 'receiver'. "+
+			"If prefix is not provided, it is considered as a Knative service.")
 
 	for _, p := range config.GlobalConfig.SinkMappings() {
 		//user configration might override the default configuration
--- a/pkg/kn/commands/flags/sink.go
+++ b/pkg/kn/commands/flags/sink.go
@@ -38,13 +38,11 @@ func (i *SinkFlags) Add(cmd *cobra.Command) {
 		"s",
 		"",
 		"Addressable sink for events. "+
-			"You can specify broker, service or URI. "+
+			"You can specify broker, knative service or URI. "+
 			"Examples: '--sink broker:nest' for 'nest' broker, "+
 			"'--sink https://event.receiver.uri' for an URI (identified by 'http' or 'https'), "+
-			"'--sink service:receiver' for 'receiver' knative service, "+
-			"you can also use prefix alias 'svc' for 'service'. "+
-			"If do not specify a prefix, it is considered as knative service, "+
-			"i.e. '--sink receiver' is same as '--sink svc:receiver'.")
+			"'--sink ksvc:receiver' or '--sink receiver' for 'receiver' knative service. "+
+			"If prefix is not provided, it is considered as knative service.")
 
 	for _, p := range config.GlobalConfig.SinkMappings() {
 		//user configration might override the default configuration
--- a/pkg/kn/commands/flags/sink.go
+++ b/pkg/kn/commands/flags/sink.go
@@ -33,7 +33,18 @@ type SinkFlags struct {
 }
 
 func (i *SinkFlags) Add(cmd *cobra.Command) {
-	cmd.Flags().StringVarP(&i.sink, "sink", "s", "", "Addressable sink for events")
+	cmd.Flags().StringVarP(&i.sink,
+		"sink",
+		"s",
+		"",
+		"Addressable sink for events. "+
+			"You can specify broker, service or URI. "+
+			"Examples: '--sink broker:nest' for 'nest' broker, "+
+			"'--sink https://event.receiver.uri' for an URI (identified by 'http' or 'https'), "+
+			"'--sink service:receiver' for 'receiver' knative service, "+
+			"you can also use prefix alias 'svc' for 'service'. "+
+			"If do not specify a prefix, it is considered as knative service, "+
+			"i.e. '--sink receiver' is same as '--sink svc:receiver'.")
 
 	for _, p := range config.GlobalConfig.SinkMappings() {
 		//user configration might override the default configuration
--- a/pkg/kn/commands/flags/sink.go
+++ b/pkg/kn/commands/flags/sink.go
@@ -61,13 +61,8 @@ var sinkMappings = map[string]schema.GroupVersionResource{
 		Group:    "eventing.knative.dev",
 		Version:  "v1beta1",
 	},
-	"service": {
-		Resource: "services",
-		Group:    "serving.knative.dev",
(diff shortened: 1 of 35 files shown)


pkg/kn/commands/flags/sink.go
```diff
@@ -33,7 +33,18 @@ type SinkFlags struct {
...
+		"sink",
+		"s",
+		"",
+		"Addressable sink for events. "+
+			"You can specify broker, service or URI. "+
+			"Examples: '--sink broker:nest' for 'nest' broker, "+
+			"'--sink https://event.receiver.uri' for an URI (identified by 'http' or 'https'), "+
+			"'--sink service:receiver' for 'receiver' knative service, "+
+			"you can also use prefix alias 'svc' for 'service'. "+
+			"If do not specify a prefix, it is considered as knative service, "+
```
dsimansk, line 46:
```suggestion
			"If you do not specify a prefix, it is considered as knative service, "+
```

dsimansk: Or maybe the other way around "If prefix is not specified/provided".

rhuss:
Sounds good, but I'm generally not so happy with the duality of `--sink service` and `--sink svc:service` because how would we add a K8s service here ?

A suggestion would be `--sink kservice` and `--sink ksvc:kservice` + `--sink svc:service` for a k8s service. However this would not be backwards compatible :(

navidshaikh:
> However this would not be backwards compatible :(

Yes. Despite supporting prefix for k8s service, I agree that knative service prefix should be called `ksvc` OR `kservice`.

How about deprecating 'svc' and 'service' prefixes in favor of 'ksvc' and 'kservice' ?

<hr>

> how would we add a K8s service here ?

should we support prefixes for non-knative addressable resources ?

Giving URI or defining sink prefix for non-knative addressable should do.

It should explicit about GVKs for non-knative addressable than hardcoding GVKs in kn and giving a prefix for it ?

navidshaikh:
/hold 

for conclusion at https://github.com/knative/client/issues/898

pkg/kn/commands/flags/sink.go
navidshaikh: thanks @dsimansk! fixed.

navidshaikh:
/retest

    Error: ReconcileIngressFailed: Ingress reconciliation failed

rhuss:
Looks good but I wonder whether we should keep `svc:` for one or two release to point to knative service for the sake of backwards compatibility ? I mean we could remove it from all help message, but still evaluate it if present and print a deprecation warning that this will be removed.

Also we should allow overriding `svc:` with the config (but only this specific prefix because we will free it up later anyway).

Does this make sense ?

navidshaikh:
This can be done, however I wonder if you should provide this (temporary) UX for allowing to override the default prefix via config values. Our eventing support is experimental, we may choose to put hard stop for default  `svc` prefix and clear the path for k8s service without bringing any confusion/change-of-UX.

rhuss:
Ok, let's do it that way for now. We will learn anyway if we get some backslash, but otherwise being overcautios is not good as well.

We make Eventing support as experimental, so I think that's fine.
 
But maybe we can add to error message when a user uses SVC: that this has been moved to ksvc: ?

navidshaikh:
> But maybe we can add to error message when a user uses SVC: that this has been moved to ksvc: ?

@rhuss : sure, PTAL https://github.com/knative/client/pull/896/commits/ad8c47eda4ee20d1d4ff09470d61718cbf59abd0

pkg/kn/commands/flags/sink.go
```diff
@@ -33,7 +33,16 @@ type SinkFlags struct {
...
 
 func (i *SinkFlags) Add(cmd *cobra.Command) {
-	cmd.Flags().StringVarP(&i.sink, "sink", "s", "", "Addressable sink for events")
+	cmd.Flags().StringVarP(&i.sink,
+		"sink",
+		"s",
+		"",
+		"Addressable sink for events. "+
+			"You can specify broker, knative service or URI. "+
+			"Examples: '--sink broker:nest' for 'nest' broker, "+
```
rhuss, line 42:
```suggestion
			"Examples: '--sink broker:nest' for a broker 'nest', "+
```

pkg/kn/commands/flags/sink.go
```diff
@@ -33,7 +33,16 @@ type SinkFlags struct {
...
+			"'--sink https://event.receiver.uri' for an URI (identified by 'http' or 'https'), "+
```
rhuss, line 43:
```suggestion
			"'--sink https://event.receiver.uri' for an URI with an 'http://' or 'https://' schema, "+
```

pkg/kn/commands/flags/sink.go
```diff
@@ -33,7 +33,16 @@ type SinkFlags struct {
...
+			"'--sink ksvc:receiver' or '--sink receiver' for 'receiver' knative service. "+
```
rhuss, line 44:
```suggestion
			"'--sink ksvc:receiver' or simply '--sink receiver' for a knative service ''receiver'. "+
```

pkg/kn/commands/flags/sink.go
```diff
@@ -33,7 +33,16 @@ type SinkFlags struct {
...
+			"If prefix is not provided, it is considered as knative service.")
```
rhuss, line 45:
```suggestion
			"If prefix is not provided, it is considered as a Knative service.")
```

pkg/kn/commands/flags/sink.go
```diff
@@ -84,7 +88,10 @@ func (i *SinkFlags) ResolveSink(knclient clientdynamic.KnDynamicClient, namespac
 	}
 	typ, ok := sinkMappings[prefix]
 	if !ok {
-		return nil, fmt.Errorf("unsupported sink type: %s", i.sink)
+		if prefix == "svc" || prefix == "service" {
```
rhuss, line 91: I have not tested it but does this also throw an error if a user configures that prefix in its configuration file ? (that should not be the case).

navidshaikh: Nope, this check is after key lookup in `sinkMapping` (`	typ, ok := sinkMappings[prefix]`) just a couple lines ago.

rhuss:
Thanks !

/lgtm

PR #848, opened 2020-05-19
Fix Panic for kn source ping describe with Sink URI

danielhelfand:
## Description

This pull request moves all `Sink.Ref` properties from being referenced for `kn source ping describe` until after it has been checked if `Sink.Ref` is nil due to using a URI instead of a Ref. 

It also fixes tests for `kn source ping describe`, which are not asserting whether command output is successfully returned or not. 

## Reference

Fixes #847 

/lint

PR #901, opened 2020-06-22
Fix Panic for kn source apiserver and kn source binding describe with Sink URI

danielhelfand:
## Description

Following up on #848 and #900, this pull request updates `kn source apiserver` and `kn source binding describe` to avoid a panic when using a Sink that has a URI instead of a Ref. 

This pull request also updates unit tests for `apiserver` and `binding` as some tests are not asserting whether output of commands is correct. 

/lint

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -19,7 +19,7 @@
 | | Description | PR
 
 | 🐛
-| Fix Panic for `kn source apiserver` and `kn source binding` describe with Sink URI
+| Fix panic for `kn source apiserver` and `kn source binding` describe with Sink URI
 | https://github.com/knative/client/pull/901[#901]
 
 | 🐛
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -19,6 +19,10 @@
 | | Description | PR
 
 | 🐛
+| Fix Panic for `kn source apiserver` and `kn source binding` describe with Sink URI
+| https://github.com/knative/client/pull/901[#901]
+
+| 🐛
 | Fix Panic for `kn trigger describe` with Sink URI
 | https://github.com/knative/client/pull/900[#900]
 
--- a/pkg/kn/commands/source/apiserver/apiserver_test.go
+++ b/pkg/kn/commands/source/apiserver/apiserver_test.go
@@ -19,6 +19,7 @@ import (
 
 	"k8s.io/client-go/tools/clientcmd"
 	v1alpha2 "knative.dev/eventing/pkg/apis/sources/v1alpha2"
+	"knative.dev/pkg/apis"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
 
 	kndynamic "knative.dev/client/pkg/dynamic"
@@ -29,8 +30,25 @@ import (
 
 const testNamespace = "default"
 
-// Helper methods
-var blankConfig clientcmd.ClientConfig
+var (
+	sinkRef = duckv1.Destination{
+		Ref: &duckv1.KReference{
+			Kind:       "Service",
+			Name:       "testsvc",
+			APIVersion: "serving.knative.dev/v1",
+			Namespace:  "default",
+		},
+	}
+
+	sinkURI = duckv1.Destination{
+		URI: &apis.URL{
+			Scheme: "https",
+			Host:   "foo",
+		},
+	}
+
+	blankConfig clientcmd.ClientConfig
+)
 
 // TODO: Remove that blankConfig hack for tests in favor of overwriting GetConfig()
 func init() {
@@ -83,19 +101,11 @@ func cleanupAPIServerMockClient() {
 	apiServerSourceClientFactory = nil
 }
 
-func createAPIServerSource(name, resourceKind, resourceVersion, serviceAccount, mode, service string, ceOverrides map[string]string) *v1alpha2.ApiServerSource {
+func createAPIServerSource(name, resourceKind, resourceVersion, serviceAccount, mode string, ceOverrides map[string]string, sink duckv1.Destination) *v1alpha2.ApiServerSource {
 	resources := []v1alpha2.APIVersionKindSelector{{
 		APIVersion: resourceVersion,
 		Kind:       resourceKind,
 	}}
-
-	sink := duckv1.Destination{
-		Ref: &duckv1.KReference{
-			Kind:       "Service",
-			Name:       service,
-			APIVersion: "serving.knative.dev/v1",
-			Namespace:  "default",
-		}}
 
 	return clientv1alpha2.NewAPIServerSourceBuilder(name).
 		Resources(resources).
--- a/pkg/kn/commands/source/apiserver/apiserver_test.go
+++ b/pkg/kn/commands/source/apiserver/apiserver_test.go
@@ -19,7 +19,6 @@ import (
 
 	"k8s.io/client-go/tools/clientcmd"
 	v1alpha2 "knative.dev/eventing/pkg/apis/sources/v1alpha2"
-	"knative.dev/pkg/apis"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
 
 	kndynamic "knative.dev/client/pkg/dynamic"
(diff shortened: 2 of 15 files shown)


pkg/kn/commands/source/apiserver/describe.go
```diff
@@ -110,10 +109,10 @@ func writeResources(dw printers.PrefixWriter, apiVersionKindSelectors []v1alpha2
 
 func writeSink(dw printers.PrefixWriter, sink duckv1.Destination) {
 	subWriter := dw.WriteAttribute("Sink", "")
-	subWriter.WriteAttribute("Name", sink.Ref.Name)
-	subWriter.WriteAttribute("Namespace", sink.Ref.Namespace)
 	ref := sink.Ref
 	if ref != nil {
+		subWriter.WriteAttribute("Name", sink.Ref.Name)
+		subWriter.WriteAttribute("Namespace", sink.Ref.Namespace)
 		subWriter.WriteAttribute("Kind", fmt.Sprintf("%s (%s)", sink.Ref.Kind, sink.Ref.APIVersion))
```
danielhelfand, line 116: Should this be `"Resource"` instead? Every other source describe command uses `"Resource"` except apiserver.

maximilien: I can comment out lines 120 and 122 and still pass tests...

```go
func writeSink(dw printers.PrefixWriter, sink duckv1.Destination) {
	subWriter := dw.WriteAttribute("Sink", "")
	ref := sink.Ref
	if ref != nil {
		subWriter.WriteAttribute("Name", sink.Ref.Name)
		subWriter.WriteAttribute("Namespace", sink.Ref.Namespace)
		subWriter.WriteAttribute("Kind", fmt.Sprintf("%s (%s)", sink.Ref.Kind, sink.Ref.APIVersion))
	}
	uri := sink.URI
	// if uri != nil {
		subWriter.WriteAttribute("URI", uri.String())
	// }
}
```

pkg/kn/commands/source/apiserver/describe.go
```diff
@@ -110,10 +109,10 @@ func writeResources(dw printers.PrefixWriter, apiVersionKindSelectors []v1alpha2
 
 func writeSink(dw printers.PrefixWriter, sink duckv1.Destination) {
 	subWriter := dw.WriteAttribute("Sink", "")
-	subWriter.WriteAttribute("Name", sink.Ref.Name)
-	subWriter.WriteAttribute("Namespace", sink.Ref.Namespace)
 	ref := sink.Ref
 	if ref != nil {
+		subWriter.WriteAttribute("Name", sink.Ref.Name)
+		subWriter.WriteAttribute("Namespace", sink.Ref.Namespace)
```
maximilien, line 115: I can comment out line 115 (and / or 116) and still pass all tests...

```go
func writeSink(dw printers.PrefixWriter, sink duckv1.Destination) {
	subWriter := dw.WriteAttribute("Sink", "")
	ref := sink.Ref
	if ref != nil {
		subWriter.WriteAttribute("Name", sink.Ref.Name)
		// subWriter.WriteAttribute("Namespace", sink.Ref.Namespace)
		// subWriter.WriteAttribute("Kind", fmt.Sprintf("%s (%s)", sink.Ref.Kind, sink.Ref.APIVersion))
	}
	uri := sink.URI
	if uri != nil {
		subWriter.WriteAttribute("URI", uri.String())
	}
}
```

pkg/kn/commands/source/binding/describe.go
```diff
@@ -93,12 +93,12 @@ func writeSinkBinding(dw printers.PrefixWriter, binding *v1alpha2.SinkBinding, p
...
 	subWriter := dw.WriteAttribute("Sink", "")
-	if sink.Ref.Namespace != "" && sink.Ref.Namespace != namespace {
-		subWriter.WriteAttribute("Namespace", sink.Ref.Namespace)
-	}
-	subWriter.WriteAttribute("Name", sink.Ref.Name)
 	ref := sink.Ref
 	if ref != nil {
+		subWriter.WriteAttribute("Name", sink.Ref.Name)
+		if sink.Ref.Namespace != "" && sink.Ref.Namespace != namespace {
+			subWriter.WriteAttribute("Namespace", sink.Ref.Namespace)
```
maximilien, line 100: I can comment line 100 and still pass all tests...

```go
func writeSink(dw printers.PrefixWriter, namespace string, sink *duckv1.Destination) {
	subWriter := dw.WriteAttribute("Sink", "")
	ref := sink.Ref
	if ref != nil {
		subWriter.WriteAttribute("Name", sink.Ref.Name)
		if sink.Ref.Namespace != "" && sink.Ref.Namespace != namespace {
			// subWriter.WriteAttribute("Namespace", sink.Ref.Namespace)
		}
		subWriter.WriteAttribute("Resource", fmt.Sprintf("%s (%s)", sink.Ref.Kind, sink.Ref.APIVersion))
	}
	uri := sink.URI
	if uri != nil {
		subWriter.WriteAttribute("URI", uri.String())
	}
}
```

pkg/kn/commands/source/apiserver/describe.go
danielhelfand: That would make sense wouldn't it? 

This would be the output of `TestSimpleDescribe`:

```
Actual output: Name:                testsource
        Namespace:           
        Age:                 
        ServiceAccountName:  testsa
        EventMode:           Reference
        
        Sink:         
          Name:       testsvc
          Namespace:  default
          Kind:       Service (serving.knative.dev/v1)
          URI:        
        
        CloudEvent Overrides:  
          foo:                 bar
        Resources:             
          Kind:                Event (v1)
        
        Conditions:  
          OK TYPE    AGE REASON
```

Basically, `uri.String()` is nil protected. The point of the `if` is to not display `URI`, which it will not if the `if uri != nil` is left in place.

pkg/kn/commands/source/apiserver/describe.go
danielhelfand: Sure, I'll need to update `createAPIServerSource`, which doesn't allow for setting a namespace currently. I don't think the original tests were covering this.

pkg/kn/commands/source/binding/describe.go
danielhelfand: The namespace for the sink ref `myservicenamespace` as well as the kind `Service ()` have been added to the original test to increase its coverage. 

Curious about the need for `if sink.Ref.Namespace != "" && sink.Ref.Namespace != namespace {` though?

pkg/kn/commands/source/apiserver/describe.go
danielhelfand: Was able to just add `default` as namespace and kind as `Service (serving.knative.dev/v1)` to original test, which was only using `Service`. Doesn't cover since `ServiceAccount` is in the output.

pkg/kn/commands/source/apiserver/describe.go
danielhelfand: Added a ContainsNone assertion to make sure `URI` is not present when using Sink Ref in test cases.

pkg/kn/commands/source/binding/describe.go
navidshaikh: > Curious about the need for if sink.Ref.Namespace != "" && sink.Ref.Namespace != namespace { though?

@danielhelfand: We dont support referencing sink in a different namespace yet, the namespace would be same as the one `kn` operating in.
This would be helpful in printing the sink namespaces if the source is created say via yaml or `kn service import` / `kn service apply`(in future).

pkg/kn/commands/source/apiserver/update_test.go
```diff
@@ -21,13 +21,34 @@ import (
...
+	}
+
+	sinkRefsvc2 = duckv1.Destination{
+		Ref: &duckv1.KReference{
+			Kind:       "Service",
+			Name:       "svc2",
+			APIVersion: "serving.knative.dev/v1",
+			Namespace:  "default",
+		},
+	}
```
navidshaikh, line 49: how about function `sinkv1(service, namespace, string)` and use that inline wherever needed ?

(could be done in a follow up)

danielhelfand: I have added a `createSinkv1` func in `apiserver_test.go`. Turns out this same func exists for binding already as `createServiceSink`. If there is a location to recommend, we could use the same func for both, but let me know what's preferred with regard to that.

pkg/kn/commands/source/apiserver/apiserver_test.go
```diff
@@ -29,8 +30,25 @@ import (
...
+			Namespace:  "default",
+		},
+	}
+
+	sinkURI = duckv1.Destination{
+		URI: &apis.URL{
+			Scheme: "https",
+			Host:   "foo",
+		},
+	}
```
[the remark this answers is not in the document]
danielhelfand: Removed.

CHANGELOG.adoc
```diff
@@ -18,6 +18,10 @@
 |===
 | | Description | PR
 
+| 🐛
+| Fix Panic for `kn source apiserver` and `kn source binding` describe with Sink URI
```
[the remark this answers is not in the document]
danielhelfand: Uncapitalized it.

maximilien:
Please address @navidshaikh comments. Otherwise LGTM

danielhelfand:
> Please address @navidshaikh comments. Otherwise LGTM

Should all be addressed, but can wait for @navidshaikh to clarify if I've missed anything.

PR #908, opened 2020-06-25
Add Missing Assertions for Command Unit Tests

danielhelfand:
## Description

This pull request adds assertions for unit tests for verifying command output. The commands missing these assertions in unit tests were `source ping`, `trigger`, and `broker`. 

It also updates unit tests for `source ping` since `Ping` is capitalized in command output for `delete` and `list`.

/lint

### DIFF
--- a/pkg/kn/commands/broker/delete_test.go
+++ b/pkg/kn/commands/broker/delete_test.go
@@ -36,7 +36,7 @@ func TestBrokerDelete(t *testing.T) {
 
 	out, err := executeBrokerCommand(eventingClient, "delete", brokerName)
 	assert.NilError(t, err, "Broker should be deleted")
-	util.ContainsAll(out, "Broker", brokerName, "deleted", "namespace", "default")
+	assert.Assert(t, util.ContainsAll(out, "Broker", brokerName, "deleted", "namespace", "default"))
 
 	eventingRecorder.Validate()
 }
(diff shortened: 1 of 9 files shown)


pkg/kn/commands/broker/delete_test.go
```diff
@@ -36,7 +36,7 @@ func TestBrokerDelete(t *testing.T) {
 
 	out, err := executeBrokerCommand(eventingClient, "delete", brokerName)
 	assert.NilError(t, err, "Broker should be deleted")
-	util.ContainsAll(out, "Broker", brokerName, "deleted", "namespace", "default")
+	assert.Assert(t, util.ContainsAll(out, "Broker", brokerName, "deleted", "namespace", "default"))
```
maximilien, line 39: hmm, nice catches... it’s easy read this as `util.ContainAll(...)` doing the checks... but it’s not. Well done.

danielhelfand: Yeah, ran into this when writing some tests for earlier pull requests. Maybe original thought process was that assertion was taking place in ContainsAll(...) as you said.

danielhelfand:
> The go-coverage should've added the coverage table as comment? 🤔 OR it didnt since there wasnt any delta ?

Yeah, there shouldn't be any change in actual coverage as there was coverage from test, just nothing to fail the test if output didn't match expectations.

PR #913, opened 2020-07-02
feat: Add --filename flag to service create command

### LINKED ISSUE #550: Support for yaml deployments
Is it a non-goal to support declarative deployments like `gcloud alpha run services replace` command with `kn`. Honestly, kubectl works fine but `kn` has a nice "wait for readiness" feature.

As a user, I observe that  I'm not entirely convinced that `kn service create` has options for pretty much everything I need, but the resulting command looks, well, not so declarative or easy to store in a codebase:

```
kn service create hello --force \
--image gcr.io/google-samples/hello-app:1.0  \
--annotation "networking.knative.dev/ingress.class=kourier.ingress.networking.knative.dev" \
--limits-cpu 100m \
--limits-memory=256Mi \
--requests-cpu 50m \
--requests-memory=128Mi \
--env-from cm:mysql-options
```

Instead of checking in a script to this to my codebase,  I might as well opt in for the YAML+kubectl route  (and get consistency with the rest of my Kubernetes deployments), at the risk of losing "readiness check" capabilities of `kn`.

Also Knative currently only supports a single container in `RevisionSpec.containers[]` field but there are open proposals for sidecar containers, and I'm not sure how would that look like in the fully imperative CLI.

I'm curious:

- if many of Knative users would still prefer to have access to YAML via `kn`?
- can `kn` possibly offer readiness check for latest deployed revision as a separate feature?

dsimansk:
## Description

Allow users to create Services from provided YAML or JSON file.

There're couple of open questions:
- Is it necessary to keep `NAME` parameter for the create command?
  - Should the `NAME` parameter take priority over file provided name?
- Should the multi-item (services separated by `---`) YAML be supported as well? 

## Changes

* Add new flag `--filename` to `service create` command

## Reference

Fixes #550

### DIFF
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -275,7 +275,7 @@ func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
 	command.Flags().BoolVar(&p.ForceCreate, "force", false,
 		"Create service forcefully, replaces existing service if any.")
 	command.Flags().StringVarP(&p.Filename, "filename", "f", "", "Create a service from file.")
-	command.MarkFlagFilename("file")
+	command.MarkFlagFilename("filename")
 }
 
 // Apply mutates the given service according to the flags in the command.
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -274,8 +274,7 @@ func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
 	p.addSharedFlags(command)
 	command.Flags().BoolVar(&p.ForceCreate, "force", false,
 		"Create service forcefully, replaces existing service if any.")
-	//command.MarkFlagRequired("image")
-	command.Flags().StringVarP(&p.File, "file", "", "", "Create service from file.")
+	command.Flags().StringVarP(&p.File, "filename", "f", "", "Create service from file.")
 	command.MarkFlagFilename("file")
 }
 
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -68,6 +68,8 @@ type ConfigurationEditFlags struct {
 	GenerateRevisionName bool
 	ForceCreate          bool
 
+	File string
+
 	// Bookkeeping
 	flags []string
 }
@@ -272,7 +274,9 @@ func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
 	p.addSharedFlags(command)
 	command.Flags().BoolVar(&p.ForceCreate, "force", false,
 		"Create service forcefully, replaces existing service if any.")
-	command.MarkFlagRequired("image")
+	//command.MarkFlagRequired("image")
+	command.Flags().StringVarP(&p.File, "file", "", "", "Create service from file.")
+	command.MarkFlagFilename("file")
 }
 
 // Apply mutates the given service according to the flags in the command.
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -68,7 +68,7 @@ type ConfigurationEditFlags struct {
 	GenerateRevisionName bool
 	ForceCreate          bool
 
-	File string
+	Filename string
 
 	// Bookkeeping
 	flags []string
@@ -274,7 +274,7 @@ func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
 	p.addSharedFlags(command)
 	command.Flags().BoolVar(&p.ForceCreate, "force", false,
 		"Create service forcefully, replaces existing service if any.")
-	command.Flags().StringVarP(&p.File, "filename", "f", "", "Create service from file.")
+	command.Flags().StringVarP(&p.Filename, "filename", "f", "", "Create a service from file.")
 	command.MarkFlagFilename("file")
 }
 
--- a/pkg/kn/commands/service/create.go
+++ b/pkg/kn/commands/service/create.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 
 	"knative.dev/client/pkg/kn/commands"
 	servinglib "knative.dev/client/pkg/serving"
@@ -28,6 +29,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/yaml"
+
 	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
 
 	clientservingv1 "knative.dev/client/pkg/serving/v1"
@@ -79,7 +82,7 @@ func NewServiceCreateCommand(p *commands.KnParams) *cobra.Command {
 				return errors.New("'service create' requires the service name given as single argument")
 			}
 			name := args[0]
-			if editFlags.Image == "" {
+			if editFlags.Image == "" && editFlags.File == "" {
 				return errors.New("'service create' requires the image name to run provided with the --image option")
 			}
 
@@ -88,7 +91,12 @@ func NewServiceCreateCommand(p *commands.KnParams) *cobra.Command {
 				return err
 			}
 
-			service, err := constructService(cmd, editFlags, name, namespace)
+			var service *servingv1.Service
+			if editFlags.File == "" {
+				service, err = constructService(cmd, editFlags, name, namespace)
+			} else {
+				service, err = constructServiceFromFile(cmd, editFlags, name, namespace)
+			}
 			if err != nil {
 				return err
 			}
@@ -97,8 +105,7 @@ func NewServiceCreateCommand(p *commands.KnParams) *cobra.Command {
 			if err != nil {
 				return err
 			}
-
-			serviceExists, err := serviceExists(client, name)
+			serviceExists, err := serviceExists(client, service.Name)
 			if err != nil {
 				return err
 			}
@@ -108,7 +115,7 @@ func NewServiceCreateCommand(p *commands.KnParams) *cobra.Command {
 				if !editFlags.ForceCreate {
 					return fmt.Errorf(
 						"cannot create service '%s' in namespace '%s' "+
-							"because the service already exists and no --force option was given", name, namespace)
+							"because the service already exists and no --force option was given", service.Name, namespace)
 				}
 				err = replaceService(client, service, waitFlags, out)
 			} else {
@@ -258,5 +265,34 @@ func constructService(cmd *cobra.Command, editFlags ConfigurationEditFlags, name
 	if err != nil {
 		return nil, err
 	}
+	return &service, nil
+}
+
+// constructServiceFromFile creates struct from provided file
+func constructServiceFromFile(cmd *cobra.Command, editFlags ConfigurationEditFlags, name, namespace string) (*servingv1.Service, error) {
+	var service servingv1.Service
+	file, err := os.Open(editFlags.File)
+	if err != nil {
+		return nil, err
+	}
+	decoder := yaml.NewYAMLOrJSONDecoder(file, 512)
+
+	err = decoder.Decode(&service)
+	if err != nil {
+		return nil, err
+	}
+	if service.Name != name {
+		return nil, fmt.Errorf("Provided service name '%s' doesn't match name from file '%s'.", name, service.Name)
+	}
+	// Set namespace in case it's specified as --namespace
+	service.ObjectMeta.Namespace = namespace
+
+	// We need to generate revision to have --force replace working
+	revName, err := servinglib.GenerateRevisionName(editFlags.RevisionName, &service)
+	if err != nil {
+		return nil, err
+	}
+	service.Spec.Template.Name = revName
+
 	return &service, nil
 }
--- a/pkg/kn/commands/service/create.go
+++ b/pkg/kn/commands/service/create.go
@@ -82,7 +82,7 @@ func NewServiceCreateCommand(p *commands.KnParams) *cobra.Command {
 				return errors.New("'service create' requires the service name given as single argument")
 			}
 			name := args[0]
-			if editFlags.Image == "" && editFlags.File == "" {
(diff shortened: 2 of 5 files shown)


pkg/kn/commands/service/create_test.go
```diff
@@ -707,3 +710,99 @@ func TestServiceCreateWithClusterLocal(t *testing.T) {
...
+  "apiVersion": "serving.knative.dev/v1",
+  "kind": "Service",
+  "metadata": {
+    "name": "foo"
+  },
+  "spec": {
+    "template": {
+		"spec": {
+		"containers": [
+			{
```
mattmoor, line 740: Remove trailing whitespace:
```suggestion
			{
```

test/e2e/service_file_test.go
```diff
@@ -0,0 +1,122 @@
...
+  "apiVersion": "serving.knative.dev/v1",
+  "kind": "Service",
+  "metadata": {
+    "name": "foo-json"
+  },
+  "spec": {
+    "template": {
+		"spec": {
+		"containers": [
+			{
```
mattmoor, line 58: Remove trailing whitespace:
```suggestion
			{
```

pkg/kn/commands/service/create.go
```diff
@@ -260,3 +267,32 @@ func constructService(cmd *cobra.Command, editFlags ConfigurationEditFlags, name
...
+		return nil, err
+	}
+	decoder := yaml.NewYAMLOrJSONDecoder(file, 512)
+
+	err = decoder.Decode(&service)
+	if err != nil {
+		return nil, err
+	}
+	if service.Name != name {
+		return nil, fmt.Errorf("Provided service name '%s' doesn't match name from file '%s'.", name, service.Name)
```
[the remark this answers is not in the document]
dsimansk: Adressed in https://github.com/knative/client/pull/913/commits/228610a353d1a5487cd0f86f4a688117e38872ca.

maximilien:
>Is it necessary to keep NAME parameter for the create command?

Yes. I would suggest just to have it as priority. And if `--file` is passed then ignore it. So the following all work:

`kn service create NAME ...`
`kn service create --file ...`

>Should the NAME parameter take priority over file provided name?

Yes. If both provided use `NAME` and/or error

>Should the multi-item (services separated by ---) YAML be supported as well

Yes. But I can easily see an argument for supporting one service in first pass and adding this after.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -68,6 +68,8 @@ type ConfigurationEditFlags struct {
 	GenerateRevisionName bool
 	ForceCreate          bool
 
+	File string
```
maximilien, line 71: I always debate using `Filename` instead of `File` since there is invariably a `File` object created somewhere later. Of course, `File` is shorter as a flag name so makes sense too... Argh.

test/e2e/service_file_test.go
```diff
@@ -0,0 +1,122 @@
...
+	it, err := test.NewKnTest()
+	assert.NilError(t, err)
+	defer func() {
+		assert.NilError(t, it.Teardown())
+	}()
+
+	r := test.NewKnRunResultCollector(t, it)
+	defer r.DumpIfFailed()
+
+	tempDir, err := ioutil.TempDir("", "kn-file")
```
maximilien, line 85: Cleanup? Easy with a `defer os.RemoveAll(tempDir)` or similar.

pkg/kn/commands/service/create.go
```diff
@@ -260,3 +267,32 @@ func constructService(cmd *cobra.Command, editFlags ConfigurationEditFlags, name
...
+	file, err := os.Open(editFlags.File)
+	if err != nil {
+		return nil, err
+	}
+	decoder := yaml.NewYAMLOrJSONDecoder(file, 512)
+
+	err = decoder.Decode(&service)
+	if err != nil {
+		return nil, err
+	}
```
maximilien, line 283: I can comment these lines and pass all tests. Add a test for bad input file.

pkg/kn/commands/service/create.go
```diff
@@ -260,3 +267,32 @@ func constructService(cmd *cobra.Command, editFlags ConfigurationEditFlags, name
...
+		return nil, fmt.Errorf("provided service name '%s' doesn't match name from file '%s'", name, service.Name)
+	}
+	// Set namespace in case it's specified as --namespace
+	service.ObjectMeta.Namespace = namespace
+
+	// We need to generate revision to have --force replace working
+	revName, err := servinglib.GenerateRevisionName(editFlags.RevisionName, &service)
+	if err != nil {
+		return nil, err
+	}
```
maximilien, line 294: No tests for problem generating revision as I can comment these and pass tests. This is less important coverage than the comment above.

pkg/kn/commands/service/configuration_edit_flags.go
dsimansk: Following the `kubectl` example we could settle on `-f, --filename` as `--force` is not using that.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -272,7 +274,8 @@ func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
 	p.addSharedFlags(command)
 	command.Flags().BoolVar(&p.ForceCreate, "force", false,
 		"Create service forcefully, replaces existing service if any.")
-	command.MarkFlagRequired("image")
+	command.Flags().StringVarP(&p.File, "filename", "f", "", "Create service from file.")
```
navidshaikh, line 277:
```suggestion
	command.Flags().StringVarP(&p.File, "filename", "f", "", "Create a service from file.")
```

pkg/kn/commands/service/configuration_edit_flags.go
navidshaikh, line 71:
```suggestion
	Filename string
```

pkg/kn/commands/service/create.go
```diff
@@ -260,3 +267,32 @@ func constructService(cmd *cobra.Command, editFlags ConfigurationEditFlags, name
...
+
+	err = decoder.Decode(&service)
+	if err != nil {
+		return nil, err
+	}
+	if service.Name != name {
+		return nil, fmt.Errorf("provided service name '%s' doesn't match name from file '%s'", name, service.Name)
+	}
+	// Set namespace in case it's specified as --namespace
+	service.ObjectMeta.Namespace = namespace
```
navidshaikh, line 288: should the namespace also follow check as done above for service name ?

pkg/kn/commands/service/create.go
```diff
@@ -260,3 +267,32 @@ func constructService(cmd *cobra.Command, editFlags ConfigurationEditFlags, name
...
+
+	// We need to generate revision to have --force replace working
+	revName, err := servinglib.GenerateRevisionName(editFlags.RevisionName, &service)
```
navidshaikh, line 291: I am thinking if the client-side revision name generation could be turned off if using files mode?

pkg/kn/commands/service/configuration_edit_flags.go
dsimansk: https://github.com/knative/client/pull/913/commits/d10ecd121af60f0cc2ca1cc8dfea5c797fac7586

pkg/kn/commands/service/configuration_edit_flags.go
dsimansk: https://github.com/knative/client/pull/913/commits/d10ecd121af60f0cc2ca1cc8dfea5c797fac7586

rhuss:
> > Is it necessary to keep NAME parameter for the create command?
> 
> Yes. I would suggest just to have it as priority. And if `--file` is passed then ignore it. So the following all work:
> 

I would keep the NAME argument and make it non-mandatory when `--file` is given, too, but I also would never ignore anything that a user provides on the command line as this is not what they expect.

My suggestion:

* no name arg, name in file --> Take name in file
* name arg, no name in file --> Take name arg
* name arg and name in file given, both equal --> Take that name
* name arg and name in file given, but different --> Error

dsimansk:
@maximilien @rhuss 
Based on your suggestions I've refactored the initial name param restrictions in commit https://github.com/knative/client/pull/913/commits/d1c8059950ff610955056fa40eceb6f646e0439e.

Please note that I've renamed initial flag name to `-f, --filename` per Michael's suggestion and also to match `kubectl` naming.

Now it's possible to use with [extended conditions](https://github.com/knative/client/pull/913#issuecomment-653435884) suggested by Roland.
```
kn service create --filename ...
kn service create NAME --filename ... 

kn service create -f ..
kn service create NAME -f ... 
```

pkg/kn/commands/service/create.go
dsimansk: Done in https://github.com/knative/client/pull/913/commits/9a3c1b92bd85e16386ee9613296481877f774cb6

test/e2e/service_file_test.go
dsimansk: Done in https://github.com/knative/client/pull/913/commits/9a3c1b92bd85e16386ee9613296481877f774cb6.

pkg/kn/commands/service/configuration_edit_flags.go
dsimansk: Changed the name in https://github.com/knative/client/pull/913/commits/9a3c1b92bd85e16386ee9613296481877f774cb6.

pkg/kn/commands/service/create.go
dsimansk:
```
➜  client git:(create-from-file) ✗ cat /tmp/hello.yaml
kind: Service
metadata:
  name: helloworld-go
spec:
  template:
    spec:
      containers:
        - image: gcr.io/knative-samples/helloworld-go
          env:
            - name: TARGET
              value: "Go Sample v1"
➜  client git:(create-from-file) ✗ kn service create helloworld-go -f /tmp/hello.yaml 
Creating service 'helloworld-go' in namespace 'default':
  0.028s The Route is still working to reflect the latest desired specification.
  0.061s Configuration "helloworld-go" is waiting for a Revision to become ready.
  0.106s ...
  3.751s ...
  3.795s Ingress has not yet been reconciled.
  3.868s Waiting for load balancer to be ready
  4.190s Ready to serve.
Service 'helloworld-go' created to latest revision 'helloworld-go-5j5mp' is available at URL:
http://helloworld-go.default.example.com
➜  client git:(create-from-file) ✗ kn service create helloworld-go -f /tmp/hello.yaml --force
Replacing service 'helloworld-go' in namespace 'default':
```

When I was testing the flag locally I've experienced the above behaviour. When trying to -recreate the service with `--force` the update hangs and timeouts eventually. However it's not happening for `kn service update` flow or imperative `kn service create` due to the fact that we generate revision name by default. May it a serving bug?

maximilien:
@rhuss will you take a pass? I can do a final review tomorrow after you or if you don’t? I like your suggestion.

pkg/kn/commands/service/create.go
rhuss: I think for the namespace we should just allow an override from the CLI. A typical use case is that you create your service maybe for a different environment in the file but want to apply it here in a different context (that's different from the "name" which is an identifier that does not change across environments).

pkg/kn/commands/service/create.go
rhuss: yes could be, but we should be sure what it actually happen when using `--force`. Could you do a debug session and check where exactly it hangs ?

For opening an issue for serving we would need to recreate the steps via yaml files, I think.

pkg/kn/commands/service/create.go
```diff
@@ -260,3 +270,42 @@ func constructService(cmd *cobra.Command, editFlags ConfigurationEditFlags, name
...
+	if err != nil {
+		return nil, err
+	}
+	service.Spec.Template.Name = revName
+
```
rhuss, line 309: Where does the override from all other options from the CLI happen ? I mean when someone uses

```
kn service create -f mysvc.yml --service-account mysa
```

then `mysa` should be set as service account in the generated service. For array/map like options that's a bit more tricky as we have to decided whether they override those or just work on the ones from the file like with an update (i would opt for the latter). I.e. I would think adding options in addition to the file would act like an update as if the service in the file has been already present on the cluster.

wdyt ?

As this looks a bit more involved I would be happy to add this in a second PR.

pkg/kn/commands/service/create.go
dsimansk: Well, the issue is limited to a corner case when the same source is used twice, in other words to replace the same already created same Service. In that case such an update op never replaces the original Service content, causing no set of event to be produced in the cluster for wait to finish without timing out.
There's only one synthetic event `ADDED` [received](https://github.com/knative/client/blob/master/pkg/wait/wait_for_ready.go#L179), but nothing else as the server-side doesn't update anything at all.
The same behaviour can be reproduced with `kubectl replace`, meaning that original service isn't actually replaced, is still the same with original timestamp etc., but there's no wait for ready that could be blocking the execution.

That kind of issue isn't present in imperative `create` flow due to the fact that [generated revision name](https://github.com/knative/client/blob/master/pkg/kn/commands/service/configuration_edit_flags.go#L334-L341) is enough of a change to trigger the update/replace mechanism, therefore desired set of events is received in wait function.


I'd suggest the following:

- Implement @rhuss's suggestion from [this comment  ](https://github.com/knative/client/pull/913#discussion_r450685124)
- Mark `filename` flag with `markFlagMakesRevision` to enforce revision generation
  - Adding modification from other flag would always trigger new revision name generation, therefore we'll benefit from it for filename as well (although it might be seen as hack as well)
- => Profit.

pkg/kn/commands/service/create.go
dsimansk: Initially I've left it out, because of potential clashes. However from my limited tests so far it looks well good. E.g. env vars are merged, service account added etc. I'll add a few tests to cover those use cases and update the PR. Plus there's another benefit [outlined here](https://github.com/knative/client/pull/913#discussion_r451481222).

pkg/kn/commands/service/create.go
rhuss: For a quick win, I would go for the second option. We then can implement that improvement to pick up other options later. wdyt ?

dsimansk: OK, I'm on board with quick win. :) Anyway it'll better to have improvement PR with proper amount of tests and bake time for options combination. I'll create a new issue for it.

dsimansk: Follow-up enhancement issue https://github.com/knative/client/issues/923.

maximilien:
/hold 
/lgtm

@rhuss feel free to remove hold when ready

PR #914, opened 2020-07-02
Added the option to use --scale for setting MinScale and MaxScale to the same value

### LINKED ISSUE #822: support --scale
I find myself often needing to do: `--min-scale=X --max-scale=x`, usually when x=1.

It might be nice to support a short-cut here: `--scale=x` to mean "set both min & max".

Just a thought...

mpetason:
## Description

This allows a user to set --scale which will update MinScale and MaxScale to the same value. 

## Changes

* add --scale as a CLI option

## Reference

#822

### DIFF
--- a/docs/cmd/kn_service_update.md
+++ b/docs/cmd/kn_service_update.md
@@ -71,7 +71,7 @@ kn service update NAME
       --requests-cpu string           DEPRECATED: please use --request instead. The requested CPU (e.g., 250m).
       --requests-memory string        DEPRECATED: please use --request instead. The requested memory (e.g., 64Mi).
       --revision-name string          The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
-      --scale int                     Minimal and Maximal number of replicas.
+      --scale int                     Minimum and maximum number of replicas.
       --service-account string        Service account name to set. An empty argument ("") clears the service account. The referenced service account must exist in the service's namespace.
       --tag strings                   Set tag (format: --tag revisionRef=tagName) where revisionRef can be a revision or '@latest' string representing latest ready revision. This flag can be specified multiple times.
       --traffic strings               Set traffic distribution (format: --traffic revisionRef=percent) where revisionRef can be a revision or a tag or '@latest' string representing latest ready revision. This flag can be given multiple times with percent summing up to 100%.
--- a/docs/cmd/kn_service_update.md
+++ b/docs/cmd/kn_service_update.md
@@ -71,7 +71,7 @@ kn service update NAME
       --requests-cpu string           DEPRECATED: please use --request instead. The requested CPU (e.g., 250m).
       --requests-memory string        DEPRECATED: please use --request instead. The requested memory (e.g., 64Mi).
       --revision-name string          The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
-      --scale                         Set min and max scale as the same value.
+      --scale int                     Minimal and Maximal number of replicas.
       --service-account string        Service account name to set. An empty argument ("") clears the service account. The referenced service account must exist in the service's namespace.
       --tag strings                   Set tag (format: --tag revisionRef=tagName) where revisionRef can be a revision or '@latest' string representing latest ready revision. This flag can be specified multiple times.
       --traffic strings               Set traffic distribution (format: --traffic revisionRef=percent) where revisionRef can be a revision or a tag or '@latest' string representing latest ready revision. This flag can be given multiple times with percent summing up to 100%.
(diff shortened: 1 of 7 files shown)


test/e2e/service_options_test.go
```diff
@@ -72,6 +72,11 @@ func TestServiceOptions(t *testing.T) {
 	validateServiceMinScale(r, "svc2", "1")
 	validateServiceMaxScale(r, "svc2", "3")
 
+	t.Log("create and validate service with scale options ")
+	serviceCreateWithOptions(r, "svc2", "--scale", "5")
+	validateServiceMinScale(r, "svc2", "5")
+	validateServiceMaxScale(r, "svc2", "5")
+
 	t.Log("update and validate service with max scale option")
 	test.ServiceUpdate(r, "svc2", "--max-scale", "2")
```
navidshaikh, line 81: e2e failure:
service `svc2` exists (created at line no 73), try updating the service or create a fresh one after line:85

mpetason: Thank you for the help. I missed that.

docs/cmd/kn_service_update.md
```diff
@@ -71,6 +71,7 @@ kn service update NAME
       --requests-cpu string           DEPRECATED: please use --request instead. The requested CPU (e.g., 250m).
       --requests-memory string        DEPRECATED: please use --request instead. The requested memory (e.g., 64Mi).
       --revision-name string          The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
+      --scale int                     Minimal and Maximal number of replicas.
```
maximilien, line 74: Wonder if “Minimum” and “maximum” are better words here?

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -434,6 +438,17 @@ func (p *ConfigurationEditFlags) Apply(
 		}
 	}
 
+	if cmd.Flags().Changed("scale") {
```
maximilien, line 441: I can comment lines 443-445 and/or lines 447-449 and all tests still pass.

This tells me you don’t have tests for `servinglib.UpdateMaxScale(...)` and `servinglib.UpdateMinScale(...)` failings. Might be easy to cover with a test with negative or 0 scale (perhaps).

test/e2e/service_options_test.go
```diff
@@ -72,6 +72,11 @@ func TestServiceOptions(t *testing.T) {
 	validateServiceMinScale(r, "svc2", "1")
 	validateServiceMaxScale(r, "svc2", "3")
 
+	t.Log("create and validate service with scale options ")
+	serviceCreateWithOptions(r, "svc2a", "--scale", "5")
```
maximilien, line 76: I wonder if you want to try to change scale value for both direction up and down... Thanks.

docs/cmd/kn_service_update.md
rhuss: :+1: (if I look at the result of Google hits for "minimum number" and "minimal number"). Also please use lower case, something like

"Minimum and maximum number of replicas"

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -185,6 +186,9 @@ func (p *ConfigurationEditFlags) addSharedFlags(command *cobra.Command) {
 	command.Flags().IntVar(&p.MaxScale, "max-scale", 0, "Maximal number of replicas.")
 	p.markFlagMakesRevision("max-scale")
 
+	command.Flags().IntVar(&p.Scale, "scale", 0, "Minimal and Maximal number of replicas.")
```
rhuss, line 189:
```suggestion
	command.Flags().IntVar(&p.Scale, "scale", 0, "Minimum and maximum number of replicas.")
```

pkg/kn/commands/service/create_test.go
```diff
@@ -544,6 +544,43 @@ func TestServiceCreateMaxMinScale(t *testing.T) {
 	}
 }
 
+func TestServiceCreateScale(t *testing.T) {
+	action, created, _, err := fakeServiceCreate([]string{
+		"service", "create", "foo", "--image", "gcr.io/foo/bar:baz",
+		"--scale", "5",
+		"--concurrency-target", "10", "--concurrency-limit", "100",
```
rhuss, line 551: Why are these options important here ? I would just focus on `--scale` to make this test more focussed and easier to maintain (e.g. when something around concurrency-target would change in the future)

pkg/kn/commands/service/update_test.go
```diff
@@ -355,6 +355,46 @@ func TestServiceUpdateMaxMinScale(t *testing.T) {
 
 }
 
+func TestServiceUpdateScale(t *testing.T) {
+	original := newEmptyService()
+
+	action, updated, _, err := fakeServiceUpdate(original, []string{
+		"service", "update", "foo",
+		"--scale", "5", "--concurrency-target", "10", "--concurrency-limit", "100", "--concurrency-utilization", "50", "--no-wait"})
```
rhuss, line 363: Same as above, please only test for `--scale`

pkg/kn/commands/service/create_test.go
rhuss: As mentioned by @maximilien you should also add a test for the error case, i.e. when `--scale` is negative.

rhuss: And also a test for what happens when `--scale` and `--min-scale` & `--max-scale` are all provided (the more specific options should take precedence or an error should be thrown if both set of options are used)

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -434,6 +438,17 @@ func (p *ConfigurationEditFlags) Apply(
 		}
 	}
 
+	if cmd.Flags().Changed("scale") {
+		err = servinglib.UpdateMaxScale(template, p.Scale)
```
rhuss, line 442: What is if `--max-scale` is also given on the command line ? Which one takes precedence ? Or should an error be thrown ? 

IMO we should return an error if both `--scale` and either `--min-scale` or `--max-scale` is provied.

mpetason: I need to add this functionality and also tests for it.

docs/cmd/kn_service_update.md
mpetason: I had "Set min and max scale as the same value" but something recommended those changes when I submitted the first PR? I'll make this update.

mpetason:
@rhuss this doesn't "fix" the other issue, but I'm not sure how to remove the "fixes" unless I can just edit the comment. I will update this to range in another PR, just wanted to get the functionality in and learn how the client works with this PR.

mpetason:
Still work in progress. I need to address the @maximilien issue + testing.

mpetason:
I think I have addressed all of the comments/change requests. Ready for another review.

PR #917, opened 2020-07-07
Use broker injection label key from trigger types

navidshaikh:
## Description
Uses the broker injection key from eventing  and not hard code, as there is a change coming it in eventing v0.16 release with sugar controller.

## Reference
see https://github.com/knative/eventing/pull/3459


navidshaikh:
logged https://github.com/knative/client/issues/918 to ensure we change the key name for v0.16 release for backward compatibility.

PR #919, opened 2020-07-07
Fixed the error message for min/max scale tests

mpetason:
## Description

Fixed an error message in the create_test.go file

## Changes

* Changed 1000 to 100 to match the value used in the test.


mpetason:
I'm not sure why the test if failing, i only updated two values in an error message.

PR #920, opened 2020-07-07
Setup eventing extension: sugar controller

navidshaikh:
## Description
 We'll need sugar controller for reconciling broker for tests
 which labels namespace and annotates the triggers.

 Currently, sugar controller is not part of the eventing nightly release setup
 by test-infra libs. This changeset adds install of same for nightly
 CI setup.


n3wscott:
thanks, and sorry

PR #937, opened 2020-07-16
feat: Add support to combine service create --filename with other options

### LINKED ISSUE #923: Add support to combine service create --filename with other options
### Feature request

Per PR #913 review [discussion](https://github.com/knative/client/pull/913#discussion_r451481222) and [comment](https://github.com/knative/client/pull/913#discussion_r450685124). There's a possibility to extend the initial functionality of `kn service create --filename` with other options. In a nutshell to view provided file as a baseline that can be further modified but other option flags.
There should be a reasonable amount of tests added to verify common usage pattern works, e.g. command line options are reflected, take priority and can override existing values, lists are merged etc.

### Use case

Create a service from file and add env variable:
```
kn service create mysvc --filename mysvc.yaml --env VERSION="v2"
```

Create a service from file and specify service account:
```
kn service create mysvc --filename mysvc.yaml --service-account specialSA
```

#### Implementation notes

The following code block should be added and suggestion points from [this PR](https://github.com/knative/client/pull/913#discussion_r451481222) reflected.

```go
err = editFlags.Apply(&service, nil, cmd)
if err != nil {
    return nil, err
}
```

dsimansk:
## Description

Add enhancement to the command `kn service create -f` to accept other options to modify the created service.

Per discussion in #923 we should decide how `name` is handled:
- Provided cmdline `name` overrides the name in the file
- Keep the current implementation that if `name` is provided on cmdline it must match the name in the file.

## Changes

*  Add support to combine service create --filename with other options

## Reference

Fixes #923

### DIFF
--- a/docs/cmd/kn_service_create.md
+++ b/docs/cmd/kn_service_create.md
@@ -64,7 +64,7 @@ kn service create NAME --image IMAGE
       --concurrency-utilization int   Percentage of concurrent requests utilization before scaling up. (default 70)
   -e, --env stringArray               Environment variable to set. NAME=value; you may provide this flag any number of times to set multiple environment variables. To unset, specify the environment variable name followed by a "-" (e.g., NAME-).
       --env-from stringArray          Add environment variables from a ConfigMap (prefix cm: or config-map:) or a Secret (prefix secret:). Example: --env-from cm:myconfigmap or --env-from secret:mysecret. You can use this flag multiple times. To unset a ConfigMap/Secret reference, append "-" to the name, e.g. --env-from cm:myconfigmap-.
-  -f, --filename string               Create a service from file.
+  -f, --filename string               Create a service from file. The created service can be futher modified by combining with other options.Example: -f /path/to/file --env NAME=value will also add environment variable.
       --force                         Create service forcefully, replaces existing service if any.
   -h, --help                          help for create
       --image string                  Image to run.
--- a/docs/cmd/kn_service_create.md
+++ b/docs/cmd/kn_service_create.md
@@ -64,7 +64,7 @@ kn service create NAME --image IMAGE
       --concurrency-utilization int   Percentage of concurrent requests utilization before scaling up. (default 70)
   -e, --env stringArray               Environment variable to set. NAME=value; you may provide this flag any number of times to set multiple environment variables. To unset, specify the environment variable name followed by a "-" (e.g., NAME-).
       --env-from stringArray          Add environment variables from a ConfigMap (prefix cm: or config-map:) or a Secret (prefix secret:). Example: --env-from cm:myconfigmap or --env-from secret:mysecret. You can use this flag multiple times. To unset a ConfigMap/Secret reference, append "-" to the name, e.g. --env-from cm:myconfigmap-.
-  -f, --filename string               Create a service from file. The created service can be further modified by combining with other options.Example: -f /path/to/file --env NAME=value will also add environment variable.
+  -f, --filename string               Create a service from file. The created service can be further modified by combining with other options. For example -f /path/to/file --env NAME=value will also add environment variable.
       --force                         Create service forcefully, replaces existing service if any.
   -h, --help                          help for create
       --image string                  Image to run.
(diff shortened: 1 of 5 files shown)


docs/cmd/kn_service_create.md
```diff
@@ -64,7 +64,7 @@ kn service create NAME --image IMAGE
       --concurrency-utilization int   Percentage of concurrent requests utilization before scaling up. (default 70)
   -e, --env stringArray               Environment variable to set. NAME=value; you may provide this flag any number of times to set multiple environment variables. To unset, specify the environment variable name followed by a "-" (e.g., NAME-).
       --env-from stringArray          Add environment variables from a ConfigMap (prefix cm: or config-map:) or a Secret (prefix secret:). Example: --env-from cm:myconfigmap or --env-from secret:mysecret. You can use this flag multiple times. To unset a ConfigMap/Secret reference, append "-" to the name, e.g. --env-from cm:myconfigmap-.
-  -f, --filename string               Create a service from file.
+  -f, --filename string               Create a service from file. The created service can be further modified by combining with other options.Example: -f /path/to/file --env NAME=value will also add environment variable.
```
maximilien, line 67: "options.Example:" => "options, e.g.," OR "options. For example:"

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -278,8 +278,11 @@ func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
 	p.addSharedFlags(command)
 	command.Flags().BoolVar(&p.ForceCreate, "force", false,
 		"Create service forcefully, replaces existing service if any.")
-	command.Flags().StringVarP(&p.Filename, "filename", "f", "", "Create a service from file.")
+	command.Flags().StringVarP(&p.Filename, "filename", "f", "", "Create a service from file. "+
+		"The created service can be further modified by combining with other options."+
+		"Example: -f /path/to/file --env NAME=value will also add environment variable.")
```
maximilien, line 283: Same as above. Use same resolution

pkg/kn/commands/service/create.go
```diff
@@ -303,12 +303,11 @@ func constructServiceFromFile(cmd *cobra.Command, editFlags ConfigurationEditFla
 	// Set namespace in case it's specified as --namespace
 	service.ObjectMeta.Namespace = namespace
 
-	// We need to generate revision to have --force replace working
-	revName, err := servinglib.GenerateRevisionName(editFlags.RevisionName, &service)
+	// Apply options provided from cmdline
+	err = editFlags.Apply(&service, nil, cmd)
 	if err != nil {
 		return nil, err
 	}
```
maximilien, line 310: I could comment these lines and all tests are still passing. So this tells me no test for errors when `editFlags.Apply(...)`

rhuss: But isn't `editFlags.Apply()` already covered ? Ideally, there is a dedicated unit test only for this method, and should not be tested indirectly within tests that test that intend to test other stuff.

rhuss: If I look at the overall test suite I see, that `editFlags.Apply()` is already tested to a high degree. Looking at that method though it looks like that it deserves some refactoring, it is far too large.

docs/cmd/kn_service_create.md
rhuss: "also add environment variable" --> "also add an environment variable"

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -278,8 +278,11 @@ func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
 	p.addSharedFlags(command)
 	command.Flags().BoolVar(&p.ForceCreate, "force", false,
 		"Create service forcefully, replaces existing service if any.")
-	command.Flags().StringVarP(&p.Filename, "filename", "f", "", "Create a service from file.")
+	command.Flags().StringVarP(&p.Filename, "filename", "f", "", "Create a service from file. "+
+		"The created service can be further modified by combining with other options. "+
+		"For example -f /path/to/file --env NAME=value will also add environment variable.")
```
rhuss, line 283:
```suggestion
		"For example, -f /path/to/file --env NAME=value adds also an environment variable.")
```

pkg/kn/commands/service/create.go
```diff
@@ -303,12 +303,11 @@ func constructServiceFromFile(cmd *cobra.Command, editFlags ConfigurationEditFla
 	// Set namespace in case it's specified as --namespace
 	service.ObjectMeta.Namespace = namespace
 
-	// We need to generate revision to have --force replace working
-	revName, err := servinglib.GenerateRevisionName(editFlags.RevisionName, &service)
```
rhuss, line 307: Why has this been removed ?

dsimansk: It's superseded by combination of `markFlagMakesRevision` and `editFlags.Apply` combination, per my suggestion/proposal from [previous PR](https://github.com/knative/client/pull/913#discussion_r451481222).

https://github.com/knative/client/pull/937/files#diff-74b189b0b506d256ede4322475d950f5R285

rhuss: I see, and I think it makes sense.

PR #944, opened 2020-07-21
Upgrade spf13/cobra dep to current master

navidshaikh:
## Description
- Pin cobra at  v1.0.1-0.20200715031239-b95db644ed1c
- Parse args without invoking a separate command
  
     Do not define and execute extractCommand before running actual root
     command to parse all the args without flags. This was creating issues
     with completion utils to generate additional shell completion directive.
     Now uses cobra's inbuilt utilities to parse the command args without flags.


cmd/kn/main.go
```diff
@@ -64,21 +64,26 @@ func run(args []string) error {
...
 
-	// Strip of all flags to get the non-flag commands only
-	commands, err := stripFlags(args)
+	pluginManager := plugin.NewManager(config.GlobalConfig.PluginsDir(), config.GlobalConfig.LookupPluginsInPath())
+
+	// Create kn root command and all sub-commands
+	rootCmd, err := root.NewRootCommand(pluginManager.HelpTemplateFuncs())
 	if err != nil {
 		return err
 	}
```
maximilien, line 73: This error is not covered in tests...

cmd/kn/main.go
```diff
@@ -64,21 +64,26 @@ func run(args []string) error {
...
-	// Find plugin with the commands arguments
-	pluginManager := plugin.NewManager(config.GlobalConfig.PluginsDir(), config.GlobalConfig.LookupPluginsInPath())
-	plugin, err := pluginManager.FindPlugin(commands)
+	// temporary setting to parse all flags
+	rootCmd.FParseErrWhitelist = cobra.FParseErrWhitelist{UnknownFlags: true}
+	// Strip of all flags to get the non-flag commands only
+	commands, err := stripFlags(rootCmd, args)
 	if err != nil {
 		return err
 	}
```
maximilien, line 81: I can comment and all tests pass. Since `stripFlags` is added by you, perhaps a test is warranted?

cmd/kn/main.go
```diff
@@ -64,21 +64,26 @@ func run(args []string) error {
...
+	// reset the temporary setting
+	rootCmd.FParseErrWhitelist = cobra.FParseErrWhitelist{UnknownFlags: false}
 
-	// Create kn root command and all sub-commands
-	rootCmd, err := root.NewRootCommand(pluginManager.HelpTemplateFuncs())
+	// Find plugin with the commands arguments
+	plugin, err := pluginManager.FindPlugin(commands)
 	if err != nil {
 		return err
 	}
```
maximilien, line 89: Also not covered

cmd/kn/main.go
navidshaikh: thats for creating rootCommand, while we could add (duplicate) test for it here I wonder if its worth as its tested in `pkg/kn/root/root_test.go`.

cmd/kn/main.go
navidshaikh: `stripFlags` been around, this PR changes the implementation of it, the function explicitly takes argument with (temporary) settings added on the command as `rootCmd.FParseErrWhitelist = cobra.FParseErrWhitelist{UnknownFlags: true}` to allow passing unknown flags (unknown to core kn, to provision plugins). I will try and add a case to cover this. thanks!

cmd/kn/main.go
navidshaikh: not changed in the PR.

cmd/kn/main.go
```diff
@@ -106,46 +111,12 @@ func run(args []string) error {
...
-	extractCommand.FParseErrWhitelist = cobra.FParseErrWhitelist{UnknownFlags: true}
-
-	// Execute to get to the command args
-	err := extractCommand.Execute()
-	if err != nil {
-		return nil, err
+// Get only the args provided but no options
+func stripFlags(rootCmd *cobra.Command, args []string) ([]string, error) {
+	if err := rootCmd.ParseFlags(filterHelpOptions(args)); err != nil {
+		fmt.Errorf("error while parsing flags from args %v: %s", args, err.Error())
```
daisy-ycguo, line 117: Do you need a return statement here ? What's the usage of the error object created by `fmt.Errorf` ? I guess maybe you want to:
```
return nil, fmt.Errorf("error while parsing flags from args %v: %s", args, err.Error())
```

navidshaikh: good catch, fixed.

cmd/kn/main.go
navidshaikh: Added a unit test without aforementioned temp setting on command under test.

navidshaikh:
/retest

          result_collector.go:75: ERROR: Error: ReconcileIngressFailed: Ingress reconciliation failed

maximilien:
OK thanks.

/lgtm

PR #948, opened 2020-07-23
List inbuilt sources if CRD access is restricted

### LINKED ISSUE #947: kn source list and list-types return error if CRD access is restricted
### Bug report
`kn source list` and `kn source list-types` uses dynamic client and a common operation is to query CRDs and filter on a particular label. It finds all the sources CRDs with label `duck.knative.dev/source: true`, and able to list eventing inbuilt and non inbuilt sources.
A user of cluster might be restricted to list the CRDs in the cluster causing `kn source list` and `kn source list-types` commands to fail, though the same user is able to read the particular GVK of the source (kn source ping list).
We can identify these restricted access error and fallback to listing only eventing inbuilt sources as we know their GVKs.

navidshaikh:
## Description
- Identify restricted access error
- If server returns restricted access error, fallback to listing
  only eventing inbuilt sources using their GVKs.
- List every typed source and read the error
  to know if eventing is installed for `kn source list-types`.

## Changes
-  Add `isForbiddenError` in kn error factory
-  Add `ListSourcesUsingGVKs` to dynamic client
-  Add  `BuiltInSourcesGVKs` to `pkg/sources/v1alpha2/client.go` which returns inbuilt source GVKs

## Reference
Fixes #947 

/lint

### DIFF
--- a/pkg/dynamic/client.go
+++ b/pkg/dynamic/client.go
@@ -15,7 +15,7 @@
 package dynamic
 
 import (
-	"fmt"
+	"errors"
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -121,7 +121,7 @@ func (c *knDynamicClient) ListSources(types ...WithType) (*unstructured.Unstruct
 	}
 
 	if sourceTypes == nil || len(sourceTypes.Items) == 0 {
-		return nil, fmt.Errorf("404: no sources found on the backend, please verify the installation")
+		return nil, errors.New("no sources found on the backend, please verify the installation")
 	}
 
 	namespace := c.Namespace()
--- a/pkg/dynamic/client.go
+++ b/pkg/dynamic/client.go
@@ -15,6 +15,8 @@
 package dynamic
 
 import (
+	"strings"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
@@ -47,6 +49,9 @@ type KnDynamicClient interface {
 
 	// ListSources returns list of available source objects
 	ListSources(types ...WithType) (*unstructured.UnstructuredList, error)
+
+	// ListSources returns list of available source objects using given list of GVKs
+	ListSourcesUsingGVKs(*[]schema.GroupVersionKind, ...WithType) (*unstructured.UnstructuredList, error)
 
 	// RawClient returns the raw dynamic client interface
 	RawClient() dynamic.Interface
@@ -107,7 +112,7 @@ func (c *knDynamicClient) ListSources(types ...WithType) (*unstructured.Unstruct
 	var (
 		sourceList               unstructured.UnstructuredList
 		options                  metav1.ListOptions
-		numberOfsourceTypesFound int
+		numberOfSourceTypesFound int
 	)
 	sourceTypes, err := c.ListSourcesTypes()
 	if err != nil {
@@ -141,14 +146,56 @@ func (c *knDynamicClient) ListSources(types ...WithType) (*unstructured.Unstruct
 
 		if len(sList.Items) > 0 {
 			// keep a track if we found source objects of different types
-			numberOfsourceTypesFound++
+			numberOfSourceTypesFound++
 			sourceList.Items = append(sourceList.Items, sList.Items...)
 			sourceList.SetGroupVersionKind(sList.GetObjectKind().GroupVersionKind())
 		}
 	}
 	// Clear the Group and Version for list if there are multiple types of source objects found
 	// Keep the source's GVK if there is only one type of source objects found or requested via --type filter
-	if numberOfsourceTypesFound > 1 {
+	if numberOfSourceTypesFound > 1 {
+		sourceList.SetGroupVersionKind(schema.GroupVersionKind{Group: "", Version: "", Kind: "List"})
+	}
+	return &sourceList, nil
+}
+
+// ListSources returns list of available source objects using given list of GVKs
+func (c *knDynamicClient) ListSourcesUsingGVKs(gvks *[]schema.GroupVersionKind, types ...WithType) (*unstructured.UnstructuredList, error) {
+	if gvks == nil {
+		return nil, nil
+	}
+
+	var (
+		sourceList               unstructured.UnstructuredList
+		options                  metav1.ListOptions
+		numberOfSourceTypesFound int
+	)
+	namespace := c.Namespace()
+	filters := WithTypes(types).List()
+
+	for _, gvk := range *gvks {
+		if len(filters) > 0 && !util.SliceContainsIgnoreCase(filters, gvk.Kind) {
+			continue
+		}
+
+		gvr := gvk.GroupVersion().WithResource(strings.ToLower(gvk.Kind) + "s")
+
+		// list objects of source type with this GVR
+		sList, err := c.client.Resource(gvr).Namespace(namespace).List(options)
+		if err != nil {
+			return nil, err
+		}
+
+		if len(sList.Items) > 0 {
+			// keep a track if we found source objects of different types
+			numberOfSourceTypesFound++
+			sourceList.Items = append(sourceList.Items, sList.Items...)
+			sourceList.SetGroupVersionKind(sList.GetObjectKind().GroupVersionKind())
+		}
+	}
+	// Clear the Group and Version for list if there are multiple types of source objects found
+	// Keep the source's GVK if there is only one type of source objects found or requested via --type filter
+	if numberOfSourceTypesFound > 1 {
 		sourceList.SetGroupVersionKind(schema.GroupVersionKind{Group: "", Version: "", Kind: "List"})
 	}
 	return &sourceList, nil
--- a/pkg/dynamic/client.go
+++ b/pkg/dynamic/client.go
@@ -15,6 +15,7 @@
 package dynamic
 
 import (
+	"fmt"
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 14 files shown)


pkg/errors/factory.go
```diff
@@ -38,6 +39,10 @@ func isEmptyConfigError(err error) bool {
 	return strings.Contains(err.Error(), "no configuration has been provided")
 }
 
+func IsForbiddenError(status api_errors.APIStatus) bool {
```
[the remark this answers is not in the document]
navidshaikh: changed to unexported method

navidshaikh:
@rhuss @maximilien @daisy-ycguo: I will add the tests if the approach looks good.

pkg/kn/commands/source/list.go
```diff
@@ -58,8 +61,15 @@ func NewListCommand(p *commands.KnParams) *cobra.Command {
 			}
 			sourceList, err := dynamicClient.ListSources(filters...)
 			if err != nil {
-				return err
+				if strings.HasPrefix(knerrors.GetError(err).Error(), "403") {
```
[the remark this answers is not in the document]
navidshaikh: We are processing the error at https://github.com/knative/client/pull/948/files#diff-1a1baf9d1fe0b4fa3d8d4e515233df48R42 where we're prefixing api status code (403 in case of forbidden) to message.
Though we can change harcoded `403` here and fetch that from lib.

pkg/dynamic/client.go
```diff
@@ -141,14 +146,56 @@ func (c *knDynamicClient) ListSources(types ...WithType) (*unstructured.Unstruct
...
-			numberOfsourceTypesFound++
+			numberOfSourceTypesFound++
+			sourceList.Items = append(sourceList.Items, sList.Items...)
+			sourceList.SetGroupVersionKind(sList.GetObjectKind().GroupVersionKind())
+		}
+	}
+	// Clear the Group and Version for list if there are multiple types of source objects found
+	// Keep the source's GVK if there is only one type of source objects found or requested via --type filter
+	if numberOfSourceTypesFound > 1 {
+		sourceList.SetGroupVersionKind(schema.GroupVersionKind{Group: "", Version: "", Kind: "List"})
```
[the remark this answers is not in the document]
navidshaikh: This was done to show proper values if json or yaml format of list is requested and the list contains multiple types of source objects.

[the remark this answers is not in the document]
navidshaikh: We'd need GVK set for the (list) object here as (json/yaml) printers require it. The question was which version/group to set if there are different types of sources found, so we set cleared group and version values and set `List` for kind. 
If there is only one type of source object found OR user requested single type using `--type` filter, we keep the GVK intact.

pkg/dynamic/client.go
```diff
@@ -107,12 +113,17 @@ func (c *knDynamicClient) ListSources(types ...WithType) (*unstructured.Unstruct
...
 		options                  metav1.ListOptions
-		numberOfsourceTypesFound int
+		numberOfSourceTypesFound int
 	)
 	sourceTypes, err := c.ListSourcesTypes()
 	if err != nil {
 		return nil, err
 	}
+
+	if sourceTypes == nil || len(sourceTypes.Items) == 0 {
```
maximilien, line 123: I can comment this condition (both parts around ||) and still pass all tests...

pkg/kn/commands/source/list.go
```diff
@@ -52,15 +54,25 @@ func NewListCommand(p *commands.KnParams) *cobra.Command {
...
-			if err != nil {
-				return err
+
+			switch {
+			case knerrors.IsForbiddenError(err):
+				gvks := sourcesv1alpha2.BuiltInSourcesGVKs()
+				if sourceList, err = dynamicClient.ListSourcesUsingGVKs(&gvks, filters...); err != nil {
+					return knerrors.GetError(err)
+				}
+			case err != nil:
```
maximilien, line 71: I can comment lines 71-73 and still pass all tests...

pkg/kn/commands/source/list_types.go
```diff
@@ -47,13 +52,17 @@ func NewListTypesCommand(p *commands.KnParams) *cobra.Command {
 			}
 
 			sourceListTypes, err := dynamicClient.ListSourcesTypes()
-			if err != nil {
-				return err
+			switch {
+			case knerrors.IsForbiddenError(err):
```
maximilien, line 56: I can comment lines 56-59 and still pass all tests

pkg/kn/commands/source/list_types.go
```diff
@@ -47,13 +52,17 @@ func NewListTypesCommand(p *commands.KnParams) *cobra.Command {
...
+					return knerrors.GetError(err)
+				}
+			case err != nil:
+				return knerrors.GetError(err)
 			}
 
-			if len(sourceListTypes.Items) == 0 {
-				fmt.Fprintf(cmd.OutOrStdout(), "No sources found.\n")
-				return nil
+			if sourceListTypes == nil || len(sourceListTypes.Items) == 0 {
```
maximilien, line 64: I can comment either parts around the || and still have all tests green...

pkg/dynamic/client.go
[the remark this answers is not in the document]
rhuss: Yes, but imagine the situation that by accident you have only one source of one type. If you list those, you the GVK set. Some time later another source of another type is added. Suddenly know the GVK gets removed or changed to an artificial "List". I think as we are dealing with a heterogenous list, that, by accident, can be also homogenous, we should treat it as a heterogenous list always. I would not mind to introduce a new type here, like we did for the `Export`, with an `client.knative.dev` group and a `v1alpha1` version for now, but then use it all the time.

pkg/kn/commands/source/list_types.go
navidshaikh: added https://github.com/knative/client/issues/952 to track adding mock client for dynamic client, with which we should be able to pass the errors to expect and verify this in unit tests, can be added in a separate PR, sounds good?

pkg/kn/commands/source/list.go
navidshaikh: added tests

pkg/dynamic/client.go
navidshaikh: added tests

pkg/kn/commands/source/list_types.go
navidshaikh: added tests

pkg/dynamic/client.go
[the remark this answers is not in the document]
navidshaikh: > I would not mind to introduce a new type here, like we did for the Export, with an client.knative.dev group and a v1alpha1 version for now, but then use it all the time.

yea, we could do that, however for the other point about list of single source type, I think setting exact source type for single-type-of-source list is explicit and we could scope the client custom type for list *only* if there multiple types of sources.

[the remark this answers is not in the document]
navidshaikh: Added https://github.com/knative/client/issues/953 to track adding setting this custom client type for list. We can refine in next iteration.

pkg/dynamic/client.go
```diff
@@ -141,14 +152,56 @@ func (c *knDynamicClient) ListSources(types ...WithType) (*unstructured.Unstruct
...
+
+	var (
+		sourceList               unstructured.UnstructuredList
+		options                  metav1.ListOptions
+		numberOfSourceTypesFound int
+	)
+	namespace := c.Namespace()
+	filters := WithTypes(types).List()
+
+	for _, gvk := range *gvks {
```
daisy-ycguo, line 182: Just wonder what's the better behavior when `gvks` is `nil`: a) return a `nil` list or b) ignore `gvks` and use `types ...WithType` to filter.
I see when `types` is nil, the behavior is to b) ignore `types` and use `gvks` to filter, not a) return a `nil` list. Maybe we use the same logic to handle either of these two parameters is nil.

pkg/kn/commands/source/list_test.go
```diff
@@ -71,6 +79,20 @@ func TestSourceListTypesNoHeaders(t *testing.T) {
 	assert.Check(t, util.ContainsAll(output[0], "PingSource"))
 }
 
+func TestListBuiltInSources(t *testing.T) {
```
daisy-ycguo, line 82: I will suggest to use `TestListBuiltInSourceTypes` other than `TestListBuiltInSources` because the method is to test listing of builtin source types, other than sources.

pkg/dynamic/client.go
```diff
@@ -141,14 +152,56 @@ func (c *knDynamicClient) ListSources(types ...WithType) (*unstructured.Unstruct
...
+	// Clear the Group and Version for list if there are multiple types of source objects found
+	// Keep the source's GVK if there is only one type of source objects found or requested via --type filter
+	if numberOfSourceTypesFound > 1 {
+		sourceList.SetGroupVersionKind(schema.GroupVersionKind{Group: "", Version: "", Kind: "List"})
+	}
+	return &sourceList, nil
+}
+
+// ListSourcesUsingGVKs returns list of available source objects using given list of GVKs
+func (c *knDynamicClient) ListSourcesUsingGVKs(gvks *[]schema.GroupVersionKind, types ...WithType) (*unstructured.UnstructuredList, error) {
```
daisy-ycguo, line 169: I wonder what is the best function behavior when `gvks` is `nil`: a) to return a `nil` list or b) to ignore `gvks` and use `types` to filter. The current behavior is a).
Yet when `types` is `nil`, the current behavior is b) to ignore `types` and use `gvks` to filter, not a) to return a `nil` list. It makes me a little weird. 
Of course it doesn't affect the current functions. For future usage, maybe change to the same behavior when either of these two parameters is `nil`.

navidshaikh: We use `types` only to subset the found source types. If gvks list is nil, it returns nil. Does this answer your question ?

pkg/kn/commands/source/list_test.go
navidshaikh: updated

navidshaikh:
/retest

    24.839s Ingress has not yet been reconciled.
    Error: timeout: service 'hello' not ready after 600 seconds

navidshaikh:
/retest

    27.413s Ingress has not yet been reconciled.

PR #958, opened 2020-07-28
Deprecates --max-scale/--min-scale for more consistent naming

mpetason:
## Description

This PR updates --max-scale and --min-scale to --scale-max and --scale-min

## Changes

* Updated --max-scale to --scale-max
* Updated --min-scale to --scale-min

## Reference

https://github.com/knative/client/issues/822

### DIFF
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -182,14 +182,14 @@ func (p *ConfigurationEditFlags) addSharedFlags(command *cobra.Command) {
 		"DEPRECATED: please use --limit instead. The limits on the requested memory (e.g., 1024Mi).")
 	p.markFlagMakesRevision("limits-memory")
 
-	command.Flags().IntVar(&p.MinScale, "min-scale", 0, "Minimal number of replicas.")
-	p.markFlagMakesRevision("min-scale")
-
-	command.Flags().IntVar(&p.MaxScale, "max-scale", 0, "Maximal number of replicas.")
-	p.markFlagMakesRevision("max-scale")
-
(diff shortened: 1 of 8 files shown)


mpetason:
This may require additional work. I wasn't sure if I should be changing some of the function names and others to match the renaming.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -182,15 +182,15 @@ func (p *ConfigurationEditFlags) addSharedFlags(command *cobra.Command) {
 		"DEPRECATED: please use --limit instead. The limits on the requested memory (e.g., 1024Mi).")
 	p.markFlagMakesRevision("limits-memory")
 
-	command.Flags().IntVar(&p.MinScale, "min-scale", 0, "Minimal number of replicas.")
-	p.markFlagMakesRevision("min-scale")
-
-	command.Flags().IntVar(&p.MaxScale, "max-scale", 0, "Maximal number of replicas.")
-	p.markFlagMakesRevision("max-scale")
-
```
sixolet, line 190: Can we keep this, but hidden, for back compat?

mpetason: Do you mean the option "min-scale" and "max-scale"? I moved the functionality for scale-min/scale-max below scale so that it would be alphabetical (prob should have put max first.) The functionality for min-scale, max-scale has been removed.

mpetason: I added them back with MarkHidden

mpetason:
I think this is ready for another review. Do I need to resolve previous conversations before that happens?

PR #964, opened 2020-08-03
Feature: "kn service apply"

rhuss:
This implementation reuses the `kubectl.kubernetes.io/last-applied-configuration` for storing the original configuration so that kubectl tooling can be applied, too.

Only client-side patching is implemented as it's not clear to me whether server-side patching is implemented as part of the Knative API contract. 

An initial `kn service apply` for creating a service works as expected, but a subsequent `kn service apply` fails with a patch conflict error

The following challenges are still open:

* As we are using a structured approach when building up a `KService`, when serializing this so a JSON string to be added to the annotation mentioned above, fields that are typed to a struct (not a pointer to struct) are always present. This includes `createdTimestamp` or `resources` in the container spec. `kubectl` which uses always `Unstructured` doesn't suffer from this problem.
* The `name` of a container in the `PodSpec` is marked as mandatory for Kubernetes, but is used optional in Knative (it will be autonamed "user-container" if not provided. So it is always present (even as empty string) when deserializing to JSON from a `Service` struct, indicating the intention of the user to set it to an empty string (which actually is not the case as we never specify that name). That very likely is one of the causes of the patch conflict (as the backend set it to "user-container").

The idea is that we converted a Service to an `Unstructured` object and remove all map entries from which we know that they have not been specified. This work is still on-going.

I added the PR nevertheless as `Draft` as I'm now on PTO for the next two weeks and very likely can't continue on this PR until then. If someone feels fancy and might take a look, feel free to work on this PR.  

Also no tests have been implemented yet.


duglin:
@rhuss what's the diff between this and `kn service create --force`? They both seem to want to update an existing service with new metadata

kwiesmueller:
Hey there,
this came up in a GitHub search for Server-Side Apply.
I see you are using the `kubectl.kubernetes.io/last-applied-configuration` annotation which is provided by client-side apply.
We are working on Server-Side Apply which might resolve some of the issues i read in the description above.
If you need any help or got questions around apply, feel free to get in touch with us (wg-api-expression) on Slack or ping me here.
https://kubernetes.slack.com/archives/C0123CNN8F3

wslyln:
@duglin 'kn service create --force' will completely override the existing service resource. (akin to git push --force) while 'kn service apply' will do a three way merge of the configuration (aka git merge).

rhuss:
@kwiesmueller Thanks for the offer ! I'm just back from PTO and will continue on this issue soon.

rhuss:
@kwiesmueller Finally I got around to continue on working on this PR. I am still trying to get client side patching to work, before switching to server side, but there are still some stumbling block with defaulting, serializing and also Knative backend handling. At the moment I'm a bit stuck to understand the following error message:

```
✦ at 20:00 ❯ kn service apply random --revision-name "" --image rhuss/random:2.0
Error: patch:
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"serving.knative.dev/v1","kind":"Service","metadata":{"name":"random","namespace":"default"},"spec":{"template":{"metadata":{"annotations":{"client.knative.dev/user-image":"rhuss/random:2.0"}},"spec":{"containers":[{"image":"rhuss/random:2.0","name":"user-container"}]}}}}
spec:
  template:
    metadata:
      annotations:
        client.knative.dev/user-image: rhuss/random:2.0
    spec:
      $setElementOrder/containers:
      - name: user-container
      containers:
      - image: rhuss/random:2.0
        name: user-container

conflicts with changes made from original to current:
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"serving.knative.dev/v1","kind":"Service","metadata":{"name":"random","namespace":"default"},"spec":{"template":{"metadata":{"annotations":{"client.knative.dev/user-image":"rhuss/random:1.0"}},"spec":{"containers":[{"image":"rhuss/random:1.0","name":"user-container"}]}}}}
    serving.knative.dev/creator: minikube-user
    serving.knative.dev/lastModifier: minikube-user
  generation: 1
  managedFields:
  - apiVersion: serving.knative.dev/v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .: {}
          f:kubectl.kubernetes.io/last-applied-configuration: {}
      f:spec:
        .: {}
        f:template:
          .: {}
          f:metadata:
            .: {}
            f:annotations:
              .: {}
              f:client.knative.dev/user-image: {}
            f:creationTimestamp: {}
          f:spec:
            .: {}
            f:containers: {}
    manager: kn
    operation: Update
    time: "2020-09-09T18:00:46Z"
  - apiVersion: serving.knative.dev/v1
    fieldsType: FieldsV1
    fieldsV1:
      f:status:
        .: {}
        f:address:
          .: {}
          f:url: {}
        f:conditions: {}
        f:latestCreatedRevisionName: {}
        f:latestReadyRevisionName: {}
        f:observedGeneration: {}
        f:traffic: {}
        f:url: {}
    manager: controller
    operation: Update
    time: "2020-09-09T18:00:54Z"
  resourceVersion: "1473596"
  selfLink: /apis/serving.knative.dev/v1/namespaces/default/services/random
  uid: a37d3619-302e-4a5f-98c5-c4889faf6654
spec:
  template:
    spec:
      containerConcurrency: 0
      containers:
      - name: user-container
        readinessProbe:
          successThreshold: 1
          tcpSocket:
            port: 0
      timeoutSeconds: 300
  traffic:
  - latestRevision: true
    percent: 100
```

@kwiesmueller can you see, where the actual conflict is ? I though I fixed everything, but can't really see what is still conflicting.

kwiesmueller:
If you're using SSA then the conflict error should give you the exact fields you get a conflict with.

Did I read correctly, that it's not SSA yet?
If not, I'm not yet sure what the conflict could be. Will take another look once I have time tomorrow.

kwiesmueller:
It could be the CSA annotation, we should have a fix for that in the latest release.

rhuss:
> Did I read correctly, that it's not SSA yet?

true, it's a true client side patch for now. But happy to explore SSA, too (thought client-side patch was easier to start with)

rhuss:
@kwiesmueller do you have a good entrypoint/sample showing how to use SSA so that I can find out what's going on ? Currently I'm a bit lost.

kwiesmueller:
I think it makes sense to also support CSA as SSA is still in beta and only available since v1.16.
That said, to find out more about using SSA we have a bit of documentation here: https://kubernetes.io/docs/reference/using-api/api-concepts/#server-side-apply

On explicit usage examples, I recommend looking at our integration tests. A very basic one being this one: https://github.com/kubernetes/kubernetes/blob/master/test/integration/apiserver/apply/apply_test.go#L803

We have a lot of improvements in the works right now as well, especially working on a good client implementation.
If you have input on your use-case or need dedicated help with this, feel free to stop by on [Slack](https://kubernetes.slack.com/archives/C0123CNN8F3) or our [Bi-Weekly tomorrow (15. Sept 2020 @ 06:30pm GMT+2)](https://docs.google.com/document/d/1CSpNaicbEqKJoW306qaQEBIkwC1mGIcKl3yiB_C0HZk/edit).

docs/cmd/kn_service_apply.md
```diff
@@ -0,0 +1,110 @@
...
+
+```
+kn service apply NAME
+```
+
+### Examples
+
+```
+
+  # Create a service 's0' using image knativesamples/helloworld
```
maximilien, line 17: the examples don’t seem to match the command in question `service apply`?

maximilien (changes requested):
I can see a lot is missing. So let me wait to do a thorough review when ready.
/retest

rhuss:
let me squash the whole PR, looks like that a rebase messed that up.

rhuss:
@maximilien the tests are missing, I will work on that in this week. But beside this the PR is considered to be complete, so if you like you can already start to review. I will add the tests as extra commits to make an incremental review easier.

rhuss:
the next steps are now to add test to complete this initial implementation and continue to work on a more sophisticated way for a CSA that merges also arrays properly.

pkg/kn/commands/service/apply.go
```diff
@@ -0,0 +1,110 @@
...
+			namespace, err := p.GetNamespace(cmd)
+			if err != nil {
+				return err
+			}
+
+			var service *servingv1.Service
+			applyFlags.RevisionName = ""
+			if applyFlags.Filename == "" {
+				service, err = constructService(cmd, applyFlags, name, namespace)
+			} else {
```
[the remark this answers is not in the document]
maximilien: Also it might be indicative of dead code since early on there is a check for that field value.

pkg/kn/commands/service/apply_mock_test.go
```diff
@@ -0,0 +1,152 @@
...
+	"knative.dev/client/pkg/wait"
+
+	"knative.dev/client/pkg/util"
+)
+
+func TestServiceApplyCreateMock(t *testing.T) {
+	// New mock client
+	client := knclient.NewMockKnServiceClient(t)
+
+	//service := createServiceWithImage("foo", "gcr.io/foo/bar:baz")
```
maximilien, line 40: Is this intended or forgotten? The commented line.

pkg/kn/commands/service/create_mock_test.go
```diff
@@ -681,8 +681,12 @@ func TestServiceCreateWithAutoScaleServiceAnnotationsError(t *testing.T) {
...
 	service := servingv1.Service{}
-	url, _ := apis.ParseURL(urlName)
-	service.Status.URL = url
 	service.Name = name
+	setUrl(&service, urlName)
 	return &service
 }
+
+func setUrl(service *servingv1.Service, urlName string) {
+	url, _ := apis.ParseURL(urlName)
```
maximilien, line 690: why are ignoring errors here? the parse could fail, no?

pkg/serving/v1/apply.go
```diff
@@ -0,0 +1,293 @@
...
+// (yet) for a custom resource definition.
+var patchStrategy = types.MergePatchType
+
+func (cl *knServingClient) patch(modifiedService *servingv1.Service, currentService *servingv1.Service, uOriginalService []byte) (bool, error) {
+	uModifiedService, err := getModifiedConfiguration(modifiedService, true)
+	if err != nil {
+		return false, err
+	}
+	hasChanged, err := cl.patchSimple(currentService, uModifiedService, uOriginalService)
+	for i := 1; i <= 5 && apierrors.IsConflict(err); i++ {
```
maximilien, line 50: “5” is a bit magical :) perhaps use a named constant that gives more meaning?

pkg/serving/v1/apply.go
```diff
@@ -0,0 +1,293 @@
...
+			time.Sleep(1 * time.Second)
+		}
+		currentService, err = cl.GetService(currentService.Name)
+		if err != nil {
+			return false, err
+		}
+		hasChanged, err = cl.patchSimple(currentService, uModifiedService, uOriginalService)
+	}
+
+	// TODO: Introduce a "force" parameter that would do the same as `kn service create --force`
```
maximilien, line 61: Commented code...

pkg/serving/v1/apply.go
```diff
@@ -0,0 +1,293 @@
...
+	if patchStrategy == types.MergePatchType {
+		patch, err = jsonmergepatch.CreateThreeWayJSONMergePatch(uOriginalService, uModifiedService, uCurrentService)
+	} else if patchStrategy == types.StrategicMergePatchType {
+		// Compute a three way strategic merge patch to send to server.
+		lookupPatchMeta, err := strategicpatch.NewPatchMetaFromStruct(servingv1.Service{})
+		if err != nil {
+			return false, err
+		}
+		// TODO: Allow "overwrite" to be configured
+		patch, _ = strategicpatch.CreateThreeWayMergePatch(uOriginalService, uModifiedService, uCurrentService, lookupPatchMeta, false)
```
maximilien, line 88: ignoring errors....

pkg/serving/v1/apply_test.go
```diff
@@ -0,0 +1,101 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	clienttesting "k8s.io/client-go/testing"
+	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
+
+	"knative.dev/client/pkg/util"
+)
+
+// Copyright © 2020 The Knative Authors
```
maximilien, line 16: Copyright should be at the top :)

pkg/serving/v1/apply.go
```diff
@@ -0,0 +1,293 @@
...
+	if resources == nil {
+		return
+	}
+	resourcesMap := resources.(map[string]interface{})
+	if len(resourcesMap) == 0 {
+		delete(uContainer, "resources")
+	}
+}
+
+func extractUserContainer(uService map[string]interface{}) map[string]interface{} {
```
maximilien, line 240: Would be cool to have tests for each of these... or at minimum to have some files fixtures that in tests that could be used to verify different “typical YAMLs”

maximilien (changes requested):
Good start for an important feature. Left various feedback (some easy and some harder). Please try to address and tag me and I’d be happy to do another pass.

Thanks for the contribution. Cheers 🍻

docs/cmd/kn_service_apply.md
```diff
@@ -0,0 +1,80 @@
...
+Apply a service declaration
+
+```
+kn service apply NAME
+```
+
+### Examples
+
+```
+
```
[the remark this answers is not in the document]
rhuss: yep, right on it.

pkg/kn/commands/service/apply.go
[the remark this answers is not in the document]
rhuss: It's definitely no dead code, let me try to improve the tests. Its more or less taken over how 'service create' handles file based definitions and it should have been tested over there as well.

pkg/kn/commands/service/apply_mock_test.go
rhuss: it's a left over from the test development. will clean up, good find

pkg/kn/commands/service/create_mock_test.go
rhuss: this is just extracted from create test and this is a test we have full control of the input as well so I think it perfectly valid in this case to omit error handling when don't use this method with an invalid parameter (as mentioned, this is our own test code not exposed for public usage)

pkg/serving/v1/apply.go
rhuss: We have the same retry mechanism for update as well (also 5 times). It's kind of a heuristic and if it turns out that needs to be tuned, happy to expose this as an option too. but for now I would just keep it as an internal implementation optimisation until we found some issues with this approach. Also not keen on introducing a global constant for this as this would introduce a coupling among different packages that is not necessary.

pkg/serving/v1/apply.go
rhuss: I'm not sure how to deal with this, as there is still a way how we can improve this initial apply implementation that has some limitations. But only then a --force option can be implemented, which is not possible right now. Before forgetting this I think its fair to keep some comments, although Im not a fan of TODO comments in general, but in this case it might be justified as the implementation is blocked by the way how the merge is done. See also my general comments above about the merge algorithm chosen.

pkg/serving/v1/apply.go
rhuss: The whole code section is commented out and will never be executed. I will remove that part and find an other way how we can memorize the better alternative way for doing the merge itself.

kwiesmueller:
@rhuss I'm still reading along. If you need anything from me regarding apply, chat me up :-)

rhuss:
@maximilien thanks for the review, I adjusted the code according to your suggestions. You might want to have a look to the latest commit where I added my changes fixing the issues that you have pointed out

maximilien (approved):
Thanks for addressing feedback. LGTM. 

@navidshaikh feel free to do your review and merge

rhuss:
@navidshaikh I think we are good to merge if you don't have any objections.

pkg/serving/v1/apply.go
```diff
@@ -0,0 +1,268 @@
...
+	original := annots[v1.LastAppliedConfigAnnotation]
+	delete(annots, v1.LastAppliedConfigAnnotation)
+	service.Annotations = annots
+
+	uModifiedService, err := encodeService(service)
+	if err != nil {
+		return nil, err
+	}
+
+	if annotate {
```
navidshaikh, line 131: will there be a case when we'd want to control whether to annotate the service?

pkg/serving/v1/apply.go
```diff
@@ -0,0 +1,268 @@
...
+func extractUserContainer(uService map[string]interface{}) map[string]interface{} {
+	tSpec := extractTemplateSpec(uService)
+	if tSpec == nil {
+		return nil
+	}
+	containers := tSpec["containers"]
+	if len(containers.([]interface{})) == 0 {
+		return nil
+	}
+	return containers.([]interface{})[0].(map[string]interface{})
```
navidshaikh, line 224: I am not sure about the position of the user-container in the array, is it always at 0 index (thinking about multi-container)?

pkg/serving/v1/client.go
```diff
@@ -68,6 +69,17 @@ type KnServingClient interface {
 	// place.
 	UpdateServiceWithRetry(name string, updateFunc ServiceUpdateFunc, nrRetries int) error
 
+	// Apply a service's definition to the cluster. The full service declaration needs to be provided,
+	// which is different to UpdateService which can also do a partial update. If the given
+	// service does not already exists (identified by name) then the service is create.
```
navidshaikh, line 74:
```suggestion
	// service does not already exists (identified by name) then the service is created.
```

pkg/serving/v1/client.go
```diff
@@ -68,6 +69,17 @@ type KnServingClient interface {
...
+	// If the service exists, then a three-way merge will be performed between the original
+	// configuration given (from the last "apply" operation), the new configuration as given ]
```
navidshaikh, line 76:
```suggestion
	// configuration given (from the last "apply" operation), the new configuration as given
```

test/e2e/service_apply_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"gotest.tools/assert"
+
+	pkgtest "knative.dev/pkg/test"
+
+	"knative.dev/client/lib/test"
+	"knative.dev/client/pkg/util"
+)
+
+func TestServiceApply(t *testing.T) {
```
navidshaikh, line 31: we can also add another one e2e for apply with file

navidshaikh (approved):
/lgtm

thank you!
I've added a few questions and a few nits suggestions which can be addressed in a subsequent PR along with CHANGELOG.

pkg/kn/commands/service/apply.go
```diff
@@ -0,0 +1,123 @@
...
+# Read the service declaration from a file
+kn service apply s0 --filename my-svc.yml
+`
+
+func NewServiceApplyCommand(p *commands.KnParams) *cobra.Command {
+	var applyFlags ConfigurationEditFlags
+	var waitFlags commands.WaitFlags
+
+	serviceApplyCommand := &cobra.Command{
+		Use:     "apply NAME",
```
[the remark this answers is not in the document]
rhuss: true, we should think about a common scheme here. (e.g using `[ ]`  to describe optional positional arguments)

pkg/serving/v1/apply.go
rhuss: Yes, there is. IIUR when you do the diff, then you don't want to have that annotation present. I would have to look a bit closer again, but there is a good reason. I also more or less took over the way how kubectl does the client side apply.

pkg/serving/v1/apply.go
rhuss: For the moment we are assuming only single, container, that's true. Maybe we should iterate over the list and just pick the container with name `user-container` ?

test/e2e/service_apply_test.go
rhuss: 👍

pkg/serving/v1/apply.go
navidshaikh: Should this only remove the name of the user-container? or for all the containers in array? If its later, we can simply range over the array and remove the names. I need to catch up on the multi container support to understand this better.

PR #976, opened 2020-08-11
Pin knative/pkg to release-0.17

Harwayne:
## Proposed Changes

- Pin knative/pkg to `release-0.17`.
    - In preparation for `release-0.17` of knative/client in a week.

**Release Note**

```release-note
NONE
```


hack/update-deps.sh
```diff
@@ -28,7 +28,7 @@ VERSION="master"
 # The list of dependencies that we track at HEAD and periodically
 # float forward in this repository.
 FLOATING_DEPS=(
-  "knative.dev/pkg@${VERSION}"
+  "knative.dev/pkg@release-0.17"
```
mattmoor, line 31: I believe that line 50 will need a similar change, or the mix of variable vs. constant should be juggled.

Harwayne: Done.

PR #980, opened 2020-08-18
Set client custom GVK for source list for machine readable output

### LINKED ISSUE #953: [source list] Add custom type for different types of source objects
### Feature request

`kn source list` gives you objects of different types of sources. While printing the JSON/YAML format of list, we'd need 
proper values for `apiVersion` and `kind` fields of the list printed, since the objects list contains are of different types.

Current behavior, sets empty string for `apiVersion` and `List` for `kind`.
This is to track setting these fields to for eg: `apiVersion: client.knative.dev/v1alpha1` 
ref: https://github.com/knative/client/pull/948#discussion_r461362683

There is also a discussion about what `apiVersion` and `kind` values should be if, the source list is of single type source OR `--type` filter is given with single type to `kn source list`, in this scenario, only single type of source objects are printed.

navidshaikh:
## Description
 - Source list may contain different source types CO and machine readable output (using -o)
   requires List object to have GVK set, since the list contains different types of source COs,
   we set a custom client GVK on it.

## Changes
 - Use custom GVK {Group: client.knative.dev, Version: v1alpha1, Kind: SourceList}

## Reference

Fixes #953

/lint

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -45,6 +45,10 @@
 | 🐛
 | Fix client side volume name generation
 | https://github.com/knative/client/pull/975[#975]
+
+| ✨
+| `kn source list` output now has client custom GVK set as `{Group: client.knative.dev, Version: v1alpha1, Kind: SourceList}`
+| https://github.com/knative/client/pull/980[#980]
 |===
 
 ## v0.16.0 (2020-07-14)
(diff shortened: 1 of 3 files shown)


navidshaikh:
> Maybe you could add to the CHANGELOG.adoc the concrete values which you use for the list ?

done

rhuss:
thanks ! 

/lgtm
/retest

PR #981, opened 2020-08-18
Add kafka source desc for kn source list-types

navidshaikh:
also update the description for other source types to refer 'addressable' instead of 'sink'

### DIFF
--- a/pkg/kn/commands/source/human_readable_flags.go
+++ b/pkg/kn/commands/source/human_readable_flags.go
@@ -27,10 +27,11 @@ import (
 )
 
 var sourceTypeDescription = map[string]string{
-	"ApiServerSource": "Watch and send Kubernetes API events to a sink",
-	"SinkBinding":     "Binding for connecting a PodSpecable to a sink",
-	"PingSource":      "Send periodically ping events to a sink",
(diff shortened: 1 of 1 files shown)


navidshaikh:
/retest


        result_collector.go:75: ERROR: Error: ReconcileIngressFailed: Ingress reconciliation failed

navidshaikh:
/retest

      result_collector.go:75: ERROR: Error: ReconcileIngressFailed: Ingress reconciliation failed
 
and

         result_collector.go:75: ERROR: Error: cannot delete broker 'foo2' in namespace 'kne2etests19' because: timeout: broker 'foo2' not ready after 600 seconds

dsimansk:
Lets check if broker test is still flaky.

/test pull-knative-client-integration-tests

rhuss:
as discussed yesterday in the WG call, i'm not sure that we add a reference to a non-builtin source in kn proper, even when it's only in the description. We could do it now for a workaround, but simultaneously also think about how to query source plugins for their meta data. 

Please add at least a comment describing the situation. I would be fine then with this temporary solution.

navidshaikh:
> Please add at least a comment describing the situation. I would be fine then with this temporary solution.


@rhuss : will do a subsequent PR to address this! thanks!

PR #990, opened 2020-08-25
Add changes for --scale-init support

### LINKED ISSUE #959: Add support for --scale-init
Now that we have support for defining the initial scale for a service, it would be nice if `kn` supported it via a top-level option instead of just as an annotation.

e.g.
kn service create echo --image duglin/echo --annotation autoscaling.knative.dev/initialScale=0
vs
kn service create echo --image duglin/echo --scale-init=0

or --scale-initial

hemanrnjn:
## Description

Now that we have support for defining the initial scale for a service, implementing `--scale-init` with kn as a top-level option instead of just as an annotation.

## Changes

* Add support for --scale-init option

## Reference

Fixes #959


hemanrnjn:
@rhuss I'm adding test cases so till then I'll put this PR in WIP.

hemanrnjn:
> @hemanrnjn thanks for the contribution. Are you ready for some feedback or would you rather we wait for when you remove the WIP tag?

Sure. Why not! I'd also implement those feedback while the PR is in WIP.

docs/cmd/kn_service_update.md
```diff
@@ -70,6 +70,7 @@ kn service update NAME
       --requests-memory string        DEPRECATED: please use --request instead. The requested memory (e.g., 64Mi).
       --revision-name string          The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
       --scale int                     Minimum and maximum number of replicas.
+      --scale-init int                Initial scale for a service
```
maximilien, line 73: Perhaps mention here that the range is 0..some positive number, I imagine limited by vendor or is it limited in API or just the max positive int. 

Also, since max value of int would be very large, perhaps a recommended sensible value is better... Not clear but need to say here that 0..positive number is range since folks might assume 1 is minimum.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -149,6 +151,9 @@ func (p *ConfigurationEditFlags) addSharedFlags(command *cobra.Command) {
 			"any number of times to set multiple annotations. "+
 			"To unset, specify the annotation name followed by a \"-\" (e.g., name-).")
 	p.markFlagMakesRevision("annotation")
+
+	command.Flags().IntVar(&p.ScaleInit, "scale-init", 0, "Initial scale for a service")
```
maximilien, line 155: Cool that default is 0 but need to reflect this in the generated docs for the help output.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -427,6 +432,28 @@ func (p *ConfigurationEditFlags) Apply(
 		servinglib.UpdateUser(template, p.PodSpecFlags.User)
 	}
 
+	if cmd.Flags().Changed("scale-init") {
+		contains := func(s []string, e string) bool {
+			for _, a := range s {
+				if strings.Contains(a, e) {
```
maximilien, line 438: Can we better names than `a`, `s`, `e`, etc? :)

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -427,6 +432,28 @@ func (p *ConfigurationEditFlags) Apply(
...
+				}
+			}
+			return false
+		}
+
+		if cmd.Flags().Changed("annotation") && contains(p.Annotations, "autoscaling.knative.dev/initialScale") {
+			return fmt.Errorf("only one of the --scale-init or --annotation autoscaling.knative.dev/initialScale can be specified")
+		}
+		annotationsMap := make(map[string]string)
+		annotationsMap["autoscaling.knative.dev/initialScale"] = strconv.Itoa(p.ScaleInit)
```
maximilien, line 449: What's the use of this new line?

maximilien, line 449: Is there a check for negative values of `ScaleInit`?

hemanrnjn: All the validation should be done at the server side, I suppose. Even with the max and min scale option, there is no validation done on the client side.

pkg/kn/commands/service/configuration_edit_flags.go
hemanrnjn: Yes, I can definitely name it better but it was only a scope limited helper function I wrote that cannot be used outside this block so thought naming won't be necessary.

pkg/kn/commands/service/configuration_edit_flags.go
hemanrnjn: I will denote it in the description message.

pkg/kn/commands/service/configuration_edit_flags.go
hemanrnjn: That's a bit redundant. Fixed it though.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -427,6 +432,29 @@ func (p *ConfigurationEditFlags) Apply(
...
+			for _, element := range annotationList {
+				if strings.Contains(element, annotation) {
+					return true
+				}
+			}
+			return false
+		}
+
+		if cmd.Flags().Changed("annotation") && containsAnnotation(p.Annotations, "autoscaling.knative.dev/initialScale") {
+			return fmt.Errorf("only one of the --scale-init or --annotation autoscaling.knative.dev/initialScale can be specified")
```
rhuss, line 446: good catch, but you should probably refer to the constant that is used by knative serving for this annotation instead of using it literally here.

Also, I wonder whether we should throw an error here. If so we should do the same for the other scale related annotation(s).

rhuss (changes requested):
Thanks a lot for this PR ! Looks good to me, with some minor suggestions (see below)

pkg/kn/commands/service/configuration_edit_flags.go
hemanrnjn: It is indeed "1". The default "0" here in client is actually of no significance and could be any value, because it's not like 0 is going to be sent as initial Scale value if the flag `--scale-init` is not passed and in which case, the server will set 1 as default. In case you explicitly specify this option with a value, that value will be validated in the server.

docs/cmd/kn_service_update.md
hemanrnjn: Actually I thought that this option was to be used with service create only so I didn't keep in mind that update will use it too. Therefore I haven't added test cases for service update either, if you will see. I just had this discussion with Navid yesterday and I'm adding service update test cases as well.

Also, there will be some changes needed for update to work too. I will make them by EOD.

navidshaikh:
> Just out of curiosity I wonder how this works for an "update". Is it that the current number of replicas is changed to this number when the update is triggered ?

@rhuss : IIUC, yes, this would scale the number of replicas to given match given init-scale to mark the revision ready and then it'll ignore the value (for eg in no-traffic scale to zero case). This property applies to per revision, so setting this for to-be-created revision in an update should be a valid use case.

pkg/kn/commands/service/configuration_edit_flags.go
navidshaikh: yea, the value should be processed only if the flag `Changed`, so the default value is irrelevant here

hemanrnjn:
@navidshaikh and @rhuss I have made all the changes and added all test cases. Please take one final look at it, if they look fine to you guys too?

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -427,6 +432,29 @@ func (p *ConfigurationEditFlags) Apply(
 		servinglib.UpdateUser(template, p.PodSpecFlags.User)
 	}
 
+	if cmd.Flags().Changed("scale-init") {
+		containsAnnotation := func(annotationList []string, annotation string) bool {
+			for _, element := range annotationList {
+				if strings.Contains(element, annotation) {
+					return true
```
navidshaikh, line 439: we should do an exact match for the respective annotation key here, because we support unsetting an annotation by appending `-` to anno key, and that will match the case here.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -427,6 +432,29 @@ func (p *ConfigurationEditFlags) Apply(
...
+				}
+			}
+			return false
+		}
+
+		if cmd.Flags().Changed("annotation") && containsAnnotation(p.Annotations, "autoscaling.knative.dev/initialScale") {
+			return fmt.Errorf("only one of the --scale-init or --annotation autoscaling.knative.dev/initialScale can be specified")
+		}
+		annotationsMap := map[string]string{
+			"autoscaling.knative.dev/initialScale": strconv.Itoa(p.ScaleInit),
```
navidshaikh, line 449: use the constant as mentioned above

navidshaikh (changes requested):

pkg/kn/commands/service/configuration_edit_flags.go
navidshaikh: Well, in either case (set/unset), if user specifies both way, we should prompt that, so we can keep this check as is.

pkg/kn/commands/service/configuration_edit_flags.go
hemanrnjn: Pushed the changes using constant. And about the changes for other scale related issues, we can do it in a separate PR!

navidshaikh (approved):
/lgtm
/approve

thanks @hemanrnjn ! The changelog entry is missing but you can add that in a separate PR.

PR #1006, opened 2020-09-10
fix: Fix service export example doc

dsimansk:
## Description

Example of `service export` doesn't contain actual `--mode` parameters.

## Changes

* Fix parameters in the `service export` example

## Reference

Fixes #N/A

### DIFF
--- a/pkg/kn/commands/service/export.go
+++ b/pkg/kn/commands/service/export.go
@@ -57,10 +57,13 @@ func NewServiceExportCommand(p *commands.KnParams) *cobra.Command {
 		Example: `
   # Export a service in YAML format
   kn service export foo -n bar -o yaml
+
   # Export a service in JSON format
   kn service export foo -n bar -o json
+
   # Export a service with revisions
   kn service export foo --with-revisions --mode=export -n bar -o json
+
   # Export services in kubectl friendly format, as a list kind, one service item for each revision
   kn service export foo --with-revisions --mode=replay -n bar -o json`,
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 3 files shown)


maximilien (approved):
Good catch and /lgtm

How cool would it be to have a test that scrapes all commands descriptions and try to execute and validate :) them? turns out @cppforlife and I did that in knctl and I believe he kept that going on k14s tools. I never got around adding this here but not impossible.

Let me know if interested and we can add and track and execute this as a follow-up issue.

/ok-to-test

pkg/kn/commands/service/export.go
```diff
@@ -60,9 +60,9 @@ func NewServiceExportCommand(p *commands.KnParams) *cobra.Command {
   # Export a service in JSON format
   kn service export foo -n bar -o json
   # Export a service with revisions
-  kn service export foo --with-revisions --mode=resources -n bar -o json
+  kn service export foo --with-revisions --mode=export -n bar -o json
   # Export services in kubectl friendly format, as a list kind, one service item for each revision
-  kn service export foo --with-revisions --mode=kubernetes -n bar -o json`,
+  kn service export foo --with-revisions --mode=replay -n bar -o json`,
```
navidshaikh, line 65: should we add a line separator in each example with this to align with other command examples?

dsimansk: Sure, will do.

dsimansk: PTAL https://github.com/knative/client/pull/1006/commits/4aaab55d8e7456dcae6809986a9d808f6bf2d87a.

navidshaikh (approved):
/lgtm
/approve

thanks!

PR #1010, opened 2020-09-14
Added back functionality to deprecated max-scale/min-scale

### LINKED ISSUE #1008: --min-scale/max-scale are broken
the latest `kn` seems to support --min-scale and --max-scale, at least it doesn't complain when I try to use them. But nothing is set in the ksvc - so they have no effect. If we're not going to generate an error (due to them not being valid flag) then we should still support them on the backend properly.

I suspect this might catch a lot of people - took me a while to figure out why my services kept scaling down to zero despite specifying --min-scale=1  :-) 

IMO we should back-port the fix for this

mpetason:
## Description

Adds back functionality for max-scale/min-scale options. I ended up swapping over naming but never added additional sections to update. 

## Changes

* Adds back functionality for max-scale/min-scale

## Reference

Fixes #1008

### DIFF
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -336,6 +336,22 @@ func (p *ConfigurationEditFlags) Apply(
 		}
 	}
 
+	// Deprecated option
+	if cmd.Flags().Changed("min-scale") {
+		err = servinglib.UpdateMinScale(template, p.MinScale)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Deprecated option
+	if cmd.Flags().Changed("max-scale") {
+		err = servinglib.UpdateMaxScale(template, p.MaxScale)
+		if err != nil {
+			return err
+		}
+	}
+
 	if cmd.Flags().Changed("scale-min") {
 		err = servinglib.UpdateMinScale(template, p.MinScale)
 		if err != nil {
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -336,30 +336,16 @@ func (p *ConfigurationEditFlags) Apply(
 		}
 	}
 
-	// Deprecated option
-	if cmd.Flags().Changed("min-scale") {
+	// Deprecated "min-scale" in 0.19, updated to "scale-min"
+	if cmd.Flags().Changed("scale-min") || cmd.Flags().Changed("min-scale") {
 		err = servinglib.UpdateMinScale(template, p.MinScale)
 		if err != nil {
 			return err
 		}
 	}
 
-	// Deprecated option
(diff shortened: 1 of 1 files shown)


mpetason:
Issue also mentions that I need to add e2e tests, which will take me a little bit longer to address.

maximilien (approved):
/retest

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -313,6 +313,22 @@ func (p *ConfigurationEditFlags) Apply(
 		}
 	}
 
+	// Deprecated option
+	if cmd.Flags().Changed("min-scale") {
```
rhuss, line 317: you could add it to the `if` below with an "or", to avoid duplicated code. Also, I would add an comment for which release we can remove this option.

According to our policy this would be n+2 where n is the first version where this feature was deprecated (so 0.19 iiur)

rhuss (commented):
thanks for the quick fix ! As soon as we have this merged we should trigger a new minor release asap.

rhuss:
> Issue also mentions that I need to add e2e tests, which will take me a little bit longer to address.

It's totally fine to have this quick fix first (so that we can deliver a patch release asap) and then add the e2e test later. This should also not be that hard as it is just to add some extra variations for these options.

pkg/kn/commands/service/configuration_edit_flags.go
mpetason: Thank you for the suggestions. Went ahead and make the updates.

mpetason:
Finally passed the `pull-knative-client-integration-tests` test!

rhuss:
Thanks !

/lgtm
/approve

navidshaikh:
The build tests are failing complaining markdown link check failure from test-infra scripts

```
scripts/test-infra/README.md
	ERROR	../kntest/pkg/kubetest2/gke/README.md
		Stat scripts/kntest/pkg/kubetest2/gke/README.md: no such file or directory
```
cc: @coryrc @chizhg

navidshaikh:
@mpetason : you might want to try rebasing onto current master and see if the tests are through ?

rhuss:
Oh, the build issue comes from an update of the test scripts, coming in with a pr that was merged just before.

I will take care.

rhuss:
Infratest failure is because of a flake.

mpetason:
Sounds good, let me know if you end up needing me to do something on my end.

navidshaikh:
The markdown check failure only happens in this PR but not others, @mpetason a rebase against latest master should help I think.

mpetason:
Sounds good, I'll get that taken care of.

mpetason:
Awesome. I think that fixed it. Thanks for the help and suggestions. I got some help and didn't follow the git recommendations and was able to rebase without adding 100 more commits haha.

navidshaikh (approved):
/lgtm

PR #1019, opened 2020-09-18
Print logs from PipelineRun pods when Tekton tests fail

mgencur:
## Description

It often happens that the tests fail and there's no error message. This is very helpful when analyzing the failures.

## Changes

*
*
*

## Reference

Fixes #


navidshaikh (approved):
/lgtm
/approve

thanks @mgencur , indeed this should help find the error in CI logs itself!

Issue #1014, opened 2020-09-15
Add "--annotation-service" and "--annotation-revision" to kn service create/update


rhuss:
### Feature request

In order to provide the same support as for labels also for annotations, the two options mentioned in the title should be implemented with the same semantics as for labels:

```
-l, --label stringArray             Labels to set for both Service and Revision. name=value; you may provide this flag any number of times to set
                                      multiple labels. To unset, specify the label name followed by a "-" (e.g., name-).
      --label-revision stringArray    Revision label to set. name=value; you may provide this flag any number of times to set multiple labels. To unset,
                                      specify the label name followed by a "-" (e.g., name-). This flag takes precedence over "label" flag.
      --label-service stringArray     Service label to set. name=value; you may provide this flag any number of times to set multiple labels. To unset,
                                      specify the label name followed by a "-" (e.g., name-). This flag takes precedence over "label" flag.
```
### Use case

One use case is to add `sidecar.istio.io/inject: "true"` annotation to the pod template which is currently not possible. 

cc: @nak3

PR #1021, opened 2020-09-18
fix: Fix autoscaling annotations in Service metadata

### LINKED ISSUE #1020: autoscaling annotations must be put under "spec.template.metadata.annotations" to work
In client, we set (all) the annotations to service metadata annos as well as template metadata annotations.
see https://github.com/knative/client/blob/1607bf8c0c6404632d203f5a8794ac60d7c7659c/pkg/serving/config_changes.go#L457

With https://github.com/knative/serving/pull/9358 serving now validates the autoscaling related annotations, we'd need to stop setting the respective annotations at service metadata and only set for template metadata. The change in serving landed in master and would be part of v0.18.0 , so we'll need this for client v0.18.0 and probably backport to v0.17.x as well ? (cc: @rhuss )

The fix should be made in a manner to consider this related feature request https://github.com/knative/client/issues/1014

dsimansk:
## Description

This fix should mitigate usage of autoscaling annotations in Service's `metadata.Annotations` field.

Related to split annotation flag mentioned in https://github.com/knative/client/issues/1014.

## Changes

* `UpdateAnnotations` can distinguish autoscaling annotations and add them to template only
* 

## Reference

Fixes #1020

### DIFF
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -412,10 +412,23 @@ func (p *ConfigurationEditFlags) Apply(
 		}
 
 		annotationsToRemove := util.ParseMinusSuffix(annotationsMap)
-		err = servinglib.UpdateAnnotations(service, template, annotationsMap, annotationsToRemove)
+		// Service Annotations can't contain Autoscaling ones
+		serviceAnnotations := make(map[string]string)
+		for key, value := range annotationsMap {
+			if !strings.HasPrefix(key, autoscaling.GroupName) {
+				serviceAnnotations[key] = value
+			}
+		}
+		// Add all user provided annotations to RevisionTemplate
+		err = servinglib.UpdateRevisionTemplateAnnotations(template, annotationsMap, annotationsToRemove)
 		if err != nil {
 			return err
 		}
+		err = servinglib.UpdateServiceAnnotations(service, serviceAnnotations, annotationsToRemove)
+		if err != nil {
+			return err
+		}
+
 	}
 
 	if cmd.Flags().Changed("service-account") {
@@ -446,7 +459,7 @@ func (p *ConfigurationEditFlags) Apply(
 		if cmd.Flags().Changed("annotation") && containsAnnotation(p.Annotations, autoscaling.InitialScaleAnnotationKey) {
 			return fmt.Errorf("only one of the --scale-init or --annotation %s can be specified", autoscaling.InitialScaleAnnotationKey)
 		}
-
+		// Autoscaling annotations are only applicable on Revision Template, not Service
 		err = servinglib.UpdateRevisionTemplateAnnotation(template, autoscaling.InitialScaleAnnotationKey, strconv.Itoa(p.ScaleInit))
 		if err != nil {
 			return err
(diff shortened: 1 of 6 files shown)


dsimansk:
For the reviewers, going though the code I'd add definitely add some test cases with such annotations, unit and e2e.

maximilien (changes requested):
Looking forward to the tests. Will do thorough review when added. Thanks for contribution.

test/e2e/service_options_test.go
```diff
@@ -20,8 +20,7 @@ package e2e
 import (
 	"encoding/json"
 	"fmt"
-	"os"
-	"strconv"
+	"knative.dev/serving/pkg/apis/autoscaling"
```
mattmoor, line 23: Format Go code:
```suggestion

```

test/e2e/service_options_test.go
```diff
@@ -20,8 +20,7 @@ package e2e
...
 	"strings"
 	"testing"
```
mattmoor, line 25: Format Go code:
```suggestion
	"testing"

	"knative.dev/serving/pkg/apis/autoscaling"
```

mattmoor (commented):
Produced via:
  `gofmt -s -w $(find -path './vendor' -prune -o -path './third_party' -prune -o -name '*.pb.go' -prune -o -type f -name '*.go' -print)`
  `goimports -w $(find -name '*.go' | grep -v vendor | grep -v third_party | grep -v .pb.go | grep -v wire_gen.go)`

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -446,11 +446,8 @@ func (p *ConfigurationEditFlags) Apply(
 		if cmd.Flags().Changed("annotation") && containsAnnotation(p.Annotations, autoscaling.InitialScaleAnnotationKey) {
 			return fmt.Errorf("only one of the --scale-init or --annotation %s can be specified", autoscaling.InitialScaleAnnotationKey)
 		}
-		annotationsMap := map[string]string{
-			autoscaling.InitialScaleAnnotationKey: strconv.Itoa(p.ScaleInit),
-		}
 
-		err = servinglib.UpdateAnnotations(service, template, annotationsMap, []string{})
+		err = servinglib.UpdateRevisionTemplateAnnotation(template, autoscaling.InitialScaleAnnotationKey, strconv.Itoa(p.ScaleInit))
```
navidshaikh, line 450: IMO, we should have two utils
 - `UpdateRevisionAnnotations`
 - `UpdateServiceAnnotations`
Each taking respective object, toUpdate and toRemove map.
The callers of these utils can then decide if the given annotations should be set on either or both (explaining why in comments).

navidshaikh (commented):
PTAL comment about checking the annotations type at the caller level and the utils just doing the update task.

pkg/kn/commands/service/configuration_edit_flags.go
dsimansk: In the current codebase there's only one place that `UpdateAnnotations` is used and needs to be changed to filter autoscaling annotations. I tried to aim for quick win to unblock CI etc. However, I do agree that separate `Update*` util methods are likely better solution to expose filtering logic at caller level. I'll update the PR shortly. :wink:

navidshaikh (approved):
/lgtm
/approve

thank you!

PR #1038, opened 2020-10-01
Update error message when autoscaling annotation used in service

arghya88:
## Description
Update error message when autoscaling annotation used in service

## Reference

/lint


navidshaikh (approved):
/lgtm
/approve

PR #1041, opened 2020-10-04
feat: Add aliases to commands

### LINKED ISSUE #940: Proposal: Add aliases to commands
# Feature request

There has been now two features requests which can be covered by command _aliases_

* Add `ls` as alias for `list` #939 
* Add plural form like e.g.`kn services` (or maybe also `kn ksvc`) #812 

Before tackling these suggestions, let's discuss how we would like to introduce aliases.

My design proposal for adding aliases would be:

* An alias should feel tightly connected to the aliases command
* An alias should not very likely do not cover any future command "slot" for a future feature (if that can be foreseen)
* Aliases should not clutter or dilute the help message (e.g. should not repeat descriptions and listed separately on the in the help message)
* Aliases that fix obvious typos, like the plural form of given commands should possibly not be added to the help message.
* We should not go overboard with alias usage and don't take the introduction of an alias lightly. 
* Alias should be limited to command, no option aliases should be introduced.

wdyt ? The goal of this issue should be to come up with an addition to the CLI convention doc, which describes how aliases should be handled.

arghya88:
## Description

Adds aliases to commands

## Reference

Fixes #940

/lint

### DIFF
--- a/docs/cmd/kn_broker.md
+++ b/docs/cmd/kn_broker.md
@@ -1,10 +1,10 @@
 ## kn broker
 
-Manage message broker
+Manage message broker (alias: brokers)
 
 ### Synopsis
 
-Manage message broker
+Manage message broker (alias: brokers)
 
 ```
 kn broker
@@ -30,5 +30,5 @@ kn broker
 * [kn broker create](kn_broker_create.md)	 - Create a broker
 * [kn broker delete](kn_broker_delete.md)	 - Delete a broker
 * [kn broker describe](kn_broker_describe.md)	 - Describe broker
-* [kn broker list](kn_broker_list.md)	 - List brokers
+* [kn broker list](kn_broker_list.md)	 - List brokers (alias: 'ls')
 
--- a/pkg/kn/commands/broker/list.go
+++ b/pkg/kn/commands/broker/list.go
@@ -43,7 +43,8 @@ func NewBrokerListCommand(p *commands.KnParams) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:     "list",
-		Short:   "List brokers",
+		Short:   "List brokers (alias: 'ls')",
+		Aliases: []string{"ls"},
 		Example: listExample,
 		RunE: func(cmd *cobra.Command, args []string) (err error) {
 			namespace, err := p.GetNamespace(cmd)
--- a/pkg/kn/commands/service/service.go
+++ b/pkg/kn/commands/service/service.go
@@ -33,8 +33,9 @@ const (
 
 func NewServiceCommand(p *commands.KnParams) *cobra.Command {
 	serviceCmd := &cobra.Command{
-		Use:   "service",
-		Short: "Manage Knative services",
+		Use:     "service",
+		Short:   "Manage Knative services (aliases: ksvc, services)",
+		Aliases: []string{"ksvc", "services"},
 	}
 	serviceCmd.AddCommand(NewServiceListCommand(p))
 	serviceCmd.AddCommand(NewServiceDescribeCommand(p))
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -34,6 +34,10 @@
 | Add "--annotation-service" and "--annotation-revision" to kn service create/update
 | https://github.com/knative/client/pull/1029[#1029]
 
+| 🎁
+| Add alias to commands
+| https://github.com/knative/client/pull/1041[#1041]
+
 |===
 
 ## v0.17.0 (2020-08-26)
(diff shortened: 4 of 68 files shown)


arghya88:
/retest

`Ingress reconciliation failed`

rhuss (changes requested):
Thanks ! Looks good, but could you please add some tests ? This shouldn't be hard by either copying already existing tests or modifying some of the existing tests to use an alias instead of the original command.

I'd suggest to add additional test for the unit tests (possible wrapping existing tests into a loop). This does not have to be done every, some tests are fine (as I think we can confident if the alias resolution worked once that it then works everywhere)

For the integration tests please modify some of the existing tests to use aliases, too, so that you have a mix of tests. The reason why I would like to avoid to add additional e2e tests is, that this would add to the overall time for the e2e test suite, which I think is not needed for just testing that aliases are resolved properly.

pkg/kn/commands/broker/list.go
```diff
@@ -43,7 +43,8 @@ func NewBrokerListCommand(p *commands.KnParams) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:     "list",
-		Short:   "List brokers",
+		Short:   "List brokers(Alias: 'ls')",
```
danielhelfand, line 46: Doesn't Cobra usually include Aliases under `-h` without the need to include in the command description? There does appear to be something preventing it with `kn`, but it would probably be easier to maintain than documenting like this. 

A nit is that there should probably be a space between `List brokers` and the aliases if keeping this approach is preferred:

```suggestion
		Short:   "List brokers (Alias: 'ls')",
```

docs/cmd/kn_broker.md
```diff
@@ -30,5 +30,5 @@ kn broker
 * [kn broker create](kn_broker_create.md)	 - Create a broker
 * [kn broker delete](kn_broker_delete.md)	 - Delete a broker
 * [kn broker describe](kn_broker_describe.md)	 - Describe broker
-* [kn broker list](kn_broker_list.md)	 - List brokers
+* [kn broker list](kn_broker_list.md)	 - List brokers(Alias: 'ls')
```
maximilien, line 33: I wonder if "alias" (lower case) would be better/cleaner?

maximilien (changes requested):
Thanks for contribution. An integration tests that mixes aliases and non-alias commands would be nice.

pkg/kn/commands/broker/list.go
arghya88: help using cobra shows plurals? If yes then it might be confusing.

pkg/kn/commands/service/service.go
```diff
@@ -33,8 +33,9 @@ const (
 
 func NewServiceCommand(p *commands.KnParams) *cobra.Command {
 	serviceCmd := &cobra.Command{
-		Use:   "service",
-		Short: "Manage Knative services",
+		Use:     "service",
+		Short:   "Manage Knative services",
```
navidshaikh, line 37:
```suggestion
		Short:   "Manage Knative services (aliases: ksvc, services)",
```

navidshaikh: as done for leaf commands

navidshaikh (changes requested):
+1 to the point raised by @danielhelfand , but can be done in a subsequent PR once we know what preventing cobra to document that. Lets get this in today (release day), just a minor nit to follow the alias hint for command group as well.

docs/cmd/kn_broker.md
arghya88: Done

pkg/kn/commands/service/service.go
arghya88: So if there is one alias only should it be `(alias:` ?

navidshaikh: yes

arghya88: Done

rhuss:
+1 for adding this now and adding the testing + doc tuning after this PR got merged

arghya88:
@navidshaikh tests passed finally..PTAL

navidshaikh (approved):
/lgtm
/approve

thanks @arghya88 !
The CI flakes should be solved in upcoming patch release.

navidshaikh:
logged https://github.com/knative/client/issues/1045 to track the auto doc for aliases
thanks @danielhelfand

pkg/kn/commands/broker/list.go
arghya88: Will revisit this as part of #1045. Resolving this here.

PR #1044, opened 2020-10-06
fix(e2e): Let the subscription and related resource reconcile

navidshaikh:
sleep for 5 seconds after subscription create and update in e2e

### DIFF
--- a/lib/test/subscription.go
+++ b/lib/test/subscription.go
@@ -17,6 +17,8 @@ limitations under the License.
 package test
 
 import (
+	"time"
+
 	"gotest.tools/assert"
 
 	"knative.dev/client/pkg/util"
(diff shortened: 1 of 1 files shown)


lib/test/subscription.go
```diff
@@ -18,6 +18,7 @@ package test
 
 import (
```
mattmoor, line 19: Format Go code:
```suggestion
import (
	"time"

```

lib/test/subscription.go
```diff
@@ -18,6 +18,7 @@ package test
...
 	"gotest.tools/assert"
+	"time"
```
mattmoor, line 21: Format Go code:
```suggestion

```

mattmoor (commented):
Produced via:
  `gofmt -s -w $(find -path './vendor' -prune -o -path './third_party' -prune -o -name '*.pb.go' -prune -o -type f -name '*.go' -print)`
  `goimports -w $(find -name '*.go' | grep -v vendor | grep -v third_party | grep -v .pb.go | grep -v wire_gen.go)`

rhuss (approved):
/lgtm

thanks !

PR #1048, opened 2020-10-07
[0.17] Backport 1044 and 10101

navidshaikh:
- Backport:
  - #1044 
  - #1010 
- Update CHANGELOG

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -12,11 +12,15 @@
 | https://github.com/knative/client/pull/[#]
 ////
 
-## v0.17.1 (2020-10-02)
+## v0.17.1 (2020-10-07)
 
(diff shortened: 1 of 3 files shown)


rhuss:
@mpetason Could you please do a `@googlebot I consent.` to mak CLAbot happy ?

rhuss (approved):
/lgtm 

but @navidshaikh , this is only for #1010 , right ? (not #1044

navidshaikh:
> but @navidshaikh , this is only for #1010 , right ? (not #1044

Added #1044 as well, which is the fix for subscription e2e,  as that might fail in actual patch release job as well. @rhuss

navidshaikh:
@googlebot I consent

@mpetason : can you please add your consent here like above?

mpetason:
@googlebot I consent.

PR #1051, opened 2020-10-07
fix(e2e): Skip service export test

dsimansk:
## Description

Hot fix to unblock 0.18 release.

/assign @rhuss 

If it isn't needed, feel free to close the PR.

### DIFF
--- a/test/e2e/service_export_test.go
+++ b/test/e2e/service_export_test.go
@@ -19,8 +19,6 @@ package e2e
 
 import (
 	"encoding/json"
-	"os"
-	"strings"
 	"testing"
 
 	"gotest.tools/assert"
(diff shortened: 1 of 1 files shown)


rhuss:
```
        🦆 kn service update svc3a -a alpha=direwolf -a brave- --namespace kne2etests23
        ┃ Updating Service 'svc3a' in namespace 'kne2etests23':
        ┃ 
        ┃   0.059s The Configuration is still working to reflect the latest desired specification.
        ┃   0.480s Ingress reconciliation failed
        ┃ 
        🔥 Error: ReconcileIngressFailed: Ingress reconciliation failed
        🔥 Run 'kn --help' for usage
        🔥 
```

So, the ServiceOptionTest is a flake but it always breaks at the same sub-test (this one above). So it must be somehow specific of how we update the service and the error is triggered on the server-side.

Maybe the "Ingress reconciliation" is an intermediate errors (this event really comes in very quickly after the start), and we should wait a bit longer ?

PR #1065, opened 2020-10-15
feat: Add service import command

### LINKED ISSUE #654: Add "kn service import"
## Describe the feature:

Add a `kn service import` command to import a YAML/JSON description of a service. This is the equivalent of `kubectl create -f`  but can be used also via the Knative API contract. The idea is that `kn service export` (#653) and `kn service import` are complementary.

Possible options:

* `--apply` : Use an apply algorithm like with `kubectl apply` (i.e. add the applied resource as an annotation and perform a three-way diff when actually applying). This would allow also an idempotent application of exported resources, very useful in a CI context where `kn` is used as a last step for deployment/update).
* `--namespace` for specifying the target namespace

dsimansk:
## Description

The PR is missing any kind of test coverage, TBD. I'd like to gather early feedback on the implementation approach here. 

## Changes

* Add `service import` command

## Reference

Fixes #654 

/lint
/hold
/cc @rhuss @navidshaikh


itsmurugappan:
looks good. Have two questions /comments

1. Will this is only support only Export kind ? If I export in kubectl friendly style, will I not be able to use the import functionality
2. I see initial service without traffic and traffic getting updated later. IIRC traffic with missing revisions should not cause the ksvc to fail. Could save you an additional step of updating the traffic.

maximilien (commented):
/ok-to-test

Nice start. I wonder if `—dry-runs option would be nice whereby the import verifies but does not deploy. Maybe as a follow on. Anyhow, love seeing this whole export/import working out. An obvious e2e test will be something like:

```bash
kn service create testSvc —image ...
...
kn service export testSvc test.yml
...
kn service delete testSvc
...
kn service import testSvc test.yml
```

And end result is the same as first step :)

dsimansk:
>     1. Will this is only support only Export kind ? If I export in kubectl friendly style, will I not be able to use the import functionality

Per our conversation with @rhuss `kn` shouldn't duplicate functionality as such export is already usable with `kubectl`. Therefore `kn service import` should work with its own Export format. Please @rhuss correct me if I'm wrong here.

>     2. I see initial service without traffic and traffic getting updated later. IIRC traffic with missing revisions should not cause the ksvc to fail. Could save you an additional step of updating the traffic.

That's intentional to avoid error states. During my local testing I've seen intermittent failures during wait phase. Those might have been caused by waiting issue that was [recently fixed](https://github.com/knative/client/commit/878b037c1665af8c0d616d16b7690873e824590b). Thanks for the suggestion, I'll double check if there's still an issue or not.



> Nice start. I wonder if `—dry-runs option would be nice whereby the import verifies but does not deploy. Maybe as a follow on. Anyhow, love seeing this whole export/import working out. An obvious e2e test will be something like:

Thanks for the suggestion, `--dry-run` sounds useful especially in this context.

pkg/kn/commands/service/import.go
```diff
@@ -0,0 +1,150 @@
...
+			tmp := r
+			// OwnerRef ensures that Revisions are recognized by controller
+			tmp.OwnerReferences = []metav1.OwnerReference{
+				{
+					APIVersion:         currentConf.APIVersion,
+					Kind:               currentConf.Kind,
+					Name:               currentConf.Name,
+					UID:                currentConf.UID,
+					Controller:         ptr.Bool(true),
+					BlockOwnerDeletion: ptr.Bool(true),
```
itsmurugappan, line 122: pls check whether you can use the `kmeta` package to set the owner reference. Below is from the serving repo
https://github.com/knative/serving/blob/5233f0c607c5a89d6a58d6faac9eed57b4877b79/pkg/reconciler/revision/resources/deploy.go#L253

dsimansk: Nice find, thanks! It's exactly the function needed here.

rhuss:
@dsimansk please let me know when this PR is ready from your side (so no WIP anymore), I would then jump on the review, too.

dsimansk:
@rhuss ready now. :smiley_cat: 

/unhold

dsimansk:
@rhuss @navidshaikh @itsmurugappan 

There's one scenario I'm not really sure how to handle. 

- Let's have simple `hello` ksvc without additional revisions
- Exec: `kn service export hello -o yaml --mode export`

<details>
<summary>Cmd output</summary>

```
➜  client git:(service-import) kn service export hello -o yaml --mode export                                  
kind: Service
metadata:
  creationTimestamp: null
  name: hello
spec:
  template:
    metadata:
      annotations:
        client.knative.dev/user-image: docker.io/dsimansk/helloworld
      creationTimestamp: null
      name: hello-djrrc-1
    spec:
      containerConcurrency: 0
      containers:
      - env:
        - name: TARGET
          value: v1
        image: docker.io/dsimansk/helloworld
        name: user-container
        readinessProbe:
          successThreshold: 1
          tcpSocket:
            port: 0
        resources: {}
      enableServiceLinks: false
      timeoutSeconds: 300
status: {}
```
</details>

  - For this case `--mode` flag is ignored and plain ksvc output is produces
  - As a result there's no usable `Export` type object for the `import` to use. 
    - If `import` would support plain ksvc yaml than there's a duplicite functionality with `service create --filename`, `service apply` and `kubectl`.

- Exec: `kn service export hello -o yaml --mode export --with-revisions`
  - Adding `--with-revisions` will produce requested `Export` type

<details>
<summary>Cmd output</summary>

```
➜  client git:(service-import) kn service export hello -o yaml --mode export --with-revisions
apiVersion: client.knative.dev/v1alpha1
kind: Export
metadata:
  creationTimestamp: null
spec:
  revisions: null
  service:
    kind: Service
    metadata:
      creationTimestamp: null
      name: hello
    spec:
      template:
        metadata:
          annotations:
            client.knative.dev/user-image: docker.io/dsimansk/helloworld
          creationTimestamp: null
          name: hello-djrrc-1
        spec:
          containerConcurrency: 0
          containers:
          - env:
            - name: TARGET
              value: v1
            image: docker.io/dsimansk/helloworld
            name: user-container
            readinessProbe:
              successThreshold: 1
              tcpSocket:
                port: 0
            resources: {}
          enableServiceLinks: false
          timeoutSeconds: 300
    status: {}
```

</details>


It feels like inconsistent behaviour of `--mode` flag, that's actually depending on `--with-revisions` to produce expected result.

dsimansk:
```        ┣━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
        🦆 kn service import /tmp/kn-file993807706/foo-with-revisions --namespace kne2etests16
        ┃ 
        🔥 Error: configurations.serving.knative.dev "foo" not found
        🔥 Run 'kn --help' for usage
        🔥 
```
https://github.com/knative/client/pull/1065/files#diff-89bebfee01d13203ce4186956ccc919b30072d4ca985d3c5b054336ce0c735f0R105-R109

It seems that controller was still handling the subsequent creation of `Configuration` for the `Service`.
Would you mind something like `GetConfigurationWithRetry()` or watch?

itsmurugappan:
> It feels like inconsistent behaviour of --mode flag, that's actually depending on --with-revisions to produce expected result.

IMO this is a bug on kn service export, whenever the mode is export , kn should give a yaml in the Export kind instead of KSVC kind.

@rhuss @navidshaikh  please pitch in.

dsimansk:
I've opened a new issue to track export bug. https://github.com/knative/client/issues/1087

pkg/kn/commands/service/import.go
```diff
@@ -0,0 +1,143 @@
...
+			if apierrors.IsNotFound(err) && retries < 5 {
+				retries++
+				time.Sleep(time.Second)
+				continue
+			} else {
+				return err
+			}
+		}
+		break
+	}
```
dsimansk, line 123: I've added a simple retry-loop to retrieve Configuration to verify it it's causing e2e test to fail. The retry can be moved to client like `UpdateServiceWithRetry`. Let me know if there are any objections to actual retry or it could be done differently.

dsimansk: There's also `retry.OnError` from client-go [utils](https://github.com/kubernetes/client-go/blob/master/util/retry/util.go#L48) that's usable here, if there's a need for more sophisticated approach.

navidshaikh: I am with this approach; 

beside I think we can separate this retry logic in a dedicated function, but then if we'd to separate it out anyway we can consume the clint-go utils you linked. Infact we should consider using these utils wherever we find a fit to encourage consumption from these standard libs we are vendoring.  (all this can be done in a subsequent iteration if thats feasible).

pkg/kn/commands/service/import.go
```diff
@@ -0,0 +1,143 @@
...
+
+	// Create revision with current Configuration's OwnerReference
+	if len(export.Spec.Revisions) > 0 {
+		for _, r := range export.Spec.Revisions {
+			tmp := r
```
navidshaikh, line 128: I think we should do deep copy here?

pkg/kn/commands/service/import.go
```diff
@@ -0,0 +1,143 @@
...
+			// OwnerRef ensures that Revisions are recognized by controller
+			tmp.OwnerReferences = []metav1.OwnerReference{*kmeta.NewControllerRef(currentConf)}
+			if err = client.CreateRevision(&tmp); err != nil {
+				return err
+			}
+		}
+	}
+
+	err = waitIfRequested(client, serviceName, waitFlags, "Importing", "imported", out)
```
navidshaikh, line 137: a service has 5 revisions among which 2 are active (configured in traffic block); will service still report true if any of those 3 non-active revisions failed to become ready ?

pkg/kn/commands/service/import.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
```
navidshaikh, line 142:
```suggestion
	return err
```

navidshaikh (commented):
suggested a few nits and a wait for service related question, also please add CHANGELOG entry

pkg/kn/commands/service/import.go
dsimansk: The current `service export` impl doesn't produce such export output, only active revisions are [included](https://github.com/knative/client/blob/master/pkg/kn/commands/service/export.go#L245).

pkg/kn/commands/service/import.go
dsimansk: I like the idea of using standard libs. I've pushed an updated version in https://github.com/knative/client/pull/1065/commits/7b8317c366e2fb6bc1685d29ca5c6e04417dd069. We can start small here and incrementally update `UpdateServiceWithRetry` or any other instances.

pkg/kn/commands/service/import.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return err
+	}
+	decoder := yaml.NewYAMLOrJSONDecoder(file, 512)
+	err = decoder.Decode(&export)
+	if err != nil {
+		return err
+	}
+	if export.Spec.Service.Name == "" {
+		return fmt.Errorf("provided import file doesn't contain service name, please note that only kn' custom Export format is supported")
```
rhuss, line 88:
```suggestion
		return fmt.Errorf("provided import file doesn't contain service name, please note that only kn's custom export format is supported")
```

pkg/kn/commands/service/import.go
```diff
@@ -0,0 +1,143 @@
...
+	svcExists, err := serviceExists(client, serviceName)
+	if err != nil {
+		return err
+	}
+	if svcExists {
+		return fmt.Errorf("cannot import service '%s' in namespace '%s' because the service already exists",
+			serviceName, client.Namespace())
+	}
+
+	err = client.CreateService(&export.Spec.Service)
```
rhuss, line 103: when we create a service like this, then an initial revision is created. This is fine, but I think we might get into ordering issues as the generation of this revision would be 1 (and which might conflict with the revision history that we are importing, where there would be also 1 generation). At least this is what I suspect. (would need to be verified)

pkg/kn/commands/service/import.go
rhuss: I wonder whether we should allow later in `service export`, too, to export all revisions, so that the import then also would import the history. But let's leave that question for future refinement of this feature.

pkg/kn/commands/service/import.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return err
+	}
+	return err
+}
+
+func getConfigurationWithRetry(client clientservingv1.KnServingClient, name string) (*servingv1.Configuration, error) {
+	var conf *servingv1.Configuration
+	var err error
+	err = retry.OnError(retry.DefaultBackoff, func(err error) bool {
```
rhuss, line 136: good find ! I think we have several other spots in the client where do a retry, but we do it in a home grown way. I.e to retry on case of a concurrent modification error. Not sure if the k8s utility method could be used there as well.

pkg/kn/commands/service/import_test.go
```diff
@@ -0,0 +1,277 @@
...
+	assert.NilError(t, err)
+	defer os.RemoveAll(filepath.Dir(file))
+
+	client := knclient.NewMockKnServiceClient(t)
+	r := client.Recorder()
+
+	r.GetService("foo", nil, nil)
+	_, err = executeServiceCommand(client, "import", file)
+
+	assert.ErrorContains(t, err, "because the service already exists")
```
rhuss, line 55: In general, for tests, I would assert that all required context information is included, i.e. here the service name and maybe the namespace (if this is part of the error message). The exact wording is secondary.

rhuss: This is true for the other checks below, too, that check for some messages

pkg/serving/v1/client_mock.go
```diff
@@ -192,6 +192,26 @@ func (c *MockKnServingClient) GetConfiguration(name string) (*servingv1.Configur
 	return call.Result[0].(*servingv1.Configuration), mock.ErrorOrNil(call.Result[1])
 }
 
+// Create a new revision
+func (sr *ServingRecorder) CreateRevision(revision interface{}, err error) {
+	sr.r.Add("CreateRevision", []interface{}{revision}, []interface{}{err})
+}
+
+func (c *MockKnServingClient) CreateRevision(service *servingv1.Revision) error {
```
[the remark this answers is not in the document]
rhuss: Don't get that error really ...

pkg/serving/v1/client_mock.go
```diff
@@ -202,6 +202,28 @@ func (c *MockKnServingClient) GetConfiguration(name string) (*servingv1.Configur
 	return call.Result[0].(*servingv1.Configuration), mock.ErrorOrNil(call.Result[1])
 }
 
+// CreateRevision records a call CreateRevision
```
rhuss, line 205: Could you please also add a call to the recorder and the mock methods to `client_mock_test` ? This keeps the test coverage happy :)

rhuss (commented):
Thanks ! Looks good in general, but I'm not yet 100% sure how it behaves in the wild. E.g. whether the ordering is correct (which is also reflected in `kn revision list`). I'm also not sure if the export includes the latest revision generated from the service in the revision list, because in this case we do have an issue, as the initial service creation implicitly also creates a revision.

We definitely require an E2E test for the long run, to verify the full behaviour. But for now I would be happy to merge this, but also would mark the command as "experimental" (would add this to the changelog.adoc and the command description)

pkg/kn/commands/service/import.go
dsimansk: IMO the concurrent modification was the primary purpose to have such util methods, by the following [comment](https://github.com/kubernetes/client-go/blob/1eb2027cd51eefaca0f851d5a2c90f8dd9b2a71d/util/retry/util.go#L68-L75).

pkg/serving/v1/client_mock.go
dsimansk: Sure, I've missed that.

pkg/kn/commands/service/import.go
dsimansk: Well, your suspicion is exactly on point here, revision history is mangled after import. I'll try to double-check if there's an alternative to the current import order. 
```
➜  client git:(service-import) kn revision list                 
NAME            SERVICE   TRAFFIC   TAGS   GENERATION   AGE   CONDITIONS   READY   REASON
hello-xbgkn-2   hello     25%              2            60s   4 OK / 4     True    
hello-brqcf-3   hello     50%              1            60s   4 OK / 4     True    
hello-kphzr-1   hello     25%              1            60s   4 OK / 4     True   
```

dsimansk: Currently revision list is sorted by 
```
// revisionListSortFunc sorts by namespace, service,  generation and name
```
In fact it's Configuration's generation, [here](// revisionListSortFunc sorts by namespace, service,  generation and name). That's in line with revision history in my comment above. However, I don't see any clean way how to reproduce the history in the import wihout leaning towards "replay" approach actually.

maximilien (approved):
LGTM

pkg/kn/commands/service/import.go
rhuss: 👍 Let me open a separate issue for moving to the more general way for doing retries.

rhuss: https://github.com/knative/client/issues/1102

rhuss (commented):
Beside the issue of building up revision with the proper generation ordering, this PR looks good to me. 

I suggest, to merge this PR as now (it's still marked as experimental), but think about another solution after this.

E.g would it be possible to remove the revision that has been created implicitly when we created the Service and after we have added the revision list ? Also we could experiment with building up a Configuration first and only add the `Service` as a final step (with the same name as the Configuration, as this is the correlation between Service and Configuration). Also, the ownerReference needs between Service and Configuration then would be needed to be set as the last step.

wdyt, should we move forward now ?

CHANGELOG.adoc
```diff
@@ -37,6 +37,10 @@
 | 🎁
 | Add WithLabel list filter to serving client lib
 | https://github.com/knative/client/pull/1054[#1054]
+
+| 🎁
+| Add `kn service export` command (experimental)
```
rhuss, line 42:
```suggestion
| Add `kn service import` command (experimental)
```

dsimansk: Sometimes I really surprise myself. Thanks! :)

dsimansk:
@rhuss sure, sounds good to me. 

I've been experimenting with temp initial revision to recreate generation history correctly without too much luck yet. I can explore the options further, but it shouldn't prevent "experimental" import to be merged now.

rhuss:
Ok, then let's go for it

/Lgtm
/Approve

PR #1069, opened 2020-10-20
Export test refactor

itsmurugappan:
## Description

Refactor kn service export test to use helper functions from serving repo.

First and third commit - Has only vendor changes
Second commit - Has the test refactor changes


lib/test/service.go
```diff
@@ -15,18 +15,28 @@
...
+	"knative.dev/pkg/kmeta"
+	"knative.dev/pkg/ptr"
 	pkgtest "knative.dev/pkg/test"
 	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
+	servingtest "knative.dev/serving/pkg/testing/v1"
 
 	"knative.dev/client/pkg/util"
 )
 
+type ExpectedServiceListOption func(*servingv1.ServiceList)
```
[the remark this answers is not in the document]
navidshaikh: @itsmurugappan : We should add the description to all these exported types as they're in `lib` and are expected to be consumed by plugins etc, it helps in the editor.

lib/test/service.go
```diff
@@ -134,3 +144,171 @@ func GetServiceFromKNServiceDescribe(r *KnRunResultCollector, serviceName string
...
+
+	for _, fn := range options {
+		fn(knExport)
+	}
+
+	return knExport
+}
+
+//Cfg builds servingv1.Configuration with the options provided
+func Cfg(co ...servingtest.ConfigOption) *servingv1.ConfigurationSpec {
```
navidshaikh, line 204: how about renaming this to `ServiceConfig` ?

lib/test/service.go
```diff
@@ -134,3 +144,171 @@ func GetServiceFromKNServiceDescribe(r *KnRunResultCollector, serviceName string
...
+
+var revisionSpec = servingv1.RevisionSpec{
+	PodSpec: corev1.PodSpec{
+		Containers: []corev1.Container{{
+			Image: pkgtest.ImagePath("helloworld"),
+		}},
+		EnableServiceLinks: ptr.Bool(false),
+	},
+	TimeoutSeconds: ptr.Int64(300),
+}
```
navidshaikh, line 227: move the var at the top of the file ?

lib/test/service.go
```diff
@@ -134,3 +144,171 @@ func GetServiceFromKNServiceDescribe(r *KnRunResultCollector, serviceName string
...
+
+var revisionSpec = servingv1.RevisionSpec{
+	PodSpec: corev1.PodSpec{
+		Containers: []corev1.Container{{
+			Image: pkgtest.ImagePath("helloworld"),
+		}},
+		EnableServiceLinks: ptr.Bool(false),
+	},
+	TimeoutSeconds: ptr.Int64(300),
```
navidshaikh, line 226: should use `DefaultRevisionTimeoutSeconds` constant from `knative.dev/serving/pkg/apis/config/defaults.go`

lib/test/service.go
```diff
@@ -134,3 +144,171 @@ func GetServiceFromKNServiceDescribe(r *KnRunResultCollector, serviceName string
...
+}
+
+//GetSvcWithOptions returns ksvc with options provided
```
navidshaikh, line 229:
```suggestion
// GetSvcWithOptions returns ksvc with options provided
```

lib/test/service.go
```diff
@@ -134,3 +144,171 @@ func GetServiceFromKNServiceDescribe(r *KnRunResultCollector, serviceName string
...
+	svc.TypeMeta = metav1.TypeMeta{
+		Kind:       "Service",
+		APIVersion: "serving.knative.dev/v1",
+	}
+	svc.Spec.Template.Spec.Containers[0].Resources = corev1.ResourceRequirements{}
+	return svc
+}
+
+//WithTrafficSplit adds route to ksvc
+func WithTrafficSplit(revisions []string, percentages []int, tags []string) servingtest.ServiceOption {
```
navidshaikh, line 241:
```suggestion
func WithTrafficSpec(revisions []string, percentages []int, tags []string) servingtest.ServiceOption {
```

lib/test/service.go
```diff
@@ -134,3 +144,171 @@ func GetServiceFromKNServiceDescribe(r *KnRunResultCollector, serviceName string
...
+			}
+		}
+		svc.Spec.RouteSpec = servingv1.RouteSpec{
+			Traffic: trafficTargets,
+		}
+	}
+}
+
+//GetRev returns Revision object with the options provided
+func GetRev(name string, options ...servingtest.RevisionOption) servingv1.Revision {
```
navidshaikh, line 265: above, `Cfg` is used to build the configuration and here `GetRev`; all these builder utils should follow the same pattern, IMO we should avoid short names as well. 
wdyt about `BuildRevision` / `BuildConfiguration`?
I'd also avoid `Get` verb here as that denote to client GET call.

lib/test/service.go
```diff
@@ -134,3 +144,171 @@ func GetServiceFromKNServiceDescribe(r *KnRunResultCollector, serviceName string
...
+		Kind:       "Revision",
+		APIVersion: "serving.knative.dev/v1",
+	}
+	rev.Spec.PodSpec.Containers[0].Name = "user-container"
+	rev.ObjectMeta.SelfLink = ""
+	rev.ObjectMeta.Namespace = ""
+	rev.ObjectMeta.UID = ""
+	rev.ObjectMeta.Generation = int64(0)
+	rev.Spec.PodSpec.Containers[0].Resources = corev1.ResourceRequirements{}
+	return *rev
```
navidshaikh, line 277: may pass the pointer instead ? (like done for Configuration above)

navidshaikh (changes requested):
Thanks for promoting these utils to `lib`!

There was a fix landed for the failure in e2e (against master), looks like that's lost during refactor ?
```
        --- knExport
        +++ &actSvc
        [...]

        - 						EnableServiceLinks:        nil,
        + 						EnableServiceLinks:        &false,
        [...]
```

maximilien (approved):
LGTM

Wonder if these vendor changes could have been part of a separate commit/PR?. Never easy to review... Anyhow.

navidshaikh (changes requested):
a few places left in the code which can be aligned with the latest improvements made to the PR, i.e. to use full name `Revision` over `Rev`. Also, if the builder function takes a single object of a type, IMO we should consider using singular name for the function, for example `WithRevision`.

maximilien:
@itsmurugappan you need to rebase to pickup latest deps

navidshaikh (approved):
/lgtm
/approve

Thank you @itsmurugappan for the refactor! 
There are a few other parts for example the flags definition etc that we'd need regroup in a better way, lets talk about it in upcoming WG call.

navidshaikh:
@itsmurugappan : also please send another with an entry in CHANGELOG :-)

itsmurugappan:
sure @navidshaikh can include the changelog and other changes in the pr for bug fix #1087

PR #986, opened 2020-08-20
[WIP] Added range options to --scale

mpetason:
## Description

Adds range options to --scale. 

## Changes

* Adds three more options to --scale (1..5, 1.., ..5)

## Reference

Fixes #822

PR #1070, opened 2020-10-22
Add range options to --scale

### LINKED ISSUE #822: support --scale

mpetason:
## Description

Adds range options to --scale

## Changes

* Changed Scale from Int to String
* Added Options to set min/max with 1..5

## Reference

Fixes #822 

Replaces previous PR where I had issues with a rebase - #986


mpetason:
I'm not sure about the failed lint test - maybe I can get more information in a review.

When setting 1.. or ..5 it defaults to 0 <-- I still need to resolve this. I'll write an additional test to cover this as well.

docs/cmd/kn_service_create.md
```diff
@@ -88,7 +88,7 @@ kn service create NAME --image IMAGE
       --requests-cpu string               DEPRECATED: please use --request instead. The requested CPU (e.g., 250m).
       --requests-memory string            DEPRECATED: please use --request instead. The requested memory (e.g., 64Mi).
       --revision-name string              The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
-      --scale int                         Minimum and maximum number of replicas.
+      --scale string                      Set the Minimum and Maximum number of replicas. You can use this flag to set both to a single value, or set a range with min/max values, or set either min or max values without specifying the other. Example: --scale 5 or --scale 1..5 or --scale 1.. or --scale ..5. (default "1")
```
maximilien, line 91: I would not use a “.” in the last example “or —scale ..5” since it’s confusing as to whether that last character is part of the syntax

pkg/kn/commands/service/create_test.go
```diff
@@ -618,6 +618,116 @@ func TestServiceCreateScaleWithMinScaleSet(t *testing.T) {
...
 }
 
+func TestServiceCreateScaleRange(t *testing.T) {
+	action, created, _, err := fakeServiceCreate([]string{
+		"service", "create", "foo", "--image", "gcr.io/foo/bar:baz",
+		"--scale", "1..5", "--no-wait"}, false)
+
+	if err != nil {
+		t.Fatal(err)
+	} else if !action.Matches("create", "services") {
```
maximilien, line 628: I would separate them as two different checks. No need to be if/else since either execution will stop and fail the test.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -560,3 +565,33 @@ func (p *ConfigurationEditFlags) AnyMutation(cmd *cobra.Command) bool {
 	}
 	return false
 }
+
+// Helper function for --scale
+func (p *ConfigurationEditFlags) scaleConversion(scale string) (scaleMin int, scaleMax int, err error) {
+	if len(scale) <= 2 {
```
maximilien, line 571: My preference would be to divide this into smaller private functions for each of the cases... might simply things. Also for each case returning early could also simply code since once case identified the result is there. No need for big if/else if/else.

maximilien (changes requested):
Missing a test case for case when user only specifies separator or has too many separators, e.g.,

1. —scale ..
2. —scale 1..5..

Also couple more minor feedback. Thanks for contribution.

docs/cmd/kn_service_create.md
navidshaikh: with each example listed we can add what does that pattern represent in `()` ? like `'--scale 1..' (min-scale=1, max scale undefined)` and also single quote the example themselves

pkg/kn/commands/service/configuration_edit_flags.go
mpetason: If I break up the helper function into multiple private functions it will complicate the cmd.Flags().Changed("scale") section, is that ok? I was trying to minimize how large that code block was going to be.

docs/cmd/kn_service_create.md
mpetason: Sounds good. I'll make this update - have been working on breaking up the helper function at the moment. This will get added to my next commit. I've also updated the formatting so that it looks like other examples in the configuration_edit_flags.go file.

mpetason:
TODO: 
Break up helper function and update cmd.Flags() section.
Fix issues where undefined value defaults to 0.

Fixed:
Addressed the easier updates based on reviews.

rhuss:
@mpetason please let me know if I can jump on a final review.

pkg/kn/commands/service/create_test.go
mpetason: I separated these, however the original blocks were copied from other tests and I was trying to keep it consistent. Many of the other code blocks in create_test and update_test use:

```
	if err != nil {
		t.Fatal(err)
	} else if !action.Matches("create", "services") {
		t.Fatalf("Bad action %v", action)
	}
```

Should I revert this back for consistency?

mpetason:
@rhuss How can I resolve this issue? This is mostly a copy/paste from other tests. Do we need to create a function for this and just have each test call it to reduce the amount of times it's duplicated? 

 ```
pkg/kn/commands/service/update_test.go#L529

[Redundant Format] reported by reviewdog 🐶
Please consider avoiding tail format like this:
t.Fatalf("Bad action %v", action)

Raw Output:
./pkg/kn/commands/service/update_test.go:529: Please consider avoiding tail format like this:
t.Fatalf("Bad action %v", action)
```

File: pkg/kn/commands/service/configuration_edit_flags.go (hunk 3 above)
Comment (mpetason) [reply pkg/kn/commands/service/configuration_edit_flags.go:579]: Working on breaking this up next - I should be able to avoid complication in the cmd.Flags().Changed("scale") section.

File: docs/cmd/kn_service_create.md [hunk 4]
```diff
@@ -88,7 +88,7 @@ kn service create NAME --image IMAGE
       --requests-cpu string               DEPRECATED: please use --request instead. The requested CPU (e.g., 250m).
       --requests-memory string            DEPRECATED: please use --request instead. The requested memory (e.g., 64Mi).
       --revision-name string              The revision name to set. Must start with the service name and a dash as a prefix. Empty revision name will result in the server generating a name for the revision. Accepts golang templates, allowing {{.Service}} for the service name, {{.Generation}} for the generation, and {{.Random [n]}} for n random consonants. (default "{{.Service}}-{{.Random 5}}-{{.Generation}}")
-      --scale int                         Minimum and maximum number of replicas.
+      --scale string                      Set the Minimum and Maximum number of replicas. You can use this flag to set both to a single value, or set a range with min/max values, or set either min or max values without specifying the other. Example: --scale 5 (scale-min = 5, scale-max = 5) or --scale 1..5 (scale-min = 1, scale-max = 5) or --scale 1.. (scale-min = 1, scale-max = undefined) or --scale ..5 (scale-min = undefined, scale-max = 5) (default "1")
```
duglin, line 91: I think saying “undefined” isn’t as clear as saying that the current value is unchanged. And there is no default value so I would remove that.

mpetason:
Split up the helper function, still working on getting the correct values set when something isn't defined. Right now it sets the default to 0 when it isn't set, which isn't what we want for something like "..5" when using "update."

docs/cmd/kn_service_create.md
rhuss: +1 for putting literal examples in quotes, but i would suggest double quotes (") as in the example for `--revision-name` (just to be consistent).

pkg/kn/commands/service/create_test.go
rhuss: if this is the case, then typically I would recommend to move the shared code pattern into a dedicated function for validation like in a `validateAction(t, action, err, []string{ "create", "services")` to guarantee that the same code is used everywhere.

However as this is not in scope for this pr I would recommend to do this in a separate pr and for now just keep the pattern so that it can be easily detected when doing the full refactoring.

rhuss:
> @rhuss How can I resolve this issue? This is mostly a copy/paste from other tests. Do we need to create a function for this and just have each test call it to reduce the amount of times it's duplicated?

@mpetason This warning is kind of stupid and we should disable it (have to find out, how to do this). You can avoid this to use string concatenation here: `t.Fatal("Bad action %v" + action)`

docs/cmd/kn_service_create.md
rhuss: I don't think that is is "unchanged" as we are here talking about a create operation. Also for an update operation I still think that `--scale 1..` mean min-scale is set to one, but no max-scale is set or removed if set, so that there is no upper bound specified  by the user (regardless what was specified before). If you want to change *only* the minimal scale, you should use `--min-scale`.

+1 that the default value `1` is wrong as there is supposed to be no default at all, the argument is mandatory.

rhuss: So maybe instead of `scale-max = undefined`, better use `no maximum scale defined`) ?

rhuss:
> Split up the helper function, still working on getting the correct values set when something isn't defined. Right now it sets the default to 0 when it isn't set, which isn't what we want for something like "..5" when using "update."

Tbh, I think this is exactly what is should do, setting min-scale to 0. For min-scale we could say 0 as lower boundary and `..2` for unspecified, but that does not work max-scale. I would say (and that was the initial intention, too, when I suggested this feature. however rereading the original issue, I probably didn't say that properly) to consider `..` really as a range operator, that completely specifies the scale. So it could be a range with is unbounded to one side, but this means 0 or infinity, i.e. "not set". Even a `..` could be possible to completely remove all scale parameters (unbounded in both directions). If we don't allow that, there is no way to remove a scale-max once is set.

rhuss:
Ok, I probably just added some confusion with my latest comments (and sorry that I'm back to the game that late). The _range operator_ (`..`) as introduced here can be interpreted in two ways when one or both ends are not given:

* _unbounded_ at that end (e.g. `1..` means 1,2,3 .... infinity (but in reality the maximum as configured at the backend)
* _unchanged_ at that end (e.g. `1..` leave any max-scale untouched  if set previously)

We have this problem only for the scale-max end as for scale-min we can have both `0..5` and `..5` to possibly distinguish these two cases.

There are two ways how we could resolve this dilemma:

* Introduce a special character for "infinity" (but that makes it even more complicated to communicate this to the user)
* Decide between "unbounded" and "unchanged". 

I tend to the second solution and would select "unbounded" as this reflects the similar mathematical notation of a half-open sequence. Also this would allow us to remove a max-scale once set (which would not be possible if the interpretation "unchanged" is used).

@mpetason Sorry again for this late game-change, I have to re-read the argumentation of the initial issue and check the reasoning over there. 

Happy to check out also alternative solutions (like introducing a special character meaning "unbounded").

mpetason:
@rhuss Thank you for the updates. I'll work on a solution and then reach out for code reviews so we can figure out what ends up working out the best.

docs/cmd/kn_service_create.md
duglin: I'm confused. @rhuss in the past I think you said you wanted `1..` to mean only the lower bound is set, and upper bound is untouched.  That's why I commented the way I did - I thought you still wanted that.

Let's back up a bit... at one point people wanted to deprecate --scale-min and only have --scale - is that still in the cards or is that idea dead? If --scale-min/max will live on, then I'm ok with a missing value in `..` meaning unset that value (or in the 'create' case it means just leave it as the default).  And yes `--scale ..` would mean unset both :-)

mpetason:
Running build again doesn't update the docs, not sure how to resolve the current failing test.

maximilien (commented):
Flake perhaps.
/retest

rhuss:
@mpetason You need to rebase you local branch so that you get the newly introduced `kn service apply` also locally and the regenerate the docs with `hack/build.sh`. The reason for the build failure is, that the CI did the rebase but you did not when committing, and as `kn service apply` use also the scale options, it is affected by the update.

Just try a 

```
git pull upstream master; t rebase upstream/master
```

mpetason:
@rhuss that didn't work, I guess I'll close this issue and try again.

PR #1075, opened 2020-10-27
Uniform multiple writeSink(s) to DescribeSink

daisy-ycguo:
## Description

There are multiple `writeSink` in source commands to print Sink object. And there is a utility `DescribeSink` in `lib/printing`. It's better to uniform them.

## Changes

* Uniform multiple writeSink(s) to a single WriteSink

### DIFF
--- a/lib/printing/describe.go
+++ b/lib/printing/describe.go
@@ -24,7 +24,7 @@ import (
 )
 
 // DescribeSink prints the given 'sink' for the given prefix writer 'dw',
-// provide 'attribute' to print the section heading for this sink
+// provide 'attribute' to print the section heading for this sink,
 func DescribeSink(dw printers.PrefixWriter, attribute, namespace string, sink *duckv1.Destination) {
 	if sink == nil {
 		return
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -18,6 +18,10 @@
 | | Description | PR
 
 | 🐣
+| Uniform multiple writeSink to DescribeSink
+| https://github.com/knative/client/pull/1075[#1075]
+
+| 🐣
 | Add function ResolvePodSpec and move the utilities to podspec_helper
 | https://github.com/knative/client/pull/1024[#1024]
 
--- a/pkg/kn/commands/source/apiserver/describe.go
+++ b/pkg/kn/commands/source/apiserver/describe.go
@@ -21,8 +21,8 @@ import (
 
 	"github.com/spf13/cobra"
 	v1alpha2 "knative.dev/eventing/pkg/apis/sources/v1alpha2"
-	duckv1 "knative.dev/pkg/apis/duck/v1"
 
+	"knative.dev/client/lib/printing"
 	"knative.dev/client/pkg/kn/commands"
 	"knative.dev/client/pkg/printers"
 )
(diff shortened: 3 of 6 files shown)


pkg/kn/commands/flags/sink.go
```diff
@@ -147,3 +148,18 @@ func SinkToString(sink duckv1.Destination) string {
...
+	if ref != nil {
+		subWriter.WriteAttribute("Name", sink.Ref.Name)
+		subWriter.WriteAttribute("Namespace", sink.Ref.Namespace)
+		subWriter.WriteAttribute("Resource", fmt.Sprintf("%s (%s)", sink.Ref.Kind, sink.Ref.APIVersion))
+	}
+	uri := sink.URI
+	if uri != nil {
+		subWriter.WriteAttribute("URI", uri.String())
+	}
+}
```
navidshaikh, line 165: while doing the subscription CRUD, separated this util here https://github.com/knative/client/blob/abb75e4da2c45534593bb2684dc3253eef3a7d91/lib/printing/describe.go#L28

lets refer this util at following places

navidshaikh (changes requested):

pkg/kn/commands/flags/sink.go
daisy-ycguo: Thank you for the information. I updated. Please review again.

lib/printing/describe.go
```diff
@@ -29,6 +29,9 @@ func DescribeSink(dw printers.PrefixWriter, attribute, namespace string, sink *d
 	if sink == nil {
 		return
 	}
+	if len(attribute) == 0 {
```
rhuss, line 32: It looks like that this method is never called with an `attribute` that is different to "". So I just would remove that argument and maybe re-introduce it if there is ever a need for a different label than `Sink`.

rhuss (commented):
Looks good, with a minor change request.

lib/printing/describe.go
navidshaikh: @rhuss : subscription CRUD uses it for DeadLetterSink etc

https://github.com/knative/client/blob/abb75e4da2c45534593bb2684dc3253eef3a7d91/pkg/kn/commands/subscription/describe.go#L105

navidshaikh: @daisy-ycguo : can you please update the function docstring to mention this behavior ? (like `'Sink' is used for attribute value if empty string is provided`)

daisy-ycguo: sure. I will update the docstring.

rhuss: I see. But maybe it's just better to require an explicit argument ? This could then also serve as documentation when calling the function (so just changing "" everywhere to "Sink") ? Not sure about this, so feel free to keep it .

Yet an alternative would be to move `attribute` to the end of the argument list and make it an optional parameter like `attribute ...string`. But I'm not dogmatic here, feel to chose what you think fits best.

navidshaikh: > But maybe it's just better to require an explicit argument ? This could then also serve as documentation when calling the function (so just changing "" everywhere to "Sink") ? Not sure about this, so feel free to keep it .

I was thinking same about it when I first saw this. As we are exporting this util, IMO we shouldn't carry any assumption and require explicit attribute value.

maximilien (approved):
/ok-to-test

lib/printing/describe.go
daisy-ycguo: I changed the code to make mandatory.

navidshaikh (approved):
/lgtm
/approve

PR #1077, opened 2020-10-27
use the new hack repo for scripts

n3wscott:
## Proposed Changes

- test-infra/scripts -> hack

ref: https://github.com/knative/hack/pull/2


hack/update-deps.sh
```diff
@@ -14,52 +14,10 @@
...
-readonly ROOT_DIR=$(dirname $0)/..
-source ${ROOT_DIR}/scripts/test-infra/library.sh
-
 set -o errexit
 set -o nounset
 set -o pipefail
 
-cd ${ROOT_DIR}
-
-VERSION="release-0.18"
```
n3wscott, line 26: not sure why client was still building for release 0.18.

n3wscott:
it seems client _is_ still on 0.18. Drat. we are 2 weeks away from 0.19 cutting

maximilien (approved):
/ok-to-test

hack/update-deps.sh
navidshaikh: /assign

missed updating it back to master
currently the build tests are complaining, will fix in a separate PR

navidshaikh: PTAL #1082

rhuss:
> it seems client _is_ still on 0.18. Drat. we are 2 weeks away from 0.19 cutting

@n3wscott sorry, about to fix this with #1082 . When this is merged, please rebase it then on top of main again and the we merge this PR soon after.

n3wscott:
Waiting for your lead @rhuss 

Note I drop the scripts folder and use go mod again to get you the base hack dir

rhuss:
@n3wscott happy to jump to a `go.mod` dependency for `hack` as I understood that the promise is, that it won't bring in additional go dependencies ?

I'm not sure if we should continue with this pr because of the conflicts that the update on main has caused, or maybe its better to just start afresh from the current head ?

n3wscott:
> @n3wscott happy to jump to a `go.mod` dependency for `hack` as I understood that the promise is, that it won't bring in additional go dependencies ?
> 
> I'm not sure if we should continue with this pr because of the conflicts that the update on main has caused, or maybe its better to just start afresh from the current head ?


that is correct, knative.dev/hack will have no dependencies. I will continue to push that [this](https://github.com/knative/hack/blob/master/go.mod) stays empty.

PR #1082, opened 2020-10-30
Pin deps back to master

navidshaikh:
## Description
Pin deps back to master, this was left over task after 0.18 release

Steps:
1. 
```
# edit hack/update-deps.sh and update VERSION="master"
./hack/update-deps.sh --upgrade
./hack/update-codegen.sh
```

2. Fix any build errors

## Changes
* Update autoscaling annotation validation
    see ref: https://github.com/knative/serving/pull/9496
* Replace DeprecatedInjectionAnnotation with InjectionAnnotation
* Vendor deps at master


pkg/serving/config_changes.go
```diff
@@ -164,7 +166,10 @@ func UpdateServiceAnnotations(service *servingv1.Service, toUpdate map[string]st
 // UpdateRevisionTemplateAnnotations updates annotations for the given Revision Template.
 // Also validates the autoscaling annotation values
 func UpdateRevisionTemplateAnnotations(template *servingv1.RevisionTemplateSpec, toUpdate map[string]string, toRemove []string) error {
-	if err := autoscaling.ValidateAnnotations(true, toUpdate); err != nil {
+	ctx := context.Background()
+	autoscalerConfig := servingconfig.FromContextOrDefaults(ctx).Autoscaler
+	autoscalerConfig.AllowZeroInitialScale = true
+	if err := autoscaling.ValidateAnnotations(ctx, autoscalerConfig, toUpdate); err != nil {
 		return err
```
navidshaikh, line 173: this allows users to specify `scale-init=0` during service create/update ; IIRC I think the final validation would kick in based on the defaults set at serving level? /cc @markusthoemmes

pkg/serving/config_changes.go
```diff
@@ -164,7 +166,10 @@ func UpdateServiceAnnotations(service *servingv1.Service, toUpdate map[string]st
 // UpdateRevisionTemplateAnnotations updates annotations for the given Revision Template.
 // Also validates the autoscaling annotation values
 func UpdateRevisionTemplateAnnotations(template *servingv1.RevisionTemplateSpec, toUpdate map[string]string, toRemove []string) error {
-	if err := autoscaling.ValidateAnnotations(true, toUpdate); err != nil {
+	ctx := context.Background()
```
rhuss, line 169: would change here to `context.TODO()` as we don't really want to create a new context, but indicate that we just not support this when entering this library method (we could consider todo this later). Technically its the same (a new empty context is created), but semantically its a help to indicate that we are just not offering context support.

as a side note, i recently listened to https://changelog.com/gotime/143 which explained the difference nicely, too (plus a lot of other interesting information)

pkg/eventing/v1beta1/client.go
```diff
@@ -183,10 +183,10 @@ func (b *TriggerBuilder) Broker(broker string) *TriggerBuilder {
 // InjectBroker to add annotation to setup default broker
 func (b *TriggerBuilder) InjectBroker(inject bool) *TriggerBuilder {
 	if inject {
-		meta_v1.SetMetaDataAnnotation(&b.trigger.ObjectMeta, v1beta1.DeprecatedInjectionAnnotation, "enabled")
+		meta_v1.SetMetaDataAnnotation(&b.trigger.ObjectMeta, v1beta1.InjectionAnnotation, "enabled")
 	} else {
-		if meta_v1.HasAnnotation(b.trigger.ObjectMeta, v1beta1.DeprecatedInjectionAnnotation) {
-			delete(b.trigger.ObjectMeta.Annotations, v1beta1.DeprecatedInjectionAnnotation)
+		if meta_v1.HasAnnotation(b.trigger.ObjectMeta, v1beta1.InjectionAnnotation) {
```
rhuss, line 188: funny ping pong with the `Deprecated` Prefix. Really not a big fan of this way of deprecation that already breaks the API (so it's kinda funny to deprecate and break backwards compatibility at the same time :). I already mentioned that over at serving, so maybe that's the reason for reverting this renaming?

hack/update-deps.sh
```diff
@@ -23,7 +23,7 @@ set -o pipefail
 
 cd ${ROOT_DIR}
 
-VERSION="release-0.18"
+VERSION="master"
```
rhuss, line 26: Looks good. In a next step we then switch over to the hack repo ...

rhuss (approved):
Looks good to me with a small suggestion for improvements.

pkg/eventing/v1beta1/client.go
navidshaikh: how I understood the constant deprecation here is; there is a time frame (3 releases IIRC) the deprecated values are honored and during this the consumers are expected to move to newer constants. For client, we do it at the end of the time frame when the constant is no longer available in the vendored library. Beside, I think the respective controller was considering both deprecated and new values; but after a point it doesn't. So with this value, client should be able to work with earlier 3 releases, please correct me if I got this wrong.

pkg/serving/config_changes.go
navidshaikh: now using context.TODO
thanks for link! :-)

navidshaikh:
/retest

        Error: timeout: service 'hello' not ready after 600 seconds

pkg/eventing/v1beta1/client.go
n3wscott: We shipped both `DeprecatedInjectionAnnotation` and `InjectionAnnotation`, you could have continued to use `InjectionAnnotation` and it would have worked.

PR #1083, opened 2020-10-30
Cross-compile the kn binary for linux/s390x

BbolroC:
## Description

As part of the work to support the `knative` for the IBM Z system (`s390x` architecture), this PR is to make the `kn-linux-s390x` binary available in the `release`.

## Changes

* Add an additional `go build` command with `GOOS=linux GOARCH=s390x` in `hack/build.sh` and `hack/release.sh`, respectively.

## Reference

None

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -282,6 +282,8 @@ cross_build() {
   GOOS=darwin GOARCH=amd64 go build -mod=vendor -ldflags "${ld_flags}" -o ./kn-darwin-amd64 ./cmd/... || failed=1
   echo "   🎠 kn-windows-amd64.exe"
   GOOS=windows GOARCH=amd64 go build -mod=vendor -ldflags "${ld_flags}" -o ./kn-windows-amd64.exe ./cmd/... || failed=1
+  echo "   Z  kn-linux-s390x"
+  GOOS=linux GOARCH=s390x go build -mod=vendor -ldflags "${ld_flags}" -o ./kn-linux-s390x ./cmd/... || failed=1
 
   return ${failed}
 }
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -37,9 +37,11 @@ function build_release() {
   GOOS=darwin GOARCH=amd64 go build -mod=vendor -ldflags "${ld_flags}" -o ./kn-darwin-amd64 ./cmd/...
   echo "🚧 🎠 Building for Windows"
   GOOS=windows GOARCH=amd64 go build -mod=vendor -ldflags "${ld_flags}" -o ./kn-windows-amd64.exe ./cmd/...
+  echo "🚧 Z  Building for Linux(s390x)"
+  GOOS=linux GOARCH=s390x go build -mod=vendor -ldflags "${ld_flags}" -o ./kn-linux-s390x ./cmd/...
   echo "🚧 🐳 Building the container image"
   ko resolve --strict ${KO_FLAGS} -f config/ > kn-image-location.yaml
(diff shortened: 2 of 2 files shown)


hack/build.sh
```diff
@@ -284,6 +284,8 @@ cross_build() {
   GOOS=darwin GOARCH=amd64 go build -mod=vendor -ldflags "${ld_flags}" -o ./kn-darwin-amd64 ./cmd/... || failed=1
   echo "   🎠 kn-windows-amd64.exe"
   GOOS=windows GOARCH=amd64 go build -mod=vendor -ldflags "${ld_flags}" -o ./kn-windows-amd64.exe ./cmd/... || failed=1
+  echo "   Z  kn-linux-s390x"
```
rhuss, line 287:
```suggestion
  echo "   🗡️  kn-linux-s390x"
```

We love emojis ;-) What's about a dagger like the one that [Zorro](https://en.wikipedia.org/wiki/Zorro) used ?

hack/release.sh
```diff
@@ -35,9 +35,11 @@ function build_release() {
   GOOS=darwin GOARCH=amd64 go build -mod=vendor -ldflags "${ld_flags}" -o ./kn-darwin-amd64 ./cmd/...
   echo "🚧 🎠 Building for Windows"
   GOOS=windows GOARCH=amd64 go build -mod=vendor -ldflags "${ld_flags}" -o ./kn-windows-amd64.exe ./cmd/...
+  echo "🚧 Z  Building for Linux(s390x)"
```
rhuss, line 38:
```suggestion
  echo "🚧 🗡️  Building for Linux(s390x)"
```

see comment above.

rhuss (commented):
Looks good to me, thanks for the PR ! (a minor suggestion added).

Technically that's ok (and affair that's the second request for adding s390), so I tend to add this platform, too. I wonder whether we then should go all in and also add other popular platforms  like armv7 or arm64.

Also, we then should push multiarch OCI image for our binary.

Wdyt, @maximilien @navidshaikh and others ? Let's discuss this on the next WG call how wide we want to extend our arch support.

maximilien (approved):
/ok-to-test

hack/build.sh
BbolroC: Thanks for the suggestion. I love a dagger, but let me check that with my manager first if the emoji, which does not representing the Z system, can be used in the open source. (actually I spent more time on choosing a suitable emoji than the PR itself and ended up using the alphabet, which is not a proprietary asset)

Could you please hold this PR until I get an answer for this?

rhuss: sure. btw, we are using also non official emoji for all the other OS, and the emoji is not meant to represent an official product logo, but just as kind of a fancy "bullet point" in the list. We also never talk about the architecture itself here, as you see it's just the name of the binary that we produce.

BbolroC: Yop, I have the same opinion. I am just wondering if the company thinks the same. ^^

BbolroC: I got the decision from the company about the emoji: just sticking to the capital z as it is. Is it okay to go with it?

maximilien (approved):
LGTM

maximilien:
Looks like you need to rebase @BbolroC file: `hack/release.sh`

BbolroC:
> Looks like you need to rebase @BbolroC file: `hack/release.sh`

Thanks for the notice. The rebase is done.

navidshaikh (approved):
/lgtm
/approve
could've made to v0.19 release , but we can consider this for patch release as well?
/hold for @rhuss

rhuss:
> could've made to v0.19 release , but we can consider this for patch release as well?

+1 one for backporting this to 0.18 and 0.19

/unhold

PR #1094, opened 2020-11-06
Drop deprecated --async flag support

### LINKED ISSUE #1089: Drop deprecated --async flag in favor of --no-wait
Flag `--async` was deprecated but still supported in the favor of `--no-wait`, introduced in v0.13.0 release. Its been 5 releases since then and we can consider removing support for `--async` now. Please consider moving to `--no-wait` if you are using `--async`.

navidshaikh:
## Description
Drop deprecated --async flag support  in the favor of --no-wait flag

## Changes
* Drop --async flag
* Remove the respective test cases
* Add CHANGELONG

## Reference

Fixes #1089

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -18,6 +18,10 @@
 | | Description | PR
 
 | 🐣
+| Drop deprecated --async flag support
+| https://github.com/knative/client/pull/1094[#1094]
+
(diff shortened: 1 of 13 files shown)


rhuss (commented):
Thanks ! 

/lgtm
/approve

PR #1107, opened 2020-11-10
fix: Fix var initialization in lib/service function

dsimansk:
## Description

I ran into the issue with E2E execution and `--imagetemplate`, when `TestServiceExport` fails comparing the wrong image value. I've found out that actual RevisionSpec used for the expected value is initialized with wrong value, hence the fix should mitigate that.

```bash
./test/local-e2e-tests.sh -run ^TestServiceExport$ --imagetemplate docker.io/dsimansk/{{.Name}}

# fails with
...
        - 								Image:   "/helloworld:latest",
        + 								Image:   "docker.io/dsimansk/helloworld",


```

## Changes

* Add `BuildRevisionSpec` to lib/test/service.go to fix the issue

## Reference

Fixes N/A

### DIFF
--- a/lib/test/service.go
+++ b/lib/test/service.go
@@ -43,16 +43,6 @@ type ExpectedRevisionListOption func(*servingv1.RevisionList)
 // ExpectedKNExportOption enables further configuration of a Export.
 type ExpectedKNExportOption func(*clientv1alpha1.Export)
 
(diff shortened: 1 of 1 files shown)


navidshaikh (approved):
/lgtm

PR #1113, opened 2020-11-11
Fix race conditions when creating watches

### LINKED ISSUE #1109: Race condition in commands using watch
### Bug report

The `wait.Wait.Wait()` implementations create the watch they are listen to in a goroutine that can execute later than the code that emits the event the watch was intended for.

### Expected behavior

The watch is created upfront, then the wait for the event is executed in a goroutine while the main thread calls the backend that will emit the expected event.

I was able reproduce this by:
- adding 30 seconds sleep into `vi/client.go`'s `WatchRevision` method before the watch is created on the backend
- call `kn revision delete --wait <rev_name>`

### kn version

HEAD

N/A

akerekes:
## Description

Fix a race condition between creating a watch and initiating the action that emits the event it is watching for.

Currently the wait.Wait function creates the watch and is invoked in a goroutine, i.e. parallel to the command whose result the watch is targeted for. If the watch creation takes longer in the goroutine than the main execution thread's call to the backend, the event generated by the change can be emitted sooner than the watch is listening and it will wait there until timeout happens.

## Changes

* Update `wait.Wait.Wait()` to accept a watch, instead of the implementing types creating it
* Update callers of `wait.Wait.Wait()` to create the watch upfront

## Reference

Fixes #1109

/lint


navidshaikh (approved):
/assign @rhuss

rhuss:
Thanks ! looking at it tomorrow ...

pkg/serving/v1/client.go
```diff
@@ -319,12 +319,17 @@ func (cl *knServingClient) DeleteService(serviceName string, timeout time.Durati
...
+	watcher, err := cl.WatchService(serviceName, timeout)
+	if err != nil {
+		return nil
+	}
+	defer watcher.Stop()
 	go func() {
-		waitForEvent := wait.NewWaitForEvent("service", cl.WatchService, func(evt *watch.Event) bool { return evt.Type == watch.Deleted })
-		err, _ := waitForEvent.Wait(serviceName, wait.Options{Timeout: &timeout}, wait.NoopMessageCallback())
+		waitForEvent := wait.NewWaitForEvent("service", func(evt *watch.Event) bool { return evt.Type == watch.Deleted })
+		err, _ := waitForEvent.Wait(watcher, serviceName, wait.Options{Timeout: &timeout}, wait.NoopMessageCallback())
```
rhuss, line 329: This closure works as long as watcher is in scope. Not sure how it works in go, but if this variable would go out of scope (e.g. when this function where it is allocated is left), wouldn't it be garbage collected and the go routine (in case it is still running) would then access a zombie watcher object ?

I know in this case this is a theoretical question, as we block on the channel before leaving the function, and only the goroutine can send to the channel.

So it's more a theoretical question, I guess :)

rhuss (approved):
Good catch ! Looks good to me (one minor question though), and sorry for the delay in reviewing this.

/approve
/lgtm

Issue #1100, opened 2020-11-06
kn channel create does not honour `default-ch-webhook` settings


kameshsampath:
### Bug report

Using the docs https://knative.dev/docs/eventing/channels/default-channels/, we can configure the default channel, assume Kafka is installed and I configure the `default-ch-webhook` CM to be like:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: default-ch-webhook
  namespace: knative-eventing
data:
  default-ch-config: |
    clusterDefault:
      apiVersion: messaging.knative.dev/v1
      kind: InMemoryChannel
    namespaceDefaults:
      knativetutorial:
        apiVersion: messaging.knative.dev/v1beta1
        kind: KafkaChannel
        spec:
          numPartitions: 2
          replicationFactor: 1
```
__NOTE__:  Kafka is configured and running correctly and so is KafkaChannel resource

When I do `kn channel create my-events-ch -n knativetutorial`, I expect the channel to be created as KafkaChannel type but its getting defaulted to `InMemoryChannel`

e.g. 
```shell
$ kn channel -n knativetutorial ls

NAME           TYPE              URL                                                                AGE     READY   REASON
my-events-ch   InMemoryChannel   http://my-events-ch-kn-channel.knativetutorial.svc.cluster.local   4m33s   True
```

### Expected behavior

```shell
cat <<EOF | kubectl create -f  -
apiVersion: messaging.knative.dev/v1
kind: Channel
metadata:
  name: my-events-ch
  namespace: knativetutorial
EOF
```

```shell
 kn channel ls
NAME           TYPE           URL                                                                AGE   READY   REASON
my-events-ch   KafkaChannel   http://my-events-ch-kn-channel.knativetutorial.svc.cluster.local   13s   True
```

### Steps to reproduce the problem

1. Set default channel webhook settings to be like 

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: default-ch-webhook
  namespace: knative-eventing
data:
  default-ch-config: |
    clusterDefault:
      apiVersion: messaging.knative.dev/v1
      kind: InMemoryChannel
    namespaceDefaults:
      knativetutorial:
        apiVersion: messaging.knative.dev/v1beta1
        kind: KafkaChannel
        spec:
          numPartitions: 2
          replicationFactor: 1
```

2. Run the command `kn channel create -n knativetutorial my-events-ch`

### kn version

```
Version:      v0.18.0
[...]

PR #1117, opened 2020-11-12
Embed the namespace in request body while creating channels

navidshaikh:
## Description
 Embed the namespace in request body while creating channels
 since on the eventing side, defaulting for channel isnt picking
 the namespace from the context (see https://github.com/knative/eventing/issues/4514)

 workaround for #1100
 this changeset should be reverted when eventing#4514 is resolved

## Changes
- Add namespace in the channel object meta while creating the channels

## Reference
workaround for #1100 and https://github.com/knative/eventing/issues/4514

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -12,6 +12,16 @@
 | https://github.com/knative/client/pull/[#]
 ////
 
+### Unreleased
+[cols="1,10,3", options="header", width="100%"]
+|===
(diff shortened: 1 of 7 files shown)


rhuss (commented):
Looks good to me, but please add a Changelog entry, too. I think this bug is worth mentioning in there.

rhuss (approved):
/lgtm

PR #1119, opened 2020-11-12
Fix a few more occurrences of divisive language

markusthoemmes:
As per title.

- Flagging all whitespace occurrences with a TODO.
- Removing "dummy" and "sanity check".


rhuss (approved):
/lgtm
/approve

PR #1122, opened 2020-11-17
add git ops options

### LINKED ISSUE #30: Gitops mode

itsmurugappan:
## Description

Adds Gitops mode where in the changes are made in a local directory instead of a remote cluster

## Changes

* This PR only enables the option for kn service create/delete/update/list/describe
* Gitops client implementing the  knservingclient interface
* Client injection is based on the command
~* Tests pending~
* e2e pending

Need reveiwers input on the below

* When '--in-dir' flag is provided this mode is enabled
* Named the new client gitops client

## Reference

Partially fixes #30 

/hold

cc @rhuss @navidshaikh @dsimansk @maximilien

### DIFF
--- a/pkg/kn/commands/human_readable_flags.go
+++ b/pkg/kn/commands/human_readable_flags.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/duration"
 	hprinters "knative.dev/client/pkg/printers"
@@ -112,4 +113,8 @@ func TranslateTimestampSince(timestamp metav1.Time) string {
 		return "<unknown>"
 	}
 	return duration.HumanDuration(time.Since(timestamp.Time))
+}
+
+func AddGitOpsFlags(flags *pflag.FlagSet) {
+	flags.String("in-dir", "", "work on local directory instead of a remote cluster")
 }
--- a/pkg/kn/commands/human_readable_flags.go
+++ b/pkg/kn/commands/human_readable_flags.go
@@ -117,5 +117,5 @@ func TranslateTimestampSince(timestamp metav1.Time) string {
 
 // AddGitOpsFlags adds flags to enable gitops mode
 func AddGitOpsFlags(flags *pflag.FlagSet) {
-	flags.String("in-dir", "", "work on local directory instead of a remote cluster")
+	flags.String("target", "", "work on local directory instead of a remote cluster")
 }
--- a/pkg/kn/commands/human_readable_flags.go
+++ b/pkg/kn/commands/human_readable_flags.go
@@ -115,6 +115,7 @@ func TranslateTimestampSince(timestamp metav1.Time) string {
 	return duration.HumanDuration(time.Since(timestamp.Time))
 }
 
+// AddGitOpsFlags adds flags to enable gitops mode
 func AddGitOpsFlags(flags *pflag.FlagSet) {
 	flags.String("in-dir", "", "work on local directory instead of a remote cluster")
 }
--- a/pkg/kn/commands/revision/revision_test.go
+++ b/pkg/kn/commands/revision/revision_test.go
@@ -68,7 +68,7 @@ func executeRevisionCommand(client clientservingv1.KnServingClient, args ...stri
 
 	output := new(bytes.Buffer)
 	knParams.Output = output
-	knParams.NewServingClient = func(namespace string) (clientservingv1.KnServingClient, error) {
+	knParams.NewServingClient = func(namespace string, cmd *cobra.Command) (clientservingv1.KnServingClient, error) {
 		return client, nil
 	}
 	cmd := NewRevisionCommand(knParams)
--- a/pkg/kn/commands/revision/revision_test.go
+++ b/pkg/kn/commands/revision/revision_test.go
@@ -68,7 +68,7 @@ func executeRevisionCommand(client clientservingv1.KnServingClient, args ...stri
 
 	output := new(bytes.Buffer)
 	knParams.Output = output
-	knParams.NewServingClient = func(namespace string, cmd *cobra.Command) (clientservingv1.KnServingClient, error) {
+	knParams.NewServingClient = func(namespace string) (clientservingv1.KnServingClient, error) {
 		return client, nil
 	}
 	cmd := NewRevisionCommand(knParams)
--- a/pkg/kn/commands/types.go
+++ b/pkg/kn/commands/types.go
@@ -81,26 +81,21 @@ func (params *KnParams) Initialize() {
 }
 
 func (params *KnParams) newServingClient(namespace string, cmd *cobra.Command) (clientservingv1.KnServingClient, error) {
-	dir := ""
-	if cmd.Flag("in-dir") != nil {
-		dir = cmd.Flag("in-dir").Value.String()
+	// return gitops client if 'in-dir' flag is specified
+	if cmd.Flag("in-dir") != nil && cmd.Flag("in-dir").Value.String() != "" {
+		return clientservingv1.NewKnServingGitOpsClient(namespace, cmd.Flag("in-dir").Value.String()), nil
 	}
-	switch dir {
-	case "":
-		restConfig, err := params.RestConfig()
-		if err != nil {
-			return nil, err
-		}
 
-		client, err := servingv1client.NewForConfig(restConfig)
-		if err != nil {
-			return nil, err
-		}
-		return clientservingv1.NewKnServingClient(client, namespace), nil
-
-	default:
-		return clientservingv1.NewKnServingGitOpsClient(namespace, dir), nil
+	restConfig, err := params.RestConfig()
+	if err != nil {
+		return nil, err
 	}
+
+	client, err := servingv1client.NewForConfig(restConfig)
+	if err != nil {
+		return nil, err
+	}
+	return clientservingv1.NewKnServingClient(client, namespace), nil
 }
 
 func (params *KnParams) newSourcesClient(namespace string) (v1alpha2.KnSourcesClient, error) {
--- a/pkg/kn/commands/types.go
+++ b/pkg/kn/commands/types.go
@@ -20,6 +20,8 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/spf13/cobra"
+
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
@@ -43,7 +45,7 @@ type KnParams struct {
 	Output             io.Writer
 	KubeCfgPath        string
 	ClientConfig       clientcmd.ClientConfig
-	NewServingClient   func(namespace string) (clientservingv1.KnServingClient, error)
+	NewServingClient   func(namespace string, cmd *cobra.Command) (clientservingv1.KnServingClient, error)
 	NewSourcesClient   func(namespace string) (v1alpha2.KnSourcesClient, error)
 	NewEventingClient  func(namespace string) (clienteventingv1beta1.KnEventingClient, error)
 	NewMessagingClient func(namespace string) (clientmessagingv1beta1.KnMessagingClient, error)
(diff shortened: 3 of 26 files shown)


pkg/kn/commands/revision/revision_test.go
```diff
@@ -68,7 +68,7 @@ func executeRevisionCommand(client clientservingv1.KnServingClient, args ...stri
 
 	output := new(bytes.Buffer)
 	knParams.Output = output
-	knParams.NewServingClient = func(namespace string) (clientservingv1.KnServingClient, error) {
+	knParams.NewServingClient = func(namespace string, cmd *cobra.Command) (clientservingv1.KnServingClient, error) {
```
maximilien, line 71: What about typing this `CreateClient(...)` function type?

pkg/kn/commands/types.go
```diff
@@ -78,14 +80,21 @@ func (params *KnParams) Initialize() {
...
+	switch dir {
+	case "":
+		restConfig, err := params.RestConfig()
+		if err != nil {
+			return nil, err
+		}
 
-	client, _ := servingv1client.NewForConfig(restConfig)
-	return clientservingv1.NewKnServingClient(client, namespace), nil
+		client, _ := servingv1client.NewForConfig(restConfig)
```
maximilien, line 92: Why are we ignoring error here?

pkg/serving/v1/gitops.go
```diff
@@ -0,0 +1,242 @@
...
+type knServingGitOpsClient struct {
+	dir       string
+	namespace string
+	printer   printers.ResourcePrinter
+}
+
+// NewKnServingGitOpsClient returns an instance of the
+// kn service gitops client
+func NewKnServingGitOpsClient(namespace, dir string) KnServingClient {
+	yamlPrinter, _ := genericclioptions.NewJSONYamlPrintFlags().ToPrinter("yaml")
```
maximilien, line 52: Ignoring errors... I would rather propagate or log at minimum

maximilien (approved):
Good start. Important feature. Thanks for starting 💯

rhuss:
@itsmurugappan sorry that I couldn't review it yet. This week is still crazy busy for me, but let me give it a quick review at least about the UX tomorrow.

itsmurugappan:
tests are passing and unit tests are added

pkg/serving/v1/gitops.go
```diff
@@ -0,0 +1,274 @@
...
+func NewKnServingGitOpsClient(namespace, dir string) *knServingGitOpsClient {
+	// create dir , if not present
+	namespaceDir := filepath.Join(dir, namespace, ksvcKind)
+	if _, err := os.Stat(namespaceDir); os.IsNotExist(err) {
+		os.MkdirAll(namespaceDir, 0777)
+	}
+	return &knServingGitOpsClient{
+		dir:        dir,
+		namespace:  namespace,
+		fileClient: &fileClient{},
```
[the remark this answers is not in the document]
rhuss: the methods in the client all have direct equivalents to system libraries, so maybe even the functions are not needed (and you could use the system calls directly ?)

rhuss (commented):
Thanks a lot and sorry for the late reply.

I like the simplicity of the implementation but have some stylistic comment though.
There are probably some edge case, and we should "cleanup" the service from the empty / default value before writing it out (same as we do the for `kn service apply` command)

Not sure about the name of the option though, and maybe we should allow also to operate on a single file (eg. with an option that points to a file instead of a directory).

The directory option is nice because it will allow to add other objects (e.g. from eventing) into the mix.

Let's discuss tomorrow the naming a bit, I need to sleep over it ;-)

pkg/serving/v1/gitops.go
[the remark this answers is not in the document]
itsmurugappan: have this so that i could create fake file client for testing. For unit tests i dint want to write files to directories.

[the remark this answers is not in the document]
rhuss: tbh, I'm not a big fan of adapting the original code so that you can better test it (same for making private methods public or package visible in Java, just so that the test can access so internal variables). What's about adapting the test to let them write into a temporary directory, and then validate on the content of the written files? (and then delete the directory after the test)This would be also a little bit more realistic as it includes the same filesystem interaction as for the real business case. Also, I don't think its much slower.

It's maybe not a "pure" unit test, but it would also help us easily to detect issues on operating systems (e.g. windows) easier as long as we don't have full e2e test coverage for all the platforms that we support (e.g. linux, windows, macos, arm64, ....)

I really would avoid that extra abstraction introduced only for the sake of testing. (tl;dr - this is a 'leaky' abstraction is introduced to overcome technical constraints but shines through to the implementation of the business logic)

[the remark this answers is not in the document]
itsmurugappan: Sure Roland, that was my initial plan (create file and delete it) but i was not sure on where to draw the line between e2e and unit testing. You are right, if you see my initial commit it dint have this abstraction layer, it was introduced for unit testing.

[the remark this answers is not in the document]
rhuss: +1, thanks for taking care. Normally I would totally agree with you and filesystem interactions are not at the heart of unit tests. But one should not be dogmatic here, and it it takes a lot of extra efforts that leak into the code to test, then I think one should be just pragmatic ;-) With accessing the filesystem this might be often the case. Of course it would be much different when talking e.g. about databases or so.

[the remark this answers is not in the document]
itsmurugappan: have made the changes. Please review

pkg/serving/v1/gitops_test.go
```diff
@@ -0,0 +1,151 @@
...
+	servingtest "knative.dev/serving/pkg/testing/v1"
+
+	libtest "knative.dev/client/lib/test"
+	"knative.dev/pkg/ptr"
+	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
+)
+
+func TestGitOpsOperations(t *testing.T) {
+	// create clients
+	fooclient := NewKnServingGitOpsClient("foo-ns", "./tmp/kn/files")
```
dsimansk, line 37: You could also go with `ioutils` tmp that will make sure it's temporary and unique dir, we have it used [here](https://github.com/knative/client/blob/master/pkg/kn/commands/service/create_test.go#L1012-L1013). 
Regardless of the actuall tmp dir creation,  adding `defer rm` might be beneficial here as well.

```
tempDir, err := ioutil.TempDir("", "kn-file")
defer os.RemoveAll(tempDir)
```

rhuss: yes, please use the `ioutils` package as this will work on Windows then, too.

rhuss (commented):
Except for the tmpdir calculation, looks good to me ! We still need to think about a proper option name, as I think that `in-dir` is not descriptive enough.

What's about:

* `--offline-dir` (emphasizing that we are 'offline')
* `--local-dir` (emphasizing that no cluster is used, but lacks an indication that its for the output)
* `--target-dir` (emphasizing that it is generating something in the directory)

Or we could also have something like `--offline` / `--local` together with the `--target-dir` to make it even clearer that we:

* Don't connect to a cluster
* Create something in a target-dir. We could default to "." for the target dir if not given.
* Allows also to add later an `--target-file` which then would point to a single file to update.

Instead of `target` we could also use `output` (e.g. (`--output-dir`, `--output` (for a file) or `--output-file`. However, `--output` (or `-o`) is already kind of occupied by the output format for the read-only commands, so I probably wouldn't reuse it in this context. (but `--target`, `--target-dir` would be also possible)

Finally we might think if the extra namespace directory level is needed. E.g do we expect to create many services across namespaces or is the major use case to create a single service or multiple services in the same namespace. Because if the latter is the case, we might first implement a single file argument `--target mysvcl.yml` and later maybe add options for bulk operations.

wdyt ?

rhuss (commented):
Also, while trying it out, I think we need to tune the output message as this still is focussed on cluster-side creation. E.g. we see an empty URL (which is natural), and we might need to adapt / check certain output methods that they can work with an empty `status:` field.

pkg/serving/v1/gitops_test.go
dsimansk: > yes, please use the `ioutils` package as this will work on Windows then, too.

That's a great point the paths should be constructed with `filepath.Join` to account for Windows eventually.

itsmurugappan:
> Finally we might think if the extra namespace directory level is needed. E.g do we expect to create many services across namespaces or is the major use case to create a single service or multiple services in the same namespace. Because if the latter is the case, we might first implement a single file argument `--target mysvcl.yml` and later maybe add options for bulk operations.

IMO having namespaces will have the following benefits
* Organized
* Scenario where this command is used to generate files and checkin to GitHub , followed by ci process creating the services, namespaces might be handy
* Supports listing
* from kn argument perspective we still have the one argument

itsmurugappan:
> Except for the tmpdir calculation, looks good to me ! We still need to think about a proper option name, as I think that `in-dir` is not descriptive enough.
> 
> What's about:
> 
> * `--offline-dir` (emphasizing that we are 'offline')
> * `--local-dir` (emphasizing that no cluster is used, but lacks an indication that its for the output)
> * `--target-dir` (emphasizing that it is generating something in the directory)

I like offline-dir

pkg/serving/v1/gitops_test.go
itsmurugappan: have made this change

dsimansk:
My preferential list would be:
1. `--offline-dir` 
2. `--target-dir`
3. `--local-dir`

However, I wonder if we need to have `*-dir` suffix at all. E.g. to have a bit shorter flag name without dash, just `--offline` flag.  
With the following resolution:
- default: `./` - write to current dir
- `path/to/dir` - write everything to the directory
- `path/to/file.yaml` - write to single file

rhuss:
Yeah, I discussed this with @itsmurugappan on slack and my current preference would be `--target` (as "offline" sounds so negative ;-P

And also I think we can use a single option for both modes that we might want to support:

I think we should not create the top-level dir automatically and when we use `--target` something then, if something points to a directory we do it like implemented, but if not, and something ends with `.yaml` , `.yml`, or `.json` we create the appropriate file (even if it does not exist).

So for the _directory_ mode, we would require an existing directory (the structure within the directory can then be built up by us)

For the _file_ mode, we require a proper extension and create the file if needed.

dsimansk:
> Yeah, I discussed this with @itsmurugappan on slack and my current preference would be `--target` (as "offline" sounds so negative ;-P

@rhuss @itsmurugappan +1 sounds well thought through.

itsmurugappan:
@rhuss @dsimansk all comments have been addressed, please take a look.

pkg/kn/commands/human_readable_flags.go
```diff
@@ -113,3 +114,8 @@ func TranslateTimestampSince(timestamp metav1.Time) string {
 	}
 	return duration.HumanDuration(time.Since(timestamp.Time))
 }
+
+// AddGitOpsFlags adds flags to enable gitops mode
+func AddGitOpsFlags(flags *pflag.FlagSet) {
+	flags.String("target", "", "work on local directory instead of a remote cluster")
```
dsimansk, line 120: A little formatting nit to align with the rest.
```suggestion
	flags.String("target", "", "Work on local directory instead of a remote cluster.")
```

dsimansk: On a second thought I wonder if the message shouldn't be more descriptive/verbose. The following is just an example feel free to change or ignore it. :)
```
Create Knative resources in local directory or file instead of a remote cluster. Default mode is current directory, in addition path to custom direcoty or file extensions .yaml, .yml, .json are accepted. 
```
@rhuss I recall you've mentioned the feature is experimental. Therefore we should add `(experimental)` suffix at the end of description, right?

rhuss:
@itsmurugappan thanks a ton ! hope to jump on a next round of review tomorrow or the day after. End of the year is always crazy busy, but I will be there next week, too, when things are slowing down. So confident to get this landed within this year still !

pkg/kn/commands/human_readable_flags.go
rhuss: Yes, we should mark it as experimental and we should definitely include it for 0.20

rhuss: To not lose to much time, lets merge this and create a followup PR to add the indication that this feature is experimental in a followup PR.

rhuss:
/lgtm

let's merge it now to get it into 0.20 and address the remaining minor issues (marking as "experimental" in the description in a new PR (which we should also merge right before 0.20)

PR #1127, opened 2020-11-19
debug: eventing install in CI

navidshaikh:
/hold


navidshaikh:
/close

The eventing install using eventing.yaml is working as expected.

PR #1140, opened 2020-11-25
fix quoting in hack/release.sh

navidshaikh:


dsimansk:
Looks as good as the first time. :)
/lgtm

PR #1185, opened 2021-01-08
Permit revision name in apply

### LINKED ISSUE #1184: `kn apply` ignores revision name
### Bug report

Revision name is discarded -- note the revision name in the output below:

```shell
$ kn service apply --filename service.yaml --revision-name="foo-iamrevision"
Creating service 'foo' in namespace 'default':
...
Service 'foo' created to latest revision 'foo-00001' is available at URL:
http://foo.default.cluster.auth.dev
```

### Expected behavior

Expected it to behave like create in this scenario (note the revision name).

```
kn service create --filename service.yaml --revision-name="foo-iamrevision"
Creating service 'foo' in namespace 'default':
...
Service 'foo' created to latest revision 'foo-iamrevision' is available at URL:
http://foo.default.cluster.auth.dev
```

Pretty much the steps above.

### kn version

`go installed` from e7042dd15f8413cc9b82dacad9347a5e4460d8e7

```
Version:      
Build Date:   
Git Revision: 
Supported APIs:
* Serving
* Eventing
```

> 0.19.x

### LINKED ISSUE #1186: Support traffic flags for `kn service apply`
### Feature request

I'd like to be able to use `traffic` flags with `kn service apply`.

### Use case

My current goal is for my team to have CD driven by a service definition that is the source-of-truth, while supporting dark-mode logic for certain branches.

My naive assumption was I’d be able to do something like this:

- On “true” deploy: `kn service apply --filename service.yaml --image "${IMAGE}"`
- On “dark mode” deploy: `kn service apply --filename service.yaml --image "${IMAGE}" --tag foo --traffic "@latest=0,${OLD_TRAFFIC_SETTINGS}"`

However, this isn’t possible with the current iteration of the apply command because it doesn’t support the traffic flags. 

If I switch to using update for these procedure — `kn service update --image "${IMAGE}" --tag foo --traffic "@latest=0,${OLD_TRAFFIC_SETTINGS}"` — there’s no support for the file config, so we can’t atomically deploy a dark mode version with changes to the file’s contents.

This would be resolved by either:
a) adding the traffic args to `apply` or
b) adding `--filename` to `update`. 

Option (b) doesn’t make sense because the purpose of it is to be a partial update — I’m not aware of why (a) can’t happen though.

jwilner:
## Description

Permit using `--revision-name` with the `apply` command

## Changes

* Don't zero out the revision name field

## Reference

Fixes #1184

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -1112,5 +1112,9 @@ Ignore PRs:
 |===
 | | Description | PR
 
+| 🐛
+| Permit revision name in `service apply`
+| https://github.com/knative/client/pull/1185[#1185]
+
 |===
 ////
--- a/pkg/kn/commands/service/apply.go
+++ b/pkg/kn/commands/service/apply.go
@@ -66,7 +66,6 @@ func NewServiceApplyCommand(p *commands.KnParams) *cobra.Command {
 			}
 
 			var service *servingv1.Service
-			applyFlags.RevisionName = ""
 			if applyFlags.Filename == "" {
 				service, err = constructService(cmd, applyFlags, name, namespace)
 			} else {


jwilner:
@googlebot I signed it!

jwilner:
@googlebot I signed it!

jwilner:
Having trouble reproing this failure locally:

```shell
$ test/local-e2e-tests.sh -run TestServiceApply -count=1
🧪  Testing
=== RUN   TestServiceApply
=== PAUSE TestServiceApply
=== CONT  TestServiceApply
    service_apply_test.go:40: apply hello service (initially)
    service_apply_test.go:44: apply hello service (unchanged)
    service_apply_test.go:49: apply hello service (update env)
--- PASS: TestServiceApply (180.93s)
PASS
ok  	knative.dev/client/test/e2e	181.472s
```

maximilien (changes requested):
I wonder if the title of this could be more descriptive or explicative? Why was it being set to “” before? And more importantly I wonder why no tests failed or should you add a new one? Just curious.

rhuss:
As mentioned in #1184 `kn service apply` does not support BYO revision names, but only server generated revision names, because I was not able to use `kn service apply` in an idempotent way with BYO revision name. This should be rechecked, but I wonder about the use case of using client side revision names with `kn service apply` ?

Also, it looks like that the test is not passing + there  should be a e2e test that exactly verify that it possible to apply again with the same name without getting an error.

maximilien:
What’s the final decision on this? Seems to be lingering. Test are passing with two fails. Likely needs to rebase
/ok-to-test

rhuss:
The decision to not allow client side revision names  was a deliberate one (as reasoned about in the comments above). Let us close this PR, but feel free to reopen it we can identify a valid use case for leveraging client-side generated names for kn-apply.

PR #1187, opened 2021-01-08
Add the traffic flags to the apply command

### LINKED ISSUE #1186: Support traffic flags for `kn service apply`

jwilner:
## Description

Adds the traffic flags to the `apply` command.

## Changes

* Adds the traffic flags to the `apply` command.

## Reference

Fixes #1186


navidshaikh:
@jwilner : The build tests complaining about the missing auto-docs (for new traffic flags), which can be generated by running `./hack/build.sh`, please add those docs to the PR.

jwilner:
Thanks for the tip @navidshaikh -- it should be fixed now.

maximilien (changes requested):
Thanks for contribution. Can you please add a test for new feature?

rhuss:
Sorry for my late reply, I still have to wrap around how and even whether the traffic split should be included into `kn service apply`, as I'm not sure that we can provide the full state on the command line to regenerate the setup. E.g. you need at least two revisions for a traffic split, not something that we can specify in a single CLI call (remember the difference between `kn service create` and `kn service update` to `kn service  apply`, where the first is about incremental, step-wise build up of the state and the latter nees to provide always the full state.

jwilner:
Hi @rhuss. Thanks for responding.

It's certainly not beautiful to have to do a read and all of the parsing to construct the traffic split expression with old versions, but it's possible and useful. You can already control traffic splits with this command via the file; really, this is just providing parity via flags imo.

For context, my use case is really very standard I think.

I am trying to set up a simple CD pipeline. Like most CD pipelines, we want to track as much as possible the configuration of our service in source control. Using `kn service {create,update}` for this purpose is awkward. `create` is very limited in its scope; `update` fundamentally operates with deltas, so it is at odds with the notion of keeping the whole config in vcs. 
`apply` of course works for tracking and applying the whole state. 

That said, we also want to do some very basic traffic splitting -- we want the ability to "dark deploy" a feature branch of our service, and it would be very cumbersome to have to edit and check in the file definition to achieve that. Permitting the traffic flags with the `apply` command gets us there. 

I think the above is a very standard use case, and I think this change set is necessary to support it (which I can empirically confirm, b/c we're using my fork for CD).

I'm pasting below the wrapper I wrote because it my help provide additional context. The material part is the `deploy` function, but it might be helpful to understand that `PLUGIN_ROUTE_TAG` is conventionally set to a git-branch name.

```shell
#!/usr/bin/env bash

# A drone plugin which runs in two different logical modes:
# - `deploy` creates a new knative revision serving 100% of traffic; old revisions serving traffic are forgotten.
# - `dark deploy` deploys a new knative service serving 0% of traffic that's exposed via knative tagged header routing;
#    they're intended for testing. Triggered by the presence of `route_tag`.
#
# See the README for a full documentation of available parameters.

main() {
  [[ -n "${PLUGIN_CREDENTIALS}" ]] || die "credentials is required"
  [[ -n "${PLUGIN_PROJECT}" ]] || die "project is required"
  [[ -n "${PLUGIN_REGION}" ]] || die "region is required"
  [[ -n "${PLUGIN_CLUSTER}" ]] || die "cluster is required"

  [[ -z "${PLUGIN_SERVICE_YAML}" ]] && export PLUGIN_SERVICE_YAML=service.yaml
  [[ -f "${PLUGIN_SERVICE_YAML}" ]] || die "service_yaml does not exist"

  local service_name
  service_name="$(yq --exit-status e .metadata.name "${PLUGIN_SERVICE_YAML}")"
  export SERVICE_NAME="${service_name}"

  if [[ -z "${PLUGIN_DOCKER_REPO}" ]]; then
    local repo
    repo="$(cut -d/ -f2 <<<"${DRONE_REPO}")"
    export PLUGIN_DOCKER_REPO="us.gcr.io/nyt-auth-dev/${repo}"
  fi

  if [[ -z "${PLUGIN_IMAGE_TAG}" ]] && [[ -f .version ]]; then
    PLUGIN_IMAGE_TAG="$(cat .version)"
    export PLUGIN_IMAGE_TAG
  fi

  log_in

  deploy
}

die() {
  echo "${@}" >&2
  exit 1
}

log_in() {
  local credentials="${PLUGIN_CREDENTIALS}"
  local project="${PLUGIN_PROJECT}"
  local region="${PLUGIN_REGION}"
  local cluster="${PLUGIN_CLUSTER}"

  local creds_file
  creds_file="$(mktemp)"
  # shellcheck disable=SC2064
  trap "rm ${creds_file}" EXIT

  echo "${credentials}" >"${creds_file}"
  gcloud auth activate-service-account --key-file="${creds_file}"

  gcloud container clusters get-credentials --region "${region}" --project "${project}" "${cluster}"
}

deploy() {
  local service_yaml="${PLUGIN_SERVICE_YAML}"
  local service_name="${SERVICE_NAME}"
  local docker_repo="${PLUGIN_DOCKER_REPO}"
  local image_tag="${PLUGIN_IMAGE_TAG}"
  local route_tag="${PLUGIN_ROUTE_TAG}" # empty if regular release

  # Traffic Settings
  #
  # Knative traffic settings look like:
  #
  # [{"tag": "foo", "revisionName": "rev2", "percent": 0}, {"revisionName": "rev1", "percent": 100}]
  #
  # The fields
  # - Tag exposes the revision at an alternate URL
  # - Revision name is the actual revision
  # - Percent is the amount of traffic it receives
  #
  # Note that a revision does not necessarily have a tag.
  #
  # We aim to always preserve "dark" revisions. If doing a non-dark release, we:
  # - Remove all revisions serving > 0% (other non-dark revisions)
  # - Add our revision at 100%
  #
  # If you wanted to add a new dark mode revision (fizzbuzz) to the service, you need to pass the whole state of the
  # world via key-value pairs on the CLI like so:
  #   --tag @latest=fizzbuzz,rev2=foo --traffic @latest=0,rev2=0,rev1=100
  #
  # If you wanted to redeploy -- overwrite -- the tag foo, you must first untag the existing revision:
  #   --untag foo
  # And then reapply it (also omitting the old rev from traffic at this point)
  #   --tag @latest=foo --traffic @latest=0,rev1=100

  # load traffic settings, defaulting to empty if service doesn't exist yet
  local traffic_settings="[]"
  {
    local svc_desc
    if svc_desc="$(kn service describe --output json "${service_name}" 2>/dev/null)"; then
      traffic_settings="$(jq '.status.traffic' <<<"${svc_desc}")"
    fi
  }

  if [[ -z "${route_tag}" ]]; then
    traffic_settings="$(
      jq '
        [{revisionName: "@latest", percent: 100}]
        + [.[] | select(.percent == 0)]
      ' <<<"${traffic_settings}"
    )"
  else
    # Does the desired tag already exist? If so, delete it:
    if jq --exit-status --arg tag "${route_tag}" 'any(.tag == $tag) ' <<<"${traffic_settings}" >/dev/null; then
      echo "Untagging old revision with tag ${route_tag}" >&2
      kn service update --untag "${route_tag}" "${service_name}"
    fi
    # Note that we filter out any overwritten tags
    traffic_settings="$(
      jq --arg tag "${route_tag}" '
        [{tag: $tag, revisionName: "@latest", percent: 0}]
        + [.[] | select(.tag != $tag)]
      ' <<<"${traffic_settings}"
    )"
  fi

  # [{"revisionName": "@latest", "percent": 100}, ...] -> "@latest=100,..."
  local traffic
  traffic="$(jq --raw-output '[.[] | "\(.revisionName)=\(.percent)"] | join(",")' <<<"${traffic_settings}")"

  if [[ "${traffic}" == "@latest=0" ]]; then
    echo "Cannot do a deploy if no service has been created yet." >&2
    exit 1
  fi

  local args=(
    kn service apply "${service_name}"
    --filename "${service_yaml}"
    --image "${docker_repo}:${image_tag}"
    --revision-name "${service}-${image_tag}-{{.Generation}}-{{.Random 5}}"
    --traffic "${traffic}"
  )
  if [[ -n "${route_tag}" ]]; then
    args+=(--label-revision "auth.dev.nyt.net/auto-cleanup=1")
  fi

  {
    local tags
    # [{"revisionName": "@latest", "tag": "foo"}, ...] -> "@latest=foo,..."
    tags="$(jq --raw-output '[.[] | select(.tag) | "\(.revisionName)=\(.tag)"] | join(",")' <<<"${traffic_settings}")"
    if [[ -n "${tags}" ]]; then
      args+=(--tag "${tags}")
    fi
  }

  "${args[@]}"
}

main "${@}"

```

rhuss:
@jwilner do you have an example of an initial Knative service YAML that defines the traffic split already during creation time (with at least one route that is not `@latest`) ? not sure how this is supposed to work, but I could have missed something obvious for that of course.

rhuss:
Thanks for the update,  and I can see you use case to set the traffic split for an already running service. Still, this it might be trickier than just adding those flags to `kn service apply`, especially for the case when you _create_ the service. I have to test it, but I suspect we will get back an error from the server anyway in this case. If the error message is meaningful enough, then I'm happy to add this PR. if not, we might need to add some more checks to prevent the case of adding a traffic split in this situation. An alternative behaviour would be to ignore a traffic specification on creation (which btw can not really be performed if you don't have an already existing revision).

I still have to play with the flow and the edge cases, but I think for a CD setup it totally makes sense to support the traffic split (which is an odd construct wrt/ to state anyway, so we might add a 'special' behaviour for this part)

jwilner:
@rhuss I added an earlier guard that errors if you've used the flag weirdly during creation. LMK what you think.

jwilner:
Are these tests reliable? I don't immediately see how they can be related.

jwilner:
Hi @rhuss -- does this PR have a future or should I close it out?

rhuss:
sorry for being late (too much on the table), and still have to dive deeper a bit. Please let the issue open as I think it's a valid use case, but we need to find out about the corner cases (e.g. when using the traffic options when initially applying a service that does not exist yet).

I hope to work on this soonish (but probably only at the beginning of next week).

jwilner:
Ok @rhuss, but I think you may have missed the update, so to reiterate in case you did: the pr was updated to clearly error on invalid usage per your request. That handles the corner case you alluded to in your last message.

rhuss:
After checking the PR, I think it makes sense to integrate the traffic flags with apply as:

* You can perfectly provide a traffic split during creating but only the value with `@latest=100`
* Using tags referencing non-existing revisions doesn't lead to an error server side, so there is no need to check this here. No sure if this an error on the server-side (but I guess it's not a hard reference for the tags anyway, more like a symbolic link)

rhuss:
However, this error message looks strange:

```
./kn service apply random2 --image rhuss/random:1.0 --traffic "@latest=100" --tag random-0001=v4
Error: multiple traffic targets are impossible when creating a service
Run 'kn --help' for usage
```

There is only one target, but an additional tag option when creating the service. I would expect this command to succeed. Omitting the `--tag` indeed works:

```
./kn service apply random2 --image rhuss/random:1.0 --traffic "@latest=100"
Creating service 'random2' in namespace 'default':

  0.013s The Route is still working to reflect the latest desired specification.
  0.037s ...
  0.062s Configuration "random2" is waiting for a Revision to become ready.
  7.437s ...
  7.502s Ingress has not yet been reconciled.
  7.681s Waiting for load balancer to be ready
  7.972s Ready to serve.

Service 'random2' created to latest revision 'random2-00001' is available at URL:
http://random2-default.apps.rhuss-dev.devcluster.openshift.com
```

rhuss:
tl;dr - I'm positive to integrate the PR as I think it might be useful exactly for the use case that @jwilner mentioned above, and it still can be performed in an idempotent way.

The messaging needs to be improved though, as well as the error check for create has some issues.

I'm just trying to find out why my tests sometimes hang when running apply and then will jump on the review.

pkg/kn/commands/service/apply.go
```diff
@@ -75,15 +78,28 @@ func NewServiceApplyCommand(p *commands.KnParams) *cobra.Command {
...
 
 			client, err := p.NewServingClient(namespace)
 			if err != nil {
 				return err
 			}
 
-			waitDoing, waitVerb, err := examineServiceForApply(cmd, client, service.Name)
-			if err != nil {
+			waitDoing, waitVerb := "Applying", "applied"
+			if isCreate, err := examineServiceForApply(cmd, client, service.Name); err != nil {
```
maximilien, line 94: Why not is,ply this if/else/if with simply:

```go
err := examineServiceForApply(cmd, client, service.Name)
if err != nil {
  return err
}

// do as in the else / if since the ifCreate condition applies to both parts of the statement
// and just add a `if !isCreate ...` guard and return or do what's needed in that case in a helper function

maximilien (changes requested):
First, thanks for contribution. Second, wondering if the main logic could be simplified. And third, I did not see an integration test.
/ok-to-test

PR #1203, opened 2021-01-25
Making a toReference public to enable reuse from kn-plugins

### LINKED ISSUE #1202: Allow to use toReference method from outside, in ex. kn plugins
https://github.com/knative/client/blob/c503b88be21d859227d4e35471130e03c62def47/pkg/kn/commands/source/binding/binding.go#L77

Writing a kn-event plugin I wanted to parse the addressable in the same way as it is parsed in `kn source binding --subject` option. To be able to do it, this function needs to be moved to a public place. Maybe `knative.dev/client/pkg/util` is a good place?

cardil:
## Description

Moving a toReference to a public package, so that is it possible to invoke it, from kn plugins.

## Changes

* Move `toReference` function to `knative.dev/client/pkg/util.ToTrackerReference`
* Move `parseSelector` function to `knative.dev/client/pkg/util.ParseSelector`

## Reference

Fixes #1202

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -18,6 +18,9 @@
 | | Description | PR
 
 | 🐣
+| Making a toReference public to enable reuse from kn-plugins
+| https://github.com/knative/client/pull/1203[#1203]
+| 🐣
 | Do not print serviceUID and configUID labels in service export result
 | https://github.com/knative/client/pull/1194[#1194]
 |===
--- a/pkg/kn/commands/source/binding/binding.go
+++ b/pkg/kn/commands/source/binding/binding.go
@@ -15,12 +15,9 @@
 package binding
 
 import (
-	"fmt"
 	"strings"
 
 	"github.com/spf13/cobra"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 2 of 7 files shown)


maximilien (commented):
I like the idea and we definitely have every more functions to expose for plugins. However, we had discussed to move all these to the `lib/*` directory instead of leaving them in the `pkg/*` So perhaps it makes sense to create a `lib/util` package and move things there? WDYT?

cardil:
IDT, if I'm on board with such move. It seems to me like an affront to Golang standards. `pkg` directory is a Golang standard for code for external use - see: https://github.com/golang-standards/project-layout#pkg

> Library code that's ok to use by external applications (e.g., `/pkg/mypubliclib`)

Internal code should be placed in `internal` directory.

rhuss (approved):
Thanks, looks good to me ! We might move this code again later when we do a larger refactoring of shared code though, but for now this is fine for me.

/lgtm
/approve

rhuss:
> IDT, if I'm on board with such move. It seems to me like an affront to Golang standards. `pkg` directory is a Golang standard for code for external use - see: https://github.com/golang-standards/project-layout#pkg

Then we did it all wrong until know as we definitely added for-internal-use-only code into this package already. Thanks for the heads up though, we will take the golang recommendations into account when we move around the code into a shared repository.

In order to unblock your work, let's merge this PR and fix that later with the bigger move.

PR #1206, opened 2021-01-27
update to k8s 0.19.7, go v1.15, fix gnostic

vaikas:
## Changes

* In go.mod: use go 1.15, remove replace directives, use k8s 0.19.7
* fix the gnostic OpenAPIv2 vs openapiv2 mess
*

## Reference

Fixes #


rhuss:
looks good to me, the unit test failure is probably due to an assertion that needs to be updated because of a format supported by k8s 1.19 that is not available 1.18 (tbh I question the value of this test as this really tests the dependency, not the client itself).

rhuss:
@vaikas I can push a fix for the unit test fix at least (if your branch is open). For the e2e test failure I have to check.

vaikas:
@rhuss Thanks, yeah, the branch should be open push away :)

rhuss:
@googlebot I consent

docs/cmd/kn_broker_describe.md
```diff
@@ -33,7 +33,7 @@ kn broker describe NAME
       --allow-missing-template-keys   If true, ignore any errors in templates when a field or map key is missing in the template. Only applies to golang and jsonpath output formats. (default true)
   -h, --help                          help for describe
   -n, --namespace string              Specify the namespace to operate in.
-  -o, --output string                 Output format. One of: json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-file|url.
+  -o, --output string                 Output format. One of: json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-as-json|jsonpath-file|url.
```
maximilien, line 36: Should we then make `jsonpath-as-files but maybe this should be in different PR

maximilien (commented):
/ok-to-test

vaikas:
I see this test failing:
```
        === CONT  TestServiceOptions
    service_options_test.go:116: create, update and validate service with cmd and arg options
    service_options_test.go:304: assertion failed: ["/ko-app/helloworld"] (out.Stdout string) != [/ko-app/helloworld] (expected string)
    result_collector.go:107: === FAIL: =======================[[ERROR]]========================
<SNIP>
        🦆 kn service update svc1 --concurrency-limit -1 --concurrency-target 0 --namespace kne2etests21
        ┃ 
        🔥 Error: target 0 should be at least 0.01: autoscaling.knative.dev/target
        🔥 Run 'kn --help' for usage
        🔥 
```

vaikas:
But that error above looks actually like something that is expected as I see the same test in serving. Perhaps this is the real error here:
```
    service_options_test.go:304: assertion failed: ["/ko-app/helloworld"] (out.Stdout string) != [/ko-app/helloworld] (expected string)
```

rhuss:
> But that error above looks actually like something that is expected

Exactly, the error is typically printed above the dump of the test output itself (which is only printed out in case of a failure)

vaikas:
Great, thanks @rhuss for helping fix the test failures!!

vaikas:
Now failing with:
```
    service_options_test.go:116: create, update and validate service with cmd and arg options
    service_options_test.go:304: assertion failed: ["myArg1","--myArg2"] (out.Stdout string) != ["myArg1" "--myArg2"] (expected string)
```

vaikas:
Ok, progress. The integration tests pass now, the integration against latest eventing failed because the setup failed:
```
Error from server (InternalError): error when creating "https://storage.googleapis.com/knative-releases/eventing/previous/v0.20.0/eventing.yaml": Internal error occurred: failed calling webhook "config.webhook.eventing.knative.dev": Post https://eventing-webhook.knative-eventing.svc:443/config-validation?timeout=2s: no endpoints available for service "eventing-webhook"
Waiting until all pods in namespace knative-eventing are up......................................................................................................................................................
```

/test pull-knative-client-integration-tests-latest-release

docs/cmd/kn_broker_describe.md
rhuss: @maximilien this is coming directly from a dependency, we delegate all option handling to Kubernetes-client. The name comes from there, so not so much we can change here (and we also shouldn't do it if we could in order to avoid surprises to Kubernetes users)

rhuss:
> Now failing with: ...

yeah, that was just a wild guess from my side ;-) The fix to adapt the assertion is the proper way.

rhuss (approved):
/lgtm
/approve

PR #1209, opened 2021-02-01
fix: Pin go-openapi/spec version to fix conflict

dsimansk:
## Description

Latest Serving update will bring incompatible version, for more details see [the comment](https://github.com/knative/client/pull/1207#issuecomment-770866919).

## Changes

* Pin `github.com/go-openapi/spec` version to `v0.19.3`
  * The same version is used in `k8s.io/cli-runtime@v0.19.7 github.com/go-openapi/spec@v0.19.3`.

## Reference

Unblocks #1207


maximilien (approved):
/ok-to-test
/lgtm

PR #1210, opened 2021-02-02
Allow to specify build platform for test images

BbolroC:
## Description

This is an echo of the https://github.com/knative/eventing/pull/4763 for client. The test images are built and uploaded without a break on other architectures by specifying a `PLATFORM` env variable.

## Changes

* Add a description to test/README.md
* Add a local variable `platform` (empty by default) which can be set by an environment variable `PLATFORM` at test/upload-test-images.sh

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -85,6 +85,10 @@ To run the script for all end to end test images:
 A docker tag may be passed as an optional parameter. This can be useful on
 Minikube in tandem with the `--tag` [flag](#using-a-docker-tag):
 
+`PLATFORM` environment variable is optional. If it is specified, test images
+will be built for specific hardware architecture, according to its value
+(for instance,`linux/arm64`).
+
 ```bash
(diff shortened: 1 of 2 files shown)


dsimansk:
Thanks!

/ok-to-test

rhuss (approved):
/lgtm
/approve

test/README.md
```diff
@@ -85,6 +85,10 @@ To run the script for all end to end test images:
 A docker tag may be passed as an optional parameter. This can be useful on
 Minikube in tandem with the `--tag` [flag](#using-a-docker-tag):
 
+`PLATFORM` environment variable is optional. If it is specified, test images
```
maximilien, line 88: Should this be `PLATFORMS` with s since the point is there could be more than one?

maximilien (changes requested):
See if my one change comment makes sense.

PR #1217, opened 2021-02-10
Add --prune & --prune-all options to delete the unreferenced revisions

### LINKED ISSUE #317: Consider supporting multiple NAMEs on some commands

Kaustubh-pande:
## Description
As per the review comment [#317 ](https://github.com/knative/client/issues/317#issuecomment-518628152)

## Changes

- Add --prune option to remove all unreferenced revisions for a given service
- Add --prune-all option to remove all the unreferenced revisions

## Reference

Fixes #317

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -17,6 +17,10 @@
 |===
 | | Description | PR
 
+| 🎁
+| Add `--prune` & `--prune-all` options to delete the unreferenced revisions
+| https://github.com/knative/client/pull/1217[#1217]
+
 | ✨
 | Update CRDs API version to `v1`
 | https://github.com/knative/client/issues/1248[#1248]
--- a/docs/cmd/kn_revision_delete.md
+++ b/docs/cmd/kn_revision_delete.md
@@ -12,6 +12,12 @@ kn revision delete NAME [NAME ...]
 
   # Delete a revision 'svc1-abcde' in default namespace
   kn revision delete svc1-abcde
+
+  # Delete all unreferenced revisions
+  kn revision delete --prune-all
+
+  # Delete all unreferenced revisions for a given service 'mysvc'
+  kn revision delete --prune mysvc
 ```
 
 ### Options
@@ -20,6 +26,8 @@ kn revision delete NAME [NAME ...]
   -h, --help               help for delete
   -n, --namespace string   Specify the namespace to operate in.
       --no-wait            Do not wait for 'revision delete' operation to be completed. (default true)
+      --prune string       Remove unreferenced revisions for a given service in a namespace.
+      --prune-all          Remove all unreferenced revisions in a namespace.
       --wait               Wait for 'revision delete' operation to be completed.
       --wait-timeout int   Seconds to wait before giving up on waiting for revision to be deleted. (default 600)
 ```
--- a/lib/test/revision.go
+++ b/lib/test/revision.go
@@ -139,3 +139,17 @@ func RevisionListWithService(r *KnRunResultCollector, serviceNames ...string) {
 		}
 	}
 }
+
+// RevisionDeleteWithPruneOption verifies removeing all unreferenced revisions for a given service in sync mode
+func RevisionDeleteWithPruneOption(r *KnRunResultCollector, serviceName, revName string) {
+	out := r.KnTest().Kn().Run("revision", "delete", "--prune", serviceName)
+	assert.Check(r.T(), util.ContainsAll(out.Stdout, "Revision", "deleted", revName, "namespace", r.KnTest().Kn().Namespace()))
+	r.AssertNoError(out)
+}
+
+// RevisionDeleteWithPruneAllOption verifies removeing all unreferenced revision in sync mode
+func RevisionDeleteWithPruneAllOption(r *KnRunResultCollector, revName1, revName2 string) {
+	out := r.KnTest().Kn().Run("revision", "delete", "--prune-all")
+	assert.Check(r.T(), util.ContainsAll(out.Stdout, "Revision", "deleted", revName1, revName1, "namespace", r.KnTest().Kn().Namespace()))
+	r.AssertNoError(out)
+}
--- a/pkg/kn/commands/revision/delete.go
+++ b/pkg/kn/commands/revision/delete.go
@@ -21,24 +21,39 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
-
 	"knative.dev/client/pkg/kn/commands"
+	v1 "knative.dev/client/pkg/serving/v1"
+	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
 )
 
 // NewRevisionDeleteCommand represent 'revision delete' command
 func NewRevisionDeleteCommand(p *commands.KnParams) *cobra.Command {
 	var waitFlags commands.WaitFlags
-
+	// prune filter, used with "-p"
+	var pruneFilter string
+	var pruneAll bool
 	RevisionDeleteCommand := &cobra.Command{
 		Use:   "delete NAME [NAME ...]",
 		Short: "Delete revisions",
 		Example: `
   # Delete a revision 'svc1-abcde' in default namespace
-  kn revision delete svc1-abcde`,
+  kn revision delete svc1-abcde
+
+  # Delete all unreferenced revisions
+  kn revision delete --prune-all
+
+  # Delete all unreferenced revisions for a given service 'mysvc'
+  kn revision delete --prune mysvc`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			if len(args) < 1 {
+			prune := cmd.Flags().Changed("prune")
+			argsLen := len(args)
+			if argsLen < 1 && !pruneAll && !prune {
 				return errors.New("'kn revision delete' requires one or more revision name")
 			}
+			if argsLen > 0 && pruneAll {
+				return errors.New("'kn revision delete' with --prune-all flag requires no arguments")
+			}
+
 			namespace, err := p.GetNamespace(cmd)
 			if err != nil {
 				return err
@@ -46,6 +61,21 @@ func NewRevisionDeleteCommand(p *commands.KnParams) *cobra.Command {
 			client, err := p.NewServingClient(namespace)
 			if err != nil {
 				return err
+			}
+			// Create list filters
+			var params []v1.ListConfig
+			if prune {
+				params = append(params, v1.WithService(pruneFilter))
+			}
+			if prune || pruneAll {
+				args, err = getUnreferencedRevisionNames(params, client)
+				if err != nil {
+					return err
+				}
+				if len(args) == 0 {
+					fmt.Fprintf(cmd.OutOrStdout(), "No unreferenced revisions found.\n")
+					return nil
+				}
 			}
 
 			errs := []string{}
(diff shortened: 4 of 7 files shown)


maximilien (approved):
Nice. Thanks for contribution.
/ok-to-test
/lgtm
/hold to allow others a chance to review

pkg/kn/commands/revision/delete.go
```diff
@@ -48,6 +62,17 @@ func NewRevisionDeleteCommand(p *commands.KnParams) *cobra.Command {
 				return err
 			}
 
+			if all {
+				args, err = getRevisionNames(client)
+				if err != nil {
+					return err
+				}
+				if len(args) == 0 {
+					fmt.Fprintf(cmd.OutOrStdout(), "No Revisions found.\n")
```
navidshaikh, line 71:
```suggestion
					fmt.Fprintf(cmd.OutOrStdout(), "No revisions found.\n")
```

pkg/kn/commands/revision/delete.go
```diff
@@ -67,7 +92,21 @@ func NewRevisionDeleteCommand(p *commands.KnParams) *cobra.Command {
 			return nil
 		},
 	}
+	flags := RevisionDeleteCommand.Flags()
+	flags.Bool("all", false, "Delete all revisions.")
```
navidshaikh, line 96:
```suggestion
	flags.Bool("all", false, "Delete all revisions in a namespace.")
```

lib/test/revision.go
```diff
@@ -139,3 +139,10 @@ func RevisionListWithService(r *KnRunResultCollector, serviceNames ...string) {
 		}
 	}
 }
+
+// RevisionDeleteWithAllOption verifies deleting all revision in sync mode
+func RevisionDeleteWithAllOption(r *KnRunResultCollector, revName string) {
+	out := r.KnTest().Kn().Run("revision", "delete", "--all")
+	assert.Check(r.T(), util.ContainsAll(out.Stdout, "Revision", revName, "deleted", "namespace", r.KnTest().Kn().Namespace()))
+	r.AssertNoError(out)
+}
```
navidshaikh, line 148: either this should take multiple revision names to verify or no revisions names at all (could then just verify without revision names).

test/e2e/revision_test.go
```diff
@@ -59,6 +59,9 @@ func TestRevision(t *testing.T) {
 	nonexistRevision := "hello-nonexist"
 	test.RevisionMultipleDelete(r, existRevision1, existRevision2, nonexistRevision)
 
+	t.Log("delete all revision from hello service and return no error")
+	test.RevisionDeleteWithAllOption(r, "hello")
+
```
navidshaikh, line 64: At this point, following the order of commands in this test, the service will have only one revision. You can update the service yet again to increase the count to 2 and then try deleting them all.

navidshaikh (changes requested):
@rhuss : Rethinking on this command, if user deletes all the revisions, the one (or may be multiple) receiving the traffic will be recreated. Should we have this option then ? as it tries to change the desire state? 
Instead, we may consider `kn revision delete --unused` or `kn revision delete --prune` to inspect the revisions not receiving traffic and delete them. wdyt?

Kaustubh-pande:
> @rhuss : Rethinking on this command, if user deletes all the revisions, the one (or may be multiple) receiving the traffic will be recreated. Should we have this option then ? (as it tries to change the desire state)
> Instead, we may consider `kn revision delete --unused` or `kn revision delete --prune` to inspect the revisions not receiving the traffic and delete them. wdyt?

@navidshaikh, @rhuss  I think `kn revision delete --prune` is more suitable here, can I rename --all to --prune?

rhuss:
I agree that we should not unconditionally delete all revisions, as this can (a) lead to an inconsistent state (e.g. when a revision is part of a traffic split) and (b) one of the revision would be recreated anyway.

So my suggestion is to add either a `kn revision prune` command or an option `kn revision delete --prune` so that all *unreferenced* revisions are removed. IMO this is the only reasonable use case.

CHANGELOG.adoc
```diff
@@ -17,6 +17,9 @@
 |===
 | | Description | PR
 
+| ✨
+| Add --prune option to remove all the unreferenced revisions
+| https://github.com/knative/client/pull/1217[#1217]
```
navidshaikh, line 22:
```suggestion
| https://github.com/knative/client/pull/1217[#1217]

```

docs/cmd/kn_revision_delete.md
```diff
@@ -12,6 +12,9 @@ kn revision delete NAME [NAME ...]
 
   # Delete a revision 'svc1-abcde' in default namespace
   kn revision delete svc1-abcde
+
+  # Delete all revisions
```
navidshaikh, line 16:
```suggestion
  # Delete all unreferenced revisions
```

pkg/kn/commands/revision/delete.go
```diff
@@ -67,7 +92,21 @@ func NewRevisionDeleteCommand(p *commands.KnParams) *cobra.Command {
...
+	revisionList, err := client.ListRevisions()
+	if err != nil {
+		return []string{}, err
+	}
+	revisionNames := []string{}
+	for _, revision := range revisionList.Items {
+		revisionNames = append(revisionNames, revision.Name)
+	}
+	return revisionNames, nil
+}
```
navidshaikh, line 112: Where are we checking for unreferenced revisions here ?

This function needs to find out the subset of revisions present in this namespace which are not referenced by *any* service and return them.

navidshaikh (changes requested):

rhuss:
I don't know by heart but is this PR about delete all revisions in a namespace or only revisions for a specific service ? I think it really should be about deleting revisions for a given service as I think this is the main use case.

So, from UX pov I see the following options that I would love to see

* `kn revision delete --prune myservice` (so taking a service name as an argument)
* `kn revision delete --prune-all`  (extra option to avoid accidentally pruning everything when a service name is forgotten. also I think the option space is not very polluted for `kn revision delete` so I think having two options is ok)

sorry for reiterating on this, but I think this is important.

lib/test/revision.go
```diff
@@ -139,3 +139,17 @@ func RevisionListWithService(r *KnRunResultCollector, serviceNames ...string) {
...
+	out := r.KnTest().Kn().Run("revision", "delete", "--prune", serviceName)
+	assert.Check(r.T(), util.ContainsAll(out.Stdout, "Revision", "deleted", "namespace", r.KnTest().Kn().Namespace()))
+	r.AssertNoError(out)
+}
+
+// RevisionDeleteWithPruneAllOption verifies removeing all unreferenced revision in sync mode
+func RevisionDeleteWithPruneAllOption(r *KnRunResultCollector) {
+	out := r.KnTest().Kn().Run("revision", "delete", "--prune-all")
+	assert.Check(r.T(), util.ContainsAll(out.Stdout, "Revision", "deleted", "namespace", r.KnTest().Kn().Namespace()))
+	r.AssertNoError(out)
```
dsimansk, line 154: Could we add a verification step here that asserts revision list count at least? To check expected number of revision is present or not. It can be either here or a separate `Verify...` function, as the same check is valid for the above `func` as well.

pkg/kn/commands/revision/delete.go
```diff
@@ -67,7 +107,28 @@ func NewRevisionDeleteCommand(p *commands.KnParams) *cobra.Command {
 			return nil
 		},
 	}
+	flags := RevisionDeleteCommand.Flags()
+	flags.Bool("prune", false, "Remove unreferenced revisions for a given service in a namespace.")
```
navidshaikh, line 111: this needs to be of type `flags.StringVar` for accepting the service name arg
see ref:
https://github.com/knative/client/blob/270e32240152cd02218932c72238edf0fd56dd04/pkg/kn/commands/revision/list.go#L112

navidshaikh (changes requested):

rhuss:
@Kaustubh-pande @navidshaikh @dsimansk what is left for this PR to get merged ? (sorry, have not followed up the latest changes of the PR)

Kaustubh-pande:
> @Kaustubh-pande @navidshaikh @dsimansk what is left for this PR to get merged ? (sorry, have not followed up the latest changes of the PR)

sorry, I was off for a couple of days, I will push a commit today.

pkg/kn/commands/revision/delete.go
```diff
@@ -47,6 +67,22 @@ func NewRevisionDeleteCommand(p *commands.KnParams) *cobra.Command {
 			if err != nil {
 				return err
 			}
+			// Create list filters
+			var params []v1.ListConfig
+			if prune {
+				serviceName := cmd.Flag("prune").Value.String()
+				params = append(params, v1.WithService(serviceName))
```
dsimansk, line 74:
```suggestion
				params = append(params, v1.WithService(pruneFilter))
```

pkg/kn/commands/revision/delete.go
```diff
@@ -67,7 +103,27 @@ func NewRevisionDeleteCommand(p *commands.KnParams) *cobra.Command {
 			return nil
 		},
 	}
+	flags := RevisionDeleteCommand.Flags()
+	flags.StringVarP(&pruneFilter, "prune", "p", "", "Remove unreferenced revisions for a given service in a namespace.")
+	flags.Bool("prune-all", false, "Remove all unreferenced revisions in a namespace.")
```
dsimansk, line 108: I'd suggest to add new variable `var pruneAll bool` variable to store the value to make the code more uniform. Then get rid of L49.
```
pruneAll, err := cmd.Flags().GetBool("prune-all")
```

pkg/kn/commands/revision/delete.go
```diff
@@ -21,10 +21,14 @@ import (
...
 
 	"github.com/spf13/cobra"
-
 	"knative.dev/client/pkg/kn/commands"
+	v1 "knative.dev/client/pkg/serving/v1"
+	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
 )
 
+// prune filter, used with "-p"
+var pruneFilter string
```
dsimansk, line 30: Move into `NewRevisionDeleteCommand` scope? And per suggestion below as `pruneAll` as well.

pkg/kn/commands/revision/delete.go
```diff
@@ -67,7 +103,27 @@ func NewRevisionDeleteCommand(p *commands.KnParams) *cobra.Command {
 			return nil
 		},
 	}
+	flags := RevisionDeleteCommand.Flags()
+	flags.StringVarP(&pruneFilter, "prune", "p", "", "Remove unreferenced revisions for a given service in a namespace.")
```
navidshaikh, line 107:
```suggestion
	flags.StringVar(&pruneFilter, "prune", "", "Remove unreferenced revisions for a given service in a namespace.")
```

navidshaikh: we should use the flag shorthand sparingly

navidshaikh (changes requested):

pkg/kn/commands/revision/delete.go
```diff
@@ -47,6 +62,21 @@ func NewRevisionDeleteCommand(p *commands.KnParams) *cobra.Command {
...
+			if prune || pruneAll {
+				args, err = getUnreferencedRevisionNames(params, client)
+				if err != nil {
+					return err
+				}
+				if len(args) == 0 {
+					fmt.Fprintf(cmd.OutOrStdout(), "No unreferenced revisions found.\n")
+					return nil
+				}
+			}
```
navidshaikh, line 79: we can create the client first and then replace the multiple if/else with `switch`

test/e2e/revision_test.go
```diff
@@ -59,6 +59,25 @@ func TestRevision(t *testing.T) {
 	nonexistRevision := "hello-nonexist"
 	test.RevisionMultipleDelete(r, existRevision1, existRevision2, nonexistRevision)
 
+	t.Log("update hello service and increase revision count to 4")
```
navidshaikh, line 62:
```suggestion
	t.Log("update hello service and increase revision count")
```

navidshaikh: since the the revisions are being deleted in above tests

test/e2e/revision_test.go
```diff
@@ -59,6 +59,25 @@ func TestRevision(t *testing.T) {
...
+	test.ServiceUpdate(r, "hello", "--env", "TARGET=kn", "--port", "8888")
+	t.Log("delete all unreferenced revision from hello service and return no error")
+	unRefRevision := test.FindRevisionByGeneration(r, "hello", 3)
+	test.RevisionDeleteWithPruneOption(r, "hello", unRefRevision)
+
+	t.Log("update hello service and increase revision count to 5")
```
navidshaikh, line 68:
```suggestion
	t.Log("update hello service and increase revision count")
```

navidshaikh (approved):
/lgtm
/unhold
suggested a few nits which can be fixed in a follow up, thanks!

PR #1221, opened 2021-02-11
Drop pkg/errors dependency and replace with stdlib usage

markusthoemmes:
## Description

https://github.com/pkg/errors has been mostly subsumed into the stdlib, so there's no reason to use it anymore. Less dependencies, YAY!

/assign @rhuss


rhuss (approved):
Thanks !

/lgtm
/approve

PR #1223, opened 2021-02-12
Enable errorlint and fix all issues

markusthoemmes:
## Description

This enables the `errorlint` linter, which is super helpful in catching "wrong" usages of errors (i.e. unsafe casts, ensuring errors are wrapped etc.)

## Changes

* Enable the `errorlint` linter
* Make sure all errors that somehow contain a child error properly wrap the child
* Replace all comparisons and casts with deepchecks using `Is` and `As`

/assign @rhuss

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -6,6 +6,7 @@ run:
 
 linters:
   enable:
+    - errorlint
     - unconvert
     - prealloc
   disable:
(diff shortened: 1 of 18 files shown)


maximilien (approved):
/ok-to-test
/lgtm
Thanks

PR #1224, opened 2021-02-12
Enable unparam linter and fix issues

markusthoemmes:
## Description

Probably the linter that found the most bugs over in Serving. Think of it as a more aggressive deadcode linter, telling you about unused parameters and unused return types.

Tests are excluded as the noise ratio is much higher in them.

/assign @rhuss

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -7,6 +7,13 @@ run:
 linters:
   enable:
     - unconvert
+    - unparam
     - prealloc
   disable:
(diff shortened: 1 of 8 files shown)


maximilien (approved):
/ok-to-test
/lgtm
Thanks for this. Did not realize so many unnecessary params. Thx

rhuss:
Looks like we have issues in the integration test to the cluster up

/retest

PR #1227, opened 2021-02-16
Drop override of github.com/go-openapi/spec

markusthoemmes:
## Description

As per title. I was curious why this was necessary, and nothing seemed to break superficially so I thought I'd send it.

## Changes

* Drop override of github.com/go-openapi/spec to make things move in lockstep with K8s deps etc.

/assign @rhuss


markusthoemmes:
Well, nevermind :joy:

dsimansk:
I should probably add a bit of explanation why it's needed. :)

PR #1228, opened 2021-02-16
Enable gosec linter and fix existing issues

markusthoemmes:
## Description

As per title. I find this linter super useful in that it points out potential security issues and **at least** makes you think about them and document why it's not an issue (see the existing ones I fixed).

Also: Pointers to range variables are finnicky, see https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable. (These have been pointed out by gosec too)

## Changes

* Comment on existing "security issues" to document why they are not an issue in this context.
* Remove all pointers to loop variables.

/assign @rhuss

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -7,9 +7,10 @@ run:
 linters:
   enable:
     - errorlint
+    - gosec
+    - prealloc
     - unconvert
     - unparam
-    - prealloc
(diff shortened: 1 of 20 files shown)


rhuss (commented):
Thanks !

/lgtm
/approve

PR #1229, opened 2021-02-16
Reorg OWNERS and introduced OWNERS_ALIASES

rhuss:
This update is for aligning with the other Knative projects so
that a group like the TOC list can be updated with less efforts
in the future.

Also I took the occasion to clean up a bit, so removed
@cppforlife and @sixolet as they are not active in the project since
at least one year. Happy to re-add both accounts again in the future.

Thanks a lot Naomi and Dmitriy for all your contribution and kicking off the client !

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,9 +1,4 @@
 approvers:
-- sixolet
-- cppforlife
-- rhuss
 - maximilien
-- navidshaikh
-# TOC members as a fallback
-- mattmoor
-- evankanderson
+- client-wg-leads
+- toc
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -1,0 +1,10 @@
+aliases:
+  toc:
+  - evankanderson
+  - grantr
+  - markusthoemmes
+  - mattmoor
+  - rhuss
+  client-wg-leads:
+  - rhuss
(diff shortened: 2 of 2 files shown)


rhuss:
@knative/technical-oversight-committee I reorganized the OWNERS files here but also removed inactive (over 1 year) approvers without actually knowing if this is the right process. If this is not ok, please let us know and point us to the correct way for keeping the OWNERS file healthy.

/hold

markusthoemmes (approved):
/lgtm
/approve

Inactive folks can speak up, we can always add them back imo.

mattmoor:
If Naomi isn't active, then we should probably update https://github.com/knative/community/blob/master/working-groups/WORKING-GROUPS.md#client as well.  I'll add this to the agenda for Thursday.

rhuss:
> If Naomi isn't active, then we should probably update https://github.com/knative/community/blob/master/working-groups/WORKING-GROUPS.md#client as well. I'll add this to the agenda for Thursday.

I already added a PR https://github.com/knative/community/pull/456 and would hold this issue until this has been eventually be merged.

Going to add this issue to the TOC project ...

maximilien (approved):
/ok-to-test
/lgtm

PR #1236, opened 2021-02-22
update: Change default to server-side generated revision names

### LINKED ISSUE #1144: Use server-generated revisions names as default for services (instead of BYO revision name)
As discussed in https://github.com/knative/serving/issues/9544 BYO revision names that we use by default (i.e. the client set the revision name and not the serving controller), has some issues when it comes to sink bindings, but also for `kn service apply` we can't use BYO revision names. Tbh, I always have a hard time to remember the actual use case for what BYO revision names are used (and again, I forgot about it ;-(. I conclude, that the use case for BYO revision names is so rare that it does not justified to use this as a default (when contrasting this to the trouble that it brings elsewhere).

My suggestion is to make server-side naming the default and *only* when `--revision-name` is specified we switch to BYO revision names. And the moment, switching off BYO revision naming is even very obfuscated as one has to provide a `--revision-name=` (i.e. an empty name) so switch this off.

The switch should work transparently except for that one use that I don't remember anymore (@evankanderson I think you explained it to me earlier, but I forgot again. sorry :( . From a UX perspective we just would not allow anymore a `--revision-name` without or empty argument, and just use server-side revision naming by default when this option is not given.

Opinions ?

rhuss:
Fixes #1144


dsimansk:
@rhuss E2E will need updates as well.

This test fails due to "same change" isn't generating new revision. I'll look into a bit to spot other issues.
https://github.com/knative/client/blob/master/test/e2e/revision_test.go#L53-L54

maximilien (approved):
/ok-to-test
/lgtm

rhuss:
> This test fails due to "same change" isn't generating new revision. I'll look into a bit to spot other issues.
> https://github.com/knative/client/blob/master/test/e2e/revision_test.go#L53-L54

Interesting that there is a timeout. Regardless what, that should never happen.

dsimansk:
> > This test fails due to "same change" isn't generating new revision. I'll look into a bit to spot other issues.
> > https://github.com/knative/client/blob/master/test/e2e/revision_test.go#L53-L54
> 
> Interesting that there is a timeout. Regardless what, that should never happen.

Indeed the timeout is caused by out waiting function. In case of executing same `kn service update` twice there's no resource change on server side, therefore out wait will hange waiting for next `MODIFIED` event that's never received.

First successful update
```bash
➜  client git:(pr/switch-to-server-generated-revision-names-by-default) ✗ kn service update hello --env TARGET="v2"                                      
Updating Service 'hello' in namespace 'default':

Event type: ADDED
Event type: MODIFIED
  0.058s The Configuration is still working to reflect the latest desired specification.
Event type: MODIFIED
Event type: MODIFIED
  5.835s Traffic is not yet migrated to the latest revision.
Event type: MODIFIED
  5.891s Ingress has not yet been reconciled.
Event type: MODIFIED
  5.947s Waiting for load balancer to be ready
Event type: MODIFIED
  6.112s Ready to serve.

Service 'hello' updated to latest revision 'hello-00002' is available at URL:
http://hello.default.example.com

```

The second one blocked waiting:
```bash
➜  client git:(pr/switch-to-server-generated-revision-names-by-default) ✗ kn service update hello --env TARGET="v2"
Updating Service 'hello' in namespace 'default':

Event type: ADDED
```



With my naive fix as an example. 
```go
@@ -177,6 +177,13 @@ func (w *waitForReadyConfig) waitForReadyCondition(watcher watch.Interface, star
                        //  resource version. All following watch events are for all changes that occurred after the resource
                        //  version the watch started at."
                        if event.Type != watch.Modified {
+                               inSync, err := generationCheck(event.Object)
+                               if err != nil {
+                                       return false, false, err
+                               }
+                               if inSync {
+                                       return false, false, nil
+                               }
                                continue
                        }

```

Update is "instantly" executed.
```bash
➜  client git:(pr/switch-to-server-generated-revision-names-by-default) ✗ kn service update hello --env TARGET="v2"
Updating Service 'hello' in namespace 'default':

Event type: ADDED
  0.003s Ready to serve.

Service 'hello' with latest revision 'hello-00002' (unchanged) is available at URL:
http://hello.default.example.com
```

rhuss:
/hold 

On some local tests I have issues with the wait loop

rhuss:
@maximilien @dsimansk would need a lgtm here to get it merged

/unhold

dsimansk:
:crossed_fingers: 
/lgtm

rhuss:
Close, fixed by #1240

PR #1240, opened 2021-02-23
update: Change default to server-side generated revision names (extended)

### LINKED ISSUE #1144: Use server-generated revisions names as default for services (instead of BYO revision name)

rhuss:
Fixes #1144.

This is an extension of #1236 with better handling for service updates. If tests are passing then this PR should be merged instead of #1236


rhuss:
@maximilien could I get this stamped please ? This is the PR that we talked about at the WG call, but a version that should be more robust (and irons out also my corner case). It checks now the generation's first before going into a wait loop. This should be much more robust as we did before anyway.

I would need this PR to be merged to continue with the 0.21 release (i will release the plugins then tomorrow, but would like to kick of the auto-release for kn already)

rhuss:
@navidshaikh could you please approve + lgtm this PR ? When merged, I wold then kickoff the kn release.

CHANGELOG.adoc
```diff
@@ -37,6 +37,10 @@
 | 🐣
 | Do not print `serviceUID` and `configUID` labels in service export results
 | https://github.com/knative/client/pull/1194[#1194]
+
+| ✨
+| Use server-side generated revision names by default now. For BYO revision names use `--revision name` for service commands
+| https://github.com/knative/client/issues/1144[#1144]
```
navidshaikh, line 43:
```suggestion
| ✨
| Use server-side generated revision names by default now. For BYO revision names use `--revision-name` for service commands
| https://github.com/knative/client/issues/1240[#1240]
```

navidshaikh (approved):
/lgtm
/approve

The changelog fix can be done in a follow up.

rhuss:
Ah, good point. Fixed (and need another stamp now ;-)

rhuss:
@navidshaikh Also stupid clabot needs some love again :)

thanks !

navidshaikh:
@googlebot I fixed it.

navidshaikh:
@googlebot I fixed it.

rhuss:
> @googlebot I fixed it.

Maybe `@googlebot I consent` works better ?

If not, I'm going squash everything and push again

navidshaikh:
@googlebot I consent.

navidshaikh:
@rhuss : lets squash and push again

rhuss:
@navidshaikh done. Hopefully we are at the last stamp ;-)

PR #1248, opened 2021-03-03
chore: Update CRDs API version to v1

dsimansk:
## Description

The CRDs `v1beta1` version is deprecated since K8s 1.16 and will be removed in `1.22`. However, it may produce a warning on newer Kubernetes clusters already.

We should probably handle the warning as well and don't fail prematurely. 

## Changes

* Update CRDs API version to `v1`

## Reference

Fixes N/A

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -12,6 +12,17 @@
 | https://github.com/knative/client/pull/[#]
 ////
 
+## (Unreleased)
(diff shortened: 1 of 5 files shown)


maximilien (approved):
/ok-to-test
/lgtm

rhuss (approved):
Thanks !

/lgtm
/approve

PR #1256, opened 2021-03-09
Use net-istio provided version in E2E setup

dsimansk:
## Description

This PR changes Istio setup to version provided from Knative's net-istio repository instead of GKE addon.

/hold


rhuss (commented):
Thanks for taking care !

I wonder whether we should trigger a discussion also for whether to include this common code that seems to be duplicated across serving / eventing / client into the `hack` repo, which still contains a different way to setup istio via a GKE plugin (if I understood you correctly)

dsimansk:
Yep, I had the same thought to move the code to `hack` repo. It's well isolated within one bash file, so it should be straightforward move. 

Meanwhile I've cleaned the code, removed commented out lines etc. Can we use it until there's shared one available?

PR #1259, opened 2021-03-10
Add net-istio install script through go modules

dsimansk:
## Description

 PTAL @rhuss, I think it's the second best solution in comparison to `hack` repo. We pull only install scripts and files through go modules and it's going to be updated through auto-update action.

Thanks @nak3 for the idea.


rhuss (commented):
Are you sure that net-istio does not bring in any transitive dependencies that might conflict e.g. later with plugins ? 

We had that issue when included `test-infra` via go.mod dependencies to pickup the CI scripts and `test-infra` had some very hefty dependencies on docker etc. the bite us in the client. That was also the first motivation to introduce the `hack` repo.

Happy with this solution if we can rule out that this is the case for `net-istio`

dsimansk:
> Are you sure that net-istio does not bring in any transitive dependencies that might conflict e.g. later with plugins ?

That's a very good point. Since we have a working install now, let hold this PR. I'll try to verify build with plugins.


/hold

rhuss:
@dsimansk as talked about in the WG call, I'm going to close this issue and we should do it for now like serving does.

PR #1263, opened 2021-03-15
Fix memory leak on closed watch channel

ericmillin:
## Description

This fixes a serious bug in the wait logic. If the watcher's channel closes, the loop tightens as it continuously retries until the command times out. This has three negative effects:

1. The code retries even though the results channel is never recreated, meaning the wait will never succeed
1. The CPU is pegged
2. `time.After` creates a new timer on each invocation. The timers aren't garbage collected until they expire, resulting in a crippling memory leak.

It's not clear to me why the results channel is closed, but it happens regularly in our testing. I believe some of the issues were introduced by the [changes in November](https://github.com/knative/client/pull/1113). The old logic appears to have recreated watchers on retries, which presumably included a new results channel.

## Changes
Ultimately, the code needs a refactor, but the changes here fix the all the bugs other than the permanently closed results channel. I also removed some cruft in `waitForEvent::Wait`.

/lint


pkg/wait/wait_for_ready.go
```diff
@@ -236,15 +239,12 @@ func (w *waitForEvent) Wait(watcher watch.Interface, name string, options Option
 	timeout := options.timeoutWithDefault()
 	start := time.Now()
 	// channel used to transport the error
-	errChan := make(chan error)
```
ericmillin, line 239: removed as cruft

pkg/wait/wait_for_ready.go
```diff
@@ -111,20 +111,20 @@ func NoopMessageCallback() MessageCallback {
...
-		floatingTimeout = floatingTimeout - time.Since(start)
-		if timeoutReached || floatingTimeout < 0 {
+		if timeoutReached {
 			return fmt.Errorf("timeout: %s '%s' not ready after %d seconds", w.kind, name, int(timeout/time.Second)), time.Since(start)
 		}
 
 		if retry {
-			// restart loop
+			// sleep to prevent CPU pegging and restart the loop
+			time.Sleep(pollInterval)
```
ericmillin, line 127: This isn't necessary if there's no longer a chance the CPU can be pegged. However, unless you're 100% sure that can't happen, then it should remain.

pkg/wait/wait_for_ready.go
```diff
@@ -111,20 +111,20 @@ func NoopMessageCallback() MessageCallback {
 func (w *waitForReadyConfig) Wait(watcher watch.Interface, name string, options Options, msgCallback MessageCallback) (error, time.Duration) {
 
 	timeout := options.timeoutWithDefault()
-	floatingTimeout := timeout
+	timeoutTimer := time.NewTimer(timeout)
```
ericmillin, line 114: Reusing the same timer prevents the timer leak caused by `time.After`. It is also cleaner because the floatingTImeout math is no longer needed.

maximilien (commented):
Looks cleaner. I guess this is difficult to test?

/ok-to-test

ericmillin:
> Looks cleaner. I guess this is difficult to test?
> 
> /ok-to-test

Yes, I tried with the mocks, but I couldn't get it working. You need to close the channel after the watcher starts. I didn't see a way to do that. Happy to take another crack at it if you see a way to make them work.

pkg/wait/wait_for_ready.go
rhuss: This is not really clear to me. As I understand this, it's meant as a safeguard in case the retry loop would go wild again (like that `waitForReadyCondition` returning immediately with retry == true, which has happened up to now when not detecting that the watcher channel has been closed). 

Looking at the code I can't see how the CPU could be pegged, but I have no prove. Adding a sleep here however would make it hard to detect such a situation, but of course, could be 'good enough'. So I'm a bit biased here :)

However,  I would not couple this to the `pollInterval` which is used as poll interval for a special kind of watcher (i.e. when the API does not support intrinsic watching like this is the case for CloudRun). Instead, I probably would use a shorter safe-guard, maybe 200 or 500ms ? (but I would use a different constant or even a literal value here).

rhuss (commented):
Thanks a lot, very good catch ! I only have a minor question, otherwise we should merge this asap.

@maximilien I think this is really hard to test, maybe one could simulate a watch function that closes the channel but also detects if someone reads from that closed channel afterwards (waiting some time) and then fails the test if this is the casde. Not sure if this is easy to implement, I would merge the PR without it, too.

dsimansk:
> 2. The CPU is pegged

ericmillin:
We're doing more testing with the change I made and it's definitely not enough to get the correct behavior, though it does solve the CPU and memory issues. Seems like the watch really needs to be recreate on closure.

@dsimansk I've tried a different approach and reverted the [November fix for the race](https://github.com/knative/client/pull/1113). I then made some simpler changes to fix the issue. It works so far. Happy to share the changes. However, it sounds like `resourceVersion` may be the "right way" to address this.

dsimansk:
@ericmillin I don't have the complete code yet, but you can take a look [here](https://github.com/dsimansk/client/commit/646fe33aeec9462fac86ef788ace5a3ccfec9eca). Hopefully I can put it together next week and have PR ready to gather more thoughts on it. 

Anyway we should get this PR merged once @rhuss questions are resolved. :)

rhuss:
Hi @ericmillin I think we should get this fix for the next release which is scheduled for next Tuesday, and then we can still continue to work on the resource version syncing. That's what we agreed on in the last client WG call.

Would you be able to resolve the conflicts ? If not, no problem, we would then take over.

ericmillin:
@rhuss Sorry going incommunicado--I was out of town. I can resolve the conflicts, but you may need to take over if you want this merged by Tuesday. I'm still awaiting approval for the CLA.

The version syncing sounds brilliant. While my changes resolved some of the issues, we were still seeing  unexpected behavior which is likely caused by fact that the logic exits when the channel is closed, rather than retrying. Because we had to release changes for an app using v0.19.x, I did a quick-and-dirty patch on a fork. The changes are a big improvement but still have problems accurately tracking state. @duglin has done a lot of manual testing and provide more details.

This is my quick-and-dirty patch which reverts to the pre-November logic and takes a different approach to solving the race: https://github.com/ericmillin/client/commit/a523953b635c6e8272ae4cd80e5569ed6c705a36

```go
Waiting() <-chan bool
```
to the `Wait` interface as way of keeping things

rhuss:
@ericmillin no worries, I will resolve the conflict now and then merge this PR for release 0.23

@dsimansk please pick up after the release to improve with the discussed idea on syncing on revisions.

rhuss:
@googlebot I consent.

PR #1264, opened 2021-03-15
Add support for namespace for all commands that takes a --sink option

### LINKED ISSUE #1143: Add support for namespaces for all commands that take a --sink option
### Feature request

It should be possible to specify the namespace of a sink for all commands that support a `--sink` option

### Use case

Allowing the namespace for a sink allows to create cross-namespace eventing meshes, which is a not uncommon use case. It adds a new level of flexibility for creating complex eventing architectures.

### UI Example

```
kn source ping ... --sink channel:mychannel:myothernamespace
```

would create a destination which points to another namespace. the last component after the last colon should be optional, and if not provided the namespace of the source itself is used (aka the current namespace)

Kaustubh-pande:
## Description
This PR will add support for the namespace that takes a  '--sink' option currently, we only resolve sink references in the current namespace but, with this PR user can able to refer the sink in another namespace.
if users pass the namespace `--sink ksvc:mysvc:my-namespace` which creates a destination that points to the my-namespace namespace, namespace support is optional, if the user not provided the namespace then-current namespace will be used

## Changes

* Add support for namespaces for all commands that take a --sink option

## Reference

Fixes #1143


dsimansk:
Thanks!
/ok-to-test

navidshaikh:
@Kaustubh-pande We should add e2e tests for this. e2e tests under dir `test/e2e/` operate under same namespace, so you can add one in smoke tests in `test/e2e-tests.sh`

maximilien (changes requested):
Thanks for this and adding all tests. Leaving it to others to review since I did not have a need for it but I can see the value

dsimansk (approved):
/lgtm

@navidshaikh could take a final look pls?

pkg/kn/commands/flags/sink.go
```diff
@@ -46,6 +46,7 @@ func (i *SinkFlags) AddWithFlagName(cmd *cobra.Command, fname, short string) {
 		"You can specify a broker, channel, Knative service or URI. " +
 		"Examples: '" + flag + " broker:nest' for a broker 'nest', " +
 		"'" + flag + " channel:pipe' for a channel 'pipe', " +
+		"'" + flag + " ksvc:mysvc:mynamespace' you can create a destination which points to another namespace, " +
```
navidshaikh, line 49:
```suggestion
		"'" + flag + " ksvc:mysvc:mynamespace' for a Knative service 'mysvc' in another namespace 'mynamespace', " +
```

navidshaikh: to align with other examples for sink flag

navidshaikh (changes requested):
a nit otherwise lgtm

dsimansk:
It should be good to go as Navid's last comment was addressed.
/lgtm
/unhold

rhuss:
@dsimansk @navidshaikh @Kaustubh-pande I see an issue here when using a sink without a prefix (which is a ksvc) with a namespace. How to distinguish this from the case prefix + service without namespace ?

I.e. given `a:b`, is `a` a prefix and `b` the name of the object or is `a` a KService name and `b` a namespace ?

navidshaikh:
@rhuss : Correct. When referring to knative service without prefix, it has to be in the same namespace else for knative service in different namespace, `prefix:name:namespace` combination must be provided. We'd need to update the help message for the flag.

dsimansk:
@rhuss nice catch! We should probably update the error [message](https://github.com/knative/client/blob/main/pkg/kn/commands/flags/sink.go#L110) as well.

rhuss:
> @rhuss : Correct. When referring to knative service without prefix, it has to be in the same namespace else for knative service in different namespace, prefix:name:namespace combination must be provided. We'd need to update the help message for the flag.

Do we get also an error when our service is called `broker` ? I.e. for `broker:bla` (i.e. referring to a service "broker" in namespace "bla") we don't get an error even we should, right ? but that's hard to detect.

navidshaikh:
> Do we get also an error when our service is called broker ? I.e. for broker:bla (i.e. referring to a service "broker" in namespace "bla") we don't get an error even we should, right ?

In this case, it'd try to resolve broker `bla`, and upon not finding one it'd fail with not found error.

PR #1265, opened 2021-03-15
Fix linting error SA5011

dsimansk:
## Description

Since the GH actions've been updated to latest `golangci-lint` version. There's false-positive SA5011 error reported on every new PR, [for example](https://github.com/knative/client/pull/1264/checks?check_run_id=2115285161).

It seems as a false-positive report as the test code starts with not nil assert, but it seems to fall into corner case of staticcheck detection capabilities.
https://staticcheck.io/docs/checks#SA5011

## Changes

* Fix SA5011 with not nil check


maximilien (approved):
Interesting you need to wrap in `if` when the assertion to verify that `pluginList` is not nil
/approve
/lgtm

PR #1267, opened 2021-03-17
Add DomainMapping CRUD commands

### LINKED ISSUE #1197: Add support for DomainMapping resource
### Feature request

It'd be great is if we could add support for `DomainMapping` custom [resource](url), that was recently added to Serving API.

### Use case

With the use of `DomainMapping` it's possible to create custom domains per namespace and per Service reference to address the endpoints. 

The required parameters are domain name (fqdn) and reference (`Addressable`, currently only [ksvc](https://github.com/knative/serving/blob/master/pkg/apis/serving/v1alpha1/domainmapping_types.go#L75-L76)). The initial goal should be to have CRUD ops available on pair with other cmds.

### UI Example

```
kn domain create <custom-domain> --ref <Service-name> --namespace <namespace>
kn domain delete <custom-domain> 
kn domain describe <custom-domain> 
kn domain list
```

/kind good-first-issue
/kind proposal

/cc @nak3

dsimansk:
## Description

This PR adds CRUD operations for `DomainMapping` resource.

TODO:
- I've covered only Knative Service as the only target, but going through the updated docs I should probably reflect the other targets as well.
https://github.com/knative/serving/commit/8eee93877d38e7cf227dcf622a0e956e491c77ef

- I'll take a look into Servings test, if we can do a real world domain E2E test.

## Changes

* 🎁 Add top-level cmd `domain` with `create, update, describe, delete, list` ops

## Reference

Fixes #1197


dsimansk:
@julz FYI.

pkg/kn/commands/domain/create.go
```diff
@@ -0,0 +1,79 @@
...
+				Reference(*destination.Ref)
+
+			client, err := p.NewServingV1alpha1Client(namespace)
+			if err != nil {
+				return err
+			}
+			err = client.CreateDomainMapping(builder.Build())
+			if err != nil {
+				return knerrors.GetError(err)
+			}
```
itsmurugappan, line 68:
```suggestion
			if err := client.CreateDomainMapping(builder.Build()); err != nil {
				return knerrors.GetError(err)
			}
```

pkg/kn/commands/domain/delete.go
```diff
@@ -0,0 +1,61 @@
...
+
+			client, err := p.NewServingV1alpha1Client(namespace)
+			if err != nil {
+				return err
+			}
+
+			err = client.DeleteDomainMapping(name)
+			if err != nil {
+				return knerrors.GetError(err)
+			}
```
itsmurugappan, line 53:
```suggestion
			if err := client.DeleteDomainMapping(name); err != nil {
				return knerrors.GetError(err)
			}
```

pkg/kn/commands/domain/delete_test.go
```diff
@@ -0,0 +1,51 @@
...
+	servingRecorder.Validate()
+}
+func TestDomainMappingDeleteWithError(t *testing.T) {
+	client := v1alpha1.NewMockKnServiceClient(t)
+	dynamicClient := dynamicfake.CreateFakeKnDynamicClient(client.Namespace(), createService("foo"))
+
+	// No call should be recorded
+	servingRecorder := client.Recorder()
+
+	_, err := executeDomainCommand(client, dynamicClient, "delete")
```
itsmurugappan, line 46: like update can add test for "not found"

pkg/serving/v1alpha1/client.go
```diff
@@ -0,0 +1,160 @@
...
+		return nil, err
+	}
+	return dm, nil
+}
+
+func (cl *knServingClient) CreateDomainMapping(domainMapping *servingv1alpha1.DomainMapping) error {
+	_, err := cl.client.DomainMappings(cl.namespace).Create(context.TODO(), domainMapping, v1.CreateOptions{})
+	if err != nil {
+		return knerrors.GetError(err)
+	}
```
itsmurugappan, line 85:
```suggestion
	if _, err := cl.client.DomainMappings(cl.namespace).Create(context.TODO(), domainMapping, v1.CreateOptions{}); err != nil {
		return knerrors.GetError(err)
	}
```

itsmurugappan (changes requested):
I gave it a try and it looks good. Just a few nits.

itsmurugappan:
Few suggestions

1. While creation, can ksvc be made the owner of this domain mapping, so that when ksvc is deleted this also gc'ed
2. AutoTLS can be disabled per route based on annotations, in follow up PR we can support annotations to domain mapping

```
kind: DomainMapping
metadata:
 annotations:
  networking.knative.dev/disableAutoTLS: "true"
 name: domainmapping.example.com
spec:
 ref:
   name: domain-mapping-svc
   kind: Service
```

julz:
> While creation, can ksvc be made the owner of this domain mapping, so that when ksvc is deleted this also gc'ed

FWIW I'm not sure this is what we should do. If the DM is deleted when the ksvc is deleted then it's possible for someone to swoop in and grab a domain mapping that you want to reuse for another ksvc later. I would understand doing this more if we created the ksvc and domain at the same time (eg if we added a field to ksvc to spit out a domainmapping), but it seems a bit surprising to me at first glance for us to do this implicitly. OTOH if this matches the behaviour for things like --sink that'd probably change my mind.. do we automatically set the ksvc as the owner of a source when you pass --sink?

dsimansk:
> * While creation, can ksvc be made the owner of this domain mapping, so that when ksvc is deleted this also gc'ed

I have the same concern as @julz that it'd break the reuse use case. In addition to `--sink` handling question `kn` isn't adding any OwnerRef wiring for sources, channels etc., except very specific case in `service import` that's trying to recreate the relationship.

However, such flag can still optional per explicit request to add OwenerRef to the resource. It might prove to be useful elsewhere as well.


> * AutoTLS can be disabled per route based on annotations, in follow up PR we can support annotations to domain mapping
Thanks, that's good point I haven't checked the annotations yet.

pkg/kn/commands/domain/domain.go
```diff
@@ -0,0 +1,36 @@
...
+
+// NewDomainCommand to manage domain mappings
+func NewDomainCommand(p *commands.KnParams) *cobra.Command {
+	domainCmd := &cobra.Command{
+		Use:     "domain COMMAND",
+		Short:   "Manage domain mappings",
+		Aliases: []string{"domains"},
+	}
+	domainCmd.AddCommand(NewDomainMappingCreateCommand(p))
+	domainCmd.AddCommand(NewDomainMappingDescribe(p))
```
maximilien, line 31: Why is this `NewDomainMappingDescribe` not `NewDomainMappingDescribeCommand` as with the others?

pkg/kn/commands/domain/domain_test.go
```diff
@@ -0,0 +1,99 @@
...
+	servingv1alpha1 "knative.dev/serving/pkg/apis/serving/v1alpha1"
+)
+
+// Helper methods
+var blankConfig clientcmd.ClientConfig
+
+func init() {
+	var err error
+	blankConfig, err = clientcmd.NewClientConfigFromBytes([]byte(`kind: Config
+version: v1
```
maximilien, line 40: I’d move this to it’s own constant string or template.

pkg/kn/commands/domain/domain_test.go
```diff
@@ -0,0 +1,99 @@
+// Copyright © 2021 The Knative Authors
```
maximilien, line 1: This file has no `Test..()` function(s)? So essentially a no-op

pkg/kn/commands/domain/list.go
```diff
@@ -0,0 +1,65 @@
...
+			}
+			domainMappingList, err := client.ListDomainMappings()
+			if err != nil {
+				return err
+			}
+			if len(domainMappingList.Items) == 0 {
+				fmt.Fprintf(cmd.OutOrStdout(), "No domain mapping found.\n")
+				return nil
+			}
+			if namespace == "" {
```
maximilien, line 56: Why is this check here when namespace is retrieved and used in line 48? Perhaps move closer so all namespace checks and usage together?

pkg/kn/commands/domain/list_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"testing"
+
+	"gotest.tools/v3/assert"
+
+	"knative.dev/client/pkg/serving/v1alpha1"
+	"knative.dev/client/pkg/util"
+	servingv1alpha1 "knative.dev/serving/pkg/apis/serving/v1alpha1"
+)
+
+func TestDomainMappingList(t *testing.T) {
```
maximilien, line 28: What about test for empty list?

pkg/kn/commands/domain/update.go
```diff
@@ -0,0 +1,82 @@
...
+				return err
+			}
+			toUpdate, err := client.GetDomainMapping(name)
+			if err != nil {
+				return err
+			}
+			if toUpdate.GetDeletionTimestamp() != nil {
+				return fmt.Errorf("can't update domain mapping '%s' because it has been marked for deletion", name)
+			}
+
```
maximilien, line 57: Minor style: Inconsistent extra line when none is added in other sections... I would add empty line after each group of statement and be consistent

pkg/serving/v1alpha1/client.go
```diff
@@ -0,0 +1,160 @@
...
+
+// NewKnServingClient create a new client facade for the provided namespace
+func NewKnServingClient(client clientv1alpha1.ServingV1alpha1Interface, namespace string) KnServingClient {
+	return &knServingClient{
+		client:    client,
+		namespace: namespace,
+	}
+}
+
+func (cl *knServingClient) Namespace() string {
```
maximilien, line 65: Add comment doc for all public functions...

maximilien (commented):
Various comments... please address. Great contribution BTW

CC: @julz who did the work on serving side as FYI

pkg/kn/commands/domain/create.go
dsimansk: Thanks for the suggestion. However, I recall from some previous PRs that if-s with inline declarations aren't very popular in client's code. 
@maximilien or @rhuss wdyt?

pkg/kn/commands/domain/delete_test.go
dsimansk: Sure, added.

pkg/kn/commands/domain/create.go
```diff
@@ -0,0 +1,79 @@
...
+			namespace, err := p.GetNamespace(cmd)
+			if err != nil {
+				return err
+			}
+
+			dynamicClient, err := p.NewDynamicClient(namespace)
+			if err != nil {
+				return err
+			}
+			destination, err := refFlags.ResolveSink(dynamicClient, namespace)
```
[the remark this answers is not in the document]
dsimansk: I've reused the sink flag handling here, but it's not entirely correct. I'll come up with something better. :)

pkg/serving/v1alpha1/client.go
dsimansk: Done in https://github.com/knative/client/pull/1267/commits/928e2cfb657dbb8011019125942a2270fc8614e7.

pkg/kn/commands/domain/update.go
dsimansk: Done in https://github.com/knative/client/pull/1267/commits/928e2cfb657dbb8011019125942a2270fc8614e7.

pkg/kn/commands/domain/list_test.go
dsimansk: Done in https://github.com/knative/client/pull/1267/commits/928e2cfb657dbb8011019125942a2270fc8614e7.

pkg/kn/commands/domain/list.go
dsimansk: Done in https://github.com/knative/client/pull/1267/commits/928e2cfb657dbb8011019125942a2270fc8614e7.

pkg/kn/commands/domain/domain_test.go
dsimansk: Done in https://github.com/knative/client/pull/1267/commits/928e2cfb657dbb8011019125942a2270fc8614e7.

pkg/kn/commands/domain/domain_test.go
dsimansk: Done in https://github.com/knative/client/pull/1267/commits/928e2cfb657dbb8011019125942a2270fc8614e7.

pkg/kn/commands/domain/delete.go
```diff
@@ -0,0 +1,61 @@
...
+		Short: "Delete a domain mapping",
+		Example: `
+  # Delete domain mappings 'hello.example.com'
+  kn domain delete hello.example.com`,
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			if len(args) != 1 {
+				return errors.New("'kn domain delete' requires the domain name given as single argument")
+			}
+			name := args[0]
+			namespace, err := p.GetNamespace(cmd)
```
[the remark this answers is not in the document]
dsimansk: Done in https://github.com/knative/client/pull/1267/commits/928e2cfb657dbb8011019125942a2270fc8614e7.

pkg/kn/commands/domain/domain.go
dsimansk: Done in https://github.com/knative/client/pull/1267/commits/928e2cfb657dbb8011019125942a2270fc8614e7.

docs/cmd/kn_domain_delete.md
```diff
@@ -0,0 +1,37 @@
+## kn domain delete
+
+Delete a domain mapping
+
+```
+kn domain delete FQDN
```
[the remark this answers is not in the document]
dsimansk: I intended to promote that fully qualified domain name is expected. It'd be better to explain in the example description probably.

dsimansk:
I've updated target reference resolution with https://github.com/knative/client/pull/1267/commits/df9c889dea3ae42bd9b1d80b39574b696c5c4548.

Do we want to support `type:name:namespace` resolution as was added for `--sink` flag? Or rather have something like `--ref-namespace`. 
I lean more towards the first option.

navidshaikh:
> Do we want to support type:name:namespace resolution as was added for --sink flag? 

+1

dsimansk:
From the DomainMapping docs:
> // Ref specifies the target of the Domain Mapping.
	//
	// The object identified by the Ref must be an Addressable with a URL of the
	// form `{name}.{namespace}.{domain}` where `{domain}` is the cluster domain,
	// and `{name}` and `{namespace}` are the name and namespace of a Kubernetes
	// Service.
	//
	// This contract is satisfied by Knative types such as Knative Services and
	// Knative Routes, and by Kubernetes Services.


Therefore I've added the following handling for 3 reference types.

>Addressable target reference for Domain Mapping. You can specify a Knative service, a Knative soute or a Kubernetes service.Examples: '--ref ksvc:hello' or simply '--ref hello' for a Knative service 'hello', '--ref kroute:hello' for a Knative route 'hello', '--ref svc:hello' for a Kubernetes service 'hello'. If a prefix is not provided, it is considered as a Knative service.

However, per discussion with @navidshaikh and @itsmurugappan yesterday there're some doubts. The addition of Kubernetes Service wtih `svc:` prefix breaks our previous convention for `--sink` flag, that doesn't support K8s Service. Even though there're 3 types that satisfy the contract, probably Knative Service is the most popular in our context.


Should we only support reference as Knative Service for now to keep it simple. And add additional support if there's request for it? Wdyt? @rhuss @maximilien

navidshaikh:
> Should we only support reference as Knative Service for now to keep it simple. And add additional support if there's request for it?

@dsimansk : Let's support knative resources, ksvc and kroute ? 

If we're planning to support prefix for k8s service, let's do it for sink flag as well? This way we've consistent prefix across flags.

dsimansk:
> > Should we only support reference as Knative Service for now to keep it simple. And add additional support if there's request for it?
> 
> @dsimansk : Let's support knative resources, ksvc and kroute ?
> 
> If we're planning to support prefix for k8s service, let's do it for sink flag as well? This way we've consistent prefix across flags.

Update the resolve func accordingly. In addition the PR is rebased and context param added.

pkg/kn/commands/domain/create.go
maximilien: Yeah, I life the none `if err := ...` but that's just me :)

docs/cmd/kn_domain_delete.md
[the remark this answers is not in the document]
maximilien: Thx clearer the better

rhuss:
Let's get that merged now, if now one objects (we can clean up later, but we need to get this in before the release)

/approve
/lgtm
/unhold

PR #1274, opened 2021-03-24
Use context.Context in API methods

### LINKED ISSUE #1272: Expose context to client APIs
### Feature request
During upgrade to k8s 1.18 libs, around knative 0.18 release timeframe, it was required to pass additional arguments for context and create/update/delete options, see https://github.com/knative/serving/pull/9367/files and https://github.com/knative/client/pull/1025/files . We worked around by adding `context.TODO` and empty create/update options.
To be consistent with the API usage pattern like in serving, eventing, k8s, we should allow the user to provide their own context.

### Use case
- Allow user to provide context while using the client APIs for more control over request lifecycle.
- Align API usage pattern with serving/eventing/k8s.

### UI Example
```go
// user's own context 'ctx'
ctx, cancel := context.WithCancel(context.Background())
knClient := NewServiceClient(namespace)
knClient.ListServices(context, ...)
```

### Implementation
This applies to different clients exposed by the kn; dynamic, serving, eventing clients.
We'd need to update the APIs of kn client to take additional argument of context, and update all the callers (and also plugins).

matejvasek:
## Description

Make API functions accept `context.Context` as the first argument.

## Reference

Fixes #1272


matejvasek:
We also need to make changes to the eventing client interface and maybe some others.

matejvasek:
@rhuss what subcommands of `kn` typically take more that one second?

matejvasek:
@navidshaikh please another /ok-to-test

dsimansk:
@navidshaikh prow bot doesn't run the tests on `Draft` PRs?

matejvasek:
@dsimansk it did run some integration tests after @navidshaikh set the ok to test label. Latter I committed new stuff, but the integration tests were not run. :man_shrugging:

dsimansk:
Well, one ok should be enough, plus there's a label now etc.

/ok-to-test

matejvasek:
>Well, one ok should be enough, plus there's a label now etc.

I would think so.

cmd/kn/main.go
```diff
@@ -15,14 +15,18 @@
...
 	"fmt"
 	"math/rand"
 	"os"
+	"os/signal"
 	"regexp"
 	"strings"
+	"syscall"
 	"time"
 
+	"github.com/pkg/errors"
```
dsimansk, line 28: Could the error check be done without adding a new dependency?
```
if errors.Is(err, context.Canceled) {
```

matejvasek: Probably yes. Initially I used `error.Cause() == ?` which is not in the standard `errors` package. But  the `Is()` function is part of the standard package (probably since go 1.13).

pkg/kn/commands/revision/list.go
```diff
@@ -223,7 +224,7 @@ func enrichRevisionAnnotationsWithServiceData(serviceFactory serviceFactoryFunc,
 }
 
 // Create a function for being able to lookup a service for an arbitrary namespace
-func serviceLookup(serviceFactory serviceFactoryFunc) serviceGetFunc {
+func serviceLookup(serviceFactory serviceFactoryFunc, ctx context.Context) serviceGetFunc {
```
dsimansk, line 227: Shouldn't be `Context` param first?
```suggestion
func serviceLookup(ctx context.Context, serviceFactory serviceFactoryFunc) serviceGetFunc {
```

dsimansk:
@matejvasek I've gone through the clients and it seems complete. (I'll double check though :)) 

There's one small nit, otherwise great work.

pkg/kn/commands/revision/list.go
matejvasek: Good catch.

matejvasek:
@dsimansk please run test again, I did small fix.

matejvasek:
Should I care about that coverage?

dsimansk:
> Should I care about that coverage?

It'd better if there was a test for new signals handling in `main.go` at least.


I've found only one leftover `context.TODO()`:
https://github.com/knative/client/blob/main/pkg/serving/config_changes.go#L168-L169
used at
https://github.com/knative/client/blob/main/pkg/kn/commands/service/configuration_edit_flags.go#L382

maximilien (commented):
Truly appreciate the effort. But I’ll also opine how much I hate the `pollution` that entails, especially for `context.TODO()`. Oh well, I guess somethings in life you just live with.

Thanks @matejvasek for the hard work on this. Cheers 🍻

matejvasek:
@maximilien I don't like that pollution too. But if there is no `ctx` to cancel potentially long running then it forces users to do stuff like:
```go
func getFoo(ctx context.Context) ([]byte, error) {
	ch := make(chan struct {
		f []byte
		e error
	})

	go func() {
		foo, err := xyz.getFoo()
		ch <- struct {
			f []byte
			e error
		}{f: foo, e: err}
	}()

	select {
	case res := <-ch:
		return res.f, res.e
	case <-ctx.Done():
		return nil, ctx.Err()
	}
}
```
which is not nice.

In  addition many interfaces use `ctx` as the first arguments, like: `docker` or `k8s` APIs.

rhuss:
I agree that the context passing in the argument list is not nice, but it's golang way of passing context (in contrast to Java where you probably would have used a ThreadLocal for this). And as @matejvasek mentioned, Docker and Kubernetes changed to this style, too, so we are in good company.

rhuss:
Thanks @matejvasek for the PR, very nice :-) Unfortunately, I don't have spare cycles for a review this week, but that's no loss since @maximilien @navidshaikh @dsimansk are excellent reviewers :)

matejvasek:
Most of the changes here are mechanical done by `IDE`/`sed`. Only functional changes are done to `wait/watch` logic. Also I added handling for signals that's for CLI, not API part.

matejvasek:
https://github.com/kubernetes/apimachinery/issues/118

navidshaikh (changes requested):
Thanks!
 I've commented inline on a few files, but the following changes need to be done for each client:

1. For each individual clients, 
 - Remove the context arg for individual `Namespace()` as it's an accessor (the ns is set via factory while creating the respective client) - for eg: see review for file `pkg/messaging/v1beta1/channels_client.go`
 - Use `ctx := context.Background()` and pass along the `ctx` to the chain of calls in `client_mock_test.go` - for eg: see review for file `pkg/dynamic/client_mock_test.go`

2. Thanks for taking care of signal handling, can we take it off from this PR and have it in subsequent one ? :-)

3. Please add an entry in changelog https://github.com/knative/client/blob/main/CHANGELOG.adoc

navidshaikh:
> kubernetes/apimachinery # 118

@matejvasek : are you seeing this while updating `pkg/wait` ? :-)

matejvasek:
> > kubernetes/apimachinery # 118
> 
> @matejvasek : are you seeing this while updating `pkg/wait` ? :-)

@navidshaikh yes I do (sometimes depending on when cancel happened).

pkg/messaging/v1beta1/channels_client.go
```diff
@@ -29,19 +29,19 @@ import (
...
-	DeleteChannel(name string) error
+	DeleteChannel(ctx context.Context, name string) error
 
 	// ListChannel lists all Channels
-	ListChannel() (*v1beta1.ChannelList, error)
+	ListChannel(ctx context.Context) (*v1beta1.ChannelList, error)
 
 	// Namespace returns the namespace for this channel client
-	Namespace() string
+	Namespace(ctx context.Context) string
```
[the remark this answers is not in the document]
matejvasek: I did it just for consistency sake, but I am going to revert it now.

matejvasek:
> > Should I care about that coverage?
> 
> It'd better if there was a test for new signals handling in `main.go` at least.
> 
> I've found only one leftover `context.TODO()`:
> https://github.com/knative/client/blob/main/pkg/serving/config_changes.go#L168-L169
> used at
> https://github.com/knative/client/blob/main/pkg/kn/commands/service/configuration_edit_flags.go#L382

I noticed that, but since it's not in client API I decided to leave it alone.

navidshaikh (approved):
/lgtm

w00t! Thanks a lot @matejvasek for quick turn around on it and adding tests!


Next set of action items would be to update the plugins consuming client APIs, if any.
/cc: @zhanggbj @maximilien @christophd @zhangtbj @dsimansk @rhuss

PR #1278, opened 2021-04-01
Fix describe/list commands with json/yaml output format

christophd:
## Description

This PR fixes issues when using describe/list commands with output format json or yaml. The commands used to run into following error:

`missing apiVersion or kind; try GetObjectKind().SetGroupVersionKind() if you know the type`

Makes sure to always set proper GroupVersionKind on unstructured objects and lists to avoid marshalling errors when printing in json/yaml output format.

The errors are basically because of these checks on the GVK when printing in such a format: https://github.com/knative/client/blob/599f1a5203665d3b0461e73035dc6091e8238950/vendor/k8s.io/cli-runtime/pkg/printers/json.go#L71

## Changes

* Make sure to set proper GVK on unstructured objects/lists
* Unit tests on describe/list commands for json output


itsmurugappan:
@christophd is there an issue create for this PR.  If not can you please tell how to reproduce this issue ?

maximilien (approved):
Thanks for the contribution and thorough tests... yes :) Only wonder if a e2e test would also help test this in a end to end fashion?
/approve
/lgtm
/hold to allow others to chime in

christophd:
@maximilien thanks! I love tests :)

@itsmurugappan no, there is no issue for this one, yet. Do you want me to create one? you can use one of the unit tests in this PR to reproduce the error (like in https://github.com/knative/client/pull/1278/files#diff-a581ea854dba8ffbcd340e621a33a4c21c674f1ab996b315f18dd707acf06ce2)

pkg/dynamic/client.go
```diff
@@ -106,6 +108,8 @@ func (c *knDynamicClient) ListCRDs(ctx context.Context, options metav1.ListOptio
 		return nil, err
 	}
 
+	uList.SetGroupVersionKind(schema.FromAPIVersionAndKind(listVersion, listKind))
+	uList.GetObjectKind().SetGroupVersionKind(schema.FromAPIVersionAndKind(listVersion, listKind))
```
rhuss, line 112: Not sure why this is needed. When setting the GVKs on other entities in this client like in `ListSourcesUsingGVKs` the object kind is not set.

Should we update those places, too ?

pkg/eventing/v1beta1/client.go
```diff
@@ -310,6 +314,10 @@ type BrokerBuilder struct {
 // NewBrokerBuilder for building broker object
 func NewBrokerBuilder(name string) *BrokerBuilder {
 	return &BrokerBuilder{broker: &v1beta1.Broker{
+		TypeMeta: meta_v1.TypeMeta{
```
rhuss, line 317: you could use `updateEventingGVK()` here, so that we do update always the same way.

pkg/dynamic/client.go
rhuss: ok, should have read the PR description first ;-)

Still, I think there might be more places that need an update.

pkg/eventing/v1beta1/client.go
rhuss: good point. actually, I would consider an error in this call as being fatal, so I wonder whether we could bail out here with a fatal exception. alternative, ignore an error here as it might no happen (we are calling the function with no user output, just constants that should have been registered. This is probably the easiest way.

The benefit is that when eventing is moving to `v1` we only have a single spot to update (`v1` will come soon).

pkg/dynamic/client.go
rhuss: So should we postpone the update of object kind at the other places in the dynamic client to post 0.23 ? wdyt ?

rhuss:
@christophd let's get this merged today, but maybe we can do a followup PR to address the remaining parts ? (sorry for the rush, but the release train is about to depart ;-)

pkg/eventing/v1beta1/client.go
christophd: @rhuss force pushed an update to use `updateEventingGVK()`

christophd:
@maximilien @rhuss @itsmurugappan how to proceed with this one?

rhuss:
I will have a look at it tomorrow. Also, will try to reproduce the original issue (as this was not clear why it works for some and for some it doesn't work)

pkg/eventing/v1/client.go
```diff
@@ -309,11 +315,17 @@ type BrokerBuilder struct {
 
 // NewBrokerBuilder for building broker object
 func NewBrokerBuilder(name string) *BrokerBuilder {
-	return &BrokerBuilder{broker: &eventingv1.Broker{
+	broker := &eventingv1.Broker{
```
rhuss, line 318: `NewBrokerBuilder` is only called for `kn broker create` which does not offer a yaml/json output. So even when the GVK is missing there, this is not an issue as by K8s API conventions this is not harmful (and even expected) that GVK is missing on objects that are sent to the API server.

Same for `NewTriggerBuilder` above.

pkg/kn/commands/broker/describe_test.go
```diff
@@ -25,44 +26,60 @@ import (
...
+	eventingv1 "knative.dev/eventing/pkg/apis/eventing/v1"
 	"knative.dev/pkg/apis"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
 
-	clientv1beta1 "knative.dev/client/pkg/eventing/v1"
+	clientv1 "knative.dev/client/pkg/eventing/v1"
 	"knative.dev/client/pkg/util"
 )
 
 func TestBrokerDescribe(t *testing.T) {
```
rhuss, line 37: The new tests work regardless of your fix in `NewBrokerBuilder` (so with the original code without the fix in this PR).

rhuss (changes requested):
@christophd I tried to reproduce the original error, but failed, too.


What I saw however in this PR is that you add the GVK coordinates for `create` operations (and add tests for that, too). According to Kubernetes API conventions the full GVK coordinates are not expected for write operations, so they are not send over the wire. Also, for 'create' you can't have really surfaced this issue as there is no `-o yaml/json` on those commands.

I don't see where you changed the behaviour for read operations (describe/list). An indication for this is also, when you use the additional tests for describe and list without your code changes, they still pass.

For this PR I see value in the additional tests for the read use case. I would like to keep those. But I'd suggest to remove the updates in the `New..Builder` commands and the corresponding tests for those.

rhuss:
@christophd are you planning to extract the extra unit tests into a new PR ? Really would love to finish up with this PR ...

christophd:
@rhuss yep sure. end of this week you should have the new PR

pkg/eventing/v1/client.go
christophd: `NewBrokerBuilder` is also used in the tests to create fake response data for the mock client recorder. This is why the tests fail as soon as I remove the GVK.

christophd:
@rhuss I had another look and the thing is that `NewBrokerBuilder` is also used in the tests to create fake response data for the mock client recorder. We would have to add separate ways to build new broker/trigger objects for create and read operations then. I am not so happy with that.

The missing GVK in the unstructured list should really be a problem isn't it?

So I could remove the changes from `NewBrokerBuilder` and `NewTriggerBuilder` but we would also loose the tests for list broker/trigger in json output format then, too.

pkg/util/unstructured.go
```diff
@@ -20,14 +20,27 @@ import (
...
 func ToUnstructuredList(obj runtime.Object) (*unstructured.UnstructuredList, error) {
 	unstructuredList := &unstructured.UnstructuredList{}
+	unstructuredList.SetGroupVersionKind(schema.FromAPIVersionAndKind(listVersion, listKind))
+	unstructuredList.GetObjectKind().SetGroupVersionKind(schema.FromAPIVersionAndKind(listVersion, listKind))
+
 	if meta.IsListType(obj) {
-		unstructuredList.SetGroupVersionKind(obj.GetObjectKind().GroupVersionKind())
+		if !obj.GetObjectKind().GroupVersionKind().Empty() {
+			unstructuredList.SetGroupVersionKind(obj.GetObjectKind().GroupVersionKind())
+			unstructuredList.GetObjectKind().SetGroupVersionKind(obj.GetObjectKind().GroupVersionKind())
```
rhuss, line 41: Just revisiting this PR, but do these two lines do not perform the very same operation ? Looking as the code isn't 

```
func (u *UnstructuredList) GetObjectKind() schema.ObjectKind { return u }
```

just an identity operation ?

Also I just debugged the method, and if you enter it with an `runtime.Object` that has GVK set (like for all objects that we return from the client in out client API where we explicitly set the GVK coordinates), then the conversion to unstructured works as expected so that it can be used by the printer package. Only when those GVK in the incoming object are _not set_, then the original code has an issue.

rhuss:
@christophd I updated the PR to make the test run without unconditionally setting the GVK all the time (which, as mentioned is not recommended when creating objects on the server side). Also, I reverted the change for the unstructured conversion, which IMO is also not needed if the incoming object already has the GVK set.

PTAL, let me know if we could merge this PR.

rhuss:
@googlebot I consent.

@christophd you would need to consent, too, again, to make the googlebot happy.

christophd:
@googlebot I consent.

christophd:
@rhuss 
/lgtm

rhuss:
Let me update the coverage a bit

rhuss:
@dsimansk @christophd added some tests, should be good now

rhuss (approved):
/approve
/lgtm

PR #1289, opened 2021-04-14
Add number of instances to describe command

### LINKED ISSUE #842: kn service describe should show # of instances
Today `kn service describe` will show this in it's output:
```
Revisions:  
  100%  @latest (echo-yyhmp-6) [6] (4d)
```
The `[6]` is interesting. I believe that the generation number - or in user's terms, my version # of my app. That's (sort of) interesting, but at first I thought it was the # of instances of this Revision that were running... that would be far more useful to me.

Proposal: show the # of instances of each revision instead of the generation #. I'm not really sure what the generation # tells the user, or what interesting thing they can do with it. It seems to me that seeing the revision name change in that list would be enough to indicate that there are multiple revisions out there.

This may be controversial because it means using a non-knative api (e.g. pods, or deployment) but I think we need to think about what's best (and most useful) for the user, and I think being able to see how their app is scaling (perhaps out of control!) would be very useful. 

We could even make it conditional, and try to get the list number of instances and if it fails due to an known API then just not show it (and no error msg), but if the API is there then show the info.

dsimansk:
## Description

Currently the Status fields `ActualReplicas,DesiredReplicas` are set to `omitempty`. Therefore the current impl doesn't show any replica details if both values are `0`.

```
➜  client git:(main) ✗ kn service describe hello                 
Name:       hello
Namespace:  default
Age:        34s
URL:        http://hello.default.127.0.0.1.nip.io

Revisions:  
  100%  @latest (hello-00001) [1] (34s)
        Image:     docker.io/dsimansk/helloworld (pinned to 29f975)
        Replicas:  1/1

Conditions:  
  ++ Ready                  20s 
  ++ ConfigurationsReady    29s 
  ++ RoutesReady            20s 
```

## Changes

* Add number of instances (`Replicas:`) to `kn service describe`
* Add number of instances (`Replicas:`) to `kn revision describe`

## Reference

Fixes #842 

/assign @knative/client-reviewers

### DIFF
--- a/pkg/kn/commands/revision/describe.go
+++ b/pkg/kn/commands/revision/describe.go
@@ -96,6 +96,7 @@ func describe(w io.Writer, revision *servingv1.Revision, service *servingv1.Serv
 	dw := printers.NewPrefixWriter(w)
 	commands.WriteMetadata(dw, &revision.ObjectMeta, printDetails)
 	WriteImage(dw, revision)
+	WriteReplicas(dw, revision)
 	WritePort(dw, revision)
 	WriteEnv(dw, revision, printDetails)
 	WriteEnvFrom(dw, revision, printDetails)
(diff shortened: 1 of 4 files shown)


itsmurugappan (approved):
/lgtm

maximilien (approved):
Thanks for the contribution

PR #1299, opened 2021-04-30
Update Eventing Sources API to v1

dsimansk:
## Description

Follow-up on recent `v1alpha1` removal from Eventing main.
https://github.com/knative/eventing/pull/5318

## Changes

* Update to `APIServerSource, ContainerSource, PingSource & SinkBinding` API to `sources/v1`

/cc @knative/client-reviewers


dsimansk:
There's one open question though. PingSource was promoted to `v1` API just know, that why integration test are failing currently. That also implies that we can't use PingSource v1 on `0.22`, breaking the backward compatibility, therefore we need to keep PingSource at `v1beta2` for now. (applying the common client API policy keeping the lowest version until available)

@rhuss @maximilien any objections to the above? I'm mainly asking due to implications for `1.0` release etc.

rhuss:
Yeah, let's stay with PingSource v1beta2.

If Eventing goes to Knative 1.0 with v1beta2, too, I don't see any problems to use this for a 1.0 client.

rhuss:
@lionelvillard Are there any plans for Evention 1.0 to remove PingSource v1beta2, too ?

lionelvillard:
> @lionelvillard Are there any plans for Evention 1.0 to remove PingSource v1beta2, too ?

We have to keep v1beta2 for at least of couple of releases.

pkg/kn/commands/source/ping/create_test.go
```diff
@@ -18,23 +18,23 @@ import (
...
+	clientsourcesv1beta2 "knative.dev/client/pkg/sources/v1beta2"
 
 	"knative.dev/client/pkg/util"
 )
 
 func TestSimpleCreatePingSource(t *testing.T) {
 	mysvc := &servingv1.Service{
-		TypeMeta:   v1.TypeMeta{Kind: "Service", APIVersion: "serving.knative.dev/v1"},
-		ObjectMeta: v1.ObjectMeta{Name: "mysvc", Namespace: "default"},
+		TypeMeta:   metav1.TypeMeta{Kind: "Service", APIVersion: "serving.knative.dev/v1beta2"},
```
rhuss, line 32: Serving v1beta2 ?

rhuss (changes requested):
Looks good, however while cross checking it looks like that there has been a gotcha when reverting from v1 back to v1beta2 for the ping source (see comment inline). Not sure if other parts are affected, too.

pkg/kn/commands/source/ping/create_test.go
dsimansk: :see_no_evil:

maximilien (commented):
/ok-to-test
/lgtm

rhuss:
Looks like that the test indeed captured a valid issue:

```
source_list_test.go:85: assertion failed: 
        Actual output: NAME              TYPE         RESOURCE                          SINK   READY
        testpingsource0   PingSource   pingsources.sources.knative.dev          True
        
        Missing strings: ksvc:testsvc0
```

dsimansk:
> Looks like that the test indeed captured a valid issue:
> 
> ```
> source_list_test.go:85: assertion failed: 
>         Actual output: NAME              TYPE         RESOURCE                          SINK   READY
>         testpingsource0   PingSource   pingsources.sources.knative.dev          True
>         
>         Missing strings: ksvc:testsvc0
> ```
> 
> i.e. there is no output of the connected sink.

Well, this one was pretty hard to find actually. :sweat: 
The error is caused by failed unstructured conversion due to https://github.com/knative/eventing/commit/7c0b4770144e0fd3d026f1b4884242abaa9bd282.

rhuss:
> The only issue with 0.22.0 ...

That's fine I would say, as we can always say that's an issue with the backend and the fix is, to update the eventing backend to 0.22.1

PR #1301, opened 2021-05-04
Refactor waiting loop to use resourceVersion

### LINKED ISSUE #1287: Refactor watch functions to pass resourceVersion
> 2. The CPU is pegged

_Originally posted by @dsimansk in https://github.com/knative/client/issues/1263#issuecomment-802225259_

dsimansk:
## Changes

* Use `resourceVersion` in `*Delete` functions to get replay of events and prevent 
* Re-enable `retry` logic when watch channel is closed before timeout

## Reference

Fixes #1287 

/cc @rhuss

### DIFF
--- a/pkg/eventing/v1/client.go
+++ b/pkg/eventing/v1/client.go
@@ -16,6 +16,7 @@ package v1
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	apis_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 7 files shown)


pkg/wait/wait_for_ready.go
```diff
@@ -140,10 +143,14 @@ func (w *waitForReadyConfig) Wait(ctx context.Context, watcher watch.Interface,
...
-	ctx context.Context, watcher watch.Interface, start time.Time, timeoutTimer *time.Timer, errorWindow time.Duration, msgCallback MessageCallback,
-) (retry bool, timeoutReached bool, err error) {
+func (w *waitForReadyConfig) waitForReadyCondition(ctx context.Context, name string, initialVersion string, start time.Time,
+	timeoutTimer *time.Timer, errorWindow time.Duration, options Options, msgCallback MessageCallback) (retry bool, timeoutReached bool, err error) {
 
+	watcher, err := w.watchMaker(ctx, name, initialVersion, options.timeoutWithDefault())
+	if err != nil {
+		return false, false, err
+	}
+	defer watcher.Stop()
```
maximilien, line 153: I can comment lines 150-152 and all tests passing. Similar on the other changes. My guess is that your tests are not covering the cases for failed timeout or similar?

pkg/wait/poll_watcher.go
```diff
@@ -83,6 +83,18 @@ func NewWatcher(ctx context.Context, watchFunc watchF, c rest.Interface, ns stri
...
+func NewWatcherWithVersion(ctx context.Context, watchFunc watchF, c rest.Interface, ns string, resource string, name string, initialResourceVersion string, timeout time.Duration) (watch.Interface, error) {
+	native, err := nativeWatchWithVersion(ctx, watchFunc, name, initialResourceVersion, timeout)
+	if err == nil {
+		return native, nil
+	}
+	polling := &pollingWatcher{
+		c, ns, resource, name, timeout, make(chan bool), make(chan watch.Event), &sync.WaitGroup{},
+		newTickerPollInterval(time.Second), nativePoll(ctx, c, ns, resource, name)}
+	polling.start()
+	return polling, nil
```
maximilien, line 95: I can comment this line and all tests pass. So missing coverage.

pkg/eventing/v1/client.go
```diff
@@ -241,25 +242,27 @@ func (c *knEventingClient) GetBroker(ctx context.Context, name string) (*eventin
...
+	return wait.NewWatcherWithVersion(ctx, c.client.Brokers(c.namespace).Watch, c.client.RESTClient(), c.namespace, "brokers", name, initialVersion, timeout)
 }
 
 // DeleteBroker is used to delete an instance of broker and wait for completion until given timeout
 // For `timeout == 0` delete is performed async without any wait
 func (c *knEventingClient) DeleteBroker(ctx context.Context, name string, timeout time.Duration) error {
+	broker, err := c.GetBroker(ctx, name)
+	if err != nil {
+		return err
+	}
```
maximilien, line 255: I can comment lines 253-255 and all tests passing

pkg/serving/v1/client.go
```diff
@@ -314,24 +314,28 @@ func (cl *knServingClient) ApplyService(ctx context.Context, modifiedService *se
 // Param `timeout` represents a duration to wait for a delete op to finish.
 // For `timeout == 0` delete is performed async without any wait.
 func (cl *knServingClient) DeleteService(ctx context.Context, serviceName string, timeout time.Duration) error {
+	service, err := cl.GetService(ctx, serviceName)
+	if err != nil {
+		return err
+	}
```
maximilien, line 320: Same as previous...

maximilien (changes requested):
/ok-to-test
Left some feedback, especially related to some missing coverage. Of course we cannot cover all but at least see if you can solve one --- some are same in different locations.

dsimansk:
> Left some feedback, especially related to some missing coverage. Of course we cannot cover all but at least see if you can solve one --- some are same in different locations.

Thanks, sure I'll take a pass on the test. The waiting loop is pretty tricky anyway, so more coverage is definitely good goal.

rhuss:
@dsimansk Thanks a lot ! I won't be able to make review this week, but will have a look asap early next week.

maximilien:
Sounds good. Thanks @dsimansk

navidshaikh (approved):
/lgtm

PR #1306, opened 2021-05-07
Fix for serviceaccounts "default" not found flaky issue

pratham-m:
## Description
This PR intends to fix `serviceaccount "default" not found` flaky issue which is a race condition observed on some OCP4.x platforms.

## Changes
`lib/test/integration.go` now waits for `default` serviceaccount to be ready before returning a new `KnTest` object when executing e2e tests.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -20,6 +20,10 @@
 |🐛
 | Respect `-o` in `list` commands in case if no data present
 | https://github.com/knative/client/pull/1276[#1276]
+
+| 🐣
+| Fix for serviceaccounts "default" not found flaky issue
+| https://github.com/knative/client/pull/1306[#1306]
 |===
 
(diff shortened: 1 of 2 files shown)


dsimansk:
@pratham-m thanks looks good!

Can you please run `./hack/build.sh -c ` there's a formatting issue in the code. Otherwise the integrations tests passed with this change.

pratham-m:
@dsimansk done with code formatting changes

dsimansk:
/lgtm

@rhuss @navidshaikh @maximilien PTAL. 
This PR should stabilize our test suite on clusters that might have a "slower" namespace initialization, hence the description. Afaik @pratham-m was able to achieve more consistent results with additional check/wait. On the hand it should *not* affect in any way our CI.

maximilien (approved):
Only change in test and seems to be OK to me.
/ok-to-test
/approve
/lgtm

PR #1312, opened 2021-05-13
Fix replicas *int32 type in describe commands

dsimansk:
## Description

Per https://github.com/knative/serving/issues/11170 the `actualReplicas, desiredReplicas` fields were changed to `*int32`. This PR brings it to client.

## Changes

* Fix replicas *int32 type in describe commands
* Update dependencies in separate commit


dsimansk:
@maximilien could you please take a look at this PR? As it might be blocking some of the release preparations and Roland's off on public holiday today.

maximilien (commented):
I generally dislike the idea of using pointers vs actual values for primitive types but I understand the need for absence of the value from the API. 
/ok-to-test

itsmurugappan (approved):
/lgtm

navidshaikh (approved):

PR #1326, opened 2021-05-24
Fix Subscription's Channel to use KRefence type

dsimansk:
## Description

Per update on Eventing side, use the correct type in client code.

## Changes

* Fix Subscription's Channel to use KRefence type

## Reference

https://github.com/knative/eventing/pull/5412

/cc @knative/client-reviewers


navidshaikh (approved):
/lgtm
/approve

PR #1328, opened 2021-05-24
Add `env-value-from` flag & keep order of env vars in created/updated services

### LINKED ISSUE #1165: kn seems to re-order the environment variables
### Bug report

It seems that `kn` does re-order the environment variables defined for a container.

### Expected behavior

I would expect the environment variables to stay in the same order.

Deploy a Knative service with `kubectl apply`. Execute `kn service update -e N=1`.

The the order of environment variables is import, e.g. in the following case:

~~~yaml
        - name: POSTGRES_DB
          valueFrom:
            configMapKeyRef:
              key: databaseName
              name: postgres-config
        - name: POSTGRES_PASSWORD
          valueFrom:
            secretKeyRef:
              key: admin.password
              name: postgres-secret
        - name: POSTGRES_USER
          valueFrom:
            secretKeyRef:
              key: admin.username
              name: postgres-secret
        - name: DATABASE_URL
          value: postgres://$(POSTGRES_USER):$(POSTGRES_PASSWORD)@postgres/$(POSTGRES_DB)
~~~

In the order show above, the `$(DATABASE_URL)` will properly be constructed. Doing a `kn update` will destroy that order, and the evaluation of variables will fail.

### kn version

```
Version:      v0.19.1
Build Date:   2020-11-25 18:27:45
Git Revision: f128d045
Supported APIs:
* Serving
* Eventing
```

> Nightly
> 0.12.x
> 0.11.x
> 0.10.x
> 0.9.x

zroubalik:
## Description

Add `env-value-from` flag to create command & keep order of env vars in created/updated services

## Changes

* `env-value-from` flag in `create` command : enables adding environment variables from a value of key in ConfigMap or Secret. Syntax is: `--env-value-from NAME=cm:myconfigmap:key`, for a ConfigMap prefix is `config-map:` or `cm:` for a Secret it is `secret:` or `sc:`.
* When creating or updating a service, order of env variables specified as flags is being kept, users can specify multiple `--env` and/or `--env-value-from` flags and the order is kept.

## Oustanding issues
1. ~[TestServiceCreateEnvMock](https://github.com/knative/client/blob/d0327bca0766247eb49e7fb71d71c34577ec1b8f/pkg/kn/commands/service/create_mock_test.go#L83), [TestServiceCreateEnv](https://github.com/knative/client/blob/d0327bca0766247eb49e7fb71d71c34577ec1b8f/pkg/kn/commands/service/create_test.go#L212) and `TestServiceUpdateEnvMock` fail because they are trying to assign env variable without a value, I wonder whether this is something that we should allow? My opinion that we should not. (eg. `--env=EMPTY`)~
2. ~There are some tests ([TestServiceCreateEnvForce](https://github.com/knative/client/blob/main/pkg/kn/commands/service/create_test.go#L735), `TestServiceCreateFromYAMLWithOverride`, `TestServiceUpdateEnv`,...) that are failing because those tests are checking env variables in the service, but because those tests are not invoked as commands the `os.Args` array doesn't contain flags with env variables, but the actual args from the test framework. I haven't find a way how to mock this, cobra unfortunately doesn't allow this.~

## Reference

Fixes #1165


pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -199,7 +200,7 @@ func (p *ConfigurationEditFlags) Apply(
 
 	template := &service.Spec.Template
 
-	err := p.PodSpecFlags.ResolvePodSpec(&template.Spec.PodSpec, cmd.Flags())
+	err := p.PodSpecFlags.ResolvePodSpec(&template.Spec.PodSpec, cmd.Flags(), os.Args)
```
zroubalik, line 203: `os.Args` -> this is the only way how to get all orderd list of all arguments passed to the command (so we do the right order of `--env` and `--env-value-from`. Unfortunately Cobra can't do that, it works on already processed args and flags.

dsimansk:
Thanks @zroubalik!
/ok-to-test

pkg/kn/flags/podspec_test.go
```diff
@@ -114,31 +115,31 @@ func TestPodSpecResolve(t *testing.T) {
...
-			err := flags.ResolvePodSpec(podSpec, cmd.Flags())
+			err := flags.ResolvePodSpec(podSpec, cmd.Flags(), inputArgs)
 			fmt.Fprint(cmd.OutOrStdout(), "Return error: ", err)
 		},
 	}
 	testCmd.SetOut(&outBuf)
 
-	args := []string{"--mount", "123456"}
-	testCmd.SetArgs(args)
+	testCmd.SetArgs(inputArgs)
```
dsimansk, line 142: @zroubalik did you try to mock `os.Args` in here to help with various unit tests that currently fails?

https://github.com/knative/client/pull/1328/checks?check_run_id=2656432279#step:6:27

zroubalik: Yeah, I just wanted to make this testcase consistent with the others. These tests are okay, the problem is in those, where we don't invoke `ResolvePodSpec()` directly, but not using the command line for the actual invocaton. As I mentioned in the PR description.

zroubalik:
> /ok-to-test
> This is a cool feature. Is the DRAFT mode because you have more TODO and thus will change a lot or is this ready for review?

Sorry for the delay, I have been busy with other stuff recently. I need to tweak a few failing tests and then I will convert it from draft. Probably early next week.

rhuss:
@zroubalik no pressure, but what is the state of this PR ? A release is planned for tomorrow so I wonder whether we could add it still or maybe its better to postpone it until next release.

zroubalik:
@rhuss I had another priorities, but can tackle this one this week. 

The only real issue I have are those issues with tests, as mentioned in PR description. Due to the fact, that this feature needs to use `os.args` a few tests are failing. WDYT about this?

zroubalik:
This is ready for review, thanks for the patience folks 😄

pkg/kn/flags/podspec_helper_test.go
```diff
@@ -35,15 +35,149 @@ func getPodSpec() (*corev1.PodSpec, *corev1.Container) {
...
-	env := []corev1.EnvVar{
+	expected := []corev1.EnvVar{
 		{Name: "a", Value: "foo"},
 		{Name: "b", Value: "bar"},
 	}
-	found, err := util.EnvToMap(env)
+	argsEnv := []string{
+		"a=foo",
+		"b=bar",
+	}
```
dsimansk, line 45: @zroubalik could please add a test case that uses reverse order (or some alphabetical mix), that'd verify the issue with kn always ordering by key.

CHANGELOG.adoc
```diff
@@ -1351,5 +1351,9 @@ Ignore PRs:
 |===
 | | Description | PR
 
+| 🎁
+| Add `env-value-from` flag & keep order of env vars in created/updated Services
+| https://github.com/knative/client/pull/1328[#1328]
+
```
dsimansk, line 1357: This should go to the top of the changelog. The latest `main` should have the initialized `Unreleased` section now.

pkg/kn/flags/podspec_helper.go
```diff
@@ -53,15 +52,45 @@ func containerOfPodSpec(spec *corev1.PodSpec) *corev1.Container {
...
+
+	envIterator := envToUpdate.Iterator()
+	envValueFromIterator := envValueFromToUpdate.Iterator()
+
+	envKey, envValue, envExists := envIterator.NextString()
+	envValueFromKey, envValueFromValue, envValueFromExists := envValueFromIterator.NextString()
+	for _, arg := range allArgs {
+		// envs are stored as NAME=value
+		if envExists &&
+			(strings.HasPrefix(arg, envKey+"="+envValue) || strings.HasPrefix(arg, "-e="+envKey+"="+envValue) || strings.HasPrefix(arg, "--env="+envKey+"="+envValue)) {
```
dsimansk, line 70: Maybe I'd extract this condition to standalone helper function. That'd allow easier unit testing eventually. Wdyt?

pkg/kn/flags/podspec_helper_test.go
zroubalik: Done: https://github.com/knative/client/pull/1328/commits/97466727ac5bd37ce9a08e18602827fd4d4b6858#diff-88a7bb094597d334c604ecc29d3780cc31646e892d4517b3bc8c833eb833bac8R54

pkg/kn/flags/podspec_helper.go
zroubalik: Helper fn: https://github.com/knative/client/pull/1328/commits/97466727ac5bd37ce9a08e18602827fd4d4b6858#diff-5d48e95bf218a0a7a822b3f8dc2db615d9ab023ad71a01658071fca6c7a6966dR97-R108

Tests: https://github.com/knative/client/pull/1328/commits/97466727ac5bd37ce9a08e18602827fd4d4b6858#diff-88a7bb094597d334c604ecc29d3780cc31646e892d4517b3bc8c833eb833bac8R675-R765

dsimansk:
Thanks @zroubalik! It looks good to me now.

/lgtm

PR #1333, opened 2021-05-31
Refactor waiting loop to use resourceVersion

### LINKED ISSUE #1287: Refactor watch functions to pass resourceVersion

dsimansk:
## Description

@rhuss per our conversation opening the PR with refactor of waiting loop. It was reverted before `0.23` release due to unstable tests probably cause by the changes.
https://github.com/knative/client/pull/1321

## Changes

* Refactor waiting loop to use resourceVersion

## Reference

Fixes #1287


maximilien (commented):
Ping me when you are ready for a second pair of eyes review on this. Best.

rhuss:
@dsimansk any idea how we should proceed with this PR ? Do you remember the original error that we have observed ? Doesn't look like that it occurs very often ...

rhuss:
We decided in the WG call that we want to merge this or after the release, and then have 6 week time to detect and fix any potential issues that only occurred once in the past (not sure anymore whether this was a flake)

rhuss (commented):
Thanks !

Let's get the merged back in and watch the nightly builds

/lgtm
/approve

PR #1343, opened 2021-06-12
increase code cov for sources

itsmurugappan:
## Description

Increase code coverage for sources


dsimansk:
Can you rerun `Error: Please run ./hack/update-codegen.sh.` pls? Probably a conflict from `main` merge.

pkg/kn/commands/source/apiserver/apiserver_test.go
```diff
@@ -80,11 +80,15 @@ func cleanupAPIServerMockClient() {
 	apiServerSourceClientFactory = nil
 }
 
-func createAPIServerSource(name, resourceKind, resourceVersion, serviceAccount, mode string, ceOverrides map[string]string, sink duckv1.Destination) *v1.ApiServerSource {
-	resources := []v1.APIVersionKindSelector{{
-		APIVersion: resourceVersion,
-		Kind:       resourceKind,
-	}}
+func createAPIServerSource(name, serviceAccount, mode string, resourceKind, resourceVersion []string, ceOverrides map[string]string, sink duckv1.Destination) *v1.ApiServerSource {
```
maximilien, line 83: Whenever I see a function with so many parameters (in code or test) I wonder whether a struct should be created to hold the related params and the struct passed to the func or be a method?

maximilien (commented):
/ok-to-test
Left a comment. Thanks for doing this. Seems like a third_party dependency is causing some of the issue in CI.

dsimansk:
@itsmurugappan wrt the verify error, In my case I've update `go-license` tool to get the same result as CI.
```
go install github.com/google/go-licenses@latest

```

navidshaikh (approved):
/lgtm


@itsmurugappan: thanks! Are you going to add more more to the PR? else please unhold

itsmurugappan:
@navidshaikh i have pushed some more changes. This PR is good to go. I am also working on a follow up or with some unit tests to increase the coverage.
/unhold

pkg/kn/commands/source/container/create.go
```diff
@@ -76,14 +76,10 @@ func NewContainerCreateCommand(p *commands.KnParams) *cobra.Command {
 
 			b := v1.NewContainerSourceBuilder(name).Sink(*objectRef).PodSpec(*podSpec)
 			err = srcClient.CreateContainerSource(cmd.Context(), b.Build())
-			if err != nil {
-				return fmt.Errorf(
-					"cannot create ContainerSource '%s' in namespace '%s' "+
-						"because: %s", name, namespace, err)
```
dsimansk, line 82: Does the error message still have the same wording even without this wrapper?

pkg/kn/commands/source/container/create_test.go
```diff
@@ -60,3 +60,22 @@ func TestNoSinkError(t *testing.T) {
...
+func TestContainerCreatePSError(t *testing.T) {
+	testsvc := &servingv1.Service{
+		TypeMeta:   metav1.TypeMeta{Kind: "Service", APIVersion: "serving.knative.dev/v1"},
+		ObjectMeta: metav1.ObjectMeta{Name: "testsvc", Namespace: "default"},
+	}
+	dynamicClient := dynamicfake.CreateFakeKnDynamicClient("default", testsvc)
+	containerClient := v1.NewMockKnContainerSourceClient(t)
+
+	_, err := executeContainerSourceCommand(containerClient, dynamicClient, "create", "testsource", "--sink", "ksvc:testsvc", "--image", "docker.io/test/testimg", "--mount", "123456")
+	assert.Error(t, err, "cannot create ContainerSource 'testsource' in namespace 'default' because: Invalid --mount: argument requires a value that contains the \"=\" character; got \"123456\"")
```
dsimansk, line 80: You could go with `assert.Assert(t, util.ContainsAll(out, ....)`, but it's minor nit really. I'm sure @rhuss would also notice though. :)

pkg/kn/commands/source/container/create.go
itsmurugappan: no, made it consistent across sources like ping and API server. Also it was difficult to cover this portion as all possible errors are handled before this step, like sink resolution, pod spec errors etc.. If you think this adds value I can put it back.

pkg/kn/commands/source/container/create_test.go
itsmurugappan: oh instead of comparing the full error message, just check for key words ?

dsimansk: Yes exactly, so we don't have to maintain the whole string that might be prone to changes.

pkg/kn/commands/source/container/create.go
dsimansk: I was just curious if it's altering the messages significantly. Anyway, I'm glad to hear it's consistent now. :+1:

pkg/kn/commands/source/container/create_test.go
navidshaikh: yes

itsmurugappan: have made the changes.

PR #1344, opened 2021-06-14
Update k8s dependencies to v0.20.7 

dsimansk:
## Description

Update K8s dependencies to follow-up on https://github.com/knative/pkg/pull/2145. 

## Changes

* Update k8s version
* Fix dynamic client tests

## Reference

Fixes #


pkg/dynamic/client_test.go
```diff
@@ -170,14 +171,27 @@ func TestListSourcesUsingGVKs(t *testing.T) {
...
+	{Group: "eventing.knative.dev", Version: "v1", Resource: "brokers"}:                   "BrokerList",
+	{Group: "eventing.knative.dev", Version: "v1", Resource: "subscriptions"}:             "SubscriptionList",
+	{Group: "eventing.knative.dev", Version: "v1", Resource: "channels"}:                  "ChannelList",
+	{Group: "messaging.knative.dev", Version: "v1", Resource: "inmemorychannels"}:         "InMemoryChannelsList",
+}
+
 // createFakeKnDynamicClient gives you a dynamic client for testing containing the given objects.
 // See also the one in the fake package. Duplicated here to avoid a dependency loop.
 func createFakeKnDynamicClient(testNamespace string, objects ...runtime.Object) KnDynamicClient {
 	scheme := runtime.NewScheme()
```
dprotaso, line 190: You could technically drop the need for the gvrToListKind if you add the correct types to the scheme

ie. for k8s types invoke `scheme.AddToScheme(scheme)` from https://github.com/kubernetes/client-go/blob/1bccfc8c60977e2ce3235394daff128996818f7b/kubernetes/scheme/register.go#L135

You can do this with eventing and serving and save yourself from maintaining this list. Unsure if there's something that prevents you from importing those.

dprotaso: I also had some issues where I couldn't provide unstructured types to the fake dynamic client. There would be an error with the way the fake client interacts with its object tracker

So I wrote some code for the fake injection client to rewire these schemes to use unstructured.* types
https://github.com/knative/pkg/pull/2145/files#diff-8438a08a79f266761b674dbf8748084e62be792ce56e7d7b598d148c069946e5L40

Unsure if you would need to do the same - if so maybe we pull that bit of logic out out injection and into some sort of fakedynamic package in `knative.dev/pkg`

pkg/dynamic/fake/fake.go
```diff
@@ -15,26 +15,30 @@
...
-	scheme.AddKnownTypeWithName(schema.GroupVersionKind{Group: "messaging.knative.dev", Version: "v1", Kind: "Channel"}, &messagingv1.Channel{})
-	client := dynamicfake.NewSimpleDynamicClient(scheme, objects...)
-	return dynamic.NewKnDynamicClient(client, testNamespace)
+	servingv1.AddToScheme(scheme)
+	eventingv1.AddToScheme(scheme)
+	messagingv1.AddToScheme(scheme)
+	sourcesv1.AddToScheme(scheme)
+	sourcesv1beta2.AddToScheme(scheme)
+	apiextensionsv1.AddToScheme(scheme)
+	_, dynamicClient := dynamicclientfake.With(context.TODO(), scheme, objects...)
```
dsimansk, line 42: @dprotaso @navidshaikh wdyt about the above change? It works for our test setup, I've gone through and changed various initialization of fake dynamic client to use the common function. 

Indeed I had to use `knative/pkg` for unstructured object to get them converted properly.

Finally, Navid do you see any problem in bringing `apiextensions-apiserver` dependency for the proper scheme?

dprotaso: Yeah looks good - I might PR a change to k8s to fix this problem but it'll take weeks/months for us to switch over to it

navidshaikh (approved):
/lgtm
/approve

PR #1353, opened 2021-06-24
Prettify printing of webhook warnings

### LINKED ISSUE #1350: Investigate Kubernetes warnings support
### Feature request

Since Kubernetes 1.19, webhooks can add warnings to the request that `kubectl` shows by default, see https://kubernetes.io/blog/2020/09/03/warnings/. It'd be awesome to also support this in `kn` in a nice way, so that we can warn users of potentially making mistakes without straight out failing requests.

### Use case

Well, see above, warning users about things that might be okay but we still want to tell them it might not be :joy:
 
/kind proposal

markusthoemmes:
## Description

Adds a proper warning printer to have warnings being printed in a similar format as "other messages". Color could be easily supported but it seemed like `kn` doesn't color in general. Let me know if that aint the case.

## Changes

Output before:

```
W0624 13:35:20.413073  372781 warnings.go:70] This is a test!
Replacing service 'foo' in namespace 'default':
...
```

Output after:

```
Warning: This is a test!
Replacing service 'foo' in namespace 'default':
...
```

## Reference

Fixes #1350


rhuss (commented):
thanks !

Looks good to me, but not sure how make codecov happy. I don't think codecov is a mandatory check, so lets try to merge it.

/lgtm
/approve

PR #1360, opened 2021-06-29
update versions to 0.24

dprotaso:


dprotaso:
/hold 

need to update the integration test file

PR #1364, opened 2021-07-05
Add an update timestamp to trigger the creation of a revision when needed.

### LINKED ISSUE #1318: can no longer force a re-download of an image via "kn update" w/o extra stuff
Before v0.21.0 I used to be able to do:
```
$ kn service update echo --image duglin/echo
```
and it would always force a new revision to be created. This was really handy when all I wanted was for the image to be re-downloaded because I did a "docker build" to update it. Notice, I don't have a tag on the image name.

It would be nice if this feature was added back in. W/o it, it's not clear to me how to force this to happen.... at least not w/o some (non-obvious to a newbie) hack, like specifying a revision name or adding `--env foo=$RANDOM` to the cmd line.

rhuss:
Add an update timestamp anytime a new revision should be triggered. Previously client-side revision names were this trigger but since we switched to server-side revision naming by default that didn't work anymore (and hasn't worked with server-side revision naming before, too).

Fixes #1318.

/lint


pkg/serving/revision_template.go
```diff
@@ -33,13 +33,49 @@ func ContainerOfRevisionTemplate(template *servingv1.RevisionTemplateSpec) (*cor
...
 
+// ContainerIndexOfRevisionSpec returns the index of the "main" container if
+// multiple containers are present. The main container is either the single
+// container when there is only ony container in the list or the first container
+// which has a ports declaration (validation guarantees that there is only one
+// such container)
+// If no container could be found (list is empty or no container has a port declaration)
+// then -1 is returned
+// This method's logic is taken from RevisionSpec.GetContainer()
+func ContainerIndexOfRevisionSpec(revisionSpec *servingv1.RevisionSpec) int {
```
dsimansk, line 55: Can we have a bit more unit test coverage for this function and later one as well?

pkg/serving/config_changes.go
```diff
@@ -91,31 +92,46 @@ func UpdateConcurrencyLimit(template *servingv1.RevisionTemplateSpec, limit int6
...
+func UpdateUserImageAnnotation(template *servingv1.RevisionTemplateSpec) {
 	// If the current image isn't by-digest, set the user-image annotation to it
 	// so we remember what it was.
-	currentContainer, _ := ContainerOfRevisionTemplate(template)
-	ui := currentContainer.Image
-	if strings.Contains(ui, "@") {
-		prev, ok := template.Annotations[UserImageAnnotationKey]
+	currentContainer, err := ContainerOfRevisionTemplate(template)
+	if err != nil {
+		return
```
dsimansk, line 106: Shouldn't we propagate the actual error?

rhuss: Indeed, if an error is returned it should be propagated. However, I think returning an error from `ContainerOfRevisionTemplate` is fishy anyway, as a nil-container is a valid outcome anyway and the only case, when error is returned, is when the template does not has a container. I remover the error slot therefore and did some addition minor refactorings.

rhuss:
@dsimansk PTAL, added some tests and addressed the error return value issue.

pkg/serving/revision_template.go
rhuss: 👍

dsimansk:
@rhuss looks good. However lint check isn't happy.

```
  Error: non-wrapping format verb for fmt.Errorf. Use `%w` to format errors (errorlint)
  Error: SA5011: possible nil pointer dereference (staticcheck)
  Error: SA5011(related information): this check suggests that the pointer can be nil (staticcheck)
  Error: SA5011: possible nil pointer dereference (staticcheck)
  Error: SA5011(related information): this check suggests that the pointer can be nil (staticcheck)
  Error: SA5011: possible nil pointer dereference (staticcheck)
  Error: SA5011(related information): this check suggests that the pointer can be nil (staticcheck)
```

rhuss:
@dsimansk thanks for the review. Really would love to catch those lint errors with our build script. Any idea ?

dsimansk:
> @dsimansk thanks for the review. Really would love to catch those lint errors with our build script. Any idea ?

Agreed, we should add it to the script. The initial run may take a bit longer ~20s, but then cache will speed it up significantly. And it's a matter of executing:
```
golangci-lint run -v 
```

dsimansk:
Thanks!

/approve
/lgtm

PR #1380, opened 2021-07-13
make command as array instead of string

### LINKED ISSUE #890: make 'cmd' an stringArray instead of string
Technically, the "command" in a podSpec can be an array of strings, not just a single string. So, it might make sense to support that from `kn`.  I've needed to use this before and had to split things across 'cmd' and 'args', which gave the same net result but wasn't what I wanted to do when trying to get my podSpec to align with the `kn` cmd line.

itsmurugappan:
## Description

make command as string array instead of string

## Reference

Fixes #890

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -18,7 +18,7 @@
 | | Description | PR
 
 | ✨
-| make cmd flag as an array instead of string
+| make --cmd flag as an array instead of string
 | https://github.com/knative/client/pull/1380[#1380]
 |===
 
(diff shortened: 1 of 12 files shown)


CHANGELOG.adoc
```diff
@@ -12,6 +12,16 @@
 | https://github.com/knative/client/pull/[#]
 ////
 
+## Unreleased
+[cols="1,10,3", options="header", width="100%"]
+|===
+| | Description | PR
+
+| ✨
+| make cmd flag as an array instead of string
```
dsimansk, line 21:
```suggestion
| Make `--cmd` flag as an array instead of string
```

dsimansk:
Just a small nit in a changelog entry, otherwise lgtm. Thanks @itsmurugappan!

PR #1382, opened 2021-07-13
Add support for multi-containers

### LINKED ISSUE #997: Add CLI support to specify multiple containers
### Feature request

Knative Serving release [v0.16.0](https://github.com/knative/serving/releases/tag/v0.16.0) has support for multiple containers as an alpha feature.

To propose multi container feature to beta user experience is one of the important aspects to be considered as part of CLI(kn).

The samples has been added as part of [PR](https://github.com/knative/docs/pull/2732)

For more info refer serving [issue](https://github.com/knative/serving/issues/8665) 

### Use case

User will get the flexibility to provide multiple containers information as part of command line tool.
### UI Example

```
```

/kind proposal

dsimansk:
## Description

This PR adds support for multiple containers through file or piped cmds.

The `container create` is not truly create cmd, rather it's printing `PodSpec.Containers` to `stdout` in yaml for now. Afterwards it can be combined through unix pipe or  stored in file and used as e.g. `--containers containers.yaml` .

UX examples:
```
kn container create --image foo.bar | \
kn container create --image bar.foo | \
kn service create --image yet.another --containers -
```

```
kn service create --image yet.another --containers containers.yaml
```

/cc @rhuss @dprotaso 

## Changes

* Add `container create` command
* Add `--containers` flag to `service create/update`
* TBD tests tests tests

## Reference

Fixes #997


pkg/kn/flags/podspec_helper.go
```diff
@@ -255,6 +258,27 @@ func UpdateImagePullSecrets(spec *corev1.PodSpec, pullsecrets string) {
 	}
 }
 
+// UpdateContainers updates the containers array with additional ones provided from file or os.Stdin
+func UpdateContainers(spec *corev1.PodSpec, containers string) error {
```
dsimansk, line 262: Looking at the changes and how to add test, I'd like to move the parsing logic to a separate function and make this one just accept `[]Containers` array.

pkg/kn/flags/podspec_helper.go
```diff
@@ -255,6 +258,11 @@ func UpdateImagePullSecrets(spec *corev1.PodSpec, pullsecrets string) {
 	}
 }
 
+// UpdateContainers updates the containers array with additional ones provided from file or os.Stdin
+func UpdateContainers(spec *corev1.PodSpec, containers []corev1.Container) {
+	spec.Containers = append(spec.Containers, containers...)
+}
+
```
dsimansk, line 265: @rhuss the current mode is append additional containers. However, that might not work very well for update operation. I wonder if replacing whatever `spec.Containers[1:]` is a good solution.
Additionally we could require container's `name` parameter and match those if Containers array is already populated.

rhuss: That's a good question. Updating a list is always tricky, but I like the idea with correlating via name. "update" is always an incremental operation, you don't have to specify the full state (that is different to "apply"). But if we would like to update based on position in the list, you would need to be able either to specify the index of the container within this list or provide always all containers to update. I think it is fair to require the container name (and I would even enforce this on creation). If we update a list with containers where not all containers have a name (e.g. on existing services that thas not been created with kn), then we would just ignore those containers on an update.

wdyt ? is this a fair solution ? @dprotaso as kind of stakeholder for this feature ;-) would it be ok to require a name for a container that is managed by kn in a multi-container setup ?

pkg/kn/commands/container/create.go
```diff
@@ -0,0 +1,89 @@
...
+			}
+			// Add container's name to current one
+			if name != "" {
+				podSpec.Containers[0].Name = name
+			}
+			b, err := sigyaml.Marshal(podSpec)
+			if err != nil {
+				return err
+			}
+			fmt.Fprintf(cmd.OutOrStdout(), "%s", b)
```
maximilien, line 71: What is this for? Debug and left out by mistake?

pkg/kn/commands/container/create.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	machineReadablePrintFlags.AddFlags(cmd)
+	podSpecFlags.AddFlags(cmd.Flags())
+	// Volume is not part of ContainerSpec
+	cmd.Flag("volume").Hidden = true
+
+	return cmd
+}
+
+func detectPipeInput(f *os.File) bool {
```
maximilien, line 83: Could this be a public util func?

pkg/kn/commands/container/create_test.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"os"
+	"testing"
+
+	"knative.dev/client/pkg/util"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestContainerCreate(t *testing.T) {
```
maximilien, line 26: What about some "negative" tests... ?

pkg/kn/commands/container/create_test.go
```diff
@@ -0,0 +1,51 @@
...
+	output, err := executeContainerCommand("create", "foo", "--image", "registry.foo:bar")
+	assert.NilError(t, err)
+	assert.Assert(t, len(output) > 0)
+	assert.Assert(t, util.ContainsAllIgnoreCase(output, "containers", "image", "foo", "registry.foo:bar"))
+}
+
+func TestDetectPipeInput(t *testing.T) {
```
maximilien, line 33: Also add test passing `--containers`

pkg/kn/flags/podspec_test.go
```diff
@@ -126,6 +132,121 @@ func TestPodSpecResolve(t *testing.T) {
 	testCmd.Execute()
 }
 
+func TestPodSpecResolveContainers(t *testing.T) {
+	rawInput := `
```
maximilien, line 136: Add negative test for empty file and for bad file format?

maximilien (changes requested):
Various questions and feedback. Happy to take a second look when you are ready to remove the WIP tag.

And of course, many thanks for the contribution. Cheers 🍻

pkg/kn/commands/container/container_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	output := new(bytes.Buffer)
+	knParams.Output = output
+
+	cmd := NewContainerCommand(knParams)
+	cmd.SetArgs(args)
+	cmd.SetOut(output)
+
+	cmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
+		return knflags.ReconcileBoolFlags(cmd.Flags())
```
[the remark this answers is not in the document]
dsimansk: It's copied from other test setups, I'll check if it's needed.

pkg/kn/commands/container/create.go
dsimansk: Actually it core `container create` functionality. The idea is to feed containers snippets in YAML to `service create/update`. And additionally to chain multiple `container create`. 
From that point of view the `container` cmd is not traditional input -> Knative backend -> response, rather it's producing those container YAML snippets to `stdout`.
Hence the other comment about `|` (pipe) symbol in the description, the pipe relies on that idea that user is able to design own container snippet with a series of kn invocations and then feed it into `kn service create ... --containers -`.

Note that Roland has suggested `kn container add`, that seems more appropriate/descriptive for the operation.

docs/cmd/kn_container.md
```diff
@@ -0,0 +1,29 @@
+## kn container
+
+Manage containers
```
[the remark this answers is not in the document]
dsimansk: Per my latter comment, even the `Manage` isn't describing the purpose and utility of `kn container`. I'm open to suggestions here. :)

pkg/kn/commands/container/create.go
maximilien: OK. Thx

docs/cmd/kn_container_create.md
```diff
@@ -0,0 +1,61 @@
...
+
+  The 'container create' represents utility command that prints YAML container spec to standard output. It's useful for
+  multi-container use cases to create definition with help of standard 'kn' option flags. The command can be chained through
+  Unix pipes to create multiple containers at once.
+
+  # Create a container 'sidecart' from image 'docker.io/example/sidecart'
+  kn container create sidecart --image docker.io/example/sidecart
+
+  # Create command can be chained by standard Unix pipe symbol '|'
+  kn container create sidecart --image docker.io/example/sidecart | \
```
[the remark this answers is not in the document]
maximilien: I can see why a pipe | is useful.

dsimansk:
Unrelated sink test failed.
/retest

dsimansk:
@maximilien @itsmurugappan @rhuss I'd like to kindly ask for another review cycle so we could aim this release for release tomorrow. :crossed_fingers:

docs/cmd/kn_container_add.md
```diff
@@ -0,0 +1,61 @@
...
+kn container add NAME
+```
+
+### Examples
+
+```
+
+  The 'container add' represents utility command that prints YAML container spec to standard output. It's useful for
+  multi-container use cases to create definition with help of standard 'kn' option flags. The command can be chained through
+  Unix pipes to create multiple containers at once.
```
rhuss, line 15: Maybe mention here that this command supports all container relevant options like `kn service create` ?

docs/cmd/kn_container_add.md
```diff
@@ -0,0 +1,61 @@
...
+  -o, --output string                 Output format. One of: json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-as-json|jsonpath-file.
+  -p, --port string                   The port where application listens on, in the format 'NAME:PORT', where 'NAME' is optional. Examples: '--port h2c:8080' , '--port 8080'.
+      --pull-secret string            Image pull secret to set. An empty argument ("") clears the pull secret. The referenced secret must exist in the service's namespace.
+      --request strings               The resource requirement requests for this Service. For example, 'cpu=100m,memory=256Mi'. You can use this flag multiple times. To unset a resource request, append "-" to the resource name, e.g. '--request cpu-'.
+      --service-account string        Service account name to set. An empty argument ("") clears the service account. The referenced service account must exist in the service's namespace.
+      --template string               Template string or path to template file to use when -o=go-template, -o=go-template-file. The template format is golang templates [http://golang.org/pkg/text/template/#pkg-overview].
```
rhuss, line 44: Not sure what this is about ?

docs/cmd/kn_container_add.md
```diff
@@ -0,0 +1,61 @@
...
+      --user int                      The user ID to run the container (e.g., 1001).
+```
+
+### Options inherited from parent commands
+
+```
+      --cluster string      name of the kubeconfig cluster to use
```
rhuss, line 51: It's probably out of scope for this PR, but cluster/config configuration doesn't make any sense for this offline-online command (and might even confuse users)

pkg/kn/commands/container/add.go
```diff
@@ -0,0 +1,93 @@
...
+	return cmd
+}
+
+// IsPipeInput determines the input is created using unix '|' pipe
+func IsPipeInput(f *os.File) bool {
+	fi, err := f.Stat()
+	if err != nil {
+		return false
+	}
+	return fi.Mode()&os.ModeCharDevice == 0
```
rhuss, line 92: cool trick :-)

docs/cmd/kn_container_add.md
```diff
@@ -0,0 +1,61 @@
...
+  multi-container use cases to create definition with help of standard 'kn' option flags. The command can be chained through
+  Unix pipes to create multiple containers at once.
+
+  # Add a container 'sidecar' from image 'docker.io/example/sidecar'
+  kn container add sidecar --image docker.io/example/sidecar
+
+  # Add command can be chained by standard Unix pipe symbol '|'
+  kn container add sidecar --image docker.io/example/sidecar | \
+  kn container add second --image docker.io/example/sidecar:second
+```
```
rhuss, line 23: I would add here also the final `kn service create|update|apply` command, so that this example is kind of self-contained

hack/tools.go
```diff
@@ -23,4 +23,6 @@ import (
 	_ "k8s.io/code-generator"
 	_ "k8s.io/code-generator/cmd/deepcopy-gen"
 	_ "knative.dev/serving/test/test_images/grpc-ping"
+	_ "knative.dev/serving/test/test_images/multicontainer/servingcontainer"
+	_ "knative.dev/serving/test/test_images/multicontainer/sidecarcontainer"
```
rhuss, line 27: just curious, why is this needed ?

pkg/kn/commands/container/container_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"gotest.tools/v3/assert"
+	"knative.dev/client/pkg/kn/commands"
+)
+
+func TestContainerCommand(t *testing.T) {
+	knParams := &commands.KnParams{}
+	containerCmd := NewContainerCommand(knParams)
+	assert.Equal(t, containerCmd.Name(), "container")
+	assert.Equal(t, containerCmd.Use, "container COMMAND")
```
rhuss, line 29: this is more testing cobra :-) But I think its important to get to the required code coverage. One reason, why I think that code coverage alone is not the holy grail.

pkg/kn/flags/podspec_helper.go
```diff
@@ -255,6 +258,14 @@ func UpdateImagePullSecrets(spec *corev1.PodSpec, pullsecrets string) {
 	}
 }
 
+// UpdateContainers updates the containers array with additional ones provided from file or os.Stdin
+func UpdateContainers(spec *corev1.PodSpec, containers []corev1.Container) {
+	c := containerOfPodSpec(spec)
+	spec.Containers = []corev1.Container{}
+	spec.Containers = append(spec.Containers, *c)
+	spec.Containers = append(spec.Containers, containers...)
```
rhuss, line 266: I haven't dived deep enough into the code, but shouldn't you correlated via the container name instead of just appending ? I.e. lookup whether the extra containers' name are already present in the existing containers and replace those in case instead of just appending ?

rhuss (commented):
Finished a first round of reviews. The only thing that I think is critical is that the update should really correlate via the name over existing containers, so that e.g. `kn service apply` stays idempotent when the same command is issued twice.

Maybe also add test with updating containers with the same name ?

hack/tools.go
dsimansk: Those are used in e2e tests. There's `./test/test-images/multicontainer/release/yaml` thats points to that code.

pkg/kn/commands/container/container_test.go
dsimansk: Yep, entirely for the sake of test coverage. :)

docs/cmd/kn_container_add.md
dsimansk: It's leftover that was aiming to add output format flag. I'll remove it for now.

pkg/kn/flags/podspec_helper.go
dsimansk: Currently we're in `drop&replace` as I was aiming to have a solution to replace all sidecars at once. E.g. when `Ksvc` is created from file without named containers etc.


Is it better now? I can confirm that `kn apply` works with the change.
https://github.com/knative/client/pull/1382/commits/21567d37634c3eaface64afb2464e92e68a2d308

pkg/kn/commands/container/add.go
```diff
@@ -0,0 +1,90 @@
...
+				return err
+			}
+			// Add container's name to current one
+			podSpec.Containers[0].Name = name
+
+			b, err := sigyaml.Marshal(podSpec)
+			if err != nil {
+				return err
+			}
+			fmt.Fprintf(cmd.OutOrStdout(), "%s", b)
```
rhuss, line 72: I think you could just write to the stream directly without printf ?

pkg/kn/flags/podspec_helper.go
```diff
@@ -255,6 +258,28 @@ func UpdateImagePullSecrets(spec *corev1.PodSpec, pullsecrets string) {
...
+			for j, toUpdate := range containers {
+				if container.Name == toUpdate.Name {
+					spec.Containers[i] = containers[j]
+					matched = append(matched, toUpdate.Name)
+				}
+			}
+		}
+		for _, container := range containers {
+			if !util.SliceContainsIgnoreCase(matched, container.Name) {
+				spec.Containers = append(spec.Containers, container)
```
rhuss, line 277: Maybe you could a test for this line ? (e.g. an update which a new container that is not previously known ?)

pkg/kn/flags/podspec_helper.go
```diff
@@ -638,3 +663,22 @@ func reviseVolumesToRemove(volumeMounts []corev1.VolumeMount, volumesToRemove []
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	podSpec := &corev1.PodSpec{}
+	decoder := yaml.NewYAMLOrJSONDecoder(f, 512)
+	if err = decoder.Decode(podSpec); err != nil {
+		return nil, err
+	}
+	return podSpec, f.Close()
```
rhuss, line 683: Not sure if `Stdin` should be closed and what side-effect that might have. tbh, an error when closing a file when reading only from it could be also ignored imo

docs/cmd/kn_container_add.md
```diff
@@ -0,0 +1,59 @@
...
+
+```
+kn container add NAME
+```
+
+### Examples
+
+```
+
+  The 'container add' represents utility command that prints YAML container spec to standard output. It's useful for
```
rhuss, line 13: Could we mark the command as "experimental" ? Still not sure if people like the UX, so this could be a hint to them.

pkg/kn/flags/podspec.go
```diff
@@ -123,6 +125,11 @@ func (p *PodSpecFlags) AddFlags(flagset *pflag.FlagSet) []string {
 			"You can use this flag multiple times.")
 	flagNames = append(flagNames, "arg")
 
+	flagset.StringVarP(&p.ExtraContainers, "extra-containers", "", "",
+		"Specify path to file including definition for additional containers, alternatively use '-' to read from stdin. "+
+			"Example: --containers ./containers.yaml or --containers -.")
```
rhuss, line 130:
```suggestion
			"Example: --extra-containers ./containers.yaml or --extra-containers -.")
```

rhuss (commented):
thanks !

/approve
/lgtm

rhuss:
Looks good, but I think some test needs to be updated, that still uses `--containers`.

dsimansk:
> Looks good, but I think some test needs to be updated, that still uses `--containers`.

Yep, sorry I should have checked it better first time. Updated.

pkg/kn/flags/podspec_test.go
```diff
@@ -126,6 +132,139 @@ func TestPodSpecResolve(t *testing.T) {
...
+			origStdin := os.Stdin
+
+			fileName := tc.mockInput(tc.rawInput)
+			if fileName == "-" {
+				defer func() { os.Stdin = origStdin }()
+			} else {
+				defer os.RemoveAll(fileName)
+			}
+
+			inputArgs := []string{"--image", "repo/user/imageID:tag", "--containers", fileName}
```
rhuss, line 235: here's another `--containers`
```suggestion
			inputArgs := []string{"--image", "repo/user/imageID:tag", "--extra-containers", fileName}
```

itsmurugappan:
sorry @dsimansk  I could not get to this as there was a power outage in my area for most of the day and just got it back.

Issue #1352, opened 2021-06-24
Allow for running E2E tests by users with reduced privileges


mgencur:
### Feature request

It should be possible to run E2E tests by users without cluster-admin permissions. This will be useful in restricted environments where the container runtime is managed by someone else or the end user is simply given reduced permissions. It should be possible to verify the basic functionality of Knative Client even without cluster-admin permissions.

The current state is following. Some resources might not be created/applied with a regular user (say a user with "project admin" privileges):
* cannot run "oc adm policy add-scc-to-group anyuid" [here](https://github.com/openshift/knative-client/blob/release-v0.19.1/openshift/e2e-common.sh#L92)
* cannot run oc adm policy add-scc-to-group anyuid system:authenticated
* for each test, namespaces are created dynamically, kn client e2e TS creates a new namespace dynamically using [this code](https://github.com/openshift/knative-client/blob/release-v0.19.1/lib/test/integration.go#L227), i.e. "kubectl create ns", then deletes namespace with "kubectl delete ns" . This is already tracked in https://github.com/knative/client/issues/1304

If the above issues are worked-around, there are still failures:
5 tests fail
In general, they fail because the user does not have sufficient permissions for creating/patching resources:
```
TestSink
broker.go:31: error executing 'label namespace kne2etests6 eventing.knative.dev/injection=enabled': stderr: Error from server (Forbidden): namespaces "kne2etests6" is forbidden: User "mvinkler" cannot patch resource "namespaces" in API group "" in the namespace "kne2etests6"

TestSourceList
source_apiserver_test.go:117: assertion failed: error is not nil: stderr: Error from server (Forbidden): clusterroles.rbac.authorization.k8s.io is forbidden: User "mvinkler" cannot create resource "clusterroles" in API group "rbac.authorization.k8s.io" at the cluster scope

TestSourceListTypes
result_collector.go:75: ERROR: Error: missing apiVersion or kind; try GetObjectKind().SetGroupVersionKind() if you know the type

TestSourceApiServer
source_apiserver_test.go:117: assertion failed: error is not nil: stderr: Error from server (Forbidden): clusterroles.rbac.authorization.k8s.io is forbidden: User "mvinkler" cannot create resource "clusterroles" in API group "rbac.authorization.k8s.io" at the cluster scope

TestBrokerTrigger
[...]

PR #1383, opened 2021-07-14
Introduce -reusenamespace test flag

mgencur:
* allows for reusing test namespaces that were created in advance
(possibly by a cluster admininstrator)

Fixes one sub-issue from https://github.com/knative/client/issues/1352 which was originally tracked as https://github.com/knative/client/issues/1304

## Description

Introduce -reusenamespace flag which skips creating test namespaces. They're supposed to be created in advance by a user with sufficient privileges (cluster-admin).

This PR has a follow-up here: https://github.com/knative/client/pull/1384

## Changes

*

## Reference

Fixes #

Issue #1304, opened 2021-05-06
Allow for tests to run in a pre-created NS


mvinkler:
### Feature request

In some environments, it is not possible for the testing infrastructure to easily create new namespaces at will, so it would be nice if there was a way to run the e2e tests all within a single, pre-created, namespace.

### Use case

Running the kn client e2e TS on Openshift Dedicated cluster as a regular user.

rhuss:
So you mean to separate the part where the test namespaces are created and where the tests are run and make the 'namespace create' step optional ?  Sounds like a good idea to me.

rhuss:
Can we track the issue together with #1352 ? Closing this issue, for now, feel free to reopen it if needed.

PR #1384, opened 2021-07-14
Run tests as project admin

mgencur:
## Description

This is a follow up to https://github.com/knative/client/pull/1383 (this PR depends on it) and fixes the remaining sub-issues from https://github.com/knative/client/issues/1304
As a result, it will be possible to run the test suite as a project admin (vs. cluster-admin).

## Changes

* Create Broker explicitly rather than by labeling namespace
* Use Role instead of ClusterRole to work with Sources
* Separate SourceList test for cluster admin and regular user
* Instructions for running E2E as project admin

## Reference

Fixes https://github.com/knative/client/issues/1352

### DIFF
--- a/test/e2e/source_list_crd_test.go
+++ b/test/e2e/source_list_crd_test.go
@@ -1,0 +1,45 @@
+// Copyright 2019 The Knative Authors
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+
+//     http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or im
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build e2e
+// +build !serving
+// +build !project_admin
+
+package e2e
+
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+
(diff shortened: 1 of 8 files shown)


test/e2e/source_list_crd_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or im
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build e2e
+// +build !serving
+// +build !project_admin
```
rhuss, line 17: I would include some comment to the test, that this test only works for cluster admin and essentially the comment that you added to the issue. That way people are more aware that this test is special. 

We should also open an issue to fix this in the kn code and add a link to the issue here.

rhuss (commented):
Thanks @mgencur !

Looks good to me with a minor comment to add a comment ;-)

mgencur:
Thanks @rhuss , pushed.

test/e2e/source_list_crd_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"gotest.tools/v3/assert"
+
+	"knative.dev/client/lib/test"
+	"knative.dev/client/pkg/util"
+)
+
+// This test requires cluster admin permissions due to working with CustomResourceDefinitions.
+// It can be excluded from test execution by using the project_admin build tag.
+// See https://github.com/knative/client/issues/1385
```
rhuss, line 32: 👍

rhuss (commented):
/approve
/lgtm

PR #1390, opened 2021-07-19
Skip non modified event first while waiting for ready

### LINKED ISSUE #1389: WaitForReady should check for event type before trying to read generation number
### Bug report

The code at location supplied behaves badly in a situation when it receives an event of type different from _MODIFIED_, in which the underlying object doesn't contain a status field yet.

That situation is completely normal, as _ADDED_ event can be received first, and the resource at that point may not possess a _status_ field yet.

https://github.com/knative/client/blob/e26d5f27cab3c554963ea0f69507f89d766cd9b0/pkg/wait/wait_for_ready.go#L183-L207

Such situation I've observed while developing wait on ready based on dynamic client at https://github.com/knative-sandbox/kn-plugin-event/pull/38

The actual error that is being returned is similar to:

```
cannot extract status from &{map[metadata:map[creationTimestamp:<nil> generation:1 name:test-service] spec:map[template:map[metadata:map[creationTimestamp:<nil>] spec:map[containers:<nil>]]]]}
```

### Expected behavior

When _WaitForReady_ receives an event of type different from MODIFIED, it should skip it first, before checking anything in the underlying object structure.

https://github.com/cardil/knative-client/commit/7ca076911e098b1a68ab742f1a11bb1492c1d7c8

> Nightly
> 0.24

cardil:
## Description

Allow using wait for ready with dynamic kube client, as before this change, an error was received on _Added_ event with unstructured object, which by default doesn't have `.status` field.

## Changes

* 🐛 Bug fix of #1389 - Skipping on non-modified event while waiting for ready

## Reference

Fixes #1389

### DIFF
--- a/pkg/wait/wait_for_ready_test.go
+++ b/pkg/wait/wait_for_ready_test.go
@@ -25,10 +25,12 @@ import (
 	"gotest.tools/v3/assert"
 	"gotest.tools/v3/assert/cmp"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
 	"knative.dev/client/pkg/util"
 	"knative.dev/pkg/apis"
+	duckv1 "knative.dev/pkg/apis/duck/v1"
 	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
 )
 
@@ -89,9 +91,7 @@ func TestAddWaitForReady(t *testing.T) {
 			func(ctx context.Context, name string, initialVersion string, timeout time.Duration) (watch.Interface, error) {
 				return fakeWatchApi, nil
 			},
-			func(obj runtime.Object) (apis.Conditions, error) {
-				return apis.Conditions(obj.(*servingv1.Service).Status.Conditions), nil
-			})
+			conditionsFor)
 		fakeWatchApi.Start()
 		var msgs []string
 		err, _ := waitForReady.Wait(context.Background(), "foobar", "", Options{Timeout: &tc.timeout}, func(_ time.Duration, msg string) {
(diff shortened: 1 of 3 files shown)


pkg/wait/wait_for_ready_test.go
```diff
@@ -325,6 +322,25 @@ func peNormal(name string) ([]watch.Event, int) {
 	}, len(messages)
 }
 
+func peUnstructured(name string) ([]watch.Event, int) {
+	events, msgLen := peNormal(name)
+	for i, event := range events {
+		unC, err := runtime.DefaultUnstructuredConverter.ToUnstructured(event.Object)
+		if err != nil {
+			panic(err)
```
maximilien, line 330: Why not just fail test here with a message?

maximilien (commented):
/ok-to-test

itsmurugappan:
lgtm, but to pass the code cov test, the lines that were changed need to have more than 76.51 % coverage.

pkg/wait/wait_for_ready_test.go
cardil: Yes, I could. But, that requires passing down the `testing.T`. Panic in test isn't nice, because if it fires tests in that package will stop and not be reported properly.

I thought maybe it's not worth it, as this mapping to unstructured data should always pass. And to pass `testing.T` it would require additional code changes.

dsimansk:
> lgtm, but to pass the code cov test, the lines that were changed need to have more than 76.51 % coverage.

@cardil can you please take a look at the coverage? We've discussed on WG call that if it's too cumbersome to reach some of the statements it can be even merged without reaching the required number, just let us know please. :)

cardil:
@itsmurugappan @dsimansk I did raise the coverage on reported method, although it's a synthetic example that shouldn't occur with duck elements, which always should have generation and observed generation. Now, this PR passes all the gates.

itsmurugappan (approved):
/lgtm

itsmurugappan:
cool, coverage also jumping by 1.68%  👍

dsimansk:
Since it's addressing all the review comments, imo it's good to go.
/approve

cardil:
This PR needs _approve_ label also.

PR #1395, opened 2021-07-21
Fix issue with Windows pathes not being escaped when applying a regexp.

rhuss:
## Description

There have been issues when a full Windows path has been inserted literally to create a regexp, since the backslashes used as path separators are confusing the regexp engine. This fix will quote any input coming from the outside.

### DIFF
--- a/cmd/kn/main_test.go
+++ b/cmd/kn/main_test.go
@@ -335,6 +335,15 @@ func TestRunWithExit(t *testing.T) {
 	}
 }
 
+func TestExtractCommandPathFromErrorMessage(t *testing.T) {
+	for _, d := range []struct{ arg0, errMsg, expected string }{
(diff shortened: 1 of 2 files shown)


cmd/kn/main_test.go
```diff
@@ -335,6 +335,15 @@ func TestRunWithExit(t *testing.T) {
 	}
 }
 
+func TestExtractCommandPathFromErrorMessage(t *testing.T) {
```
dsimansk, line 338: :+1:

dsimansk:
Thanks! Looks good to me.

@rhuss fancy adding changelog entry for the fix?

maximilien (approved):
No easy why to really test this myself. Assuming you did.
/ok-to-test
/approve
/lgtm

PR #1396, opened 2021-07-22
Use namespaced kubectl for sa,role,rolebinding in tests

mgencur:
## Description

The resources are now namespaced-scoped so we should be using a kubectl with configured namespace

## Changes

*
*
*

## Reference

Fixes #


rhuss (commented):
thanks ! 'wonder why it worked before, maybe by accident because the context was already pointing to the right namespace

/lgtm
/approve

mgencur:
@rhuss you're probably right about the context.

PR #1402, opened 2021-07-27
Adding class option to broker create

### LINKED ISSUE #1269: kn broker create with broker class
### Feature request

It would be nice to be able to specify the broker class when creating a broker, in order to create brokers with classes different from default. For more info: https://knative.dev/docs/eventing/broker/#configuring-broker-classes

### Use case

Create a Kafka Broker without setting it as default broker class: https://knative.dev/docs/eventing/broker/kafka-broker/#create-a-kafka-broker

### UI Example

```
kn create broker mybroker --broker-class="Kafka"
```

vyasgun:
## Description
Adding --class option to broker create

## Changes

* Added the flag to broker create command
* Added unit test 
* Added e2e test 

## Reference

Fixes #1269

### DIFF
--- a/lib/test/broker.go
+++ b/lib/test/broker.go
@@ -32,6 +32,13 @@ func BrokerCreate(r *KnRunResultCollector, name string) {
 	assert.Check(r.T(), util.ContainsAllIgnoreCase(out.Stdout, "Broker", name, "created", "namespace", r.KnTest().Kn().Namespace()))
 }
 
+// BrokerCreateWithClass creates a broker with the given name and class.
+func BrokerCreateWithClass(r *KnRunResultCollector, name, class string) {
+	out := r.KnTest().Kn().Run("broker", "create", name, "--class", class)
+	r.AssertNoError(out)
+	assert.Check(r.T(), util.ContainsAllIgnoreCase(out.Stdout, "Broker", name, "created", "namespace", r.KnTest().Kn().Namespace()))
+}
+
 // BrokerDelete deletes a broker with the given name.
 func BrokerDelete(r *KnRunResultCollector, name string, wait bool) {
 	args := []string{"broker", "delete", name}
--- a/pkg/kn/commands/broker/create.go
+++ b/pkg/kn/commands/broker/create.go
@@ -74,6 +74,6 @@ func NewBrokerCreateCommand(p *commands.KnParams) *cobra.Command {
 		},
 	}
 	commands.AddNamespaceFlags(cmd.Flags(), false)
-	cmd.Flags().StringVarP(&className, "class", "c", "", "Broker class name.")
+	cmd.Flags().StringVar(&className, "class", "", "Broker class like 'MTChannelBasedBroker' or 'Kafka' (if available)")
 	return cmd
 }
--- a/pkg/kn/commands/broker/create.go
+++ b/pkg/kn/commands/broker/create.go
@@ -36,6 +36,8 @@ var createExample = `
 // NewBrokerCreateCommand represents command to create new broker instance
 func NewBrokerCreateCommand(p *commands.KnParams) *cobra.Command {
 
+	var className string
+
 	cmd := &cobra.Command{
 		Use:     "create NAME",
 		Short:   "Create a broker",
@@ -58,7 +60,8 @@ func NewBrokerCreateCommand(p *commands.KnParams) *cobra.Command {
 
 			brokerBuilder := clientv1beta1.
 				NewBrokerBuilder(name).
-				Namespace(namespace)
+				Namespace(namespace).
+				Class(className)
 
 			err = eventingClient.CreateBroker(cmd.Context(), brokerBuilder.Build())
 			if err != nil {
@@ -71,5 +74,6 @@ func NewBrokerCreateCommand(p *commands.KnParams) *cobra.Command {
 		},
 	}
 	commands.AddNamespaceFlags(cmd.Flags(), false)
+	cmd.Flags().StringVarP(&className, "class", "c", "", "Broker class name.")
 	return cmd
 }
--- a/pkg/kn/commands/broker/create.go
+++ b/pkg/kn/commands/broker/create.go
@@ -30,8 +30,9 @@ var createExample = `
   # Create a broker 'mybroker' in the current namespace
   kn broker create mybroker
 
-  # Create a broker 'mybroker' in the 'myproject' namespace
-  kn broker create mybroker --namespace myproject`
+  # Create a broker 'mybroker' in the 'myproject' namespace and with a broker class of 'Kafka'
+  kn broker create mybroker --namespace myproject --broker Kafka
+`
 
 // NewBrokerCreateCommand represents command to create new broker instance
 func NewBrokerCreateCommand(p *commands.KnParams) *cobra.Command {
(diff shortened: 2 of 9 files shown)


itsmurugappan (approved):
/lgtm

pkg/kn/commands/broker/create.go
```diff
@@ -71,5 +74,6 @@ func NewBrokerCreateCommand(p *commands.KnParams) *cobra.Command {
 		},
 	}
 	commands.AddNamespaceFlags(cmd.Flags(), false)
+	cmd.Flags().StringVarP(&className, "class", "c", "", "Broker class name.")
```
rhuss, line 77: I would add some example to the description (and ommit the 'name').

Also we should add the `--class` to the overall examples sections of this command.

Not sure about the `-c` shortcut, as I don't know if this is needed (i don't expect people to call the option often). So I propose to commit this.

vyasgun: Should I mention the default broker class name (MTChannelBasedBroker) in case none is specified using the option?

rhuss:
```suggestion
	cmd.Flags().StringVarP(&className, "class", "c", "", "Broker class like \'MTChannelBasedBroker\' or \'KafkaBroker\' (if available)")
```

something like this.

pkg/kn/commands/broker/create.go
```diff
@@ -71,5 +74,6 @@ func NewBrokerCreateCommand(p *commands.KnParams) *cobra.Command {
 		},
 	}
 	commands.AddNamespaceFlags(cmd.Flags(), false)
+	cmd.Flags().StringVar(&className, "class", "", "Broker class like 'MTChannelBasedBroker' or 'Kafka' (if available)")
```
rhuss, line 77:
```suggestion
	cmd.Flags().StringVar(&className, "class", "", "Broker class like 'MTChannelBasedBroker' or 'KafkaBroker' (if available)")
```

sorry, my bad, I looked up the wrong value.

pkg/kn/commands/broker/create.go
```diff
@@ -36,6 +36,8 @@ var createExample = `
 // NewBrokerCreateCommand represents command to create new broker instance
 func NewBrokerCreateCommand(p *commands.KnParams) *cobra.Command {
 
+	var className string
```
rhuss, line 39: 3 lines above there is an example added. could we add one with a `--class KafkaBroker` example, something like:

```
# Create a broker 'mybroker' in the 'myproject' namespace and with a broker class of 'KafkaBroker'
kn broker create mybroker --namespace myproject --broker KafkaBroker
````

(extending the last example, not adding a new one)

File: lib/test/broker.go [hunk 4]
```diff
@@ -32,6 +32,13 @@ func BrokerCreate(r *KnRunResultCollector, name string) {
 	assert.Check(r.T(), util.ContainsAllIgnoreCase(out.Stdout, "Broker", name, "created", "namespace", r.KnTest().Kn().Namespace()))
 }
 
+// BrokerCreateWithClass creates a broker with the given name and class.
```
maximilien, line 35: Why is these not named `CreateBroker…`?

test/e2e/broker_test.go
```diff
@@ -59,6 +59,12 @@ func TestBroker(t *testing.T) {
 	test.BrokerDelete(r, "foo4", true)
 	verifyBrokerNotfound(r, "foo3")
 	verifyBrokerNotfound(r, "foo4")
+
+	t.Log("create broker with class")
+	test.BrokerCreateWithClass(r, "foo5", "foo-class")
```
maximilien, line 64: Are there naming conventions for the class that we should be verifying… e.g., cannot start with numbers or symbols?

maximilien (changes requested):
Left some comments. Thanks for contributions

lib/test/broker.go
vyasgun: The convention in this file is BrokerCreate, BrokerDelete etc so I am just following that.

test/e2e/broker_test.go
vyasgun: There seems to be no server side validation for this so I am not sure.

dsimansk:
Please re-run `./hack/build.sh -c`.

lib/test/broker.go
rhuss: The convention was already established in the file, not change by this PR. I propose to keep it as it and if we want to change it do it in a separate move.

pkg/kn/commands/broker/create.go
```diff
@@ -30,12 +30,15 @@ var createExample = `
   # Create a broker 'mybroker' in the current namespace
   kn broker create mybroker
 
-  # Create a broker 'mybroker' in the 'myproject' namespace
-  kn broker create mybroker --namespace myproject`
+  # Create a broker 'mybroker' in the 'myproject' namespace and with a broker class of 'Kafka'
+  kn broker create mybroker --namespace myproject --broker Kafka
```
rhuss, line 34:
```suggestion
  # Create a broker 'mybroker' in the 'myproject' namespace and with a broker class of 'KafkaBroker'
  kn broker create mybroker --namespace myproject --broker KafkaBroker
```

Sorry my bad again, I think it should be KafkaBroker (so with a trailing Broker suffix.

vyasgun: I am following the [KNative Kafka Broker documentation](https://knative.dev/docs/eventing/broker/kafka-broker/) and it is just Kafka as per it:

```
kind: Broker
metadata:
  annotations:
    # case-sensitive
    eventing.knative.dev/broker.class: Kafka
  name: default
  namespace: default
spec:
  # Configuration specific to this broker.
  config:
    apiVersion: v1
    kind: ConfigMap
    name: kafka-broker-config
    namespace: knative-eventing
  # Optional dead letter sink, you can specify either:
  #  - deadLetterSink.ref, which is a reference to a Callable
  #  - deadLetterSink.uri, which is an absolute URI to a Callable (It can potentially be out of the Kubernetes cluster)
  delivery:
    deadLetterSink:
      ref:
        kind: Service
        name: dlq-service
```

rhuss: Ok, I was referring to https://knative.dev/docs/eventing/broker/example-mtbroker/ where it used as `KafkaBroker`. @matzew which is correct now ?

vyasgun: I think it's just `Kafka` because I tried creating 2 brokers with the different classes and the Kafka controller only picked up when the annotation was set to `Kafka` and not when it was set to `KafkaBroker` but we can confirm this once before we decide to push.

rhuss: Let me open an issue over there at the docs repo, to find out and get confirmation from the experts ;-)

rhuss: I just asked over there to clarify this https://github.com/knative/docs/pull/4090 (would really prefer "Kafka" but then also "MTChannel" instead of "MTChannelBroker" ;-)

dsimansk: @rhuss I think the current example message is correct. Any objections to proceed and merge this PR?

rhuss: +1 to merge it.

PR #1406, opened 2021-07-29
Server errors should not be tainted by client error messages

### LINKED ISSUE #1342: kn is presenting API server errors incorrectly
image (5)
image (4)

export WORKLOAD_NS=serverless-demo
kubectl create ns ${WORKLOAD_NS}

then When i run this - kn service create helloworld-go -n ${WORKLOAD_NS} \
  --image gcr.io/knative-samples/helloworld-go --env TARGET="from Serverless"

Error: Internal error occurred: failed calling webhook "webhook.serving.knative.dev": Post https://webhook.knative-serving.svc:443/defaulting?timeout=10s: context deadline exceeded

vyasgun:
## Description
In some cases, if an error  is from server but contains a substring that matches a client side error condition, the error returned by the server is being accompanied by the client side error message. This should not be the case.

## Changes

* Moved the server error parsing case up so it gets processed before we breakdown the error string for client level message.
* Added test cases.

## Reference

Fixes #1342

### DIFF
--- a/pkg/errors/factory.go
+++ b/pkg/errors/factory.go
@@ -40,6 +40,27 @@ func isEmptyConfigError(err error) bool {
 	return strings.Contains(err.Error(), "no configuration has been provided")
 }
 
+func isStatusError(err error) bool {
+	var errAPIStatus api_errors.APIStatus
+	return errors.As(err, &errAPIStatus)
+}
+
+func newStatusError(err error) error {
+	var errAPIStatus api_errors.APIStatus
+	errors.As(err, &errAPIStatus)
+
+	if errAPIStatus.Status().Details == nil {
+		return err
+	}
+	var knerr *KNError
+	if isCRDError(errAPIStatus) {
+		knerr = newInvalidCRD(errAPIStatus.Status().Details.Group)
+		knerr.Status = errAPIStatus
+		return knerr
+	}
+	return err
+}
+
 //Retrieves a custom error struct based on the original error APIStatus struct
 //Returns the original error struct in case it can't identify the kind of APIStatus error
 func GetError(err error) error {
@@ -47,26 +68,16 @@ func GetError(err error) error {
 		return nil
 	}
 
-	var errAPIStatus api_errors.APIStatus
 	switch {
 	case isEmptyConfigError(err):
 		return newNoKubeConfig(err.Error())
-	case errors.As(err, &errAPIStatus):
-		if errAPIStatus.Status().Details == nil {
-			return err
-		}
-		var knerr *KNError
-		if isCRDError(errAPIStatus) {
-			knerr = newInvalidCRD(errAPIStatus.Status().Details.Group)
-			knerr.Status = errAPIStatus
-			return knerr
-		}
+	case isStatusError(err):
+		return newStatusError(err)
 	case isNoRouteToHostError(err):
 		return newNoRouteToHost(err.Error())
 	default:
 		return err
 	}
-	return err
 }
 
 // IsForbiddenError returns true if given error can be converted to API status and of type forbidden access else false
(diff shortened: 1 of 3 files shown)


pkg/errors/factory.go
```diff
@@ -47,16 +47,11 @@ func GetError(err error) error {
...
 	case isEmptyConfigError(err):
 		return newNoKubeConfig(err.Error())
-	case isNoRouteToHostError(err):
-		return newNoRouteToHost(err.Error())
-	default:
-		var errAPIStatus api_errors.APIStatus
-		if !errors.As(err, &errAPIStatus) {
-			return err
-		}
+	case errors.As(err, &errAPIStatus):
```
dsimansk, line 54: I'd suggest to add 2 new functions `isStatusError()` and then `newStatusError()` to wrap the handling of that case into it. I think it'd make the cases more self-explanatory.

pkg/errors/factory_test.go
```diff
@@ -154,14 +165,49 @@ func TestIsForbiddenError(t *testing.T) {
...
 	assert.NilError(t, GetError(nil), nil)
 }
+
+func TestIsInternalError(t *testing.T) {
+	cases := []struct {
+		Name     string
+		Error    error
+		Internal bool
+	}{
+		{
```
dsimansk, line 185: Let's add `api_errors.NewTimeoutError` just to be sure.

dsimansk: It can be a separate test function or maybe a mix of different `StatusError` kinds.

dsimansk (changes requested):
Please add a changelog entry. Sorry I haven't noticed sooner. :)

vyasgun:
@dsimansk sorry I keep forgetting about adding the entry 😆

dsimansk:
> @dsimansk sorry I keep forgetting about adding the entry laughing

Well, it happens to me all the time. :)

dsimansk:
Thanks looks good to me. I'll leave open for a while for other to take a look.

/approve

pkg/errors/factory.go
```diff
@@ -50,22 +71,11 @@ func GetError(err error) error {
 	switch {
 	case isEmptyConfigError(err):
 		return newNoKubeConfig(err.Error())
+	case isStatusError(err):
```
maximilien, line 74: Should this be moved up and considered first?

pkg/errors/factory_test.go
```diff
@@ -154,14 +167,82 @@ func TestIsForbiddenError(t *testing.T) {
...
+			Name:     "internal error with i/o timeout",
+			Error:    api_errors.NewInternalError(fmt.Errorf("failed calling webhook \"webhook.serving.knative.dev\": Post https://webhook.knative-serving.svc:443/defaulting?timeout=10s: i/o timeout")),
+			Internal: true,
+		},
+		{
+			Name:     "not internal error",
+			Error:    mockErrType{},
+			Internal: false,
+		},
```
maximilien, line 206: What about unknown errors? Or the previous covers it?

maximilien (commented):
Left a few comments and questions. Thanks for the contribution.

pkg/errors/factory.go
dsimansk: +1
I'd say that `no configuration has been provided` message is quiet distinctive from k8s api erros, but it's good idea to have it at first place to check.

pkg/errors/factory_test.go
vyasgun: The previous function covers the non k8s api errors. This one covers a mock one which implements the k8s api interface.

dsimansk: @maximilien any more comments to the above please?

dsimansk:
As I don't see any new comments, I'd say it's done. :)

Thanks @vyasgun!
/approve
/lgtm

PR #1407, opened 2021-07-29
Use revision-name flag in kn service apply

### LINKED ISSUE #1184: `kn apply` ignores revision name

vyasgun:
## Description
`kn service apply --revision-name ` should use the revision name argument to generate revision name.

## Changes

* Removed the line setting the revision name to empty in apply flag.
* Tested that applying the same revision name to the service would not result in error as follows:
```
No changes to apply to service 'xyzsvc'.
Service 'xyzsvc' with latest revision 'xyzsvc-abcde' (unchanged) is available at URL:
http://xyzsvc.knativetutorial.192.168.49.2.nip.io

```

## Reference

Fixes #1184

### DIFF
--- a/pkg/kn/commands/service/apply.go
+++ b/pkg/kn/commands/service/apply.go
@@ -66,7 +66,6 @@ func NewServiceApplyCommand(p *commands.KnParams) *cobra.Command {
 			}
 
 			var service *servingv1.Service
-			applyFlags.RevisionName = ""
 			if applyFlags.Filename == "" {
 				service, err = constructService(cmd, applyFlags, name, namespace)
 			} else {


dsimansk:
@rhuss we've discussed with Gunjan in chat, that actual issue is waiting on `user-input`. AFAIK, there was no additional movement on it. 
Just to confirm. Can we close the PR for now until there's more info it's needed?

rhuss:
The point is, that `kn service apply` das not support client-side revision naming. If you check the code, the name is always erased, even when provided on the command line.

My take is, that a client-side provided revision name does not fit to the _declararive model_ of `kn service apply`, which should be idempotent (e.g. if a apply command is given twice with the same arguments, the second command is a no-op)

Using client side revision naming is an _imperative way_ where the client dictates the name to use and you can't have idempotency because you have to always pick a new revision name.

So I'm in favor to close this PR.

maximilien (changes requested):
Would it make sense to have a test change for this?

dsimansk:
Following on our discussion with @rhuss and @vyasgun, I'll close this PR. Any further comments can be added to original issue #1184.

/close

PR #1412, opened 2021-08-03
Look up plugins on $PATH by default

### LINKED ISSUE #1399: [Proposal] Have `kn` look up plugins on `$PATH` by default
### Feature request

The `kn` client by default will look in `$PATH` for plugins. This would mirror the performance of `kubectl` and provide an easier way for users to install plugins using tools like `brew`. 

### Use case

This arose out of discussions about the [quickstart plugin](https://github.com/knative-sandbox/kn-plugin-quickstart) we're building. The goal of that plugin was to provide a streamlined experience for new users to get up and running on Knative by running something like:

```bash
brew install knative/client/kn
brew install knative-sandbox/kn-plugins/quickstart
kn quickstart kind
```

However, the last command does not work, as `kn` by default does not look up plugins from `$PATH` (which is where `brew` installs them).

See also disucssion in [Slack](https://knative.slack.com/archives/CE4MVFVAQ/p1626964104136700) and in the [homebrew-kn-plugins repo](https://github.com/knative-sandbox/homebrew-kn-plugins/issues/23). 

/kind proposal
cc @omerbensaadon @csantanapr

psschwei:
## Description

Look for knative plugins on the user's $PATH by default

## Changes

* Set default value of `lookupPluginsInPath` to `true`

This results in the following behavior (copying from [Roland's comment](https://github.com/knative/client/issues/1399#issuecomment-888182293)):

    * Lookup plugins from both, the path and `~/.config/kn/plugins`.
    * Same named plugin in `~/.config/kn/plugins` take precedence over plugins found on the PATH.

This would also allow for system-wide plugins as installed by brew in `/usr/local/bin`, but allow individual users to override plugins in their `~/.config/kn/plugins` directory. Also, for people that currently collect their plugins in `~/.config/kn/plugins` nothing would change. It's just as if you would add this directory to the front of your $PATH.

## Reference

Fixes #1399 

/lint

### DIFF
--- a/pkg/kn/config/config.go
+++ b/pkg/kn/config/config.go
@@ -178,7 +178,7 @@ func initDefaults() *defaultConfig {
 	return &defaultConfig{
 		configFile:          defaultConfigLocation("config.yaml"),
 		pluginsDir:          defaultConfigLocation("plugins"),
-		lookupPluginsInPath: false,
+		lookupPluginsInPath: true,
 	}
 }
 


psschwei:
/hold
Removing the `lookupPluginsInPath` variable was causing trouble with some of the tests, so as a temporary step this PR just sets the value to `true`.

@rhuss @maximilien  any preference as to whether we do this all in one shot (remove the config variable now) or if it's done piecemeal (change default value in this PR, remove variable in follow up PR)? (my thought was the latter to be sure the change would land in the next release)

psschwei:
> test/e2e: TestServiceImport
> kubetest2: Test

:thinking: Don't think changing the default value of `lookupPluginsInPath` would cause these to fail...

maximilien (changes requested):
You may need to change some tests @psschwei

psschwei:
> You may need to change some tests @psschwei

Interestingly, when I was testing locally, the `TestServiceImport` e2e test fails when run against the main branch (i.e. with no changes) using either the latest `kn` release and the binary built directly from source, with the same error message:

```
Error: cannot lookup plugin kn-service-import-/tmp/kn_file315602552/foo_with_revisions in directory /home/paulschw/.config/kn/plugins (lookup in path: true): error for looking up kn-service-import-/tmp/kn_file315602552/foo_with_revisions in path: exec: "kn-service-import-/tmp/kn_file315602552/foo_with_revisions": stat kn-service-import-/tmp/kn_file315602552/foo_with_revisions: no such file or directory
```

Let me look at the test itself...

(I wasn't able to figure out what exactly the `kubetest2` test was, so  haven't been able to check that one locally)

psschwei:
Ok, looks like it's not a test issue, but rather something with how `kn` is processing the command...

I ran the following in the terminal:

```
$ kn service import ./hello-world-ksvc.yaml 
Error: cannot lookup plugin kn-service-import-./hello_world_ksvc.yaml in directory
 /home/paulschw/.config/kn/plugins (lookup in path: true): error for looking up
 kn-service-import-./hello_world_ksvc.yaml in path: exec: 
"kn-service-import-./hello_world_ksvc.yaml": stat kn-service-import-./hello_world_ksvc.yaml: 
no such file or directory
Run 'kn --help' for usage

```

... and got the same error I saw in the test. It appears it's translating `kn service import file` to `kn-service-import-file`, which seems to be what's causing the error. And if I don't use `./` before the file name...

```
$ kn service import hello-world-ksvc.yaml 
Error: provided import file doesn't contain service name, please note that 
only kn's custom export format is supported
Run 'kn --help' for usage
```

So I'm somewhat confident that the test failure I'm seeing isn't related to my PR. Before I go too far down the rabbit hole looking at `kn service import`, @rhuss or @maximilien does my explanation here sound reasonable? (and if it is `kn service import`, should that fix be a separate PR?)

dsimansk:
Indeed we fail to look up in path when there's slash in argument,

https://github.com/knative/client/blob/c18a55f92610fbc5a24bd7b800ee16c1b9e15734/pkg/kn/plugin/manager.go#L468-L477

A simple skip of such argument seems to be enough, before it's added to suspected name. 

```diff --git a/pkg/kn/plugin/manager.go b/pkg/kn/plugin/manager.go
index f9b9e284..ca7a8c89 100644
--- a/pkg/kn/plugin/manager.go
+++ b/pkg/kn/plugin/manager.go
@@ -403,6 +403,9 @@ func findMostSpecificPluginInPath(dir string, parts []string, lookupInPath bool)
                for _, p := range parts[0:i] {
                        // Subcommands with "-" are translated to "_"
                        // (e.g. a command  "kn log-all" is translated to a plugin "kn-log_all")
+                       if strings.Contains(p, "/") || strings.Contains(p, "\\") {
+                               continue
+                       }
                        nameParts = append(nameParts, convertDashToUnderscore(p))
                        commandParts = append(commandParts, p)
                }
```

dsimansk:
@psschwei that's a nice bug catch! :)

psschwei:
Thanks @dsimansk !

Can confirm that your fix does solve the test problem:

```
$ test/local-e2e-tests.sh -run TestServiceImport -v
🧪  Testing
=== RUN   TestServiceImport
=== PAUSE TestServiceImport
=== CONT  TestServiceImport
    service_import_test.go:46: import service foo with revision
    service_import_test.go:56: import existing service foo error
    service_import_test.go:59: import service from missing file error
--- PASS: TestServiceImport (196.99s)
PASS
ok      knative.dev/client/test/e2e     197.003s
```
(run with `kn` built from source using fix above)

I opened #1415 as a fix

psschwei:
/hold cancel
(#1415 merged)

rhuss (approved):
Looks good, but maybe let's already add a deprecation for this configuration variable ? (so that we can remove it earlier in 0.29 ?)

Let's get this merged and I'll send a followup PR.

/approve
/lgtm

psschwei:
> Looks good, but maybe let's already add a deprecation for this configuration variable ? (so that we can remove it earlier in 0.29 ?)

Sorry @rhuss , I totally spaced on the docs :frowning_face: ... thanks for catching that!

PR #1414, opened 2021-08-04
Add lint check to build.sh script

dsimansk:
## Description

@rhuss per one of our discussion on PR review.

## Changes

* Add `golangci-lint run` to default build flow
* Remove our `run_go_tool` in favour of `knative.dev/hack`

/cc @rhuss

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+source $(dirname $0)/../vendor/knative.dev/hack/library.sh
+
 set -o pipefail
 
 source_dirs="cmd pkg test lib tools"
@@ -98,6 +100,9 @@ codegen() {
   # Format source code and cleanup imports
   source_format
 
+  # Lint source code
+  source_lint
+
   # Check for license headers
   check_license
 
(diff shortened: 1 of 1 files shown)


hack/build.sh
```diff
@@ -98,6 +100,9 @@ codegen() {
   # Format source code and cleanup imports
   source_format
 
+  # Lint source code
+  (( ! IS_PROW )) && source_lint
```
dsimansk, line 104: I've added skip on CI, since that check is part of GH actions. Any objections?

hack/build.sh
```diff
@@ -110,30 +115,19 @@ go_fmt() {
   find $(echo $source_dirs) -name "*.go" -print0 | xargs -0 gofmt -s -w
 }
 
-# Run a go tool, get it first if necessary.
-run_go_tool() {
```
rhuss, line 114: This function has been removed but is still referenced below ?

dsimansk: I've opted for the common one from hack repo, https://github.com/knative/hack/blob/main/library.sh#L489-L514.

rhuss (commented):
/approve
/lgtm

PR #1415, opened 2021-08-04
Fixing bug when looking up plugins on path with slash in argument

psschwei:
## Description

Per discussion in [PR 1412](https://github.com/knative/client/pull/1412#issuecomment-892857273), found a bug where we fail to look up in path when there's slash in arguments. (This bug was also causing the service import test to fail). 

## Changes

* fixing bug when looking up plugins on path with slash in argument

(changelog entry to follow)

/lint

### DIFF
--- a/pkg/kn/plugin/manager.go
+++ b/pkg/kn/plugin/manager.go
@@ -401,6 +401,10 @@ func findMostSpecificPluginInPath(dir string, parts []string, lookupInPath bool)
 		var nameParts []string
 		var commandParts []string
 		for _, p := range parts[0:i] {
+			// skip arguments with slashes
+			if strings.Contains(p, "/") || strings.Contains(p, "\\") {
+				continue
+			}
 			// Subcommands with "-" are translated to "_"
 			// (e.g. a command  "kn log-all" is translated to a plugin "kn-log_all")
 			nameParts = append(nameParts, convertDashToUnderscore(p))
--- a/pkg/kn/plugin/manager.go
+++ b/pkg/kn/plugin/manager.go
@@ -401,9 +401,10 @@ func findMostSpecificPluginInPath(dir string, parts []string, lookupInPath bool)
 		var nameParts []string
 		var commandParts []string
 		for _, p := range parts[0:i] {
-			// skip arguments with slashes
-			if strings.Contains(p, "/") || strings.Contains(p, "\\") {
-				continue
+			// for arguments that contain the path separator,
+			// stop the loop once the separator appears
+			if strings.Contains(p, string(os.PathSeparator)) {
+				break
 			}
 			// Subcommands with "-" are translated to "_"
 			// (e.g. a command  "kn log-all" is translated to a plugin "kn-log_all")
--- a/pkg/kn/plugin/manager_test.go
+++ b/pkg/kn/plugin/manager_test.go
@@ -301,6 +301,33 @@ func TestPluginList(t *testing.T) {
 	}
 }
 
+func TestNoSlashInPlugin(t *testing.T) {
+	ctx := setup(t)
+	defer cleanup(t, ctx)
+
+	// Prepare PATH
+	tmpPathDir, cleanupFunc := preparePathDirectory(t)
+	defer cleanupFunc()
+
+	fmt.Println(tmpPathDir)
+
+	createTestPluginInDirectory(t, "kn-path-test", tmpPathDir)
+	pluginCommands := []string{"path", "test", "/withslash"}
+
+	// args with slash are not returned
+	ctx.pluginManager.lookupInPath = true
+	plugin, err := ctx.pluginManager.FindPlugin(pluginCommands)
+	assert.NilError(t, err)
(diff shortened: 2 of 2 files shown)


dsimansk:
@psschwei thansk!
Could you please try to add unit test that would verify plugin with slash in arguments won't have that slash part in name maybe? If it's too cumbersome to reach that meaningful coverage, I'd happy to go with it.


@rhuss as a lesson here, maybe it isn't a good idea to have file to path in `args`. I recall some review feedback when working on `kn service import`. :)

psschwei:
As I'm writing a test, I have a question about what the assumptions are for plugin arguments.

Take this example:

```
kn plugin test /with/slash argument
```

According to the logic in this PR, that would "find" a plugin called `kn-plugin-test-argument` (i.e. it would simply drop the argument containing a slash). Is that the behavior we'd expect? Or do we expect that `kn` would drop ALL terms after and including one with a slash (i.e. our plugin would be `kn-plugin-test` and `/with/slash` and `argument` are parameters for the plugin)?

If it's the latter, then I think we'd need to update the logic some to parse the [initial `parts` slice](https://github.com/knative/client/blob/180ecd5057f5d3da0dd8f812b6dc68dcd89849e3/pkg/kn/plugin/manager.go#L398) before processing.

dsimansk:
>  Or do we expect that kn would drop ALL terms after and including one with a slash (i.e. our plugin would be kn-plugin-test and /with/slash and argument are parameters for the plugin)?

Yes, I believe this is already true with current implementation.

I hope we may safely assume that there won't plugin sub-command with slashes in name.

Handling of plugin name and arguments is done by the [outer for-loop ](https://github.com/knative/client/blob/main/pkg/kn/plugin/manager.go#L398) that's moving through params array and reducing it.
Until the plugin can recognized/found, the plugin object is returned. In addition, original args are preserved for the [execution time](https://github.com/knative/client/blob/main/cmd/kn/main.go#L100). Therefore dropping arg with slashes in look-up function shouldn't result in missing arg in execution.

I believe it should go like this.
```
kn plugin test arg0 /arg1/slash arg2
```

```
/arg1/slash => always dropped due to slashes breaking on path lookup
kn plugin test arg0 arg2 => plugin not found
kn plugin test arg0  => plugin not found
kn plugin test  => found 

//the rest of original array [arg0, /arg1/slash/, arg2] are arguments to the plugin cmd "kn plugin test"
```

psschwei:
Got it -- I think I was confusing the plugin lookup and actually executing the plugin :)

dsimansk:
Thanks for the fix & the test coverage @psschwei!

@rhuss @maximilien can you please take a look?

/lgtm

pkg/kn/plugin/manager.go
```diff
@@ -401,6 +401,10 @@ func findMostSpecificPluginInPath(dir string, parts []string, lookupInPath bool)
 		var nameParts []string
 		var commandParts []string
 		for _, p := range parts[0:i] {
+			// skip arguments with slashes
+			if strings.Contains(p, "/") || strings.Contains(p, "\\") {
```
rhuss, line 405: I don't think that you just skip it bit completely stop the loop when the first part appears that can not be used as a plugin part, otherwise your could call plugins accidentally, e.g.

```
kn myplugin not/possible bla
```

would also call an existing plugin `kn-myplugin-bla`, but this is not following the plugin contract.

Also, you should ensure that you only filter characters that are not possible on the current's OS filesystem. E.g. on UNIX you can have a backslash as part of a file name, so a plugin like `kn-myplugin-ext\ra` is possible on UNIX (but not on Windows). So you should filter on the concrete file separator.

dsimansk: > Also, you should ensure that you only filter characters that are not possible on the current's OS filesystem. E.g. on UNIX you can have a backslash as part of a file name, so a plugin like `kn-myplugin-ext\ra` is possible on UNIX (but not on Windows). So you should filter on the concrete file separator.

+1 that's a great point.

maximilien (commented):
/ok-to-test

pkg/kn/plugin/manager_test.go
```diff
@@ -301,6 +301,33 @@ func TestPluginList(t *testing.T) {
...
+	defer cleanup(t, ctx)
+
+	// Prepare PATH
+	tmpPathDir, cleanupFunc := preparePathDirectory(t)
+	defer cleanupFunc()
+
+	fmt.Println(tmpPathDir)
+
+	createTestPluginInDirectory(t, "kn-path-test", tmpPathDir)
+	pluginCommands := []string{"path", "test", "/withslash"}
```
rhuss, line 315: The test should also use a `os.PathSeparator` otherwise it will fail on Windows.

Also I would suggest to use a loop and multiple tests, e.g.

```
path test /withslash
path test /withslash extra arg
path test ext\ra  (using something like `os.PathSeparator == '/' ? "\\" : "/"`)
```

rhuss (approved):
/approve
/lgtm

thanks !

PR #1421, opened 2021-08-06
Fix unit tests to use NAMESPACE value in OpenShift CI env

dsimansk:
## Description

Executing unit tests inside the K8s pod may result in false positive failures due to in-cluster kubeconfig taking precedence over mocked one.

## Changes

* Fix unit tests to use NAMESPACE value in k8s cluster env 

/cc @rhuss

### DIFF
--- a/pkg/kn/commands/namespaced_test.go
+++ b/pkg/kn/commands/namespaced_test.go
@@ -143,34 +143,50 @@ func TestGetNamespaceFallback(t *testing.T) {
 	defer os.RemoveAll(tempDir)
 	assert.NilError(t, err)
 
-	tempFile := filepath.Join(tempDir, "mock")
-	err = ioutil.WriteFile(tempFile, []byte(BASIC_KUBECONFIG), test.FileModeReadWrite)
-	assert.NilError(t, err)
+	t.Run("MockConfig", func(t *testing.T) {
+		tempFile := filepath.Join(tempDir, "mock")
+		err = ioutil.WriteFile(tempFile, []byte(BASIC_KUBECONFIG), test.FileModeReadWrite)
+		assert.NilError(t, err)
 
-	testCmd := testCommandGenerator(true)
-	kp := &KnParams{KubeCfgPath: tempFile}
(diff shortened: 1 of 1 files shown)


pkg/kn/commands/namespaced_test.go
```diff
@@ -143,71 +143,111 @@ func TestGetNamespaceFallback(t *testing.T) {
...
+			// In-cluster config overrides the mocked one in OpenShift CI
+			assert.Equal(t, actual, os.Getenv("NAMESPACE"))
+		} else {
+			assert.Equal(t, actual, "default")
+		}
+	})
+}
+
+func isOpenShiftCI() bool {
+	return os.Getenv("OPENSHIFT_CI") == "true"
```
rhuss, line 252: I wonder whether if the issue wouldn't happen in any in-cluster run of the unit tests (not necessarily only on OpenShift), so that we check for the same file that the client lib looks up to decided whether it runs in-cluster ?

That would make the tests more robust in other environments, too, and we could avoid a reference to OpenShift here in Knative.

dsimansk: Reworked to client-go `Possible()` style.

pkg/kn/commands/namespaced_test.go
```diff
@@ -143,71 +143,116 @@ func TestGetNamespaceFallback(t *testing.T) {
...
+	})
+}
+
+// Inspired by client-go function
+// https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/client_config.go#L600-L606
+func isInCluster() bool {
+	fi, err := os.Stat("/var/run/secrets/kubernetes.io/serviceaccount/token")
+	return os.Getenv("KUBERNETES_SERVICE_HOST") != "" &&
+		os.Getenv("KUBERNETES_SERVICE_PORT") != "" &&
+		err == nil && !fi.IsDir()
```
rhuss, line 257: when can `fi` be a directory ? (academic question ;-)

Looks good to me.

/approve
/lgtm

PR #1419, opened 2021-08-05
Adding --tls option to domain create command

vyasgun:
## Description
Support for TLS optionn in domain create using `--tls` flag

## Changes

* Added the new flag
* Added unit tests for the new flag

## Reference

Fixes #1408

PR #1422, opened 2021-08-10
Deprecate `lookup-path` option and updated relevant documentation

rhuss:
Lookup in path is now the default for all plugins.
This option will be removed eventually in a future version,
when path lookup is enabled uncoditinally and can't be turned of.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -17,20 +17,20 @@
 |===
 | | Description | PR
 
+| ✨
+| Lookup plugins in `$PATH` by default
+| https://github.com/knative/client/pull/1412[#1412]
+
 | 🐛
 | Show server error messages without any taints
 | https://github.com/knative/client/pull/1406[#1406]
 
 | 🎁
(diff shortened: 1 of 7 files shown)


CHANGELOG.adoc
```diff
@@ -17,20 +17,20 @@
 |===
 | | Description | PR
```
dsimansk, line 18: Should we have another changelog entry for deprecation `path-lookup`?

pkg/kn/config/config_test.go
```diff
@@ -29,8 +29,7 @@ func TestBootstrapConfig(t *testing.T) {
 	configYaml := `
 plugins:
   directory: /tmp
-  path-lookup: true
-
+
```
dsimansk, line 32: There's a leftover whitespace. TBH I hate those errors. :)

dsimansk:
@rhuss and you'll probably have conflict in changelog once `--tls` PR https://github.com/knative/client/pull/1419 is merged.

PR #1423, opened 2021-08-10
Prepare changelog and version for 0.25 release

dsimansk:
## Changes

* Add missing entries to changelog
* Bump `version` strings to 0.25

/cc @rhuss 

FYI @knative/knative-release-leads, that's part of client release checklist, the remaining items are:
* Version bump in `go.mod`
* Update the version numbers of Serving and Eventing in test/presubmit-integration-tests-latest-release.sh so that the integration test is already running against the just released serving and eventing versions.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -12,7 +12,7 @@
 | https://github.com/knative/client/pull/[#]
 ////
 
-## Unreleased
+## v0.25.0 (2021-08-10)
 [cols="1,10,3", options="header", width="100%"]
 |===
 | | Description | PR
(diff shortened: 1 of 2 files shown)


vyasgun:
@dsimansk should this file also be changed? https://github.com/knative/client/blob/main/test/presubmit-integration-tests-latest-release.sh

dsimansk:
> @dsimansk should this file also be changed? https://github.com/knative/client/blob/main/test/presubmit-integration-tests-latest-release.sh

Eventually, but right now it would fail `pull-knative-client-integration-tests-latest-release` since Serving repos it not yet cut, only Eventing.

vyasgun:
> > @dsimansk should this file also be changed? https://github.com/knative/client/blob/main/test/presubmit-integration-tests-latest-release.sh
> 
> Eventually, but right now it would fail `pull-knative-client-integration-tests-latest-release` since Serving repos it not yet cut, only Eventing.

But the version number for eventing and serving has changed in the file version.go. Is that different?

dsimansk:
> But the version number for eventing and serving has changed in the file version.go. Is that different?

Yes, `kn version` only prints the version string itself without any additional checks etc. It can be changed to any value, any time.

However, the `presubmit-integration-tests-latest-release.sh` controls the version used in the CI check `*-latest-release`. That version can be changed once the release is cut, otherwise the setup phase would fail on missing `.yamls` etc.

vyasgun:
@dsimansk  okay, got it. Thanks! /lgtm

PR #1427, opened 2021-08-12
Fix release script to published darwin/arm64 bits

dsimansk:
## Changes

* Fix release script to published darwin/arm64 bits 

We can probably upload the artifact manually to the release, but that'd require updated checksum file as well. Or we could probably release `0.25.1.` once we get a few more bugs to backport?

/cc @rhuss @maximilien


rhuss (commented):
/lgtm
/approve

PR #1428, opened 2021-08-14
fix kn trigger list command to show correct v1/service sink output

Abirdcfly:
## Description
If the sink of the trigger is a normal service, the  `kn trigger list` command will show it as ksvc, which is incorrect

## Changes

*
*
*

## Reference

Fixes #

### DIFF
--- a/pkg/kn/commands/flags/sink.go
+++ b/pkg/kn/commands/flags/sink.go
@@ -149,7 +149,7 @@ func parseSink(sink string) (string, string, string) {
 // SinkToString prepares a sink for list output
 func SinkToString(sink duckv1.Destination) string {
 	if sink.Ref != nil {
-		if sink.Ref.Kind == "Service" && strings.Contains(sink.Ref.APIVersion, sinkMappings["ksvc"].Group) {
+		if sink.Ref.Kind == "Service" && strings.HasPrefix(sink.Ref.APIVersion, sinkMappings["ksvc"].Group) {
 			return fmt.Sprintf("ksvc:%s", sink.Ref.Name)
 		} else {
 			return fmt.Sprintf("%s:%s", strings.ToLower(sink.Ref.Kind), sink.Ref.Name)
--- a/pkg/kn/commands/flags/sink.go
+++ b/pkg/kn/commands/flags/sink.go
@@ -149,7 +149,7 @@ func parseSink(sink string) (string, string, string) {
 // SinkToString prepares a sink for list output
 func SinkToString(sink duckv1.Destination) string {
 	if sink.Ref != nil {
-		if sink.Ref.Kind == "Service" {
+		if sink.Ref.Kind == "Service" && strings.Contains(sink.Ref.APIVersion, sinkMappings["ksvc"].Group) {
 			return fmt.Sprintf("ksvc:%s", sink.Ref.Name)
 		} else {
 			return fmt.Sprintf("%s:%s", strings.ToLower(sink.Ref.Kind), sink.Ref.Name)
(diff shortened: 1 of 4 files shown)


pkg/kn/commands/flags/sink.go
```diff
@@ -149,7 +149,7 @@ func parseSink(sink string) (string, string, string) {
 // SinkToString prepares a sink for list output
 func SinkToString(sink duckv1.Destination) string {
 	if sink.Ref != nil {
-		if sink.Ref.Kind == "Service" {
+		if sink.Ref.Kind == "Service" && sink.Ref.APIVersion == "serving.knative.dev/v1" {
```
itsmurugappan, line 152: Instead of api version, is it enough to check for the group, like sink.Ref.Group  = "serving.knative.dev" ?

Abirdcfly: I'm afraid not. For example, a trigger with sink is a ksvc will have sink.Ref.Group = ""

pkg/kn/commands/source/duck/multisourcelist_test.go
```diff
@@ -83,8 +83,9 @@ func newSourceUnstructuredObjWithSink(name, apiVersion, kind string) *unstructur
 			"spec": map[string]interface{}{
 				"sink": map[string]interface{}{
 					"ref": map[string]interface{}{
-						"kind": "Service",
-						"name": "foo",
```
Abirdcfly, line 87: Only kind and name are not enough here. An error will appear: "Unable to get the Subscriber's URI"

pkg/kn/commands/source/list_test.go
```diff
@@ -188,8 +188,9 @@ func newSourceUnstructuredObj(name, apiVersion, kind string) *unstructured.Unstr
 			"spec": map[string]interface{}{
 				"sink": map[string]interface{}{
 					"ref": map[string]interface{}{
-						"kind": "Service",
-						"name": "foo",
```
Abirdcfly, line 192: Only kind and name are not enough here. An error will appear: "Unable to get the Subscriber's URI"

pkg/kn/commands/flags/sink.go
itsmurugappan: looks like Group is not stabilized, more details here -> https://github.com/knative/eventing/issues/5086. But if you include the version in the check, it would need to be changed when a new knative serving version comes.

itsmurugappan: looks like Group is not stabilized, more details here -> knative/eventing#5086. But if you include the version in the check, it would need to be changed when a new knative serving version comes.

itsmurugappan:
looks like Group is not stabilized, more details here -> knative/eventing#5086. But if you include the version in the check, it would need to be changed when a new knative serving version comes.

pkg/kn/commands/flags/sink.go
```diff
@@ -149,7 +149,7 @@ func parseSink(sink string) (string, string, string) {
 // SinkToString prepares a sink for list output
 func SinkToString(sink duckv1.Destination) string {
 	if sink.Ref != nil {
-		if sink.Ref.Kind == "Service" {
+		if sink.Ref.Kind == "Service" && strings.Contains(sink.Ref.APIVersion, sinkMappings["ksvc"].Group) {
```
itsmurugappan, line 152:
```suggestion
		if sink.Ref.Kind == "Service" && strings.HasPrefix(sink.Ref.APIVersion, sinkMappings["ksvc"].Group) {
```

itsmurugappan (commented):
everything looks good, one small suggestion.

dsimansk (approved):

PR #1433, opened 2021-08-22
Add option to allow enviornment variables using file

### LINKED ISSUE #278: Add option to allow enviornment variables using file
/kind feature

## Describe the feature:

Currently the `kn` command allows specifying the `environment` variables using `--env KEY=VALUE` pattern, it will be nice to have an option to specify the value via environment file.

e.g. 

env-file
--------
foo=bar
com.example.data=one

And the command can then have option like `--env-file=foo.env`  while doing create/update etc.,

boaz0:
## Description

This PR gives the option to set environment variables from a file by specifying its path in `--env-file`.

## Changes

* Add` GetEnvsFromFile` to `util`
* Add `UpdateEnvsFromFile` to `podspec_helpers`
* Add to `podspec` flags `--env-file` and `EnvFile` to `PodSpec`.
* Run `UpdateEnvsFromFile` if `env-file` was changed.

## Reference

Closes #278


boaz0:
@googlebot I fixed it.

itsmurugappan (commented):
Thanks for the PR. At first glance, I noticed that unit tests were missing. Can you please add some tests while we review the PR. Thanks

dsimansk:
Can you please rerun codegen and commit the changes? And as Muru mentioned that'd nice to have unit tests, otherwise the PR won't pass Codecov checks.

```console
./hack/build.sh -c
```

pkg/kn/flags/podspec_helper.go
```diff
@@ -97,6 +97,35 @@ func UpdateEnvVars(spec *corev1.PodSpec,
...
+            Name:  envKey,
+            Value: envValue,
+        })
+        envKey, envValue, done = envIterator.NextString()
+    }
+
+    updated := updateEnvVarsFromMap(container.Env, allEnvsToUpdate)
+    updated = removeEnvVars(updated, rmEnvFromFile)
+	container.Env = updated
+
```
itsmurugappan, line 125: is there a way you can use this func `UpdateEnvVars` instead of duplicating the same logic here ?

rhuss:
@boaz0 thanks a lot for the PR and your contribution. There are some minor comments and suggestions to the PR. Would you have time to address those? No problem if not, we would then help you here.

maximilien (commented):
Please address the go format changes as it makes reviewing a bit painful (first world problem I know) and please add some tests (unit and integration)

Thanks for contribution

pkg/kn/flags/podspec_helper.go
boaz0: Thanks @itsmurugappan for your feedback and sorry for the late response. I think I solved this but maybe you can find a more elegant way. 
You can see the updates here:  https://github.com/knative/client/blob/fa9ef7f886ff6cefcc6866b151e1f068ce582157/pkg/kn/flags/podspec.go#L177-L192

pkg/kn/flags/podspec.go
```diff
@@ -85,6 +86,9 @@ func (p *PodSpecFlags) AddFlags(flagset *pflag.FlagSet) []string {
 			"To unset, specify the environment variable name followed by a \"-\" (e.g., NAME-).")
 	flagNames = append(flagNames, "env")
 
+	flagset.StringVarP(&p.EnvFile, "env-file", "", "", "Path to a file containing environment variables.")
```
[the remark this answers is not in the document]
boaz0: Sure no problem.

boaz0:
@maximilien @rhuss @itsmurugappan @dsimansk I updated the PR as you requested. Feel free to give a second code-review.
Thanks.

itsmurugappan:
@rhuss @dsimansk why are we iterating the whole array of args here -> https://github.com/knative/client/blob/dedcad7c7ec82377c3de4d777ad4f222f8cc9904/pkg/kn/flags/podspec_helper.go#L70

It could simplify this pr if its not there.

pkg/kn/flags/podspec.go
```diff
@@ -169,6 +173,25 @@ func (p *PodSpecFlags) AddFlags(flagset *pflag.FlagSet) []string {
...
+		for _, val := range envsFromFile {
+			fileEnvsAsArgs = append(fileEnvsAsArgs, "-e="+val)
+		}
+		err = UpdateEnvVars(podSpec, fileEnvsAsArgs, envsFromFileToUpdate, envsFromFileToRemove, util.NewOrderedMap(), []string{})
+		if err != nil {
+			return err
+		}
+	}
+
 	if flags.Changed("env") || flags.Changed("env-value-from") {
```
itsmurugappan, line 195: i was thinking we could add another || here and add env-file logic here.

itsmurugappan (commented):
except for some small things, looks good. Lets wait for @rhuss  or @dsimansk to get back on my question.

pkg/kn/flags/podspec.go
boaz0: Yeah, I think I can work it out. Thanks.

rhuss:
> @rhuss @dsimansk why are we iterating the whole array of args here ->
> 
> https://github.com/knative/client/blob/dedcad7c7ec82377c3de4d777ad4f222f8cc9904/pkg/kn/flags/podspec_helper.go#L70
> 
> It could simplify this pr if its not there.

afair its so to keep the order of the environment variables as specified (a simple map is not good enough). haven't check this PR yet, but we should also take care that mixing `--env` with `--env-from` does *also* respect the order as provided on the commandline. (i.e. `-e A=1 -env-from myenv.txt -e B=2` should keep the there order A, all from the file, B)

itsmurugappan:
@boaz0 sorry for the delay. Here is the summary of changes

* add another || and handle env-file logic [here](https://github.com/knative/client/blob/82a21a5773be7157f9b46c4c7f7946adbb23457b/pkg/kn/flags/podspec.go#L172)
* Pass the envfile values to update and remove to this [func](https://github.com/knative/client/blob/82a21a5773be7157f9b46c4c7f7946adbb23457b/pkg/kn/flags/podspec_helper.go#L59)
* In the above func, the args are iterated to keep the order, so pls add an additional check similar to [this](https://github.com/knative/client/blob/82a21a5773be7157f9b46c4c7f7946adbb23457b/pkg/kn/flags/podspec_helper.go#L72) for env file. When this condition is satisfied, flush out all the env variables from env file to update iterator. (to keep the order)

Please let me know if you have any questions.

boaz0:
@itsmurugappan thanks for the guidance. I updated the PR. I hope it's good enough.
Thanks.

itsmurugappan:
changes look good to me, the only thing is code coverage is reduced so the pr merge might not succeed. @rhuss @maximilien @dsimansk can you please approve this pr, so it would allow the additional workflows to run.

rhuss:
@itsmurugappan the error looks strange, like a commit with an improper email. Would it be possible to squash the PR manually and force push it, to get rid of this ? 

--> 
image

dsimansk:
There was once a very similar error in the kamelet repository. rebase to `main` should get rid of it.

boaz0:
Sorry for the late response (holidays and family...).
Thanks for the opportunity to work on this and for the guidance @rhuss @dsimansk @itsmurugappan. Now moving to the next issue.

PR #1435, opened 2021-08-24
TestServiceExport works with images specified by digest

mgencur:
## Description

Images specified by digest are not set in user-image annotation. This PR corrects the test to work with image digest..

## Changes

* Do not expect user-image annotation to contain the image if the image is specified by digest.
* Do not expect image name in the output of source list as it might not be available when images are passed as image digest.

## Reference

Fixes #


itsmurugappan (approved):
/lgtm

dsimansk:
Thanks @mgencur!

/approve
/lgtm

PR #1441, opened 2021-08-30
Adding retry loop to update

### LINKED ISSUE #1102: Reuse conflict retry loop from client-go/util/retry
Instead of our homegrown algorithm for retrying updates in case of a conflict, we should reuse https://github.com/kubernetes/client-go/blob/1eb2027cd51eefaca0f851d5a2c90f8dd9b2a71d/util/retry/util.go#L68-L75 from the kubernetes client lib.

The goal of this refactoring task is to identify the spots where we are doing retries and switch over to this utility method, much like the `kn service import` pr does.

vyasgun:
## Description
Added retry loop from client go to service and source update commands.

## Changes

* Added retry loop from client go to service update command.
* Added a retry updated method to ping source APIs.

## Reference

Fixes #1102

### DIFF
--- a/pkg/eventing/v1/client.go
+++ b/pkg/eventing/v1/client.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"time"
 
+	"k8s.io/client-go/util/retry"
+
 	apis_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -32,6 +34,8 @@ import (
 	"knative.dev/client/pkg/util"
 	"knative.dev/client/pkg/wait"
 )
+
+type TriggerUpdateFunc func(origSource *eventingv1.Trigger) (*eventingv1.Trigger, error)
 
 // KnEventingClient to Eventing Sources. All methods are relative to the
 // namespace specified during construction
@@ -48,6 +52,8 @@ type KnEventingClient interface {
 	ListTriggers(ctx context.Context) (*eventingv1.TriggerList, error)
 	// UpdateTrigger is used to update an instance of trigger
 	UpdateTrigger(ctx context.Context, trigger *eventingv1.Trigger) error
+	// UpdateTriggerWithRetry is used to update an instance of trigger
+	UpdateTriggerWithRetry(ctx context.Context, name string, updateFunc TriggerUpdateFunc, nrRetries int) error
 	// CreateBroker is used to create an instance of broker
 	CreateBroker(ctx context.Context, broker *eventingv1.Broker) error
 	// GetBroker is used to get an instance of broker
@@ -135,6 +141,31 @@ func (c *knEventingClient) UpdateTrigger(ctx context.Context, trigger *eventingv
 		return kn_errors.GetError(err)
 	}
 	return nil
+}
+
+func (c *knEventingClient) UpdateTriggerWithRetry(ctx context.Context, name string, updateFunc TriggerUpdateFunc, nrRetries int) error {
+	return updateTriggerWithRetry(ctx, c, name, updateFunc, nrRetries)
+}
+
+func updateTriggerWithRetry(ctx context.Context, c KnEventingClient, name string, updateFunc TriggerUpdateFunc, nrRetries int) error {
+	b := retry.DefaultRetry
+	b.Steps = nrRetries
+	err := retry.RetryOnConflict(b, func() error {
+		source, err := c.GetTrigger(ctx, name)
+		if err != nil {
+			return err
+		}
+		if source.GetDeletionTimestamp() != nil {
+			return fmt.Errorf("can't update trigger %s because it has been marked for deletion", name)
+		}
+		updatedSource, err := updateFunc(source.DeepCopy())
+		if err != nil {
+			return err
+		}
+
+		return c.UpdateTrigger(ctx, updatedSource)
+	})
+	return err
 }
 
 // Return the client's namespace
--- a/pkg/eventing/v1/client.go
+++ b/pkg/eventing/v1/client.go
@@ -151,21 +151,25 @@ func updateTriggerWithRetry(ctx context.Context, c KnEventingClient, name string
 	b := retry.DefaultRetry
 	b.Steps = nrRetries
 	err := retry.RetryOnConflict(b, func() error {
-		source, err := c.GetTrigger(ctx, name)
-		if err != nil {
-			return err
-		}
-		if source.GetDeletionTimestamp() != nil {
-			return fmt.Errorf("can't update trigger %s because it has been marked for deletion", name)
-		}
-		updatedSource, err := updateFunc(source.DeepCopy())
-		if err != nil {
-			return err
-		}
-
-		return c.UpdateTrigger(ctx, updatedSource)
+		return updateTrigger(ctx, c, name, updateFunc)
 	})
 	return err
+}
+
+func updateTrigger(ctx context.Context, c KnEventingClient, name string, updateFunc TriggerUpdateFunc) error {
+	source, err := c.GetTrigger(ctx, name)
+	if err != nil {
+		return err
+	}
+	if source.GetDeletionTimestamp() != nil {
+		return fmt.Errorf("can't update trigger %s because it has been marked for deletion", name)
+	}
+	updatedSource, err := updateFunc(source.DeepCopy())
+	if err != nil {
+		return err
+	}
+
+	return c.UpdateTrigger(ctx, updatedSource)
 }
 
 // Return the client's namespace
--- a/pkg/eventing/v1/client.go
+++ b/pkg/eventing/v1/client.go
@@ -157,14 +157,14 @@ func updateTriggerWithRetry(ctx context.Context, c KnEventingClient, name string
 }
 
 func updateTrigger(ctx context.Context, c KnEventingClient, name string, updateFunc TriggerUpdateFunc) error {
-	source, err := c.GetTrigger(ctx, name)
+	trigger, err := c.GetTrigger(ctx, name)
 	if err != nil {
 		return err
 	}
-	if source.GetDeletionTimestamp() != nil {
+	if trigger.GetDeletionTimestamp() != nil {
 		return fmt.Errorf("can't update trigger %s because it has been marked for deletion", name)
 	}
-	updatedSource, err := updateFunc(source.DeepCopy())
+	updatedSource, err := updateFunc(trigger.DeepCopy())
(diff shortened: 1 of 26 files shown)


pkg/kn/commands/source/ping/update.go
```diff
@@ -19,10 +19,14 @@ import (
 	"fmt"
 
 	"github.com/spf13/cobra"
+	"k8s.io/client-go/util/retry"
+	"knative.dev/client/pkg/kn/commands/service"
```
rhuss, line 23: Why do you need this dependency ? Normally commands should stay independent and don't refer to each other to avoid cyclic dependency and allow them easily to be refactored/moved. 

If there is some code that can be reused, I'd propose to move it to a shared package and use it from there in service and here.

pkg/eventing/v1/client.go
```diff
@@ -137,6 +143,31 @@ func (c *knEventingClient) UpdateTrigger(ctx context.Context, trigger *eventingv
...
 }
 
+func (c *knEventingClient) UpdateTriggerWithRetry(ctx context.Context, name string, updateFunc TriggerUpdateFunc, nrRetries int) error {
+	return updateTriggerWithRetry(ctx, c, name, updateFunc, nrRetries)
+}
+
+func updateTriggerWithRetry(ctx context.Context, c KnEventingClient, name string, updateFunc TriggerUpdateFunc, nrRetries int) error {
+	b := retry.DefaultRetry
+	b.Steps = nrRetries
+	err := retry.RetryOnConflict(b, func() error {
```
rhuss, line 153: If you have such a long inner function, for better readability, I'd suggest to move the function to a top-level, named function. If you need some closures arguments, you could pass it as argument, like e.g. in

```golang
err := retry.RetryOnConflict(b, func() error { 
      return myInnerFunc(closureArg);
}
```

If no closure is needed, you could just use the function name

```golang
err := retry.RetryOnConflict(b, myInnerFunc);
```

rhuss (changes requested):
Looks good to me, thanks @vyasgun for jumping on this.

I have some proposals and questions as comments, but I think we are then still missing a bunch of update operations.

```
ls docs/cmd/*update* | cat -

docs/cmd/kn_domain_update.md
docs/cmd/kn_service_update.md
docs/cmd/kn_source_apiserver_update.md
docs/cmd/kn_source_binding_update.md
docs/cmd/kn_source_container_update.md
docs/cmd/kn_source_ping_update.md
docs/cmd/kn_subscription_update.md
docs/cmd/kn_trigger_update.md
```

Also, I would change our own update mechanism in `kn service update` to reuse the tool-based version.

Finally, don't forget to update the CHANGELOG ;-)

pkg/kn/commands/source/ping/update.go
vyasgun: This is because I imported the MaxUpdateRetries from service package. I will change this and create a variable in the same directory.

rhuss: Thanks !

dsimansk:
@vyasgun is the PR complete? Or are you still looking into some leftover update functions to add retry loop?

vyasgun:
@dsimansk I pushed a few more commits. It's done now and once reviewed we can merge.

pkg/eventing/v1/client_mock.go
```diff
@@ -115,6 +115,10 @@ func (c *MockKnEventingClient) UpdateTrigger(ctx context.Context, trigger *event
 	return mock.ErrorOrNil(call.Result[0])
 }
 
+func (c *MockKnEventingClient) UpdateTriggerWithRetry(ctx context.Context, name string, updateFunc TriggerUpdateFunc, nrRetries int) error {
```
dsimansk, line 118: Using the function in `client_mock_test.go` should be sufficient to get rid of codecov warning. Also for the other mock functions as well.

pkg/kn/commands/subscription/update.go
```diff
@@ -62,32 +66,28 @@ func NewSubscriptionUpdateCommand(p *commands.KnParams) *cobra.Command {
...
-			sb.Reply(rep)
-
-			ds, err := dlsFlag.ResolveSink(cmd.Context(), dynamicClient, namespace)
-			if err != nil {
-				return err
-			}
-			sb.DeadLetterSink(ds)
-
-			err = client.UpdateSubscription(cmd.Context(), sb.Build())
+			err = client.UpdateSubscriptionWithRetry(cmd.Context(), name, func(origSub *messagingv1.Subscription) (*messagingv1.Subscription, error) {
```
dsimansk, line 69: Maybe I'd opt for the same `updateFunctions := func(){}` definition as per Roland's suggestions and done in other parts of code. But it's really a minor detail.

dsimansk:
There's one more comment, but that's not super important. In addition there're several codecov warnings, but those can be addressed as part of improving coverage story.

IMO it looks very good, nice work!
/lgtm

@rhuss if you can make another please.

pkg/kn/commands/subscription/update.go
vyasgun: Done

rhuss:
not sure that I can make it today, but will review it early (European) morning at last.

pkg/messaging/v1/subscriptions_client.go
```diff
@@ -94,6 +102,35 @@ func (c *subscriptionsClient) UpdateSubscription(ctx context.Context, subscripti
...
+func (c *subscriptionsClient) UpdateSubscriptionWithRetry(ctx context.Context, name string, updateFunc SubscriptionUpdateFunc, nrRetries int) error {
+	return updateSubscriptionWithRetry(ctx, c, name, updateFunc, nrRetries)
+}
+
+func updateSubscriptionWithRetry(ctx context.Context, c KnSubscriptionsClient, name string, updateFunc SubscriptionUpdateFunc, nrRetries int) error {
+	b := retry.DefaultRetry
+	b.Steps = nrRetries
+	err := retry.RetryOnConflict(b, func() error {
+		return updateSubscription(ctx, c, name, updateFunc)
+	})
```
rhuss, line 114:
```suggestion
	err := retry.RetryOnConflict(b, updateSubscription)
	})
```

Same as above. I haven't checked every occurrence of this pattern, so you might want to update the other locations, too.

pkg/messaging/v1/subscriptions_client.go
```diff
@@ -94,6 +102,35 @@ func (c *subscriptionsClient) UpdateSubscription(ctx context.Context, subscripti
...
+	return err
+}
+
+func updateSubscription(ctx context.Context, c KnSubscriptionsClient, name string, updateFunc SubscriptionUpdateFunc) error {
+	sub, err := c.GetSubscription(ctx, name)
+	if err != nil {
+		return err
+	}
+	if sub.GetDeletionTimestamp() != nil {
+		return fmt.Errorf("can't update subscription %s because it has been marked for deletion", name)
```
rhuss, line 124: I wonder if in this case the retries will still apply ? I mean, this is a terminal error which won't be fixed when trying more often. But I guess that the retry support only retries in case of conflicts, right ?

pkg/serving/v1alpha1/client.go
```diff
@@ -98,6 +106,35 @@ func (cl *knServingClient) UpdateDomainMapping(ctx context.Context, domainMappin
...
 
+func (cl *knServingClient) UpdateDomainMappingWithRetry(ctx context.Context, name string, updateFunc DomainUpdateFunc, nrRetries int) error {
+	return updateDomainMappingWithRetry(ctx, cl, name, updateFunc, nrRetries)
+}
+
+func updateDomainMappingWithRetry(ctx context.Context, cl KnServingClient, name string, updateFunc DomainUpdateFunc, nrRetries int) error {
+	b := retry.DefaultRetry
+	b.Steps = nrRetries
+	err := retry.RetryOnConflict(b, func() error {
+		return updateDomain(ctx, cl, name, updateFunc)
```
rhuss, line 117: see comment above how to simplify

pkg/sources/v1beta2/ping_client.go
```diff
@@ -94,6 +101,35 @@ func (c *pingSourcesClient) UpdatePingSource(ctx context.Context, pingSource *so
...
 }
 
+func (c *pingSourcesClient) UpdatePingSourceWithRetry(ctx context.Context, name string, updateFunc PingSourceUpdateFunc, nrRetries int) error {
+	return updatePingSourceWithRetry(ctx, c, name, updateFunc, nrRetries)
+}
+
+func updatePingSourceWithRetry(ctx context.Context, c KnPingSourcesClient, name string, updateFunc PingSourceUpdateFunc, nrRetries int) error {
+	b := retry.DefaultRetry
+	b.Steps = nrRetries
+	err := retry.RetryOnConflict(b, func() error {
```
rhuss, line 111: dito

rhuss (commented):
Thanks a lot. I have a minor suggestion to simplify the call to a function (avoiding one unnecessary inner function) and some other minor comments.

Overall looks very good and I think we are close to merge it.

pkg/eventing/v1/client.go
```diff
@@ -137,6 +143,35 @@ func (c *knEventingClient) UpdateTrigger(ctx context.Context, trigger *eventingv
...
+		return updateTrigger(ctx, c, name, updateFunc)
+	})
```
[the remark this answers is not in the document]
vyasgun: The signature for the Retry function has to be `func() error`. I could do something like follows if it looks neater?

```
updateTriggerFunc := func() error {
	return updateTrigger(ctx, c, name, updateFunc)
}
err := retry.RetryOnConflict(b, updateTriggerFunc)
```

pkg/messaging/v1/subscriptions_client.go
vyasgun: Yes, it will only retry on conflict

pkg/eventing/v1/client.go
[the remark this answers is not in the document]
rhuss: ah, no worries. I thought the signatures were the same, so, I'm fine with it as it is now. my bad.

pkg/messaging/v1/subscriptions_client.go
rhuss: Again, please ignore my comment if the signature do no match. Sorry, should have seen that.

pkg/serving/v1alpha1/client.go
rhuss: ditp, please ignore.

pkg/messaging/v1/subscriptions_client.go
rhuss: thanks!

rhuss:
thanks a lot !

/approve
/lgtm

PR #1457, opened 2021-09-21
Fix windows e2e plugin tests

vyasgun:
## Description
Added a few windows specific changes so the tests work on the OS

## Changes

* Changed the kn binary extension to exe on windows
* Added OS specific path separators
* Added batch code for plugin tests on windows

## Reference

Fixes #

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -129,6 +129,11 @@ source_lint() {
 go_build() {
   echo "🚧 Compile"
   go build -mod=vendor -ldflags "$(build_flags $(basedir))" -o kn ./cmd/...
+  WIN="Windows"
+  FILE=$(file ./kn)
+  if grep -q "$WIN" <<< "$FILE"; then
+    mv ./kn ./kn.exe
+  fi
 }
 
 go_test() {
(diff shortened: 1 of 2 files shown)


hack/build.sh
```diff
@@ -129,6 +129,11 @@ source_lint() {
 go_build() {
   echo "🚧 Compile"
   go build -mod=vendor -ldflags "$(build_flags $(basedir))" -o kn ./cmd/...
+  WIN="Windows"
+  FILE=$(file ./kn)
+  if grep -q "$WIN" <<< "$FILE"; then
+    mv ./kn ./kn.exe
+  fi
```
rhuss, line 136:
```suggestion
  if $(file kn | grep -q -i "Windows"); then
    mv kn kn.exe
  fi
```

This is a more canonical way for this checking (and also does not pollute with global variables.

rhuss (commented):
Thanks! Looks good to me with a minor proposal for improvement.

hack/build.sh
vyasgun: Done

PR #1459, opened 2021-09-22
Update versions for 0.26.0

rhuss:


pkg/kn/config/config_test.go
```diff
@@ -29,7 +29,7 @@ func TestBootstrapConfig(t *testing.T) {
 	configYaml := `
 plugins:
   directory: /tmp
-  path-lookup: true
+
```
rhuss, line 32: not sure why this slipped in, but should not harm (as this ensure that path-lookup is the default mode)

pkg/kn/config/config_test.go
```diff
@@ -28,8 +28,7 @@ import (
 func TestBootstrapConfig(t *testing.T) {
 	configYaml := `
 plugins:
-  directory: /tmp
-  path-lookup: true
+  directory: /tmp
```
vyasgun, line 31:
```suggestion
  directory: /tmp
```

rhuss:
@vyasgun sorry, had to rebase. Could I get another `/lgtm` please ? :)

PR #1463, opened 2021-09-24
Changing the Default retry value

vyasgun:
## Description

Modifying default retry value for backoff during resource updates
## Changes

* Changing the factor to 1.5 
* Changing MaxUpdateRetries to 5

## Reference

Fixes #

### DIFF
--- a/pkg/eventing/v1/client.go
+++ b/pkg/eventing/v1/client.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"time"
 
+	"knative.dev/client/pkg/config"
+
 	"k8s.io/client-go/util/retry"
 
 	apis_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -148,7 +150,7 @@ func (c *knEventingClient) UpdateTriggerWithRetry(ctx context.Context, name stri
 }
 
 func updateTriggerWithRetry(ctx context.Context, c KnEventingClient, name string, updateFunc TriggerUpdateFunc, nrRetries int) error {
-	b := util.DefaultRetry
+	b := config.DefaultRetry
 	b.Steps = nrRetries
 	updateTriggerFunc := func() error {
 		return updateTrigger(ctx, c, name, updateFunc)
--- a/pkg/eventing/v1/client.go
+++ b/pkg/eventing/v1/client.go
@@ -148,7 +148,7 @@ func (c *knEventingClient) UpdateTriggerWithRetry(ctx context.Context, name stri
 }
 
 func updateTriggerWithRetry(ctx context.Context, c KnEventingClient, name string, updateFunc TriggerUpdateFunc, nrRetries int) error {
-	b := retry.DefaultRetry
+	b := util.DefaultRetry
 	b.Steps = nrRetries
 	updateTriggerFunc := func() error {
 		return updateTrigger(ctx, c, name, updateFunc)
--- a/pkg/kn/commands/service/create.go
+++ b/pkg/kn/commands/service/create.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"os"
 
+	"knative.dev/client/pkg/config"
 	"knative.dev/client/pkg/kn/commands"
 	servinglib "knative.dev/client/pkg/serving"
 
@@ -208,7 +209,7 @@ func prepareAndUpdateService(ctx context.Context, client clientservingv1.KnServi
 		changed, err := client.UpdateService(ctx, service)
 		if err != nil {
 			// Retry to update when a resource version conflict exists
-			if apierrors.IsConflict(err) && retries < MaxUpdateRetries {
+			if apierrors.IsConflict(err) && retries < config.DefaultRetry.Steps {
 				retries++
 				continue
 			}
--- a/pkg/kn/commands/service/create.go
+++ b/pkg/kn/commands/service/create.go
@@ -178,12 +178,7 @@ func waitIfRequested(ctx context.Context, client clientservingv1.KnServingClient
 }
 
 func prepareAndUpdateService(ctx context.Context, client clientservingv1.KnServingClient, service *servingv1.Service) (bool, error) {
-	var retries = 0
-	for {
-		existingService, err := client.GetService(ctx, service.Name)
-		if err != nil {
-			return false, err
-		}
+	updateFunc := func(origService *servingv1.Service) (*servingv1.Service, error) {
 
 		// Copy over some annotations that we want to keep around. Erase others
 		copyList := []string{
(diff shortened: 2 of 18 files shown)


pkg/util/compare.go
```diff
@@ -17,10 +17,25 @@ package util
...
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"gotest.tools/v3/assert/cmp"
 )
 
+var (
+	// DefaultRetry is the default set of rules
+	// to be followed when retrying for conflicts during
+	// resource update
+	DefaultRetry = wait.Backoff{
```
rhuss, line 31: I would not put the default to that (unrelated) util package but move it more closely to the API where it is used. I know, it is used in several different clients where this is used, so finding a proper place is tricky. But this package is for sure not the correct place ;-) For example, if we would have an intermediate package "clients" in `pkg/clients` where all our clients to the backend are collected, then this would be the proper package (we don't have that yet, but maybe that would be a good idea ?)

For now, I would not hesitate to introduce a `pkg/config` with a single file `defaults.go` in a package "config". Might sound like overkill at first, but will help future refactoring if we clearly separate the concerns.

(my general rule of thumb: If one does not know where to put a shared constant, a global `util` package is usually a bad idea (and `util` should be the last choice considered anyway in any case :). Utilities should be independent of a concrete domain and generally reusable for a wide range of scenarios)

rhuss (commented):
Thanks a lot for the quick turnaround.

Looks good, but see my comment where to place the default. Also, I think we could remove "MaxUpdateRetries" everywhere in favor of setting this in the default directly. This would also reduce the footprint of the method signatures as we don't have to carry over the retry (which typically is not changed anyway (alternatively change the `nrRetries` parameter to use a full backoff. Maybe make the backoff an optional parameter (via `...`) ?

maximilien (commented):
Thanks for contribution, however, can you also please add a bit more info (rationale) why these new values are better. Perhaps you have tests that shows increasing from 3 to 5 is better. Otherwise, looks like we are just guessing...

rhuss:
5 is the default value chosen by Kubernetes --> xhttps://github.com/knative/client/blob/ad4a7b18a2e8724d9ca08311a64f28a0adb73c47/vendor/k8s.io/client-go/util/retry/util.go#L26-L33

So the question is more, why we haven chosen 3 ;-) Actually the reason we area fiddling around with the values is that we still see CI flakes with conflict errors.

pkg/util/compare.go
vyasgun: I wasn't sure about the file to put in. I have moved the variable to the suggested location. I put it in util because the DefaultRetry variable is client-go package is in k8s.io/client-go/util/retry/util.go so why is it incorrect in our case?

vyasgun:
@rhuss @maximilien I have added a comment for the variable to specify that all fields except `duration` and `factor` have the same value as the client-go's `DefaultRetry`.

vyasgun:
@rhuss If we refactor the retry functions to accept optional number of nrRetries, we will have to add a validation step for not accepting more than 1 arg. Would it be better to just change the signature so it doesn't accept any nrRetries at all?

pkg/kn/commands/source/ping/update.go
```diff
@@ -93,7 +94,7 @@ func NewPingUpdateCommand(p *commands.KnParams) *cobra.Command {
 				updatedSource := b.Build()
 				return updatedSource, nil
 			}
-			backoff := retry.DefaultRetry
+			backoff := config.DefaultRetry
 			backoff.Steps = MaxUpdateRetries
```
dsimansk, line 98: Do we need to actually set the `.Steps` on top of default value, since `MaxUpdateRetries` is changed to `5` everywhere.

vyasgun:
@dsimansk Thanks for the review. I have changed the argument to `DefaultRetry.Steps` for now and removed `MaxUpdateRetries` constants from all packages.

pkg/kn/commands/service/create.go
```diff
@@ -208,7 +209,7 @@ func prepareAndUpdateService(ctx context.Context, client clientservingv1.KnServi
 		changed, err := client.UpdateService(ctx, service)
 		if err != nil {
 			// Retry to update when a resource version conflict exists
-			if apierrors.IsConflict(err) && retries < MaxUpdateRetries {
+			if apierrors.IsConflict(err) && retries < config.DefaultRetry.Steps {
```
dsimansk, line 212: I know this's only changed due to removal of variable, but looking at the code itself I wonder if shouldn't be just `.UpdateServiceWithRetry` without that additional conflict handling block.

pkg/kn/commands/source/ping/ping.go
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 const (
-	MaxUpdateRetries = 3
+	MaxUpdateRetries = 5
```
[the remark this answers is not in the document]
dsimansk: @vyasgun one more leftover const, otherwise it's OK.

pkg/kn/commands/trigger/trigger.go
```diff
@@ -22,7 +22,7 @@ import (
 
 const (
 	// How often to retry in case of an optimistic lock error when replacing a trigger (--force)
-	MaxUpdateRetries = 3
+	MaxUpdateRetries = 5
```
dsimansk, line 25: Same here.

pkg/kn/commands/source/ping/ping.go
dsimansk, line 28: @vyasgun one more leftover const, otherwise it's OK.

dsimansk (changes requested):

pkg/kn/commands/source/ping/ping.go
vyasgun: Thanks for pointing out. Fixed the remaining consts

dsimansk:
Thanks @vyasgun!

/approve
/lgtm

Adding hold for @rhuss @maximilien to take a look.
/hold

pkg/eventing/v1/client.go
```diff
@@ -148,7 +150,7 @@ func (c *knEventingClient) UpdateTriggerWithRetry(ctx context.Context, name stri
 }
 
 func updateTriggerWithRetry(ctx context.Context, c KnEventingClient, name string, updateFunc TriggerUpdateFunc, nrRetries int) error {
-	b := retry.DefaultRetry
+	b := config.DefaultRetry
```
rhuss, line 153: Looks good to me, except that I probably would change the signature/API instead of taking a retry number to take a full backoff struct for more flexibility.

But happy to merge now, too, we could adapt later as well.

vyasgun: Since the UpdateServiceWithRetry function is being used in the func plugin as well, I think we can change the signature later?

dsimansk:
After chat with @rhuss, it should be good to go. With a minor future improvement stated in the comment.
/unhold

PR #1483, opened 2021-10-18
Calculate traffic split when N-1 revisions are specified

vyasgun:
## Changes

* Determine traffic split when N-1 revisions are specified
* In case `@latest` is specified, cannot determine the split unless it sums to 100

## Reference

Fixes #


vyasgun:
thanks, @dsimansk!

dsimansk:
Since all the review comments are addressed, lets get it in. :)

Thanks!
/approve
/lgtm

PR #1497, opened 2021-10-29
Reword missing API error to mention client update

dsimansk:
## Description

Add a mention of client update to common missing API error message. In case of older client with newer backend with removed API, it might confusing why there's a failure.

## Changes

* Reword missing API error to mention client update

## Reference

Fixes #

### DIFF
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -25,7 +25,7 @@ func NewInvalidCRD(apiGroup string) *KNError {
 	parts := strings.Split(apiGroup, ".")
 	name := parts[0]
 	msg := fmt.Sprintf("no or newer Knative %s API found on the backend, please verify the installation or "+
-		"update the 'kn' client to latest version", firstCharToUpper(name))
+		"update the 'kn' client", firstCharToUpper(name))
 	return NewKNError(msg)
 }
 
--- a/pkg/errors/errors.go
+++ b/pkg/errors/errors.go
@@ -24,7 +24,7 @@ import (
 func NewInvalidCRD(apiGroup string) *KNError {
 	parts := strings.Split(apiGroup, ".")
 	name := parts[0]
-	msg := fmt.Sprintf("no Knative %s API found on the backend, please verify the installation or "+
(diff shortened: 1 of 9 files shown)


pkg/errors/errors.go
```diff
@@ -17,12 +17,15 @@ package errors
...
+	"unicode/utf8"
 )
 
-func newInvalidCRD(apiGroup string) *KNError {
+func NewInvalidCRD(apiGroup string) *KNError {
 	parts := strings.Split(apiGroup, ".")
 	name := parts[0]
-	msg := fmt.Sprintf("no Knative %s API found on the backend, please verify the installation", name)
+	msg := fmt.Sprintf("no Knative %s API found on the backend, please verify the installation or "+
+		"update the 'kn' client to latest version", firstCharToUpper(name))
```
dsimansk, line 28: Might be better to call it `matching version`?

rhuss: I wonder whether 'no Knative API' found is appropriate as there could be a Knative API available but too new. So maybe rephrase to "no or newer Knative %s API found on the backend, please verify the installation or update this client"

I just would say update, as it's kind of clear that it should be moved to a matching client.

pkg/errors/errors.go
```diff
@@ -37,3 +40,11 @@ func newNoRouteToHost(errString string) *KNError {
 func newNoKubeConfig(errString string) *KNError {
 	return NewKNError("no kubeconfig has been provided, please use a valid configuration to connect to the cluster")
 }
+
+func firstCharToUpper(s string) string {
```
rhuss, line 44: +1

Actually, I would use this function, too, before emitting an error message to the console, so that error messages that start with a lowercase letter (golang convention) will be upcased. There should be a single exit-point for errors to do the massaging of the error message to print. But that's of course another story ;-)

rhuss (commented):
Loogs good, one minor suggestion wrt/ wording.

pkg/errors/errors.go
dsimansk: @rhuss addressed in latest update.

rhuss (commented):
/approve
/lgtm

PR #1499, opened 2021-11-01
Rename `--extra-containers` to `--containers`

rhuss:
and make `--extra-containers deprecated`


dsimansk:
Changelog entry is missing, but we can add that during final preps for the release.

/approve
/lgtm

PR #1508, opened 2021-11-02
update status of commands

itsmurugappan:
## Description

Update status of Export/Import/Gitops to Beta

cc @rhuss @dsimansk

### DIFF
--- a/pkg/kn/commands/domain/create.go
+++ b/pkg/kn/commands/domain/create.go
@@ -33,8 +33,8 @@ func NewDomainMappingCreateCommand(p *commands.KnParams) *cobra.Command {
 		Use:   "create NAME",
 		Short: "Create a domain mapping",
 		Example: `
-  # Create a domain mappings 'hello.example.com' for Knative service 'hello' with TLS secret set
-  kn domain create hello.example.com --ref hello --tls my-cert-secret (Beta)`,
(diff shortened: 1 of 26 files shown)


dsimansk:
I haven't found `kn container` in the mix. That should be `Beta` as well.

pkg/kn/commands/domain/create.go
```diff
@@ -33,7 +33,7 @@ func NewDomainMappingCreateCommand(p *commands.KnParams) *cobra.Command {
 		Use:   "create NAME",
 		Short: "Create a domain mapping",
 		Example: `
-  # Create a domain mappings 'hello.example.com' for Knative service 'hello' with TLS secret set
+  # Create a domain mappings 'hello.example.com' for Knative service 'hello' with TLS secret set (Beta)
```
rhuss, line 36: I wonder whether we should mark the domain management as "Beta" or whether we say that it's already GA ? I think we can move it to GA level as the UI is straightforward and not really contentious.

rhuss (commented):
Thanks, looks good to me, but I wonder whether we should still mark the domain management as 'beta' ?

pkg/kn/commands/domain/create.go
dsimansk: +1 to GA

dsimansk:
@itsmurugappan thanks!

Looks good now, let's get it in.
/approve
/lgtm

PR #1510, opened 2021-11-03
Fix the most common E2E flakes

dsimansk:
## Changes

* Add simple wait for URL to domain mapping E2E test
* Disable sync delete of brokers used in defer

### DIFF
--- a/test/e2e/domain_mapping_test.go
+++ b/test/e2e/domain_mapping_test.go
@@ -108,6 +108,16 @@ func domainList(r *test.KnRunResultCollector, domainName string) {
 }
 
 func domainDescribe(r *test.KnRunResultCollector, domainName string, tls bool) {
+	k := test.NewKubectl(r.KnTest().Kn().Namespace())
+	// Wait for Domain Mapping URL to be populated
+	for i := 0; i < 10; i++ {
+		out, err := k.Run("get", "domainmapping", domainName, "-o=jsonpath='{.status.url}'")
+		assert.NilError(r.T(), err)
+		if len(out) > 0 {
+			break
+		}
+		time.Sleep(time.Second)
+	}
 	out := r.KnTest().Kn().Run("domain", "describe", domainName)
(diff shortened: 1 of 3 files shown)


test/e2e/sink_test.go
```diff
@@ -38,7 +38,7 @@ func TestSink(t *testing.T) {
 
 	// create broker
 	test.BrokerCreate(r, "default")
-	defer test.BrokerDelete(r, "default", true)
+	defer test.BrokerDelete(r, "default", false)
```
dsimansk, line 41: There's a specific set of Broker tests that cover both --wait/--no-wait cases. Usually the flake appears in Sink or Trigger tests when used in `defer` as a cleanup. Even with added sleep to wait-loop I couldn't reproduce the behavior with  a consecutive create/delete cycles the events always got through. 

Since the defer is called after tests and finally the whole namespace is deleted, I tend to believe we don't need to wait for broker delete, nor cause the tests to fail falsely.

test/e2e/domain_mapping_test.go
```diff
@@ -108,6 +108,16 @@ func domainList(r *test.KnRunResultCollector, domainName string) {
...
+	k := test.NewKubectl(r.KnTest().Kn().Namespace())
+	// Wait for Domain Mapping URL to be populated
+	for i := 0; i < 10; i++ {
+		out, err := k.Run("get", "domainmapping", domainName, "-o=jsonpath='{.status.url}'")
+		assert.NilError(r.T(), err)
+		if len(out) > 0 {
+			break
+		}
+		time.Sleep(time.Second)
+	}
```
dsimansk, line 120: Checking the domain mapping content immediately after creation leads to false-positive URL missing fails. The following simple look should make sure URL is there to be checked by the assert later.

dsimansk:
Seems to be flake free. :)

/test all

maximilien (approved):
/ok-to-test
/approve
/lgtm

PR #1514, opened 2021-11-08
Add --tag flag to service create and allow traffic split <100 when @latest is specified

vyasgun:
## Description

Add `--tag` flag to service create and interpret just value arguments to the flag as `@latest=value`
When `@latest` is specified as argument to `--traffic` and sum < 100, calculate the remaining revision

## Changes

* The arguments that don't follow `key=value` format and just `value` format will be treated as `@latest=value`
* `service create` will have a `--tag` flag so revision is created with a tag
* Updating traffic split for a service when `@latest` is specified and the total sum is less than 100 for a service with N revisions will work as follows:
     * If the update leads to a new revision creation, total N `key=value` arguments need to be specified for `--traffic`
     *  If the update does not create a new revision, total N-1 `key=value` arguments need to be specified for `--traffic`

## Reference

Fixes #

### DIFF
--- a/pkg/kn/commands/service/create_test.go
+++ b/pkg/kn/commands/service/create_test.go
@@ -1091,3 +1091,9 @@ func TestServiceCreateTag(t *testing.T) {
 	assert.Equal(t, len(created.Spec.Traffic), 1)
 	assert.Equal(t, created.Spec.Traffic[0].Tag, "foo")
 }
+
+func TestServiceCreateTagWithError(t *testing.T) {
+	_, _, _, err := fakeServiceCreate([]string{
+		"service", "create", "foo", "--image", "gcr.io/foo/bar:baz", "--tag", "foo,bar"}, false)
+	assert.Error(t, err, "repetition of identifier @latest is not allowed, use only once with --tag flag")
+}
--- a/pkg/kn/commands/service/create_test.go
+++ b/pkg/kn/commands/service/create_test.go
@@ -1082,3 +1082,12 @@ func TestServiceCreateFromYAMLWithOverrideError(t *testing.T) {
 	assert.Assert(t, err != nil)
 	assert.Assert(t, util.ContainsAll(err.Error(), "expected", "0", "<=", "2147483647", "autoscaling.knative.dev/maxScale"))
 }
+
+func TestServiceCreateTag(t *testing.T) {
+	action, created, _, err := fakeServiceCreate([]string{
+		"service", "create", "foo", "--image", "gcr.io/foo/bar:baz", "--tag", "foo"}, false)
+	assert.NilError(t, err)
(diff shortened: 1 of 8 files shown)


pkg/kn/commands/service/create_test.go
```diff
@@ -1082,3 +1082,12 @@ func TestServiceCreateFromYAMLWithOverrideError(t *testing.T) {
 	assert.Assert(t, err != nil)
 	assert.Assert(t, util.ContainsAll(err.Error(), "expected", "0", "<=", "2147483647", "autoscaling.knative.dev/maxScale"))
 }
+
```
maximilien, line 1085: What about create tests with bad (e.g., misspelled `@latst=100`) and other parsing issues?

maximilien (changes requested):
Thanks for contribution. Left one comment

pkg/kn/traffic/compute.go
```diff
@@ -44,6 +44,9 @@ func newServiceTraffic(traffic []servingv1.TrafficTarget) ServiceTraffic {
 
 func splitByEqualSign(pair string) (string, string, error) {
 	parts := strings.Split(pair, "=")
+	if len(parts) == 1 {
+		return latestRevisionRef, parts[0], nil
```
dsimansk, line 48: Can we get this part covered by tests please? It can be only specific test for the `splitByEqualSign` function.

test/e2e/traffic_split_test.go
```diff
@@ -485,6 +503,7 @@ func TestTrafficSplit(t *testing.T) {
 
 func verifyTargets(r *test.KnRunResultCollector, serviceName string, expectedTargets []TargetFields, expectErr bool) {
 	out := test.ServiceDescribeWithJSONPath(r, serviceName, targetsJsonPath)
+	r.T().Log(out)
```
dsimansk: Debug log?

pkg/kn/traffic/compute.go
vyasgun: Will add

test/e2e/traffic_split_test.go
vyasgun: Accidentally left it in the code. 😅 Just removed

pkg/kn/commands/service/create_test.go
vyasgun: Thanks, I will add those

dsimansk (commented):
Thanks @vyasgun! I can say that the code finally settled with me and I understand the motivation and changes. :)

/approve

@rhuss: Would you have time to take a look?

dsimansk (approved):
Thanks!

/approve
/lgtm

rhuss:
Sorry for being so late. I will give the whole traffic refactorings a spin today (or early next week).

PR #1516, opened 2021-11-09
Fix env, annotation and labels flags in service create/update/apply

### LINKED ISSUE #1449: Help message for `kn service apply --env` is appropriate
### Bug report

`kn service apply --help` yields:

```
-e, --env stringArray                   Environment variable to set. NAME=value; you may provide this flag any number of times to set multiple environment variables. To unset, specify the environment variable name followed by a "-" (e.g., NAME-).
```

This is not correct since you can't add or remove env vars with `apply` but just set the full set of env variables. 

Same possibly for `--labels` and `--annotations` (or any map-based option)

### Expected behavior

The help message for `--env` should be different for `kn service update` and `kn service apply` as the behaviour is different (`update` : incremental changes by adding/removing, `apply`: full state)

The same is true for `kn service create` where removing env vars also doesn't make any sense.

Thanks to @pmbanugo for reporting this issue.

boaz0:
## Description/Changes

Update descriptions to the following flags in `service create/update/apply`:
* env
* labels
* annotations

If you can not unset the label/annotation/env then don't mention it in the description of the flag.

## Reference

Fixes #1449

### DIFF
--- a/docs/cmd/kn_service_apply.md
+++ b/docs/cmd/kn_service_apply.md
@@ -29,7 +29,7 @@ kn service apply s0 --filename my-svc.yml
 ### Options
 
 ```
-  -a, --annotation stringArray            Annotations to set for both Service and Revision. name=value; you may provide this flag any number of times to set multiple annotations. To unset, specify the annotation name followed by a "-" (e.g., name-).
+  -a, --annotation stringArray            Annotations to set for both Service and Revision. name=value; you may provide this flag any number of times to set multiple annotations.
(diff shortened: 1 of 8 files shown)


dsimansk (changes requested):
It looks good to me, but it'd great if you could add unit tests to newly added `AddCreateFlags` to meet test coverage threshold as well.

boaz0:
@dsimansk sure. No problem.

dsimansk (approved):
Thanks @boaz0!

/approve
/lgtm

PR #1521, opened 2021-11-16
Move away from grpc-ping test image as symlink

dsimansk:
## Description

The former setup seems to break builds on Golang 1.17 with missing go dependencies. 
And on pair with https://github.com/knative/client/blob/main/test/test_images/multicontainer/release.yaml

## Changes

* Move away from grpc-ping test image as symlink

## Reference

Fixes #

/cc @markusthoemmes


markusthoemmes (approved):
/lgtm
/approve

PR #1547, opened 2021-12-14
Added autocompletion for service name

### LINKED ISSUE #1373: Context sensitive Bash completion
### Feature request

### UI Example

```
kn service describe my<TAB>
```

- [x] #1374 
- [x] #1547 
- [x] #1559
- [x] #1560
- [x] #1561
- [ ] #1562

vyasgun:
## Description
Added ValidArgsFunc for autocompletion during `kn service describe` and `kn service update`

## Changes

* Added ValidArgsFunc to `service describe` and `service update` commands

```
./kn service update hello-example<tab>
hello-example    hello-example-2  hello-example-3  hello-example-4
```

## Reference

Fixes #1373

### DIFF
--- a/pkg/kn/commands/completion_helper.go
+++ b/pkg/kn/commands/completion_helper.go
@@ -1,0 +1,129 @@
+// Copyright © 2021 The Knative Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package commands
+
+import (
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+type completionConfig struct {
+	params     *KnParams
+	command    *cobra.Command
+	args       []string
+	toComplete string
+	target     string
+}
+
+var (
+	resourceToFuncMap = map[string]func(config *completionConfig) ([]string, cobra.ShellCompDirective){
+		"service": completeService,
+	}
+)
+
+// ResourceNameCompletionFunc will return a function that will autocomplete the name of
+// the resource based on the subcommand
+func ResourceNameCompletionFunc(p *KnParams) func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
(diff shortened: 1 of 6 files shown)


rhuss:
Thanks a ton, @vyasgun ! I hope I can have a look still before the break and lets get it merged then. Great stuff :)

vyasgun:
@rhuss Thanks! It would be great if you could take a look. I can implement for other resources similarly

pkg/kn/commands/completion_helper.go
```diff
@@ -0,0 +1,129 @@
...
+
+var (
+	resourceToFuncMap = map[string]func(config *completionConfig) ([]string, cobra.ShellCompDirective){
+		"service": completeService,
+	}
+)
+
+// ResourceNameCompletionFunc will return a function that will autocomplete the name of
+// the resource based on the subcommand
+func ResourceNameCompletionFunc(p *KnParams) func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
rhuss, line 39: Is there a case when *not* `cobra.ShellCompDirectiveNoFileComp` is returned ? Because then you could remove that return value.

pkg/kn/commands/completion_helper.go
```diff
@@ -0,0 +1,129 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+type completionConfig struct {
+	params     *KnParams
+	command    *cobra.Command
+	args       []string
+	toComplete string
+	target     string
```
rhuss, line 28: I wonder whether `target` needs to be in an extra field. It's specific only to `service` commands and could easily be extracted from the `command` when it is already clear that we are dealing with services.

rhuss (commented):
Thanks ! (also for adding the extra tests ;-)

Looks good to me on a first view, I have some minor remarks.

pkg/kn/commands/completion_helper.go
vyasgun: yes, I think we can remove the field

pkg/kn/commands/completion_helper.go
vyasgun: `cobra.ShellCompDirectiveNoFileComp` is being returned in all cases. The function being returned needs to be of this specific signature to be used as a `ValidArgsFunc`

pkg/kn/commands/namespaced.go
```diff
@@ -43,7 +45,11 @@ func AddNamespaceFlags(flags *pflag.FlagSet, allowAll bool) {
 
 // GetNamespace returns namespace from command specified by flag
 func (params *KnParams) GetNamespace(cmd *cobra.Command) (string, error) {
-	namespace := cmd.Flag("namespace").Value.String()
+	namespaceFlag := cmd.Flag("namespace")
+	if namespaceFlag == nil {
```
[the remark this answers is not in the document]
vyasgun: thank you :)

dsimansk (approved):
Thanks!

/approve
/lgtm

PR #1556, opened 2021-12-26
Add config command to mange configuration file

### LINKED ISSUE #1293: Add "config" command to manage configuration file
### Feature request

Add a new command group `config` for managing `kn` configuration files

### Use case

`kn` can be customized by a local configuration file, that e.g. can specify how plugins are found or add additional sink mappings.

In order to provide an enhanced user experience to manages that configuration a new top-level command group, `config` is proposed. 
The details of the subcommand still need to be fleshed out, but I could imagine having commands like:

```
# Create a default configuration file, with annotated comments for possible options
kn config init

# Set a configuration option for scalar options with scalar values. Key specified as JSON path
kn config set plugins.directory=~/.kn

# Get an individual configuration value or group of values
kn config get eventing

# Show the whole configuration
kn config show
 
# Specialized commands for specific use cases that involve list commands
kn config sink-mapping add svc --group core --version v1 --resource services
kn config sink-mapping delete svc
kn config sink-mapping list
```

boaz0:
## Description
Add config command to kn CLI to read, edit or initialize knative configuration files

## Changes

* Add `config get`
* Add `config set (plugins.directory)`
* Add `config show`
* Add `config sink-mapping (list / update / add / delete)`
* Add `config channel-type-mapping (list / update / add / delete)`

## Reference

closes #1293

/lint

### DIFF
--- a/pkg/kn/commands/config/channel/add.go
+++ b/pkg/kn/commands/config/channel/add.go
@@ -1,0 +1,60 @@
+// Copyright © 2022 The Knative Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package channel
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"knative.dev/client/pkg/kn/config"
+)
+
+const ChannelTypeAliasExistsErrorMsg = "Channel type alias %q already exists in the configuration file"
+
+// NewAddCommand defines and processes `kn config channel-type-mapping add`
+func NewAddCommand() *cobra.Command {
+	ctmf := ChannelTypeMappingFlags{}
+	addChannelTypeMapCmd := &cobra.Command{
+		Use:   "add",
+		Short: "Add a new channel type mapping",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			items := []config.ChannelTypeMapping{}
+			if err := viper.UnmarshalKey(config.KeyChannelTypeMappings, &items); err != nil {
+				return err
+			}
+			channelHash := map[string]bool{}
+			for _, item := range items {
+				channelHash[item.Alias] = true
+			}
+			for _, alias := range args {
+				if _, ok := channelHash[alias]; ok {
+					return fmt.Errorf(ChannelTypeAliasExistsErrorMsg, alias)
+				}
+				ctmf.Alias = alias
+			}
+			newList := append(items, ctmf.ChannelTypeMapping)
+			viper.Set(config.KeyChannelTypeMappings, newList)
+			if err := viper.WriteConfig(); err != nil {
+				return err
+			}
+			fmt.Fprintf(cmd.OutOrStdout(), "Channel type mapping with alias '%s' was added\n", args[0])
+			return nil
+		},
+	}
+	ctmf.AddFlags(addChannelTypeMapCmd)
+	return addChannelTypeMapCmd
+}
--- a/pkg/kn/commands/config/channel_type_mapping.go
+++ b/pkg/kn/commands/config/channel_type_mapping.go
@@ -1,0 +1,34 @@
+// Copyright © 2022 The Knative Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package configCmd
+
+import (
+	"github.com/spf13/cobra"
+	"knative.dev/client/pkg/kn/commands"
+	"knative.dev/client/pkg/kn/commands/config/channel"
+)
+
+// NewConfigChannelTypeMappingCommand represents 'kn config channel-type-mapping' command
+func NewConfigChannelTypeMappingCommand(p *commands.KnParams) *cobra.Command {
+	channelTypeMappingCmd := &cobra.Command{
+		Use:   "channel-type-mapping",
+		Short: "Manage channel type mapping list",
+	}
+	channelTypeMappingCmd.AddCommand(channel.NewDeleteCommand())
+	channelTypeMappingCmd.AddCommand(channel.NewAddCommand())
+	channelTypeMappingCmd.AddCommand(channel.NewUpdateCommand())
+	channelTypeMappingCmd.AddCommand(channel.NewListCommand())
+	return channelTypeMappingCmd
+}
--- a/pkg/kn/commands/config/config.go
+++ b/pkg/kn/commands/config/config.go
@@ -1,0 +1,39 @@
+// Copyright © 2022 The Knative Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package configCmd
+
+import (
+	"github.com/spf13/cobra"
+	"knative.dev/client/pkg/kn/commands"
+)
+
+var cfgCmdExamples = ``
+
+var ConfigKeyNotFoundErrorMsg = "Could not find %q in the configuration file"
+
+func NewConfigCommand(p *commands.KnParams) *cobra.Command {
+	cfgCmd := &cobra.Command{
+		Use:     "config COMMAND",
+		Short:   "Manage Knative configuration file",
+		Aliases: []string{"cfg"},
+		Example: cfgCmdExamples,
+	}
+	cfgCmd.AddCommand(NewConfigGetCommand(p))
+	cfgCmd.AddCommand(NewConfigSetCommand(p))
+	cfgCmd.AddCommand(NewConfigShowCommand(p))
+	cfgCmd.AddCommand(NewConfigSinkMappingCommand(p))
+	cfgCmd.AddCommand(NewConfigChannelTypeMappingCommand(p))
+	return cfgCmd
+}
--- a/pkg/kn/commands/config/set_test.go
+++ b/pkg/kn/commands/config/set_test.go
@@ -1,0 +1,94 @@
+// Copyright © 2022 The Knative Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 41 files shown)


pkg/kn/commands/config/channel_type_mapping.go
```diff
@@ -0,0 +1,175 @@
...
+const (
+	ChannelTypeAliasNotFoundErrorMsg = "Channel type alias %q was not found in the configuration file"
+	ChannelTypeAliasExistsErrorMsg   = "Channel type alias %q already exists in the configuration file"
+)
+
+// NewConfigChannelTypeMappingCommand represents 'kn config channel-type-mapping' command
+func NewConfigChannelTypeMappingCommand(p *commands.KnParams) *cobra.Command {
+	channelTypeMappingCmd := &cobra.Command{
+		Use:   "channel-type-mapping",
+		Short: "Channel type mappings configuration settings",
```
maximilien, line 38: Why is the full description plural but the command is singular?

pkg/kn/commands/config/channel_type_mapping.go
```diff
@@ -0,0 +1,175 @@
...
+
+func addChannelTypeMappingFlags(flagSet *pflag.FlagSet, channelType *config.ChannelTypeMapping) {
+	flagSet.StringVar(&channelType.Kind, "kind", "", "The kind of Channel, such as InMemoryChannel or KafkaChannel, based on the default ConfigMap")
+	flagSet.StringVar(&channelType.Group, "group", "", "The API group of the Kubernetes resource")
+	flagSet.StringVar(&channelType.Version, "version", "", "The version of the Kubernetes resource")
+}
+
+func newChannelTypeUpdateCommand(_ *commands.KnParams) *cobra.Command {
+	channelType := config.ChannelTypeMapping{}
+	updateMapping := func(cmd *cobra.Command, args []string) ([]interface{}, error) {
```
maximilien, line 55: I wonder if for consistency sake whether these func definition should be placed outside or consistent order as the other commands in the client?

pkg/kn/commands/config/config.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return err
+	}
+	if _, err := fmt.Fprint(cmd.OutOrStdout(), string(b)); err != nil {
+		return err
+	}
+	return nil
+}
+
+func viperUpdater(viper *viper.Viper, cfgKey string) func([]interface{}) error {
```
maximilien, line 56: Can we have some comments to these... I know they are private but brief one liner on what function does helps reader who was not aware of that function. Tx

pkg/kn/commands/config/set_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"io/ioutil"
+	"testing"
+
+	"github.com/spf13/viper"
+)
+
+func TestConfigCommandSetSuccess(t *testing.T) {
+	cfgKey := "plugins.directory"
+	initValue := ""
+	newValue := "~/.kn/2/plugins"
```
maximilien, line 28: I'd use Go's temp io files instead of a fix name. Look at `https://pkg.go.dev/io/ioutil` and functions `TmpFile(...)` and `TmpDir(...)`

maximilien (commented):
Left a few feedback.

Thanks for contribution :)

boaz0:
thanks @maximilien your feedback was super helpful. I fixed some of your comments. Feel free to review them.

pkg/kn/commands/config/channel_type_mapping.go
dsimansk: +1 to keep the consistency with other commands. Ideally channel and sink mappings would live in their own sub-packages/directories and commands split into files.

```
./config/channel/<per_crud_cmds>.go
./config/sink/<per_crud_cmd>.go
```

dsimansk (changes requested):
Thanks!

Overall looks good. I have a few points to consider below.

And +1 to Dr. Max's structure comment. 


- Add a reference example(s) to config sub-command to demonstrate usage pattern (e.g. check `kn service update` etc. for the formatting & inspiration)
- Add more unit test to get ~80% coverage for new additions, ref. Codecov report
- Consider basic E2E tests (a follow-up PR later would be OK)

dsimansk:
I wonder if a bit unified UI for mappings wouldn't be nicer to group mappings together E.g.:

```
kn config mapping add <alias> --type {channel,sink} --gvr group:v1:KindOfResource

kn config mapping add kafka --type channel --gvr knative.messaging/v1alpha1:KafkaChannel
```

The `--gvr` flag is just another option to discuss. The separate flags are more inline with `kn` philosophy. However, it's pretty in-depth configuration to deal with.

boaz0:
@dsimansk added unit tests.
I will now work on the refactoring and move the crud functions to their own directory and change the UI as you suggested above.
Hopefully tomorrow I will update the PR with these changes.
Thanks.

pkg/kn/commands/config/channel_type_mapping.go
boaz0: @dsimansk & @maximilien DONE

pkg/kn/commands/config/channel/add.go
```diff
@@ -0,0 +1,33 @@
+package channel
```
dsimansk, line 1: Newly added files should have license headers.

boaz0: DONE! :+1: 
Thanks.

pkg/kn/commands/config/channel_type_mapping.go
```diff
@@ -0,0 +1,89 @@
...
+	channelType := config.ChannelTypeMapping{}
+	updateChannelTypeMapCmd := &cobra.Command{
+		Use:   "update",
+		Short: "Update channel type mapping by its alias",
+		Args:  cobra.ExactArgs(1),
+		RunE:  modifyList(channel.Update(&channelType), defaultChannelTypeUpdater),
+	}
+	addChannelTypeMappingFlags(updateChannelTypeMapCmd.Flags(), &channelType)
+	return updateChannelTypeMapCmd
+}
```
dsimansk, line 57: Wrt Max's and my previous comment. This whole function should move to `./channel/update.go` and then 

And then L36 would change to something like:
```
...AddCommand(NewChannelTypeUpdateCommand(p))
```
In addition I'd not hesitate to move whole `channel_type_mapping.go` to `channel` dir/package. In other words to move sub-command to its own package together with its actions.
E.g. see the structure of `kn broker` group https://github.com/knative/client/tree/main/pkg/kn/commands/broker.

pkg/kn/commands/config/show.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/spf13/viper"
+	"gopkg.in/yaml.v2"
+	"knative.dev/client/pkg/kn/commands"
+)
+
+// NewConfigSetCommand represents 'kn config show' command
+func NewConfigShowCommand(p *commands.KnParams) *cobra.Command {
+	showCmd := &cobra.Command{
+		Use:   "show",
+		Short: "Show configuration settings",
```
dsimansk, line 30: To be more aligned with rest of kn commands. I'd prefer `kn config describe`.

It might even have 2 modes:
- `kn describe describe <no-arg>` - return all settings
- `kn describe describe plugin.directory` - return specific value 

@maximilien @rhuss wdyt?

pkg/kn/commands/config/channel_type_mapping.go
```diff
@@ -0,0 +1,89 @@
...
+
+func newChannelTypeAddCommand(_ *commands.KnParams) *cobra.Command {
+	channelType := config.ChannelTypeMapping{}
+	addChannelTypeMapCmd := &cobra.Command{
+		Use:   "add",
+		Short: "Add a new channel type mapping",
```
dsimansk, line 63: In the same consistency sake. I'd like to get some opinions on `add` vs. `create`. I'd go with `create` for mappings.

@maximilien @rhuss

dsimansk:
@boaz0 I've added two comments around cmd naming of show/describe and add/create. I'd like to get @rhuss to chime in as well, but he won't be able around until next week. If you please don't mind to wait a bit for more feedback.

pkg/kn/commands/config/channel_type_mapping.go
boaz0: I see. No problem.

vyasgun (changes requested):
It would also be nice to have examples for the newly added commands.

pkg/kn/commands/config/sink/list.go
```diff
@@ -0,0 +1,48 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			items := []config.SinkMapping{}
+			if err := viper.UnmarshalKey(config.KeySinkMappings, &items); err != nil {
+				return err
+			}
+			b, err := yaml.Marshal(items)
+			if err != nil {
+				return err
+			}
+			if _, err := fmt.Fprint(cmd.OutOrStdout(), string(b)); err != nil {
```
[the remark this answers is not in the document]
dsimansk: That's a great point. IMO I'd be nicer to follow our usual table format per Gunjan's suggestion. With `--output` flag to JSON/YAML.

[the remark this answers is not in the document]
vyasgun: This should also address having an `--output` flag for specifying output type `yaml` in case the user wants to list it that way.

vyasgun:
Thanks for addressing the comments! 

The build is failing due to missing return statements in commands. While addressing that, can you also add examples to the commands like so?: 
https://github.com/knative/client/blob/main/pkg/kn/commands/broker/create.go#L47

Also, added another comment for using Resource Printer for the list sub-commands by adding `ListPrintFlags` to the them. For reference:
https://github.com/knative/client/blob/main/pkg/kn/commands/broker/list.go#L42

boaz0:
@vyasgun - I am not sure exactly what to do with `ListPrintFlags`. Should I wrap it in `Unstructured`? The link you gave is not really helping as I am trying to understand what interface BrokerList is implementing and there is a lot of magic (IMHO) so if you can provide a better example or give some useful details I will be grateful.

Thanks.

vyasgun:
@boaz0 I hope the following explanation and the eventtype command PR (specifically this commit: https://github.com/knative/client/pull/1598/commits/4077ab22a4195e595d6c7dc2f3cde95284303e8e) is helpful. Please let me know if unclear.

The [`printer pkg`](https://github.com/knative/client/blob/main/pkg/printers) is being used for printing. There are two types: machine readable (or generic) and human readable printer. The human readable printer is used by default and can be overridden with the `-o` flag. 
To register the handler and add the output related flags to the command, the struct `ListPrintFlags` is used which is defined as follows:
```
type ListPrintFlags struct {
	GenericPrintFlags  *genericclioptions.PrintFlags
	HumanReadableFlags *commands.HumanPrintFlags
	PrinterHandler     func(h hprinters.PrintHandler)
}
```
The PrinterHandler function here is responsible for registering the human readable printers, that is, map the appropriate printer functions to their respective types. To elaborate, take a look at the ListHandler function for eventtype list:

The [ListHandler function ](https://github.com/knative/client/pull/1598/commits/4077ab22a4195e595d6c7dc2f3cde95284303e8e#diff-f94645919027f1e90de023be7645cbfb4b68ac8d4b665abf700d435f82d6ef1cR87) defines the column definitions and then calls the function [TableHandler](https://github.com/knative/client/blob/main/pkg/printers/tablegenerator.go#L120) which registers the printer function + column definitions for the type of the object we are printing (channel-type-mapping and sink-mapping in the current PR's case). [Here's the code for this](https://github.com/knative/client/blob/main/pkg/printers/tablegenerator.go#L120).

To summarise, `kn eventtype list` will trigger a call to the ListPrintFlags' [`.Print`](https://github.com/knative/client/pull/1598/commits/4077ab22a4195e595d6c7dc2f3cde95284303e8e#diff-f94645919027f1e90de023be7645cbfb4b68ac8d4b665abf700d435f82d6ef1cR74) method. This kickstarts the process of calling the PrintHandler (the one we populated in the `ListPrintFlags` object) which registers the printer functions and then calls the appropriate printer based on the type of object. (EventtypeList in the example)

boaz0:
@vyasgun I did follow what you were saying but I still need to implement runtime.Object - which I am not sure what DeepCopyObject should do - is there a wrapper that does it for me? Or is there a way to avoid this?
Thanks.

rhuss:
Sorry for the delay, I plan to review the PR next week.

boaz0:
ping @rhuss any updates?

boaz0:
ping @rhuss @dsimansk any updates? :smiley:

dsimansk:
@boaz0 hi there, sorry for super late response. 

Dialing back on the listing options, since that would require to refactor our Config as a Kubernetes resource.

I'd like to see at least renaming of `show` command to `describe`. https://github.com/knative/client/pull/1556#discussion_r896667207. To get  this PR rolling again.

Wrt other UI suggestions, I'd like to reconsider the following proposal: 
https://github.com/knative/client/pull/1556#issuecomment-1152238773

/cc @rhuss

PR #1561, opened 2022-01-06
Added completion for route name

### LINKED ISSUE #1373: Context sensitive Bash completion

vyasgun:
## Description
Added completion function for autocompleting route name in `kn route describe` command

## Changes

* Added completion function to `route describe` subcommand
* Added tests for the above

## Reference

Fixes #1373


dsimansk (approved):
/approve
/lgtm

PR #1562, opened 2022-01-06
Added completion for domain command

### LINKED ISSUE #1373: Context sensitive Bash completion

vyasgun:
## Description
Added auto completion function for domain command

## Changes

* Added auto completion function for domain command
* Added unit tests for the above

### DIFF
--- a/pkg/kn/commands/completion_helper.go
+++ b/pkg/kn/commands/completion_helper.go
@@ -29,10 +29,11 @@ type completionConfig struct {
 
 var (
 	resourceToFuncMap = map[string]func(config *completionConfig) []string{
-		"service":  completeService,
-		"revision": completeRevision,
 		"broker":   completeBroker,
+		"domain":   completeDomain,
+		"revision": completeRevision,
 		"route":    completeRoute,
+		"service":  completeService,
 	}
 )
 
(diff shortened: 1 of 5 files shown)


pkg/kn/commands/completion_helper.go
```diff
@@ -30,6 +30,7 @@ type completionConfig struct {
 var (
 	resourceToFuncMap = map[string]func(config *completionConfig) []string{
 		"service": completeService,
+		"domain":  completeDomain,
```
dsimansk, line 33: Since that's going to be the last conflict. Maybe, I'd order the list alphabetically by the resource name for the better overview.

maximilien (commented):
Wondering if we really should have completion helper for `delete`? 

Also, is this worth adding some docs to README or perhaps a simple ascii recoding?

vyasgun:
@maximilien `delete` needs a resource name so I think there should be completion suggestions.

dsimansk:
> Wondering if we really should have completion helper for `delete`?
> 
> Also, is this worth adding some docs to README or perhaps a simple ascii recoding?

IMO, it should be consistent across the commands, therefore once there's `NAME` arg required such command should have completion enabled.

dsimansk (approved):

PR #1567, opened 2022-01-13
Adding completion for remaining Eventing resources

### LINKED ISSUE #1373: Context sensitive Bash completion

vyasgun:
## Description
Context specific autocompletion for eventing resources

## Changes

* Added the ValidArgs function for the following resources
    * triggers
    * source:
        * ping
        * sinkbinding
        * apiserver
        * container
    * channel
    * subscription  

## Reference

Fixes #1373

### DIFF
--- a/pkg/kn/commands/completion_helper.go
+++ b/pkg/kn/commands/completion_helper.go
@@ -18,8 +18,6 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
-	clientv1beta2 "knative.dev/client/pkg/sources/v1beta2"
-	sourcesv1beta2 "knative.dev/eventing/pkg/client/clientset/versioned/typed/sources/v1beta2"
 )
 
 type completionConfig struct {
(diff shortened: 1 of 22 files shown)


pkg/kn/commands/completion_helper.go
```diff
@@ -231,3 +240,206 @@ func completeDomain(config *completionConfig) (suggestions []string) {
...
+	if len(config.args) != 0 {
+		return
+	}
+	namespace, err := config.params.GetNamespace(config.command)
+	if err != nil {
+		return
+	}
+
+	var pingSourcesClient clientv1beta2.KnPingSourcesClient
+	if config.params.NewSourcesV1beta2Client == nil {
```
dsimansk, line 359: Is there any specific reason to initialize only `SourcesV1beta2Client`, but other completion functions don't follow this pattern?

dsimansk: Ehm, I see...it isn't properly done. Please fix that by adding it here:
https://github.com/knative/client/blob/main/pkg/kn/commands/types.go#L69

vyasgun: Thanks, fixed.

dsimansk (approved):
Thanks!

/approve
/lgtm

PR #1568, opened 2022-01-13
Fix for panic during channel list

vyasgun:
## Description
`kn channel list` was panicking when a channel in the list didn't have `.Spec.ChannelTemplate` field set to a non-nil value.

## Changes

* Added a nil check

## Reference

Fixes #


dsimansk (approved):
Thanks!

/approve
/lgtm

PR #1572, opened 2022-01-20
Remove k8s cluster version pin

dsimansk:
## Description

It's affecting latest integrations tests with net-istio failing to deploy due too low cluster version.

E.g.: https://github.com/knative/client/pull/1571

## Changes

* Remove k8s cluster version pin

/cc @rhuss @vyasgun


rhuss (commented):
/approve
/lgtm

PR #1575, opened 2022-01-24
Fix for file not found error message discrepancy in windows

vyasgun:
## Description
Added file not found error check for windows specific message

## Changes

* Added windows-specific string for non-existent file tests

## Reference

Fixes #

### DIFF
--- a/test/e2e/service_file_test.go
+++ b/test/e2e/service_file_test.go
@@ -140,5 +140,6 @@ func fileNotFoundErrorCheck(out test.KnRunResult, filePath string) cmp.Compariso
 	if result() == cmp.ResultSuccess {
 		return result
 	}
+	// Check for Windows specific error message in case file is not found
 	return util.ContainsAllIgnoreCase(out.Stderr, "system", "cannot", "find", "file", "specified", filePath)
 }
--- a/test/e2e/service_file_test.go
+++ b/test/e2e/service_file_test.go
@@ -24,6 +24,8 @@ import (
 	"path/filepath"
 	"testing"
 
+	"gotest.tools/v3/assert/cmp"
+
 	"gotest.tools/v3/assert"
 
 	"knative.dev/client/lib/test"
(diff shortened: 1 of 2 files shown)


dsimansk:
Pls rerun `./hack/build.sh -c` to regenerate code formatting.

test/e2e/service_file_test.go
```diff
@@ -124,11 +125,19 @@ func serviceCreateFromFile(r *test.KnRunResultCollector, serviceName, filePath s
...
 
 func serviceCreateFromFileNameMismatch(r *test.KnRunResultCollector, serviceName, filePath string) {
 	out := r.KnTest().Kn().Run("service", "create", serviceName, "--filename", filePath)
 	r.AssertError(out)
 	assert.Check(r.T(), util.ContainsAllIgnoreCase(out.Stderr, "provided", "'"+serviceName+"'", "name", "match", "from", "file"))
 }
+
+func fileNotFoundErrorCheck(out test.KnRunResult, filePath string) cmp.Comparison {
+	result := util.ContainsAllIgnoreCase(out.Stderr, "no", "such", "file", "directory", filePath)
+	if result() == cmp.ResultSuccess {
```
dsimansk, line 139: I wonder if that shouldn't be rather `== windows` check.

dsimansk:
@vyasgun should be used here as well. https://github.com/knative/client/blob/main/test/e2e/service_import_test.go#L93

vyasgun:
@dsimansk  I called the same function in the service_import_test.go file

test/e2e/service_file_test.go
vyasgun: I added a comment for clarity

dsimansk:
> @vyasgun should be used here as well. https://github.com/knative/client/blob/main/test/e2e/service_import_test.go#L93

I see now, sorry my bad. I kind of haven't noticed it's another file. :)

test/e2e/service_file_test.go
dsimansk: OK, thanks!

PR #1593, opened 2022-01-31
Update README.md

xtreme-sameer-vohra:
## Description
Clarify that kn plugins need `kn-` prefix to be picked up by kn. 
For example, the `func` plugin is only dected by kn if it is rename as `kn-func`


dsimansk (commented):
Thanks!

/approve
/lgtm

PR #1625, opened 2022-03-09
Fix Serving & Eventing version fetch in e2e tests

dsimansk:
## Description

Based on the idea of #1623. 

/cc @kvmware 

## Changes

* Fix Serving & Eventing version fetch in e2e tests
*
*

## Reference

Fixes #

PR #1627, opened 2022-03-10
Infer versions for E2E tests from current branch

dsimansk:
## Description

Follow-up on #1625 to introduce upper limit version inferred from current branch. The motivation is to get rid off a manual release step to update the version string, and also keep floating version limited to correct context.

## Changes

* Infer versions for E2E tests from current branch

### DIFF
--- a/test/common.sh
+++ b/test/common.sh
@@ -66,9 +66,22 @@ function knative_setup() {
     install_istio
   fi
 
-  local serving_version=${KNATIVE_SERVING_VERSION:-latest}
-  header "Installing Knative Serving (${serving_version})"
+  # Serving and Eventing 'latests' is based on branch context
+  # On `main` it means 'nightly' manifests
+  # On `release-*` it means corresponding `release-*` manifests
+  # The described behavior is achieved through hack/library.sh start_latest_knative_serving()
+  local serving_version="latest"
+  local eventing_version="latest"
 
+  # PR check variant triggering presubmit-integration-tests-latest-release
+  # should check current client main branch code on latest avaible release
+  # e.g. PR to 'main' on Serving and Eventing 1.3.0
+  if [[ -n ${LATEST_RELEASE} ]]; then
+    serving_version="$(get_latest_release_version "serving")"
+    eventing_version="$(get_latest_release_version "eventing")"
+  fi
+
+  header "Installing Knative Serving (${serving_version})"
   if [ "${serving_version}" = "latest" ]; then
     start_latest_knative_serving
   else
@@ -80,9 +93,7 @@ function knative_setup() {
       --type merge -p '{"data": {"ingress.class":"'${INGRESS_CLASS}'"}}'
   fi
 
-  local eventing_version=${KNATIVE_EVENTING_VERSION:-latest}
   header "Installing Knative Eventing (${eventing_version})"
-
   if [ "${eventing_version}" = "latest" ]; then
     start_latest_knative_eventing
     start_latest_eventing_sugar_controller
@@ -96,4 +107,24 @@ function knative_setup() {
 function test_setup() {
   echo ">> Uploading test images..."
   ${REPO_ROOT_DIR}/test/upload-test-images.sh || return 1
+}
+
+# Retrieve latest version from given Knative repository tags
+# On 'main' branch the latest released version is returned
+# On 'release-x.y' branch the latest patch version for 'x.y.*' is returned
+# Similar to hack/library.sh get_latest_knative_yaml_source()
+function get_latest_release_version() {
+    local repo_name="$1"
+    local major_minor=""
+    if is_release_branch; then
+      local branch_name="$(current_branch)"
+      major_minor="${branch_name##release-}"
+    fi
+    local version ="$(git ls-remote --tags --ref https://github.com/knative/${repo_name}.git \
+      | grep ${major_minor} \
+      | cut -d '-' -f2 \
+      | cut -d 'v' -f2 \
+      | sort -Vr \
+      | head -n 1)"
+    echo "${version}"
 }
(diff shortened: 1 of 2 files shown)


test/common.sh
```diff
@@ -66,9 +66,22 @@ function knative_setup() {
...
-  header "Installing Knative Serving (${serving_version})"
+  # Serving and Eventing 'latests' is based on branch context
+  # On `main` it means 'nightly' manifests
+  # On `release-*` it means corresponding `release-*` manifests
+  # The described behavior is achieved through hack/library.sh start_latest_knative_serving()
+  local serving_version="latest"
+  local eventing_version="latest"
+
+  # PR check variant triggering presubmit-integration-tests-latest-release
+  # should check current client main branch code on latest avaible release
```
rhuss, line 77:
```suggestion
  # should check current client main branch code on latest available release
```

test/common.sh
```diff
@@ -66,9 +66,22 @@ function knative_setup() {
...
+  # e.g. PR to 'main' on Serving and Eventing 1.3.0
+  if [[ -n ${LATEST_RELEASE} ]]; then
```
rhuss, line 79:
```suggestion
  if [ -n "${LATEST_RELEASE}" ]; then
```

Just a proposal to be POSIX compliant, but since we already went done the Bash rabbit hole, don't mind to ignore this proposal.

rhuss: But it would be consistent to the usage in the rest of the script

test/common.sh
```diff
@@ -66,9 +66,22 @@ function knative_setup() {
     install_istio
   fi
 
-  local serving_version=${KNATIVE_SERVING_VERSION:-latest}
-  header "Installing Knative Serving (${serving_version})"
+  # Serving and Eventing 'latests' is based on branch context
```
rhuss, line 69:
```suggestion
  # Serving and Eventing 'latest' is based on branch context
```

test/presubmit-integration-tests-latest-release.sh
```diff
@@ -17,7 +17,6 @@
 # This script is used in Knative/test-infra as a custom prow job to run the
 # integration tests against Knative Serving / Eventing of a specific version.
 
-export KNATIVE_SERVING_VERSION="`git ls-remote --tags --ref https://github.com/knative/serving.git | cut -d '-' -f2 | cut -d 'v' -f2 | sort -Vr | head -n 1`"
-export KNATIVE_EVENTING_VERSION="`git ls-remote --tags --ref https://github.com/knative/eventing.git | cut -d '-' -f2 | cut -d 'v' -f2 | sort -Vr | head -n 1`"
+readonly export LATEST_RELEASE="true"
```
rhuss, line 20: what would happen if you set `LATEST_RELEASE="false"` ? :-)

test/common.sh
rhuss: Or something like to also check the value of the env var:

```
if [ "xtrue" == x"${LATEST_RELEASE}" ]; then
```

test/presubmit-integration-tests-latest-release.sh
dsimansk: It's left open for next generations to interpret. :)

dsimansk: A dedicated flag seemed too much, but I can go that way. 

Or maybe?

```
LATEST_RELEASE="true" $(dirname $0)/presubmit-tests.sh --integration-tests
```

test/common.sh
```diff
@@ -97,3 +108,25 @@ function test_setup() {
...
+      local branch_name
+      branch_name="$(current_branch)"
+      major_minor="${branch_name##release-}"
+    fi
+    local version
+    version="$(git ls-remote --tags --ref https://github.com/knative/"${repo_name}".git \
+      | grep "${major_minor}" \
+      | cut -d '-' -f2 \
+      | cut -d 'v' -f2 \
+      | sort -Vr \
```
rhuss, line 129: Are you sure that `-V` works on every platform's sort ? (cool find btw, didn't know the option before)

test/common.sh
```diff
@@ -97,3 +108,25 @@ function test_setup() {
...
+      | head -n 1)"
+    echo "${version}"
```
rhuss, line 131: you could just skip the version variable by directly running the git-pipeline without executing it in a subshell

rhuss (commented):
looks good with some minor comment.

test/common.sh
dsimansk: It's used in release.sh script on a few occasions, that's my source. :)
I'll check the compatibility. However, the script isn't meant to be used outside of our CI.

test/common.sh
dsimansk: Well `shellcheck` complained about masking return value and legacy backthick.

test/common.sh
dsimansk: The `sort` seemed to work fine even without `-V` flag for the `semver` scheme.

dsimansk: And `-V` it isn't posix compatible of course, it's a bit newer. :)

rhuss: > The sort seemed to work fine even without -V flag for the semver scheme.

Nope, I don't think so for larger patcher version numbers:

```
echo -n "1.2.3\n1.2.10\n1.2.5" | sort
1.2.10
1.2.3
1.2.5
```

but I'm totally fine with using `-V` if it is already used (and working) elsewhere.

test/common.sh
rhuss: yeah, was just a cosmetic comment that you don't need backticks at all, just call git without a variable (it will echo the result anyway). but if this works better here, then I'm fine.

PR #1643, opened 2022-04-11
Add timeout option to service create

### LINKED ISSUE #1181: kn needs the ability top set a default timeout for the service
### Feature request

kn create needs a flag for setting the default timeout for the service

### Use case

I want to create my services with a default timeout of 10 seconds rather than the standard 300 - currently I can't create a service with the timeout of 10 secs and need to post-create change it (in this case using the OpenShift UI/YAML editor) which creates a new revision

There are flags for most things when creating a service but this one (timeoutSeconds) seems to be missing

/kind proposal

vyasgun:
## Description

Add `--timeout` option which will set the field, `timeoutInSeconds` in service spec. 

## Changes

* Added the flag

## Reference

Fixes #1181

### DIFF
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -220,7 +220,7 @@ func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
 			"any number of times to set multiple labels.")
 	p.markFlagMakesRevision("label")
 	command.Flags().Int64Var(&p.TimeoutSeconds, "timeout", config.DefaultRevisionTimeoutSeconds,
-		"duration in seconds that the request routing layer will wait for a request delivered to a "+""+
+		"Duration in seconds that the request routing layer will wait for a request delivered to a "+""+
 			"container to begin replying")
 	p.markFlagMakesRevision("timeout")
 }
--- a/pkg/kn/commands/service/configuration_edit_flags.go
+++ b/pkg/kn/commands/service/configuration_edit_flags.go
@@ -54,6 +54,7 @@ type ConfigurationEditFlags struct {
 	AnnotationsRevision []string
 	ClusterLocal        bool
 	ScaleInit           int
+	TimeoutSeconds      int64
 
 	// Preferences about how to do the action.
 	LockToDigest         bool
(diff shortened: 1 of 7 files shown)


pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -217,6 +219,10 @@ func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
 		"Labels to set for both Service and Revision. name=value; you may provide this flag "+
 			"any number of times to set multiple labels.")
 	p.markFlagMakesRevision("label")
+	command.Flags().Int64Var(&p.TimeoutSeconds, "timeout", config.DefaultRevisionTimeoutSeconds,
+		"duration in seconds that the request routing layer will wait for a request delivered to a "+""+
```
dsimansk, line 223:
```suggestion
		"Duration in seconds that the request routing layer will wait for a request delivered to a "+""+
```

pkg/serving/config_changes_test.go
```diff
@@ -151,6 +151,17 @@ func TestPinImageToDigestNilContainerStatus(t *testing.T) {
 	assert.NilError(t, err)
 }
 
+func TestPinImageToDigestNilContainerSpec(t *testing.T) {
```
dsimansk, line 154: Does this test belong to this PR?

vyasgun: I was trying to improve coverage to prevent the codecov project failure. Removed this test

dsimansk: IMO adding unit test to cover this flag in create and update operations would help with coverage. Since the common flag doesn't have its own test file, but are rather tested with those ops.

pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -217,6 +219,10 @@ func (p *ConfigurationEditFlags) AddCreateFlags(command *cobra.Command) {
 		"Labels to set for both Service and Revision. name=value; you may provide this flag "+
 			"any number of times to set multiple labels.")
 	p.markFlagMakesRevision("label")
+	command.Flags().Int64Var(&p.TimeoutSeconds, "timeout", config.DefaultRevisionTimeoutSeconds,
+		"Duration in seconds that the request routing layer will wait for a request delivered to a "+""+
+			"container to begin replying")
+	p.markFlagMakesRevision("timeout")
```
dsimansk, line 225: Double checking the flag, it should probably go into `addSharedFlags` functions to be present in `create, update, apply`.

vyasgun: Done

PR #1644, opened 2022-04-11
Added pull-policy flag to service

### LINKED ISSUE #578: Support a flag to set ImagePullPolicy

vyasgun:
## Description

Added `--pull-policy` option which will set the ImagePullPolicy field in the Knative service. Valid options are: Never | Always | IfNotPresent

## Changes

* Added the flat to set image pull policy

## Reference

Fixes #578


pkg/serving/config_changes.go
```diff
@@ -226,3 +240,7 @@ func updateAnnotations(annotations map[string]string, toUpdate map[string]string
 	}
 	return nil
 }
+
+func isValidPullPolicy(policy string) bool {
+	return policy == string(v1.PullAlways) || policy == string(v1.PullNever) || policy == string(v1.PullIfNotPresent)
+}
```
dsimansk, line 246: Does it need to be case sensitive (capital letter and camel case in the last `IfNotPresent`)? `--pull-policy always` seems reasonable.
I'd go with something like `getPullPolicy` to check all lower/upper and return `v1.PullPolicy` or empty/nil otherwise?

The other option would be 3 standalone `--pull-policy-*`, but that's too much clutter probably. :)

vyasgun: Yes, that makes sense. I made the change to make the flag case insensitive.

pkg/serving/config_changes.go
```diff
@@ -226,3 +254,11 @@ func updateAnnotations(annotations map[string]string, toUpdate map[string]string
 	}
 	return nil
 }
+
+func isValidPullPolicy(policy string) bool {
+	validPolicies := []string{string(v1.PullAlways), string(v1.PullNever), string(v1.PullIfNotPresent)}
+	for i := range validPolicies {
+		validPolicies[i] = strings.ToLower(validPolicies[i])
+	}
+	return sliceutil.Contains(validPolicies, strings.ToLower(policy))
```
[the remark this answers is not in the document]
vyasgun: Yes, it is. Thanks!

pkg/kn/flags/podspec.go
```diff
@@ -129,6 +130,9 @@ func (p *PodSpecFlags) AddFlags(flagset *pflag.FlagSet) []string {
 	flagset.VarP(&p.Image, "image", "", "Image to run.")
 	flagNames = append(flagNames, "image")
 
+	flagset.StringVar(&p.ImagePullPolicy, "pull-policy", "",
```
dsimansk, line 133: You can try to add unit test into resolver flags test that would improve something. :)

pkg/serving/config_changes.go
```diff
@@ -215,6 +216,32 @@ func UpdateRevisionTemplateAnnotation(template *servingv1.RevisionTemplateSpec,
...
+		return fmt.Errorf("no container found in spec")
+	}
+	if !isValidPullPolicy(imagePullPolicy) {
+		return fmt.Errorf("invalid --pull-policy %s. Valid arguments (case insensitive): Always | Never | IfNotPresent", imagePullPolicy)
+	}
+	template.Spec.Containers[idx].ImagePullPolicy = actualPolicy(imagePullPolicy)
+	return nil
+}
+
+func actualPolicy(policy string) v1.PullPolicy {
```
dsimansk, line 232: I'd rather call it `getPullPolicy`, but it's not a bit deal. :)

pkg/kn/flags/podspec_test.go
```diff
@@ -70,13 +70,14 @@ func TestPodSpecResolve(t *testing.T) {
 		"--port", "8080", "--limit", "cpu=1000m", "--limit", "memory=1024Mi",
 		"--cmd", "/app/start", "--arg", "myArg1", "--service-account", "foo-bar-account",
 		"--mount", "/mount/path=volume-name", "--volume", "volume-name=cm:config-map-name",
-		"--env-from", "config-map:config-map-name", "--user", "1001"}
+		"--env-from", "config-map:config-map-name", "--user", "1001", "--pull-policy", "always"}
```
dsimansk, line 73: Can you please add a negative test for error as well?

dsimansk:
Thanks!

/approve
/lgtm

PR #1645, opened 2022-04-11
Added flag to configure wait-window between intermediate errors durin…

### LINKED ISSUE #1023: Make error-window when waiting for a service to be ready configurable
_The use case described below in this description would be supported by a new option as described [below](https://github.com/knative/client/issues/1023#issuecomment-759986466)_

-------

> /kind question

we are running with cluster-autoscaler, and https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-configure-overprovisioning-with-cluster-autoscaler to put some low-priority pause pod in the cluster.
When the worker node cpu/memory are not nearly 99%  (30% of them are occupied by the pause pods),   we create a knative service with kn, and got error:
```
kn service create hello --image xxx --wait-timeout 300 --env TARGET=revision1 
Creating service 'hello' in namespace:
  0.380s The Route is still working to reflect the latest desired specification.
  1.253s Configuration "hello" is waiting for a Revision to become ready.
  4.249s Revision "hello-xxx" failed with message: 0/15 nodes are available: 1 Insufficient memory, 14 Insufficient cpu..
  5.077s Configuration "hello" does not have any ready Revision.
Error: RevisionFailed: Revision "hello-xxx" failed with message: 0/15 nodes are available: 1 Insufficient memory, 14 Insufficient cpu..
```

I checked with k8s scheduler team that the pod schedule will happen in 2 stage,  the 1st pod placement attempt failed and the scheduler preempted low-priority Pods; then the 2nd pod placement attempt succeed.    So as a result , the final knative service reconcile succeed. 

But  when using kn client,  the end-user got the scary `failed` msg ..    If the end-user don't have enough knowledge for the k8s reconcile,  he/she will be frightened. 

Another case is from some race condition case in knative itself.   refer to : 
https://github.com/knative/serving/issues/8675
When the error is thrown out from kn client,  the ksvc just created for 4 seconds.   Later on, with more reconcile,  the ksvc is finally ready. 

So, I am wondering whether there are a better idea for watch to reduce these intermittent errors since reconcile is a designed behaviour of k8s. 
maybe just adding a shorter wait time to see whether any condition change for the next reconcile?

vyasgun:
…g service create

## Description
Added the flag `--wait-window` for configuring the default error window between intermediate false Ready conditions when a service is coming up

## Changes

* Added the flag
* Changed existing signature for Wait function to accommodate the flag value

## Reference

Fixes #1023


pkg/serving/v1/client.go
```diff
@@ -89,7 +89,7 @@ type KnServingClient interface {
 
 	// Wait for a service to become ready, but not longer than provided timeout.
 	// Return error and how long has been waited
-	WaitForService(ctx context.Context, name string, timeout time.Duration, msgCallback wait.MessageCallback) (error, time.Duration)
+	WaitForService(ctx context.Context, name string, timeout, errorWindow time.Duration, msgCallback wait.MessageCallback) (error, time.Duration)
```
dsimansk, line 92: (Note for the future :))
I wonder if it was worth to introduce simple waitConfig structure to represent those timeout values. The number of arguments is still manageable, until it won't grow too much. :) Maybe an overall refactor of wait would help, but that's certainly outside of the PR's scope.

vyasgun: Thanks, I added the waitConfig struct for now

pkg/kn/commands/service/update.go
```diff
@@ -135,7 +136,11 @@ func NewServiceUpdateCommand(p *commands.KnParams) *cobra.Command {
 			if waitFlags.Wait && targetFlag == "" {
 				fmt.Fprintf(out, "Updating Service '%s' in namespace '%s':\n", args[0], namespace)
 				fmt.Fprintln(out, "")
-				err := waitForService(cmd.Context(), client, name, out, waitFlags.TimeoutInSeconds)
+				wconfig := clientservingv1.WaitConfig{
+					Timeout:     time.Duration(waitFlags.TimeoutInSeconds) * time.Second,
+					ErrorWindow: time.Duration(waitFlags.ErrorWindowInSeconds) * time.Second,
+				}
+				err := waitForService(cmd.Context(), client, name, out, wconfig)
```
dsimansk, line 143: `service update` cmd is missing the new flag?

``` 
waitFlags.AddErrorWindow(serviceCreateCommand, "service", "ready")
```

pkg/kn/commands/wait_flags.go
```diff
@@ -50,3 +52,10 @@ func (p *WaitFlags) AddConditionWaitFlags(command *cobra.Command, waitTimeoutDef
 	timeoutUsage := fmt.Sprintf("Seconds to wait before giving up on waiting for %s to be %s.", what, until)
 	command.Flags().IntVar(&p.TimeoutInSeconds, "wait-timeout", waitTimeoutDefault, timeoutUsage)
 }
+
+// AddErrorWindow will add a flag to configure the error window, that is,
+// the time to wait between intermediate false ready conditions
+func (p *WaitFlags) AddErrorWindow(command *cobra.Command, what, until string) {
```
dsimansk, line 58: IMHO it doesn't need to be a standalone add functions. It seems to me like a `wait` family option.

pkg/serving/v1/client.go
```diff
@@ -48,6 +48,11 @@ import (
 // or an error
 type ServiceUpdateFunc func(origService *servingv1.Service) (*servingv1.Service, error)
 
+type WaitConfig struct {
+	Timeout     time.Duration
+	ErrorWindow time.Duration
+}
+
```
dsimansk, line 55: Thanks! :100:

pkg/kn/commands/service/update.go
dsimansk: It would be resolved by my latter comment, once `--timeout` is part of `AddConditionWaitFlags`.

dsimansk (approved):
Thanks!

/approve
/lgtm

PR #1653, opened 2022-04-13
Added scale-metric flag to configure metric name

### LINKED ISSUE #1205: Add support to allow scaling metric type (concurrency, rps, cpu, memory)
### Feature request
Currently,  I can't set the autoscaler metric with rps.    How about add a new option `--autoscaler-metric` ?
### Use case

Currently, I can't achieve below knative creation with `kn`
```
cat <<EOF | kubectl apply -n default -f -
kind: Service
metadata:
  name: ceperftesthttp-rps
spec:
  template:
    metadata:
      annotations:
        autoscaling.knative.dev/metric: rps
        autoscaling.knative.dev/target: "100"
        autoscaling.knative.dev/minScale: "3"
        autoscaling.knative.dev/maxScale: "50"        
    spec:
      containerConcurrency: 100
      containers:
        - image: .....
EOF
```

### UI Example

how about using 
`kn service create --autoscaler-metric rps` for above?  The default value for `--autoscaler-metric` should be `concurrency`

vyasgun:
## Description

Added `--scale-metric` flag to configure the metric name

## Changes

* Added flag
* Added unit test
* Validation already present in serving

## Reference

Fixes #1205


pkg/kn/commands/service/configuration_edit_flags.go
```diff
@@ -93,6 +94,9 @@ func (p *ConfigurationEditFlags) addSharedFlags(command *cobra.Command) {
 	command.Flags().IntVar(&p.MaxScale, "scale-max", 0, "Maximum number of replicas.")
 	p.markFlagMakesRevision("scale-max")
 
+	command.Flags().StringVar(&p.ScaleMetric, "scale-metric", "", "Metric the PodAutoscaler should scale on.")
```
dsimansk, line 97: Can we add a few examples here, like `rps` or `concurrency` maybe? Just for better illustration.

dsimansk: Wdyt? I'd like to highlight it's "name": `Name of metric to scale on by the PodAutoscaler.`

vyasgun: Thanks, I just updated the flag usage. I also updated the usage for `--scale-target` flag. Does it read better now?

dsimansk:
Thanks!

/approve
/lgtm

PR #1654, opened 2022-04-14
Upgrade Go to 1.17

dsimansk:
## Description

Align with knative/pkg and rest of the Knative components. We're building on 1.17 toolchain for a while, it should be relatively safe.

## Changes

* Upgrade Go to 1.17


navidshaikh (approved):
/lgtm

PR #1660, opened 2022-04-19
test: use `T.TempDir` to create temporary test directory

Juneezee:
## Description

A testing cleanup. 

This pull request replaces `ioutil.TempDir` with `t.TempDir`. We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete. 

Reference: https://pkg.go.dev/testing#T.TempDir

```go
func TestFoo(t *testing.T) {
	// before
	tmpDir, err := ioutil.TempDir("", "")
	require.NoError(t, err)
	defer require.NoError(os.RemoveAll(tmpDir))

	// now
	tmpDir := t.TempDir()
}
```

## Changes

* Replace `ioutil.TempDir` with `t.TempDir`
* Remove redundant `os.RemoveAll(tmpDir)` since the temporary directory will be removed automatically once the test complete.

/lint

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -36,6 +36,10 @@
 | Added subpath functionality to --mount flag
 | https://github.com/knative/client/pull/1655[#1655]
 
+| 🐣
+| Use `T.TempDir` to create temporary test directory
+| https://github.com/knative/client/pull/1660[#1660]
+
 |===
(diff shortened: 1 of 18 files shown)


rhuss:
Thanks, looks good to me! Let's get that in right after we did the 1.4 release.

/approve
/lgtm
/hold

PR #1661, opened 2022-04-25
Mark --inject-broker flag as deprecated

### LINKED ISSUE #1629: Change to Sugar Controller will impact inject Broker functionality option of "Kn trigger create"
Once https://github.com/knative/eventing/issues/5958 is addressed, Knative users will be able to configure the LabelSelector the Sugar reconciler will use to make injection decisions rather than the current behaviour which is static.

This will impact 
`--inject-broker      Create new broker with name default through common annotation`
option of `kn trigger create`

The LabelSelector will be populated in the sugar config file and would be the new source of truth.

dsimansk:
## Description

Per Eventing change in Sugar controller config, user need to enable the behavior.

@rhuss maybe we should considere adding label flag to trigger creation instread of hardcoded name. 

## Changes

* Mark --inject-broker flag as deprecated

## Reference

Fixes #1629 

```console
➜  client git:(pr/deprecate-inject-broker) kn trigger create broker-trigger --inject-broker --sink ksvc:broker-display --filter type=dev.knative.sources.ping
Flag --inject-broker has been deprecated, effective since 1.4 and will be removed in 1.6 release. 
Please refer to 'Automatic Broker Creation' section for configuration options, https://knative.dev/docs/eventing/sugar/#automatic-broker-creation.
```

### DIFF
--- a/pkg/kn/commands/trigger/update_flags.go
+++ b/pkg/kn/commands/trigger/update_flags.go
@@ -51,6 +51,12 @@ func (f *TriggerUpdateFlags) GetUpdateFilters() (map[string]string, []string, er
 //Add is to set parameters
 func (f *TriggerUpdateFlags) Add(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&f.Broker, "broker", "default", "Name of the Broker which the trigger associates with.")
+	// The Sugar controller was integrated into main Eventing controller. With that the default behavior was changed as well.
+	// Users need to configure 'Automatic Broker Creation' per linked docs.
+	// Deprecated in 1.4, remove in 1.6.
 	cmd.Flags().BoolVar(&f.InjectBroker, "inject-broker", false, "Create new broker with name default through common annotation")
+	cmd.Flags().MarkDeprecated("inject-broker", "effective since 1.4 and will be removed in 1.6 release. \n"+
+		"Please refer to 'Automatic Broker Creation' section for configuration options, "+
+		"https://knative.dev/docs/eventing/sugar/#automatic-broker-creation.")
 	cmd.Flags().StringSliceVar(&f.Filters, "filter", nil, "Key-value pair for exact CloudEvent attribute matching against incoming events, e.g type=dev.knative.foo")
 }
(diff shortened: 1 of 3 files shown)


pkg/kn/commands/trigger/update_flags.go
```diff
@@ -51,6 +51,12 @@ func (f *TriggerUpdateFlags) GetUpdateFilters() (map[string]string, []string, er
 //Add is to set parameters
 func (f *TriggerUpdateFlags) Add(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&f.Broker, "broker", "default", "Name of the Broker which the trigger associates with.")
+	// The Sugar controller was integrated into main Eventing controller. With that the default behavior was changed as well.
+	// Users need to configure 'Automatic Broker Creation' per linked docs.
+	// Deprecated in 1.4, remove in 1.6.
 	cmd.Flags().BoolVar(&f.InjectBroker, "inject-broker", false, "Create new broker with name default through common annotation")
+	cmd.Flags().MarkDeprecated("inject-broker", "effective since 1.4 and will be removed in 1.6 release. \n"+
+		"Please refer to 'Automatic Broker Creation' section for configuration options, "+
+		"https://knative.dev/docs/eventing/sugar/#automatic-broker-creation.")
```
rhuss, line 60: Maybe add the recommendation to create a Broker directly via `kn broker create`, because that is what a user can do. I think we shouldn't really think about automatic broker creation on this level.

So maybe this a description:

```
Automatic broker creation is deprecated and might not work anymore for Knative >= 1.4 (depending on server-side configuration). Please manage brokers directly with the `kn broker` commands.
```

I wouldn't mention that this is deprecated since client 1.4, I think its good enough to know that it's deprecated for the currently used version ;-). Also wouldn't put in links in the help message if not absolutely needed as we can't be sure that this link will stay stable over time.

dsimansk: Sure, I was thinking about how stable those links might be. 

Does it look better? (the prefix is part of cobra `MarkDeprecated` function)

```console
Flag --inject-broker has been deprecated, automatic broker creation might not work anymore for Knative >= 1.4 (depending on server-side configuration). 
Please manage brokers directly with the `kn broker` commands.
```

rhuss: looks good, thanks !

/approve
/lgtm

Issue #257, opened 2019-07-11
OCI image created with ko in a release misses version informations


rhuss:
## In what area(s)?

Release 

## What version of Knative Client?

```
Version:      v20190710-local-31dc57a-dirty
Build Date:   2019-07-10 09:51:46
Git Revision: 31dc57a
Dependencies:
- serving:    v0.6.0
```

## What version of Knative Serving running on your cluster?

0.7.x

## Expected Behavior

The image created during the release in https://github.com/knative/client/blob/e1614f95e662e9e6f7f520eb2b33181d7b264a28/hack/release.sh#L36 with `ko` should also contain all version informations (propagated via `ldflags`)

## Actual Behavior

Running the created Docker image with the coordinates from the 0.2.0 release  yields:

```
docker run --rm -it gcr.io/knative-releases/github.com/knative/client/cmd/kn@sha256:393035ce99f1e639ba859d35ffb8b3f328722a2f22ab4b4ea1d889b9c015af49 version

Unable to find image 'gcr.io/knative-releases/github.com/knative/client/cmd/kn@sha256:393035ce99f1e639ba859d35ffb8b3f328722a2f22ab4b4ea1d889b9c015af49' locally
sha256:393035ce99f1e639ba859d35ffb8b3f328722a2f22ab4b4ea1d889b9c015af49: Pulling from knative-releases/github.com/knative/client/cmd/kn
155814304360: Pull complete
4003b5b92ca9: Pull complete
1f704ff7c739: Pull complete
Digest: sha256:393035ce99f1e639ba859d35ffb8b3f328722a2f22ab4b4ea1d889b9c015af49
Status: Downloaded newer image for gcr.io/knative-releases/github.com/knative/client/cmd/kn@sha256:393035ce99f1e639ba859d35ffb8b3f328722a2f22ab4b4ea1d889b9c015af49
Version:
Build Date:
Git Revision:
Dependencies:
- serving:
```

zhangtbj:
Interesting, I tried to create the image and docker run version, also get empty info:
```
jordanzt@zhangtaodembp:~/Works/workspaces/gopath/src/github.com/knative/client$ docker run --rm -it us.icr.io/knative_jordan/github.com/knative/client/cmd/kn:latest version
Version:
Build Date:
Git Revision:
Dependencies:
- serving:
```

But I checked `ko` command, it doesn't support `ldflags` flag. should use a `ldflags` similar way to add the info when `ko` creating the image? Or....

[...]

PR #1666, opened 2022-04-28
Fix OCI image build

dsimansk:
## Changes

* Fix OCI image build

## Reference

Related to #257

/cc @rhuss

### DIFF
--- a/hack/build-flags.sh
+++ b/hack/build-flags.sh
@@ -29,10 +29,10 @@ function build_flags() {
   export KN_BUILD_VERSION="${version}"
   export KN_BUILD_DATE="${now}"
   export KN_BUILD_GITREV="${rev}"
-  export KN_BUILD_LD_FLAGS="-X '${pkg}.BuildDate=${now}' \
+  KN_BUILD_LD_FLAGS="-X '${pkg}.BuildDate=${now}' \
   -X ${pkg}.Version=${version} \
   -X ${pkg}.GitRevision=${rev} \
   ${EXTERNAL_LD_FLAGS:-}"
 }
 
-build_flags "$@"
+build_flags
--- a/hack/build-flags.sh
+++ b/hack/build-flags.sh
@@ -13,13 +13,11 @@
 # limitations under the License.
 
 function build_flags() {
-  local base="${1}"
   local now rev
   now="$(date -u '+%Y-%m-%d %H:%M:%S')"
   rev="$(git rev-parse --short HEAD)"
   local pkg="knative.dev/client/pkg/kn/commands/version"
   local version="${TAG:-}"
-  local major_minor="$(echo "${version}" | cut -f1-2 -d. -n)"
   # Use vYYYYMMDD-local-<hash> for the version string, if not passed.
   if [[ -z "${version}" ]]; then
     # Get the commit, excluding any tags but keeping the "dirty" flag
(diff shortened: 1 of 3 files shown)


hack/build-flags.sh
```diff
@@ -28,29 +26,13 @@ function build_flags() {
...
-  echo "-X '${pkg}.BuildDate=${now}' \
+  export KN_BUILD_LD_FLAGS="-X '${pkg}.BuildDate=${now}' \
   -X ${pkg}.Version=${version} \
   -X ${pkg}.GitRevision=${rev} \
-  -X ${pkg}.VersionServing=${version_serving} \
-  -X ${pkg}.VersionEventing=${version_eventing}\
   ${EXTERNAL_LD_FLAGS:-}"
 }
+
+build_flags "$@"
```
dsimansk, line 38: @rhuss I went with the following call here, but I don't removing the functions completely.

rhuss, line 38: Maybe just call without arguments as the function doesn't take any argument anymore.

hack/build-flags.sh
```diff
@@ -28,29 +26,13 @@ function build_flags() {
...
-  else
-    version_serving="${technical_version_serving}"
-    version_eventing="${technical_version_eventing}"
-  fi
-  # Export as env variables to be used in `ko` OCI image build
   export KN_BUILD_VERSION="${version}"
   export KN_BUILD_DATE="${now}"
   export KN_BUILD_GITREV="${rev}"
-  echo "-X '${pkg}.BuildDate=${now}' \
+  export KN_BUILD_LD_FLAGS="-X '${pkg}.BuildDate=${now}' \
```
rhuss, line 32: I think you don't even need `export` here as you are using the variables directly and do not need to propagate them to a subprocess.

hack/release.sh
```diff
@@ -21,13 +21,9 @@ source $(dirname $0)/../vendor/knative.dev/hack/release.sh
 source $(dirname $0)/build-flags.sh
 
 function build_release() {
-  local ld_flags="$(build_flags $(dirname $0)/..)"
-  local pkg="knative.dev/client/pkg/kn/commands"
-  local version="${TAG}"
-  # Use vYYYYMMDD-<hash>-local for the version string, if not passed.
-  [[ -z "${version}" ]] && version="v${BUILD_TAG}-local"
+  # Env var exported by hack/build-flags.sh
+  local ld_flags="${KN_BUILD_LD_FLAGS:-}"
```
rhuss, line 25: An empty default doesn't work well because then you would have a `-ldflags` without value to the option which will very likely lead to an error. IMO you could directly use `KN_BUILD_FLAGS` and commit `ld_flags`. 

Or use something like 

```suggestion
  local ld_flags="-ldflags ${KN_BUILD_LD_FLAGS:-}"
```
 and then just add $ld_flags below, that should also then work with an empty env var.

hack/release.sh
```diff
@@ -21,13 +21,9 @@ source $(dirname $0)/../vendor/knative.dev/hack/release.sh
...
 
-  export GO111MODULE=on
   export CGO_ENABLED=0
   echo "🚧 🐧 Building for Linux (amd64)"
   GOOS=linux GOARCH=amd64 go build -mod=vendor -ldflags "${ld_flags}" -o ./kn-linux-amd64 ./cmd/...
```
rhuss, line 29:
```suggestion
  GOOS=linux GOARCH=amd64 go build -mod=vendor ${ld_flags} -o ./kn-linux-amd64 ./cmd/...
```

hack/build.sh
```diff
@@ -265,9 +266,8 @@ has_flag() {
 }
 
 cross_build() {
-  local basedir ld_flags
-  basedir=$(basedir)
-  ld_flags="$(build_flags $basedir)"
+  # Env var exported by hack/build-flags.sh
+  local ld_flags="${KN_BUILD_LD_FLAGS:-}"
```
rhuss, line 270: See my comment below on how to deal with an empty default.

hack/release.sh
dsimansk: The proposal doesn't work smoothly with our `date` value and quoting. On the other hand I've tested with great success that empty flags from `KN_BUILD_LD_FLAGS` doesn't break the build or produce any error.

rhuss: ok, so let's go with the original approach if this is no problem.

/approve
/lgtm

PR #1669, opened 2022-05-02
Cleanup sugar controller setup

dsimansk:
## Description

Since Knative >= 1.4 standalone Sugar controller is integrated into main Eventing one. Currently it's producing harmless 404 error during E2E tests setup, but it might be confusing or break in the future. 

Integrations tests should pass without issues.

## Changes

* :broom: Cleanup sugear controller setups

/cc @rhuss @vyasgun


navidshaikh (approved):
/lgtm

PR #1670, opened 2022-05-02
Added delivery options to broker create

### LINKED ISSUE #1557: Add delivery options for "kn broker create"
### Feature request

Actually, the support for broker options is quite limited, only the broker class can be currently specified, but none of the delivery options that are described in https://github.com/knative/eventing/blob/5e3183b552cabe35ae1015fa1baff5cd67c0b7cf/pkg/apis/duck/v1/delivery_types.go#L31-L84

These options are independent of the broker implementation and can be typed (.e.g `--dl-sink`). This is different from https://github.com/knative/client/issues/1525 where a reference to an implementation-specific configuration should be able to add.

### Use case

Creating a broker with delivery options that are different from the default values.

### UI Example

The following options are proposals, the detailed naming is up to discussion:

```
--dl-sink <sink reference>              Reference to a sink for delivering events that can not be sent
--retry <retries>.                      Number of retries before sending the event to a dead-letter sink
--timeout <duration>                    Timeout for a single request
--backoff-policy  <policy>              Backoff policy for retries, either "linear" or "exponential"
--backoff-delay <duration>              Based delay between retries
--retry-after-max <duration>            Upper bound for a duration specified in an "Retry-After" header (experimental)
```

vyasgun:
## Description
Added delivery options to `kn broker create`

## Changes

* Added `--dl-sink`, `--timeout`, --`backoff-policy`, `backoff-delay`, `retry-after-max`, `--retry` flags
* Validation is on the server side for the values
* Added unit test cases

## Reference

Fixes #1557

### DIFF
--- a/pkg/eventing/v1/client.go
+++ b/pkg/eventing/v1/client.go
@@ -37,6 +37,7 @@ import (
 )
 
 type TriggerUpdateFunc func(origTrigger *eventingv1.Trigger) (*eventingv1.Trigger, error)
+type BrokerUpdateFunc func(origBroker *eventingv1.Broker) (*eventingv1.Broker, error)
 
 // KnEventingClient to Eventing Sources. All methods are relative to the
 // namespace specified during construction
@@ -63,6 +64,10 @@ type KnEventingClient interface {
 	DeleteBroker(ctx context.Context, name string, timeout time.Duration) error
 	// ListBrokers returns list of broker CRDs
 	ListBrokers(ctx context.Context) (*eventingv1.BrokerList, error)
+	// UpdateBroker is used to update an instance of broker
+	UpdateBroker(ctx context.Context, broker *eventingv1.Broker) error
+	// UpdateBrokerWithRetry is used to update an instance of broker
+	UpdateBrokerWithRetry(ctx context.Context, name string, updateFunc BrokerUpdateFunc, nrRetries int) error
 }
 
 // KnEventingClient is a combination of Sources client interface and namespace
@@ -348,6 +353,45 @@ func (c *knEventingClient) ListBrokers(ctx context.Context) (*eventingv1.BrokerL
 	return brokerListNew, nil
 }
 
+// UpdateBroker is used to update an instance of broker
+func (c *knEventingClient) UpdateBroker(ctx context.Context, broker *eventingv1.Broker) error {
+	_, err := c.client.Brokers(c.namespace).Update(ctx, broker, meta_v1.UpdateOptions{})
+	if err != nil {
+		return kn_errors.GetError(err)
+	}
+	return nil
+}
+
+func (c *knEventingClient) UpdateBrokerWithRetry(ctx context.Context, name string, updateFunc BrokerUpdateFunc, nrRetries int) error {
+	return updateBrokerWithRetry(ctx, c, name, updateFunc, nrRetries)
+}
+
+func updateBrokerWithRetry(ctx context.Context, c KnEventingClient, name string, updateFunc BrokerUpdateFunc, nrRetries int) error {
+	b := config.DefaultRetry
+	b.Steps = nrRetries
+	updateBrokerFunc := func() error {
+		return updateBroker(ctx, c, name, updateFunc)
+	}
+	err := retry.RetryOnConflict(b, updateBrokerFunc)
+	return err
+}
+
+func updateBroker(ctx context.Context, c KnEventingClient, name string, updateFunc BrokerUpdateFunc) error {
+	broker, err := c.GetBroker(ctx, name)
+	if err != nil {
+		return err
+	}
+	if broker.GetDeletionTimestamp() != nil {
+		return fmt.Errorf("can't update broker %s because it has been marked for deletion", name)
+	}
+	updatedBroker, err := updateFunc(broker.DeepCopy())
+	if err != nil {
+		return err
+	}
+
+	return c.UpdateBroker(ctx, updatedBroker)
+}
+
 // BrokerBuilder is for building the broker
 type BrokerBuilder struct {
 	broker *eventingv1.Broker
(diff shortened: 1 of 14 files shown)


pkg/eventing/v1/client_test.go
```diff
@@ -518,6 +524,23 @@ func newBrokerWithClass(name string) *eventingv1.Broker {
...
+	testTimeout := "PT10S"
+	retry := int32(2)
+	policy := v1.BackoffPolicyType("linear")
+	return NewBrokerBuilder(name).
+		Namespace(testNamespace).
+		DlSink(sink).
+		Timeout(&testTimeout).
+		Retry(&retry).
+		BackoffDelay(&testTimeout).
+		BackoffPolicy(&policy).
```
dsimansk, line 540: Missing `RetryAfterMax()` call.

pkg/eventing/v1/client.go
```diff
@@ -387,6 +386,80 @@ func (b *BrokerBuilder) Class(class string) *BrokerBuilder {
 	return b
 }
 
+// DlSink for the broker builder
+func (b *BrokerBuilder) DlSink(dlSink *duckv1.Destination) *BrokerBuilder {
+	empty := duckv1.Destination{}
+	if dlSink == nil || *dlSink == empty {
+		return b
```
dsimansk, line 393: For the sake of managing test coverage, all those corner cases would ideally be covered. IMO `nil` cases should be fine for us.

pkg/kn/commands/broker/delivery_option_flags.go
```diff
@@ -0,0 +1,40 @@
...
+	RetryCount    int32
+	Timeout       string
+	BackoffPolicy string
+	BackoffDelay  string
+	RetryAfterMax string
+}
+
+func (d *DeliveryOptionFlags) Add(cmd *cobra.Command) {
+	d.SinkFlags.AddWithFlagName(cmd, "dl-sink", "")
+	cmd.Flag("dl-sink").Usage = "Reference to a sink for delivering events that can not be sent"
```
dsimansk, line 33: Well I'm not sure if shouldn't keep the default usage with all the explanation of sinks URIs.

pkg/kn/commands/broker/delivery_option_flags.go
```diff
@@ -0,0 +1,40 @@
...
+
+	cmd.Flags().Int32Var(&d.RetryCount, "retry", 0, "Number of retries before sending the event to a dead-letter sink")
```
dsimansk, line 35: Is there any default enforced on the Eventing side, that we could potentially follow here?

dsimansk (changes requested):
Added a few comments around coverage and help messages, but otherwise looks OK.

pkg/kn/commands/broker/delivery_option_flags.go
dsimansk: To be clear, I meant for all the flags here, not just the `retry` one.

pkg/eventing/v1/client.go
dsimansk: > IMO nil cases should be fine for us.

I used a very confusing construct in my first comment, sorry about that :)

It should really go like this. Let's add test cases with `nil` values to cover return of builder without any changes. It's for the sake of coverage mainly, but it is what it is. :)

pkg/eventing/v1/client_test.go
vyasgun: Thanks, fixed.

pkg/kn/commands/broker/delivery_option_flags.go
vyasgun: Thanks, I updated the usages.

pkg/eventing/v1/client.go
vyasgun: I added the `nil` test cases.

pkg/kn/commands/broker/delivery_option_flags.go
```diff
@@ -0,0 +1,45 @@
...
+	cmd.Flags().Int32Var(&d.RetryCount, "retry", 0, "The minimum number of retries the sender should attempt when "+
+		"sending an event before moving it to the dead letter sink.")
+	cmd.Flags().StringVar(&d.Timeout, "timeout", "", "The timeout of each single request. The value must be greater than 0.")
+	cmd.Flags().StringVar(&d.BackoffPolicy, "backoff-policy", "", "The retry backoff policy (linear, exponential).")
+	cmd.Flags().StringVar(&d.BackoffDelay, "backoff-delay", "", "The delay before retrying.")
+	cmd.Flags().StringVar(&d.RetryAfterMax, "retry-after-max", "", "An optional upper bound on the duration specified in a "+
+		"\"Retry-After\" header when calculating backoff times for retrying 429 and 503 response codes. "+
+		"Setting the value to zero (\"PT0S\") can be used to opt-out of respecting \"Retry-After\" header values altogether. "+
+		"This value only takes effect if \"Retry\" is configured, and also depends on specific implementations (Channels, Sources, etc.) "+
+		"choosing to provide this capability.")
```
dsimansk, line 44: Wrt to `--retry` + `--retry-after-maxt` is there validation provided on the webhook side?

dsimansk (commented):
Added a short comment about validation.

There's another angle of this change I haven't noticed at first. Originally broker command groups doesn't include `update`. There wasn't no options to modify, except from create/delete flow. However, it should be added probably with this addition flags.

pkg/kn/commands/broker/delivery_option_flags.go
vyasgun: Yes, the validation is present.

vyasgun:
@dsimansk Sorry about the delay. I just added the update command to this PR.

PR #1679, opened 2022-05-30
Added support for mounting EmptyDir volume type

### LINKED ISSUE #1633: Add support for different volume types
### Feature request

Add support for different volume types. In Serving we have recently added two different volume types: [Emptydir](https://github.com/knative/serving/issues/11664) and [PVCs](https://github.com/knative/serving/issues/12438).

### Use case

It would be nice to extend the cli to support volumes. Users can then use the cli to add volumes as they would normally do with
a yaml file when deploying services. 

### UI Example

This could work similarly to the `--extra-containers` flag. At minimum users should be able to pass the configuration for volumes
at pod level and at container level via json/yaml data. 

If it is found useful defaults could be added. For example if emptydir is used a default volume mount could be done under a predefined dir eg. /cache along with a predefined `emptyDir.medium` eg. memory.

Examples:
```
kn service create helloworld-go --image gcr.io/knative-samples/helloworld-go --emptydir-volumes-config json_data
kn service create helloworld-go --image gcr.io/knative-samples/helloworld-go --pvc-volumes-config json_data
```
`{emptydir, pvc}-volumes-config` refers to the pod voumes field. Regarding the container volume mounts these can be configured either with an extra parameter eg. `--emptydir-volumemounts` for the user container and rely on --extra-containers flag for the rest of the mounts targeting other containers. This should be extended in the future to support [init-containers](https://github.com/knative/serving/issues/12024) too.
 
/cc @rhuss

vyasgun:
## Description
Added support for mounting EmptyDir volume type

## Changes

* Added parsing for emptyDir volume type, medium type, and size limit

## Reference

Fixes #1633

**Release Note**

```release-note
NONE
```


vyasgun:
@dsimansk Thanks for the review. The memory and size limit syntax is according to https://github.com/knative/client/issues/1633#issuecomment-1069131224. I will add the additional functionality to the `--volume` flag and simplify `--mount `.

dsimansk:
> @dsimansk Thanks for the review. The memory and size limit syntax is according to [#1633 (comment)](https://github.com/knative/client/issues/1633#issuecomment-1069131224). I will add the additional functionality to the `--volume` flag and simplify `--mount `.

OK, I see. :) It still feels like too granular option to my taste. However, let's have it in `--volume` flag. That should be fine.

dsimansk:
@vyasgun there's one issue with current volume creation that I've stumped upon. By default we enforce `readOnly: true`, that's probably good and desired for ConfigMaps and Secrets. However, it'll be problematic for empty dir and PVC since those might be very well expected to have write access. 

https://github.com/knative/client/blob/main/pkg/kn/flags/podspec_helper.go#L504

PR #1697, opened 2022-06-30
Add --probe-liveness and --probe-readiness flags

### LINKED ISSUE #1522: Liveness- and Readiness probes for kn service commands
### Feature request

There is no way to specify liveness or readiness probes with `kn service create/update`. As this needs a bit more than a scalar value some UI design is needed to find a proper solution.

dsimansk:
## Description

Quick format example:
```go
// =======================================================================================
// Probes

// resolveProbe parses probes as a string
// It's split into two functions:
//   - resolveProbeOptions() -> common probe opts
//   - resolveProbeHandler() -> probe handler [HTTPGet, Exec, TCPSocket]
// Format:
//	- [http,https]:host:port:path;<common_opts>
//	- exec:cmd,cmd,...;<common_opts>
//  - tcp:host:port;<common_opts>
// Common opts (comma separated, case insensitive):
//	- <probe_opts>;InitialDelaySeconds=<int_value>,FailureThreshold=<int_value>,
//  	SuccessThreshold=<int_value>,PeriodSeconds==<int_value>,TimeoutSeconds=<int_value>
```

```console
kn service create hello --image docker.io/dsimansk/helloworld --probe-liveness "http:::/;timeoutSeconds=60" 
```

## Changes

* Add flag `probe-liveness`
* Add flag `probe-readiness`
* TODO: examples, more tests

## Reference

Fixes #1522 

/cc @vyasgun @rhuss 

**Release Note**

```release-note
NONE
```


vyasgun (commented):
Would it be a better idea to have separate flag(s) for probe options? The probe string seems a bit long and detailed to me and it also doesn't provide a way to update only the options

pkg/kn/flags/podspec.go
```diff
@@ -177,6 +180,18 @@ func (p *PodSpecFlags) AddFlags(flagset *pflag.FlagSet) []string {
 			"Example: --containers ./containers.yaml or --containers -.")
 	flagNames = append(flagNames, "containers")
 
+	// Probes
+	commonProbeDescription := "Supported probe types are HTTGet, Exec and TCPSocket. " +
+		"Format: [http,https]:host:port:path;<common_opts>, exec:cmd,cmd,...;<common_opts>, tcp:host:port;<common_opts>. " +
+		"Common opts (comma separated, case insensitive): InitialDelaySeconds=<int_value>, FailureThreshold=<int_value>, " +
+		"SuccessThreshold=<int_value>, PeriodSeconds==<int_value>, TimeoutSeconds=<int_value>"
```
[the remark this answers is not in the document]
dsimansk: thx!

dsimansk:
> Would it be a better idea to have separate flag(s) for probe options? The probe string seems a bit long and detailed to me and it also doesn't provide a way to update only the options

That's a good idea, I like it. The other possible improvement could be opinionated probe defaults wrt timeouts, thresholds, etc. 

@rhuss what would be your preference?

docs/cmd/kn_container_add.md
```diff
@@ -40,6 +40,8 @@ kn container add NAME
...
   -p, --port string                  The port where application listens on, in the format 'NAME:PORT', where 'NAME' is optional. Examples: '--port h2c:8080' , '--port 8080'.
+      --probe-liveness string        Add liveness probe to Service deployment. Supported probe types are HTTGet, Exec and TCPSocket. Format: [http,https]:host:port:path;<common_opts>, exec:cmd,cmd,...;<common_opts>, tcp:host:port;<common_opts>. Common opts (comma separated, case insensitive): InitialDelaySeconds=<int_value>, FailureThreshold=<int_value>, SuccessThreshold=<int_value>, PeriodSeconds=<int_value>, TimeoutSeconds=<int_value>
```
rhuss, line 43: I wonder whether we could just use an URL instead of host:port:path as seems to be more natural for the user to provide ? If we need the parts separately we still can parse them.

docs/cmd/kn_service_update.md
```diff
@@ -68,6 +68,8 @@ kn service update NAME
       --no-lock-to-digest                 Do not keep the running image for the service constant when not explicitly specifying the image. (--no-lock-to-digest pulls the image tag afresh with each new revision)
       --no-wait                           Do not wait for 'service update' operation to be completed.
   -p, --port string                       The port where application listens on, in the format 'NAME:PORT', where 'NAME' is optional. Examples: '--port h2c:8080' , '--port 8080'.
+      --probe-liveness string             Add liveness probe to Service deployment. Supported probe types are HTTGet, Exec and TCPSocket. Format: [http,https]:host:port:path;<common_opts>, exec:cmd,cmd,...;<common_opts>, tcp:host:port;<common_opts>. Common opts (comma separated, case insensitive): InitialDelaySeconds=<int_value>, FailureThreshold=<int_value>, SuccessThreshold=<int_value>, PeriodSeconds=<int_value>, TimeoutSeconds=<int_value>
```
rhuss, line 71: one general concern: Liveness and readiness probes are part of the ContainerSpec, so in case you have multiple containers you can also have multiple probes. I think we should keep it like here for the main container but also add it to the `kn container` commands (that's also one reason why we cant use pipes for the probe generation).

pkg/kn/flags/podspec.go
```diff
@@ -177,6 +180,18 @@ func (p *PodSpecFlags) AddFlags(flagset *pflag.FlagSet) []string {
...
+	commonProbeDescription := "Supported probe types are HTTGet, Exec and TCPSocket. " +
+		"Format: [http,https]:host:port:path;<common_opts>, exec:cmd,cmd,...;<common_opts>, tcp:host:port;<common_opts>. " +
+		"Common opts (comma separated, case insensitive): InitialDelaySeconds=<int_value>, FailureThreshold=<int_value>, " +
+		"SuccessThreshold=<int_value>, PeriodSeconds=<int_value>, TimeoutSeconds=<int_value>"
+	flagset.StringVarP(&p.LivenessProbe, "probe-liveness", "", "", "Add liveness probe to Service deployment. "+
+		commonProbeDescription)
+	flagNames = append(flagNames, "probe-liveness")
+	flagset.StringVarP(&p.ReadinessProbe, "probe-readiness", "", "", "Add readiness probe to Service deployment. "+
+		commonProbeDescription)
+	flagNames = append(flagNames, "probe-readiness")
```
rhuss, line 193: at some point we should also add a `probe-startup` for the startup probe (but not sure whether startup probes are supported by knative, we should check this)

pkg/kn/flags/podspec_helper.go
```diff
@@ -865,3 +889,134 @@ func decodeContainersFromFile(filename string) (*corev1.PodSpec, error) {
...
+//   - resolveProbeHandler() -> probe handler [HTTPGet, Exec, TCPSocket]
+// Format:
+//	- [http,https]:host:port:path;<common_opts>
+//	- exec:cmd,cmd,...;<common_opts>
+//  - tcp:host:port;<common_opts>
+// Common opts (comma separated, case insensitive):
+//	- <probe_opts>;InitialDelaySeconds=<int_value>,FailureThreshold=<int_value>,
+//  	SuccessThreshold=<int_value>,PeriodSeconds==<int_value>,TimeoutSeconds=<int_value>
+func resolveProbe(probeString string) (*corev1.Probe, error) {
+	parts := strings.Split(probeString, ";")
```
rhuss, line 908: Do you think that a value can contain also a ; ? Like as part of an argument for an exec probe ? I think this is quite unlikely, but not sure whether we should allows for escaping characters here. Maybe just continue like here and wait until the first bug reports occurs :-)

pkg/kn/flags/podspec_helper.go
```diff
@@ -865,3 +889,134 @@ func decodeContainersFromFile(filename string) (*corev1.PodSpec, error) {
...
+		if probeParts[2] != "" {
+			// Cosmetic fix to have default 'port: 0' instead of empty string 'port: ""'
+			handler.HTTPGet.Port = intstr.Parse(probeParts[2])
+		}
+		handler.HTTPGet.Path = probeParts[3]
+
+		probeHandler = &handler
+	case "exec":
+		if len(probeParts) != 2 {
+			return nil, fmt.Errorf("unexpected probe format, please use 'exec:<exec_command>,<exec_command>'")
```
rhuss, line 993:
```suggestion
			return nil, fmt.Errorf("unexpected probe format, please use 'exec:<exec_command>[,<exec_command>,...]'")
```

pkg/kn/flags/podspec_helper.go
```diff
@@ -865,3 +889,134 @@ func decodeContainersFromFile(filename string) (*corev1.PodSpec, error) {
...
+		}
+		if len(probeParts[1]) == 0 {
+			return nil, fmt.Errorf("at least one command parameter is required for Exec probe")
+		}
+		handler := corev1.ProbeHandler{
+			Exec: &corev1.ExecAction{},
+		}
+		cmd := strings.Split(probeParts[1], ",")
```
rhuss, line 1001: Should we allow for escaping `,` ?

rhuss (commented):
Looks good for me in general, but I think we need to add to `kn container`, too and maybe overthink how to provide the parameters for HTTP or HTTPS checks.

docs/cmd/kn_service_update.md
dsimansk: That's auto-magicly parth of `kn container add` flags because logic+flags are shared from our `PodSpec` implementation.

docs/cmd/kn_container_add.md
dsimansk: Yep, I was on the fence to have URL for HTTP probe, but I've opted for keeping the consistent format with other probe types. Not a strong preference though.

pkg/kn/flags/podspec_helper.go
dsimansk: There's a bit of trimming done later in the other two `resolve*` functions. That space wouldn't make it too far. :)

docs/cmd/kn_service_update.md
dsimansk: However, it might be a bit tricky with sidecars. 
https://github.com/knative/serving/blob/main/pkg/apis/serving/k8s_validation.go#L410-L411

pkg/kn/flags/podspec_helper.go
dsimansk: A corner case of "that should be fine..." statement? :)

pkg/kn/flags/podspec.go
dsimansk: That's not supported, as far as I've seen. 

Per: https://github.com/knative/serving/blob/main/pkg/apis/serving/k8s_validation.go#L513-L520

```
error: services.serving.knative.dev "hello" could not be patched: admission webhook "validation.webhook.serving.knative.dev" denied the request: validation failed: must not set the field(s): spec.template.spec.containers[0].startupProbe
```

dsimansk: The flag should be a quick additions once enabled on the Serving side.

rhuss:
> > Would it be a better idea to have separate flag(s) for probe options? The probe string seems a bit long and detailed to me and it also doesn't provide a way to update only the options
> 
> That's a good idea, I like it. The other possible improvement could be opinionated probe defaults wrt timeouts, thresholds, etc.
> 
> @rhuss what would be your preference?

Yes, that's sounds good to me, too. Maybe in a first step we can just omit the options in this PR (just using defaults), and then deliver the options as separate argument in a patch (or regular) release ?

docs/cmd/kn_service_update.md
rhuss: Ah, I see. So `kn container add` should *not* allow those options. Are there any other container related options that might be applied only on the main-container ? If so, we might consider to restrict the possible options on `container add` in the future. For now, I'm fine that we still have it, I guess it will lead to an error on the server side if somebody would use a probe on an extra container.

pkg/kn/flags/podspec_helper.go
rhuss: Ok, let's keep it for now like it is.

dsimansk:
> > > Would it be a better idea to have separate flag(s) for probe options? The probe string seems a bit long and detailed to me and it also doesn't provide a way to update only the options
> > 
> > 
> > That's a good idea, I like it. The other possible improvement could be opinionated probe defaults wrt timeouts, thresholds, etc.
> > @rhuss what would be your preference?
> 
> Yes, that's sounds good to me, too. Maybe in a first step we can just omit the options in this PR (just using defaults), and then deliver the options as separate argument in a patch (or regular) release ?

I'll add a new flag for the options. Since the backbone impl for parsing it is already in place, it should be fairly quick. 

How do you like the name for it? `--probe-*-opts` or `--probe*-options`?

```
--probe-liveness-opts

--probe-readiness-options
```


Just a note we will have 4 probe related flags:

```
--probe-liveness
--probe-liveness-opts
--probe-readiness
--probe-readiness-opts
```

rhuss:
> --probe-liveness
> --probe-liveness-opts
> --probe-readiness
> --probe-readiness-opts

Sounds good to me, let's keep it as short as possible (it's already quite a mouthful)

docs/cmd/kn_container_add.md
```diff
@@ -28,23 +28,27 @@ kn container add NAME
...
+  -p, --port string                   The port where application listens on, in the format 'NAME:PORT', where 'NAME' is optional. Examples: '--port h2c:8080' , '--port 8080'.
+      --probe-liveness string         Add liveness probe to Service deployment. Supported probe types are HTTGet, Exec and TCPSocket. Format: [http,https]:host:port:path, exec:cmd[,cmd,...], tcp:host:port.
+      --probe-liveness-opts string    Add common options to liveness probe. Common opts (comma separated, case insensitive): InitialDelaySeconds=<int_value>, FailureThreshold=<int_value>, SuccessThreshold=<int_value>, PeriodSeconds=<int_value>, TimeoutSeconds=<int_value>
```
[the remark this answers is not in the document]
dsimansk: Yes, they are.

vyasgun:
Thanks :)
/hold

PR #1700, opened 2022-07-04
Added --broker-config flag to broker create command

### LINKED ISSUE #1525: Create broker object with spec.config pointer
### Feature request

There is currently no option to use `kn` to create a broker and provide `config` parts (e.g. ponting to a ConfigMap or an other resource)

Something like below needs to be still created using `k apply...` :

```yaml
kind: Broker
metadata:
  annotations:
    eventing.knative.dev/broker.class: Kafka
  name: my-kafka-broker
spec:
  config:
    apiVersion: v1
    kind: ConfigMap
    name: kafka-broker-config
    namespace: knative-eventing
```

### UI Example

```

```

vyasgun:
## Description
This flag will be used to set the `.spec.config` field in the broker object

## Changes

* Added the flag
* The flag has to be used in conjunction with `--class` flag
* TODO: Add unit tests and examples

## Reference

Fixes #1525

### DIFF
--- a/docs/cmd/kn_broker_create.md
+++ b/docs/cmd/kn_broker_create.md
@@ -23,6 +23,7 @@ kn broker create NAME
 ```
       --backoff-delay string     The delay before retrying.
       --backoff-policy string    The retry backoff policy (linear, exponential).
+      --broker-config string     Broker config object like ConfigMap or RabbitMQ
       --class string             Broker class like 'MTChannelBasedBroker' or 'Kafka' (if available).
       --dl-sink string           The sink receiving event that could not be sent to a destination.
   -h, --help                     help for create
--- a/pkg/kn/commands/broker/config_flags.go
+++ b/pkg/kn/commands/broker/config_flags.go
@@ -1,0 +1,115 @@
+// Copyright © 2022 The Knative Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package broker
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"knative.dev/client/pkg/util"
+	duckv1 "knative.dev/pkg/apis/duck/v1"
+)
+
+type ConfigType int
+
+// Known config types for broker
+const (
+	ConfigMapType ConfigType = iota
+	SecretType
+	RabbitMqType
+)
+
+var (
+	KReferenceMapping = map[ConfigType]*duckv1.KReference{
+		ConfigMapType: {Kind: "ConfigMap", APIVersion: "v1"},
+		SecretType:    {Kind: "Secret", APIVersion: "v1"},
+		RabbitMqType:  {Kind: "RabbitmqCluster", APIVersion: "rabbitmq.com/v1beta1"},
+	}
+)
+
+type ConfigFlags struct {
+	BrokerConfig string
+}
+
+func (c *ConfigFlags) Add(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&c.BrokerConfig, "broker-config", "", "Broker config object like ConfigMap or RabbitMQ")
+}
+
+func (c *ConfigFlags) GetBrokerConfigReference() (*duckv1.KReference, error) {
+	config := c.BrokerConfig
+	slices := strings.SplitN(config, ":", 3)
+	if len(slices) == 1 {
+		return &duckv1.KReference{
+			Kind:       "ConfigMap",
+			Name:       slices[0],
+			APIVersion: "v1",
+		}, nil
+	} else if len(slices) == 2 {
+		kind := slices[0]
+		name := slices[1]
+		kRef := getDefaultKReference(kind)
+		if kRef.APIVersion == "" {
+			return nil, fmt.Errorf("kind %q is unknown and APIVersion could not be determined", kind)
+		}
+		kRef.Name = name
+		return kRef, nil
+	} else {
+		kind := slices[0]
+		name := slices[1]
+		kRef := getDefaultKReference(kind)
+		kRef.Name = name
+
+		params := strings.Split(slices[2], ",")
+		if len(params) == 1 && !strings.Contains(params[0], ",") {
+			kRef.Namespace = params[0]
+			return kRef, nil
+		}
+		mappedOptions, err := util.MapFromArray(params, "=")
+		if err != nil {
+			return nil, err
+		}
+		for k, v := range mappedOptions {
+			switch strings.ToLower(k) {
+			case "namespace":
+				kRef.Namespace = v
+			case "group":
+				kRef.Group = v
+			case "apiversion":
+				kRef.APIVersion = v
+			default:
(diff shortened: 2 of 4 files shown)


docs/cmd/kn_broker_create.md
```diff
@@ -23,6 +23,7 @@ kn broker create NAME
 ```
       --backoff-delay string     The delay before retrying.
       --backoff-policy string    The retry backoff policy (linear, exponential).
+      --broker-config string     Broker config object like ConfigMap or RabbitMQ
```
maximilien, line 26: Somehow this statement is missing something "Broker config object like ConfigMap or RabbitMQ" can we be more descriptive. Also is "or" the right conjunctive here?

pkg/kn/commands/broker/config_flags.go
```diff
@@ -0,0 +1,115 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"knative.dev/client/pkg/util"
+	duckv1 "knative.dev/pkg/apis/duck/v1"
+)
+
+type ConfigType int
```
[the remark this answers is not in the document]
maximilien: Please add comments to all public functions or types

pkg/kn/commands/broker/config_flags.go
```diff
@@ -0,0 +1,115 @@
...
+		for k, v := range mappedOptions {
+			switch strings.ToLower(k) {
+			case "namespace":
+				kRef.Namespace = v
+			case "group":
+				kRef.Group = v
+			case "apiversion":
+				kRef.APIVersion = v
+			default:
+				return nil, fmt.Errorf("incorrect field %q. Please specify any of the following: Namespace, Group, APIVersion", k)
```
maximilien, line 93: Since field test is on lower case names perhaps the error message should use lower case in message too?

maximilien (commented):
First thanks for contribution. Left some comments. However, is this part 1 of the PR. I don't see any tests?

docs/cmd/kn_broker_create.md
vyasgun: I updated the flag description in my latest commit. Please let me know if it can be improved further

vyasgun:
@maximilien Thanks for the review. I updated the PR with tests and more detailed docs.

pkg/kn/commands/broker/create.go
```diff
@@ -33,6 +34,15 @@ var createExample = `
...
   kn broker create mybroker --namespace myproject --class Kafka
+
+  # Create a broker 'mybroker' in the myproject namespace with config referencing configmap named spec-cm
+  kn broker create mybroker --namespace myproject --class Kafka --broker-config cm:spec-cm
+
+  # Create a broker 'mybroker' in the myproject namespace with config referencing configmap named spec-cm in test namespace
+  kn broker create mybroker --namespace myproject --class Kafka --broker-config cm:broker-spec-cm:test
+
+  # Create a broker 'mybroker' in the myproject namespace with config referencing configmap named spec-cm in test namespace
+  kn broker create mybroker --namespace myproject --class Kafka --broker-config cm:broker-spec-cm:test
```
dsimansk, line 45: It'd nice to show at least one different example than ConfigMap.

pkg/kn/commands/broker/create.go
```diff
@@ -33,6 +34,15 @@ var createExample = `
 
   # Create a broker 'mybroker' in the 'myproject' namespace and with a broker class of 'Kafka'
   kn broker create mybroker --namespace myproject --class Kafka
+
+  # Create a broker 'mybroker' in the myproject namespace with config referencing configmap named spec-cm
+  kn broker create mybroker --namespace myproject --class Kafka --broker-config cm:spec-cm
```
dsimansk, line 39: A counter proposal to the config k,v format:

```console
--broker-config rabbitmq.com/v1beta1:RabbitmqCluster:<name>:<namespace>

# With special case for core v1 resources like ConfigMaps and Secrets
--broker-config cm:spec-cm:spec-cm-ns
```

docs/cmd/kn_broker_create.md
```diff
@@ -16,13 +16,26 @@ kn broker create NAME
...
 
+  # Create a broker 'mybroker' in the myproject namespace with config referencing configmap named spec-cm
+  kn broker create mybroker --namespace myproject --class Kafka --broker-config cm:spec-cm
+
+  # Create a broker 'mybroker' in the myproject namespace with config referencing configmap named spec-cm in test namespace
+  kn broker create mybroker --namespace myproject --class Kafka --broker-config cm:broker-spec-cm:test
+
+  # Create a broker 'mybroker' in the myproject namespace with config referencing configmap named spec-cm in test namespace
+  kn broker create mybroker --namespace myproject --class Kafka --broker-config cm:broker-spec-cm:test
+
```
rhuss, line 27: The last example seems to duplicate the example before. As @dsimansk mentioned, having an example for a secret and/or a full GVK would be helpful here.

docs/cmd/kn_broker_create.md
```diff
@@ -16,13 +16,26 @@ kn broker create NAME
...
 
 ### Options
 
 ```
       --backoff-delay string     The delay before retrying.
       --backoff-policy string    The retry backoff policy (linear, exponential).
+      --broker-config string     Reference to the configuration that specifies configuration options for this Broker. For example, a pointer to a ConfigMap, Secret, RabbitmqCluster etc.The format for specifying the object is a colon separated string consisting of at most 3 substrings:
+                                 kind:object-name:namespace=?,apiVersion=?,group=?
+                                 The third substring is optional and the following is also acceptable (in case of ConfigMap, Secret, and RabbitmqCluster kinds):
+                                 kind:object-name
```
rhuss, line 38: Proposal for the wording

```
Reference pointing to the broker configuration. The value takes the format "<prefix>:<name>:<namespace>", where <prefix> is either "cm" for a ConfigMap, "secret" for a Secret or a GVK in the format "rabbitmq.com/v1beta1/RabbitmqCluster". "name" is the name of the referenced resource and "namespace" the optional namespace of that resource (which by default is the current namespace).
```

I back @dsimansk opinion that we should not invent another format to provide a full GVK (like in your proposal with additional `apiVersion` and `group`), but encode it already in the prefix. We have done that already I think, for the sink: (or at least plan for this). The prefix actually should fully specify the reference type (and we allow some shortcut for well-known types like cm and secret)

pkg/kn/commands/broker/config_flags.go
```diff
@@ -0,0 +1,126 @@
...
+			return nil, fmt.Errorf("kind %q is unknown and APIVersion could not be determined", kind)
+		}
+		return kRef, nil
+	}
+}
+
+func getDefaultKReference(kind string) *duckv1.KReference {
+	k := strings.ToLower(kind)
+	switch k {
+	case "cm", "configmap":
```
rhuss, line 117: could we make that switch more dynamic ? I.e. already add the list of shortcut prefixes to `KReferenceMapping` and then either iterate of the list to find the entry or build up a map before with the shortcut prefix as key, pointing to the KReferenceMapping entry (i.e. I would create an own Struct for the mapping with a KReference and a string array as member for KRerferenceMapping.

The advantage would be that all the relevant parts (reference + prefixes) are defined in a single place (i.e. where KReferenceMapping is defined)

docs/cmd/kn_broker_create.md
vyasgun: Thanks for the suggestion. I updated the examples accordingly. Hope they look okay now

docs/cmd/kn_broker_create.md
vyasgun: Sink is using the same delimiter (':') to get the prefix. So, instead of having the prefix as a part of the first slice and then splitting it by '/', I have added the format as follows:

`<apiVersion>:<kind>:<name>:<namespace>`

I can change this if this doesn't seem right/consistent.

pkg/kn/commands/broker/config_flags.go
vyasgun: Yes, sounds good. I'll do that.

PR #1707, opened 2022-07-21
Add check for PVC status in E2E tests

dsimansk:
## Description

Not every cluster might be ready to fulfill PVC. 

## Changes

* :broom: Add check for PVC status in E2E tests

## Reference

Fixes #

**Release Note**

```release-note
NONE
```

### DIFF
--- a/test/e2e/service_test.go
+++ b/test/e2e/service_test.go
@@ -240,11 +240,17 @@ func serviceCreateWithMount(r *test.KnRunResultCollector) {
 	assert.NilError(r.T(), err)
 	r.AssertNoError(out)
 
-	r.T().Log("update service with a new pvc mount")
-	out = r.KnTest().Kn().Run("service", "update", "test-svc", "--mount", "/mydir5=pvc:test-pvc")
-	r.AssertNoError(out)
+	_, err = kubectl.Run("wait", "--for=condition=Ready", "pvc/test-pvc", "--timeout=30s")
+	if err == nil {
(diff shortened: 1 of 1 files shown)


dsimansk:
PTAL, it's related to errors in OpenShift executed tests.
/cc @vyasgun

test/e2e/service_test.go
```diff
@@ -240,11 +240,17 @@ func serviceCreateWithMount(r *test.KnRunResultCollector) {
 	assert.NilError(r.T(), err)
 	r.AssertNoError(out)
 
-	r.T().Log("update service with a new pvc mount")
-	out = r.KnTest().Kn().Run("service", "update", "test-svc", "--mount", "/mydir5=pvc:test-pvc")
-	r.AssertNoError(out)
+	_, err = kubectl.Run("wait", "--for=condition=Ready", "pvc/test-pvc", "--timeout=30s")
```
vyasgun, line 243: The condition check will not work in the case of PVC as it doesn't have the field. Checking for `phase == Bound` should work instead?
```suggestion
	_, err = kubectl.Run("wait", "--for='jsonpath={..status.phase}'=Bound", "pvc/test-pvc", "--timeout=30s")
```

dsimansk: Yep, that's better thanks. I should have put a temp `.fail()` there first to test it out.

vyasgun:
Thanks!

/lgtm

PR #1709, opened 2022-07-27
Fix volume creation for PVC

dsimansk:
## Description

When initializing volumes for flag usage like in E2E `--mount /mydir5=pvc:test-pvc`. We shouldn't probably enforce `readOnly: true` on the created volume PVC. We may end up with service stuck in initialization phase, because of underlying storage config, e.g. some volumes are being formatted first. 

Error from volume init:
```
  Warning  FailedMount             1s (x4 over 7s)  kubelet                  MountVolume.MountDevice failed for volume "pvc-1577f113-81c1-44f9-861f-5def40d13e8b" : cannot mount unformatted disk /dev/nvme2n1 as we are manipulating it in read-only mode
```

Former behavior:
```
        volumes:
        - name: mydir5-aa14be68
          persistentVolumeClaim:
            claimName: test-pvc
            readOnly: true
```
/cc @vyasgun @rhuss @skonto 

## Changes

* Fix volume creation for PVC

## Reference

Fixes #

**Release Note**

```release-note
Fix mounted PVC not to be created as read-only
```

/kind bug

### DIFF
--- a/pkg/kn/flags/podspec_helper.go
+++ b/pkg/kn/flags/podspec_helper.go
@@ -528,7 +528,7 @@ func updateVolume(volume *corev1.Volume, info *volumeSourceInfo) error {
 	case EmptyDirVolumeSourceType:
 		volume.EmptyDir = &corev1.EmptyDirVolumeSource{Medium: corev1.StorageMedium(info.emptyDirMemoryType), SizeLimit: info.emptyDirSize}
 	case PVCVolumeSourceType:
-		volume.PersistentVolumeClaim = &corev1.PersistentVolumeClaimVolumeSource{ClaimName: info.volumeSourceName, ReadOnly: true}
+		volume.PersistentVolumeClaim = &corev1.PersistentVolumeClaimVolumeSource{ClaimName: info.volumeSourceName}
 	default:
 		return fmt.Errorf("Invalid VolumeSourceType")
 	}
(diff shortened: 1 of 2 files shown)


rhuss:
Looks good to me to change the default to not provide `readOnly: true` (what was the motivation in the first place btw ? as I think a readOnly PV doesn't make much sense for the usual use case anyway)

But would it make sense to allow options to be specified for a mount ? (like with appending a `,readOnly=true` to the spec ?) If so, we should create a followup issue for it, I'm good with merging this PR as is (but up to you)

/approve
/lgtm
/hold

vyasgun:
@rhuss I think I added the `readOnly: true` flag as default because it was the default for other types of volume mounts. There was also some discussion about adding a flag to configure the readOnly attribute.

dsimansk:
> Looks good to me to change the default to not provide `readOnly: true` (what was the motivation in the first place btw ? as I think a readOnly PV doesn't make much sense for the usual use case anyway)
> 
> But would it make sense to allow options to be specified for a mount ? (like with appending a `,readOnly=true` to the spec ?) If so, we should create a followup issue for it, I'm good with merging this PR as is (but up to you)
> 
> /approve /lgtm /hold

IMO it probably slipped through review. We had a few places that specified volumes as readOnly because of Secrets and ConfigMaps being a primary target previously. 

Indeed we should look at adding readOnly to a volume config flag to cover that case. However, most of the time I'd expect PVC needs to writable.

rhuss:
> @rhuss I think I added the `readOnly: true` flag as default because it was the default for other types of volume mounts. There was also some discussion about adding a flag to configure the readOnly attribute.

Do you know whether we have opened an issue for that ? If not, we should.

dsimansk:
Here: https://github.com/knative/client/issues/1712

/unhold

PR #1717, opened 2022-08-08
Add support for providing sink spec in GVR format

### LINKED ISSUE #1704: Add support for providing sink spec in GVR format
### Feature request
The user can specify a sink in subscription create and update (`--sink`, `--sink-reply`, `--sink-dead-letter` flags), source create and update(`--sink` flag), and broker create and update (`--dl-sink` flag). Currently, only a few known types are supported as an argument to these flags ("broker", "ksvc", and "channel"). If the user provides any other prefix outside the supported ones, `kn` returns an `unsupported sink prefix` error.

### Use case
To use sinks of custom types.

### UI Example

```
kn source ping create my-ping --schedule "*/2 * * * *" --data '{ value: "hello" }' --sink <apiVersion>:<kind>:<name>:<namespace>
```

vyasgun:
## Changes

* The user can specify the sink with custom resource types
* Follows the format: `prefix:name:namespace` where prefix is the group version resource

## Reference

Fixes #1704

**Release Note**

```release-note
Add support for providing sink spec in GVR format
```

### DIFF
--- a/pkg/kn/commands/flags/sink.go
+++ b/pkg/kn/commands/flags/sink.go
@@ -108,7 +108,21 @@ func (i *SinkFlags) ResolveSink(ctx context.Context, knclient clientdynamic.KnDy
 		if prefix == "svc" || prefix == "service" {
 			return nil, fmt.Errorf("unsupported sink prefix: '%s', please use prefix 'ksvc' for knative service", prefix)
 		}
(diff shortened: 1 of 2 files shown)


maximilien (commented):
/approve
/lgtm

My only complaint is to add more tests to cover the various corner cases to deal with bad input. Unless being tested elsewhere?

dsimansk:
@vyasgun if you could take a look at some more unit test please, in a follow-up PR.

vyasgun:
Follow up PR for unit tests: https://github.com/knative/client/pull/1724

PR #1732, opened 2022-09-14
Added config support to specify read only volumes

### LINKED ISSUE #1712: Add config option to specify `readOnly` volumes
### Feature request

Per discussion at PR #1709.

Extend current `--volume` flag to accept config parameter `readOnly=true` to limit volume accessibility.

### UI Example

```
--volume pvc:myvol:readOnly=true
```

/kind good-first-issue
/kind feature

vyasgun:
## Description
Added config support to specify read only volumes

## Changes

* User can specify if the volume is read only or not using the `--volume` flag
* Example: kn service create svc1 --image imagename/v0.0.1 --mount /mydir=vol1 --volume cm:cm1:readOnly=true

## Reference

Fixes #1712 

**Release Note**

```release-note

```


pkg/kn/flags/podspec.go
```diff
@@ -156,7 +156,8 @@ func (p *PodSpecFlags) AddFlags(flagset *pflag.FlagSet) []string {
 		"Add a volume from a ConfigMap (prefix cm: or config-map:) a Secret (prefix secret: or sc:), "+
 			"an EmptyDir (prefix ed: or emptyDir:) or a PersistentVolumeClaim (prefix pvc: or persistentVolumeClaim). "+
 			"Example: --volume myvolume=cm:myconfigmap, --volume myvolume=secret:mysecret or --volume emptyDir:myvol:size=1Gi,type=Memory. "+
-			"You can use this flag multiple times. "+
+			"You can use this flag multiple times. You can also specify if a volume is readOnly or not. "+
+			"Example: --volume myvolume=cm:myconfigmap:readOnly=True or --volume emptyDir:myvol:size=1Gi,type=Memory,readOnly=false. "+
```
rhuss, line 160:
```suggestion
			"Example: --volume myvolume=cm:myconfigmap:readOnly=true"+
```

rhuss: Maybe let's rephrase the whole help message like in 

> Add a volume. This option can be given multiple times and has the format: <name>=<prefix:objName:options>. "name" is the name of the volume to create, prefix can be "cm" or "config-map" for a ConfigMap volume, "secret" or "sc" for a Secret volume, "ed" or "emptyDir" for an emptyDir volume or "pvc:" or "persistentVolumeClaim" for a persistent volume. objName is the referenced object to use as a volume (e.g. the name of the ConfigMap). options is a comma-separated key-value list for additional options specific to the volume. Use "readOnly=true" as an option to specify the writable volumes being read-only. Example: "--volume myVolume=pvc:my-storage-claim:readOnly=true"

rhuss: tbh, I don't think you can specify `readOnly` to `--volume` but only to `--mount`. Is this correct?

rhuss (commented):
Could you please check the documentation ? I think you added it to `--volume` instead of `--mount`.

pkg/kn/flags/podspec.go
dsimansk: I think there's a specual case for `PersistenVolumeClaimVolume` that it can have `readOnly` that will enforce it on mount afterwards.

dsimansk: However, yes that should be probably `--mount` flag option.

vyasgun:
@dsimansk @rhuss Thanks for the reviews! I moved the config to `--mount` flag and modified the tests accordingly

dsimansk:
@vyasgun e2e started to fail now with the following:

```
 kn service update test-svc --mount /mydir4=ed:myvol:type=Memory,size=100Mi --namespace kne2etests11
        ┃ 
        🔥 W0922 12:36:55.592694   43907 gcp.go:120] WARNING: the gcp auth plugin is deprecated in v1.22+, unavailable in v1.25+; use gcloud instead.
        🔥 To learn more, consult https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke
        🔥 Error: unknown mount option "type"
        🔥 Run 'kn --help' for usage 
```

dsimansk:
Thanks!

/approve
/lgtm

PR #1737, opened 2022-09-21
Fix release script version calculation

dsimansk:
## Description

This should fix display version for release. IMO the `build_flags.sh` script was sourced too soon, before release values (especially `${TAG}`) were properly populated.

/cc @rhuss 

## Changes

* Fix release script version calculation

## Reference

Fixes #

**Release Note**

```release-note
Fix release script version calculation
```

/kind bug


rhuss (commented):
/approve
/lgtm

PR #1745, opened 2022-10-09
cli options to enable user, uid, and group impersonation

a7i:
## Description

CLI options to enable kubernetes user and group impersonation
```
❯ ./kn options
The following options can be passed to any command:

      --as string              username to impersonate for the operation
      --as-uid string          uid to impersonate for the operation
      --as-group stringArray   group to impersonate for the operation, this flag can be repeated to specify multiple groups
      --cluster string         name of the kubeconfig cluster to use
      --config string          kn configuration file (default: ~/.config/kn/config.yaml)
      --context string         name of the kubeconfig context to use
      --kubeconfig string      kubectl configuration file (default: ~/.kube/config)
      --log-http               log http traffic
```

## Changes

* provide string option `--as` for username impersonation
* provide string option `--as-uid` for uid impersonation
* provide stringArray option `--as-group` for group impersonation

## Reference

Fixes #

**Release Note**

```release-note
Provide cli options to enable Kubernetes user, uid, and group impersonation via `--as`, `--as-group` and `--as-uid` flags
```

- 🎁 New feature: provide cli options to enable Kubernetes user, uid, and group impersonation via `--as` and `--as-group` flags

### DIFF
--- a/pkg/kn/root/root.go
+++ b/pkg/kn/root/root.go
@@ -87,6 +87,7 @@ Find more information about Knative at: https://knative.dev`, rootName),
 	rootCmd.PersistentFlags().StringVar(&p.KubeContext, "context", "", "name of the kubeconfig context to use")
 	rootCmd.PersistentFlags().StringVar(&p.KubeCluster, "cluster", "", "name of the kubeconfig cluster to use")
 	rootCmd.PersistentFlags().StringVar(&p.KubeAsUser, "as", "", "username to impersonate for the operation")
+	rootCmd.PersistentFlags().StringVar(&p.KubeAsUID, "as-uid", "", "uid to impersonate for the operation")
 	rootCmd.PersistentFlags().StringArrayVar(&p.KubeAsGroup, "as-group", []string{}, "group to impersonate for the operation, this flag can be repeated to specify multiple groups")
 	flags.AddBothBoolFlags(rootCmd.PersistentFlags(), &p.LogHTTP, "log-http", "", false, "log http traffic")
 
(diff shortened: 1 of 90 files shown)


dsimansk:
@a7i thanks for the PR! Looks good at a first glance!

Can you pls take a look at failing unit test? https://github.com/knative/client/actions/runs/3212622694/jobs/5277761586#step:7:20

pkg/kn/root/root.go
```diff
@@ -86,6 +86,8 @@ Find more information about Knative at: https://knative.dev`, rootName),
 	rootCmd.PersistentFlags().StringVar(&p.KubeCfgPath, "kubeconfig", "", "kubectl configuration file (default: ~/.kube/config)")
 	rootCmd.PersistentFlags().StringVar(&p.KubeContext, "context", "", "name of the kubeconfig context to use")
 	rootCmd.PersistentFlags().StringVar(&p.KubeCluster, "cluster", "", "name of the kubeconfig cluster to use")
+	rootCmd.PersistentFlags().StringVar(&p.KubeAsUser, "as", "", "username to impersonate for the operation")
+	rootCmd.PersistentFlags().StringArrayVar(&p.KubeAsGroup, "as-group", []string{}, "group to impersonate for the operation, this flag can be repeated to specify multiple groups")
```
dsimansk, line 90: I wonder if we should add `--as-uid` flag as well to complete the interface with `kubectl` options.

/cc @rhuss @vyasgun

a7i: Oh interesting, I had never used that flag. That sounds like a good idea, will be adding shortly.

a7i: @dsimansk looking at kubectl options, I only see as user and as groups for impersonation:

```
kubectl options | grep impersonate
      --as='': Username to impersonate for the operation
      --as-group=[]: Group to impersonate for the operation, this flag can be repeated to specify multiple groups.
```

a7i: Just added support for `as-uid`, which seems was added in 1.23

dsimansk:
Thanks!

/approve
/lgtm

PR #1765, opened 2022-12-15
Add cli support for initContainers

ReToCode:
## Description
This PR adds `initContainer` support to Knative client. It is part of the work tracked in https://github.com/knative/serving/issues/12024.

## Changes
* Adds `--init-containers` as a flag for `kn service X`

## Reference
Partially fixes https://github.com/knative/serving/issues/12024

**Release Note**

```release-note
Add --init-containers flag to service create command options 
```

/lint

### DIFF
--- a/pkg/kn/flags/podspec.go
+++ b/pkg/kn/flags/podspec.go
@@ -45,6 +45,7 @@ type PodSpecFlags struct {
 	ReadinessProbe     string
 	ReadinessProbeOpts string
 
+	InitContainers  string
 	ExtraContainers string
 
 	Resources          ResourceOptions
@@ -172,6 +173,11 @@ func (p *PodSpecFlags) AddFlags(flagset *pflag.FlagSet) []string {
 			"Example: --arg myArg1 --arg --myArg2 --arg myArg3=3. "+
 			"You can use this flag multiple times.")
 	flagNames = append(flagNames, "arg")
+
+	flagset.StringVarP(&p.InitContainers, "init-containers", "", "",
+		"Specify path to file including definition for initContainers, alternatively use '-' to read from stdin. "+
+			"Example: --init-containers ./init-containers.yaml or --init-containers -.")
+	flagNames = append(flagNames, "init-containers")
 
 	// DEPRECATED since 1.0
 	flagset.StringVarP(&p.ExtraContainers, "extra-containers", "", "",
@@ -370,6 +376,15 @@ func (p *PodSpecFlags) ResolvePodSpec(podSpec *corev1.PodSpec, flags *pflag.Flag
 		if err != nil {
 			return err
 		}
+	}
+
+	if flags.Changed("init-containers") || p.InitContainers == "-" {
+		var fromFile *corev1.PodSpec
+		fromFile, err = decodeContainersFromFile(p.InitContainers)
+		if err != nil {
+			return err
+		}
+		UpdateInitContainers(podSpec, fromFile.InitContainers)
 	}
 
 	if flags.Changed("containers") || flags.Changed("extra-containers") || p.ExtraContainers == "-" {
--- a/pkg/kn/flags/podspec_helper.go
+++ b/pkg/kn/flags/podspec_helper.go
@@ -313,6 +313,30 @@ func UpdateContainers(spec *corev1.PodSpec, containers []corev1.Container) {
 	}
 }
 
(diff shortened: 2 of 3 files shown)


ReToCode:
Up for discussion: what would be the preferred way to generate the `initContainers` yaml?

a) extend `kn container COMMAND` with a flag to specify that it should add/update the container to produce `initContainers` instead of `containers` in the yaml
b) add a new command `kn init-container COMMAND` (would probably have lot of duplication)

Also one of those would be needed to add an e2e test. WDYT?

/cc @rhuss, @navidshaikh, @dsimansk

ReToCode:
Also is the `CHANGELOG.adoc` still relevant (as mentioned in the PR template)? Seems like it is a few versions behind https://github.com/knative/client/releases.

dsimansk:
> Up for discussion: what would be the preferred way to generate the `initContainers` yaml?
> 
> a) extend `kn container COMMAND` with a flag to specify that it should add/update the container to produce `initContainers` instead of `containers` in the yaml b) add a new command `kn init-container COMMAND` (would probably have lot of duplication)
> 
> Also one of those would be needed to add an e2e test. WDYT?
> 
> /cc @rhuss, @navidshaikh, @dsimansk
 
Under the hood `kn container add` is a builder pattern that leverages our `PodSpec` and `PodSpecFlags`. It's only producing YAML to stdout. Whereas `--containers -` flag will look for stdin input.

Is there any specific flag that needs to be present only for init containers? If not, I'd aim for reuse of `kn container add` for init containers as well. Adding it into the usage example.

In addition I wouldn't mind to have a so called smart `--containers -` flag that could process both `initContainers` and additional PodSpec `containers`.

ReToCode:
> 

Containers and InitContainers are the same `podSpec` so we basically could use the same command. I just tried that, but there are certain issues.

What I tried:
* added a `--as-init-container` to `kn container add`
* as `kn container add` always operates on `.containers` field, this parameter moves the generated container in `podSpec` from `.containers` to the `.initContainers` field
* This could then be piped like this:

```
kn container add init1 --image docker.io/example/init:first --as-init-container | \
  kn container add init2 --image docker.io/example/init:second --as-init-container | \
  kn service create myksvc --image docker.io/example/my-app:latest --init-containers -`,
```

But, `container add` also allows a mixed usage (it's not in the examples, but in that test: https://github.com/knative/client/blob/main/pkg/kn/commands/container/add_test.go#L53) as it operates on the same flags as `kn service COMMAND`. If we combine this approach with the mixed approach it might get really confusing, e.g.

`kn container add mycontainer --image docker.io/example/init:first --as-init-container --init-containers -`.
or
`kn container add mycontainer --image docker.io/example/init:first --as-init-container --containers -`.
or even
`kn container add mycontainer --image docker.io/example/init:first --as-init-container --containers=my.yaml --init-containers -`.

So the issues are
* Without an additional parameter `kn container add` cannot infer if its a `initContainer` or a `container`.
* The same applies to the `--containers` flag.
* `kn container add` allows mixed usage of commands on the cli and `--containers` and `--init-containers`.

@dsimansk Any ideas?

ReToCode:
@dsimansk gentle ping.

rhuss:
What about having a dedicated `kn init-container add` command? I think this would fit best into the UX philosophy, as initContainers and containers are two sections on the same level in the schema and should be treated on an equal footprint. The example above then would look like this:

```
kn init-container add init1 --image docker.io/example/init:first  | \
kn init-container add init2 --image docker.io/example/init:second  | \
kn container add sidecar --image docker.io/sidecar/container |\
kn service create myksvc --image docker.io/example/my-app:latest --containers -`,
```

The technical realization is probably involved as it requires changes to the schema of the data that is transported through the pipe. At the moment, it's just an array of regular containers. We could add a synthetic new field to a container definition like `type: init` when `kn init-container` is used. The `--containers -` would detangle that list for separate regular and init containers from the list coming in via stdout. When no `type:` is given, it's a regular container. This approach would be backward compatible. 

It's not perfect because of that synthetic field which makes init-containers in this list not conformant to a Container schema from Kubernets. However, if you don't use init-containers, it is still a 'pure' Container list, as you could directly copy it into a PodSpec. Still, I believe that solution is the most elegant for the problem given (i.e., keeping backward compatibility, and I doubt how useful `kn container add` in isolation is anyway).

wdyt ?

Issue #1775, opened 2023-02-13
Compile errors with Go 1.20


scottmason88:
### Bug report

Using Go 1.20 and building from source, I'm getting several errors. 

```go: downloading github.com/matttproud/golang_protobuf_extensions v1.0.1
pkg/printers/tablegenerator.go:77:2: S1011: should replace loop with `columns = append(columns, handler.columnDefinitions...)` (gosimple)
	for i := range handler.columnDefinitions {
	^
pkg/errors/factory_test.go:172:46: loopclosure: loop variable tc captured by func literal (govet)
			assert.Equal(t, IsForbiddenError(GetError(tc.Error)), tc.Forbidden)
			                                          ^
pkg/errors/factory_test.go:212:56: loopclosure: loop variable tc captured by func literal (govet)
			assert.Equal(t, api_errors.IsInternalError(GetError(tc.Error)), tc.Internal)
			                                                    ^
pkg/errors/factory_test.go:245:21: loopclosure: loop variable tc captured by func literal (govet)
			assert.Assert(t, tc.ErrorType(GetError(tc.Error)))
			                 ^
cmd/kn/main.go:33:2: SA1019: rand.Seed has been deprecated since Go 1.20 and an alternative has been available since Go 1.0: Programs that call Seed and then expect a specific sequence of results from the global random source (using functions such as Int) can be broken when a dependency changes how much it consumes from the global random source. To avoid such breakages, programs that need a specific result sequence should use NewRand(NewSource(seed)) to obtain a random generator that other packages cannot access. (staticcheck)
	rand.Seed(time.Now().UnixNano())
	^
pkg/serving/service_test.go:33:2: SA1019: rand.Seed has been deprecated since Go 1.20 and an alternative has been available since Go 1.0: Programs that call Seed and then expect a specific sequence of results from the global random source (using functions such as Int) can be broken when a dependency changes how much it consumes from the global random source. To avoid such breakages, programs that need a specific result sequence should use NewRand(NewSource(seed)) to obtain a random generator that other packages cannot access. (staticcheck)
	rand.Seed(1)
	^
[...]

PR #1779, opened 2023-02-15
Fixing errors related to Go 1.20

scottmason88:
## Description

This pull request will fix errors that came about with Go 1.20 deprecating rand.Seed.

## Changes

* Changed loop in tablegenerator.go to use the variadic function
* Removed the random number generator seed from main.go, and added a package local variable doing this to service.go, since this is the only place the seed from main.go was used.

## Reference

Issue #1775 

**Release Note**

```release-note
 Fix deprecated functions related to Go 1.20 
```

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -7,9 +7,9 @@
 // the empty line separators.
 
 ////
-| 🐛
-| Updated to work with Go 1.20
-| https://github.com/knative/client/pull/1775[#1775]
+| 🎁🐛🐣🗑️✨
+|
+| https://github.com/knative/client/pull/[#]
 ////
 ## v1.5.0 (2022-05-31)
 [cols="1,10,3", options="header", width="100%"]
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -7,9 +7,9 @@
 // the empty line separators.
 
 ////
-| 🎁🐛🐣🗑️✨
-|
-| https://github.com/knative/client/pull/[#]
+| 🐛
+| Updated to work with Go 1.20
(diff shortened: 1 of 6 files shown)


CHANGELOG.adoc
```diff
@@ -7,9 +7,9 @@
 // the empty line separators.
 
 ////
-| 🎁🐛🐣🗑️✨
-|
-| https://github.com/knative/client/pull/[#]
+| 🐛
+| Updated to work with Go 1.20
+| https://github.com/knative/client/pull/1775[#1775]
```
rhuss, line 12: Oops, I just saw that the Changelog is quite outdated 😬 (@dsimansk would it make sense to bring it up to date, covering the releases in the second half of 2022 ?)

However, @scottmason88 can you please keep the template text and move your part out of the comment (`//// .... ////`) ? Also you should add a new table for the 1.10 release (the next release), without a date yet. Then add your entry as the first row. That would be great !

rhuss (commented):
Thanks, and congrats on your first client PR ! Looks great. If you could fix the Changlog entry, that would be great.

CHANGELOG.adoc
dsimansk: Well, we don't use it anymore. I should have added a deprecation notice to that file probably. We're using release script generated release notes.

dsimansk: I'd skip the update from this PR.

scottmason88: I removed my changes to the changelog so that the comment is still there.

rhuss (commented):
/approve
/lgtm

PR #1791, opened 2023-03-21
Add kn secret command

### LINKED ISSUE #287: `kn secret` for managing secrets
## In what area(s)?

Classifications:
/kind good-first-issue
/kind feature
/kind proposal

## Describe the feature:

With us on the verge of adding support for KnEventing, it seems that we should consider adding support for basic secret creation/management in `kn`.  I say this because from a UX perspective it would be really nice if the user didn't have to switch between `kn` and `kubectl` just to do the basic KnEventing actions - like create a github event source. I like the goal of trying to keep them in the `kn` world as much as possible.  Obviously, there will always be a balance we'll need to watch for, but I think some basic secret stuff makes sense.

We should look at what you can do with `kubectl` to create secrets. Passing in name/value pairs is obvious, but reading them from files on disk is another common (and probably safer) thing to support as well.

We could then also consider adding secret volume mounting in the `service` cmd.

We might want to think about configMaps at some point too - but my more immediate concern is secrets because once Eventing support is added I think that'll be a more obvious gap in our feature set.

dsimansk:
## Changes

- 🎁 `kn secret` cmd group

## Reference

Fixes #287

**Release Note**

```release-note
Add `kn secret` commands group for managing secrets
```

PTAL do you recall this feature? :)
/cc @rhuss


pkg/kn/commands/secret/create.go
```diff
@@ -0,0 +1,102 @@
...
+			_, err = client.CoreV1().Secrets(namespace).Create(cmd.Context(), &toCreate, metav1.CreateOptions{})
+			if err != nil {
+				return err
+			}
+			fmt.Fprintf(cmd.OutOrStdout(), "Secret '%s' created in namespace '%s'.\n", name, namespace)
+			return nil
+		},
+	}
+	commands.AddNamespaceFlags(cmd.Flags(), false)
+	cmd.Flags().StringSliceVarP(&literals, "from-literal", "l", []string{}, "Specify coma separated list of key=value pairs.")
```
vyasgun, line 97: nit: spelling `coma -> comma`

pkg/kn/commands/secret/create.go
```diff
@@ -0,0 +1,102 @@
...
+func NewSecretCreateCommand(p *commands.KnParams) *cobra.Command {
+	var literals []string
+	var cert, key, secretType string
+	cmd := &cobra.Command{
+		Use:     "create NAME",
+		Short:   "Create secret",
+		Example: ``,
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			if len(args) != 1 {
+				return errors.New("'kn secret create' requires the domain name given as single argument")
```
vyasgun, line 39: Typo?: `domain name` should be `secret name`

pkg/kn/commands/secret/delete_test.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func TestSecretDeleteNotFound(t *testing.T) {
+	fakeClient := fake.NewSimpleClientset()
+
+	_, err := executeSecretCommand(fakeClient, "delete", "foo")
+	assert.ErrorContains(t, err, "not found")
+}
+
+func TestDomainMappingDeleteWithError(t *testing.T) {
```
vyasgun, line 41:
```suggestion
func TestSecretDeleteWithError(t *testing.T) {
```

pkg/kn/commands/secret/secret.go
```diff
@@ -0,0 +1,32 @@
...
+
+package secret
+
+import (
+	"github.com/spf13/cobra"
+	"knative.dev/client/pkg/kn/commands"
+)
+
+func NewSecretCommand(p *commands.KnParams) *cobra.Command {
+	domainCmd := &cobra.Command{
```
vyasgun, line 23: this should be `secretCmd`

pkg/kn/commands/secret/secret_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+func init() {
+	var err error
+	blankConfig, err = clientcmd.NewClientConfigFromBytes([]byte(kubeConfig))
+	if err != nil {
+		panic(err)
+	}
+}
+
+func TestDomainCommand(t *testing.T) {
```
vyasgun, line 55:
```suggestion
func TestSecretCommand(t *testing.T) {
```

vyasgun (changes requested):

PR #1793, opened 2023-03-22
fix: Use go_run from common lib

dsimansk:
## Changes

* ✨fix: Use go_run from common lib instead of deprecated `run_go_tool`
* Move bash flags up and expand to names

## Reference

Fixes #

**Release Note**

```release-note
NONE
```

/cc @rhuss

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -15,7 +15,8 @@
 # limitations under the License.
 
 set -o pipefail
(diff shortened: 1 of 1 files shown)


dsimansk:
@rhuss @vyasgun also ready if you can get to it. Thx!

PR #1794, opened 2023-03-22
feat: Add experimental filters to trigger describe cmd

### LINKED ISSUE #1785: kn does not show new trigger filters
I am using the Knative Eventing new trigger filters feature:
https://knative.dev/docs/eventing/experimental-features/new-trigger-filters/

```yaml
kind: Trigger
metadata:
  name: event-display-trigger
spec:
  broker: default
  filters:
    - cesql: "LOWER(type) = 'my-event-type'"
  subscriber:
    ref:
      kind: Service
      name: event-display
```

I noticed that `kn trigger describe` does not output the new `filters` data.
```
kn trigger describe event-display-trigger
Name:         event-display-trigger
Namespace:    default
Labels:       eventing.knative.dev/broker=default
Annotations:  eventing.knative.dev/creator=kubernetes-admin, eventing.knative.dev/lastModifier=ku ...
Age:          13h
Broker:       default

Sink:        
  Name:      event-display
  Resource:  Service (serving.knative.dev/v1)

Conditions:  
  OK TYPE                      AGE REASON
  ++ Ready                     13h 
  ++ BrokerReady               13h 
  ++ DeadLetterSinkResolved    13h DeadLetterSinkNotConfigured
  ++ DependencyReady           13h 
  ++ SubscriberResolved        13h 
  ++ SubscriptionReady         13h 
```

I know that this is still an alpha feature, so I don't know if this is a feature request or a bug.
I would expect when the feature is enabled to see the filters.

dsimansk:
## Changes

* 🎁  feat: Add cesql filter field to trigger describe cmd

Preview:
```shell
➜  client git:(pr/desc-cesql-filter) ✗ kn trigger describe broker-display
Name:         broker-display
Namespace:    default
Labels:       eventing.knative.dev/broker=mybroker
Annotations:  eventing.knative.dev/creator=kubernetes-admin, eventing.knative.dev/lastModifier=ku ...
Age:          12h
Broker:       mybroker
Filter:
  type:       dev.knative.demo.event.processed

Filters (experimental):
  CESQL:                 LOWER(type) = 'my-event-type'

Sink:
  Name:      event-display
  Resource:  Service (serving.knative.dev/v1)

Conditions:
  OK TYPE    AGE REASON
``` 
## Reference

Fixes #1785 

**Release Note**

```release-note
Add experimental filters to `trigger describe` cmd
```

### DIFF
--- a/pkg/kn/commands/trigger/describe.go
+++ b/pkg/kn/commands/trigger/describe.go
@@ -123,4 +123,14 @@ func writeTrigger(dw printers.PrefixWriter, trigger *v1beta1.Trigger, printDetai
 			subWriter.WriteAttribute(key, value)
 		}
 	}
+	if len(trigger.Spec.Filters) > 0 {
+		// Split Filter and Filters (experimental) with new line
+		dw.WriteLine()
+		subWriter := dw.WriteAttribute("Filters (experimental)", "")
+		for _, filter := range trigger.Spec.Filters {
+			if filter.CESQL != "" {
+				subWriter.WriteAttribute("CESQL", filter.CESQL)
+			}
+		}
+	}
 }
--- a/pkg/kn/commands/trigger/describe.go
+++ b/pkg/kn/commands/trigger/describe.go
@@ -16,11 +16,9 @@ package trigger
 
 import (
 	"errors"
-	"reflect"
 
 	"github.com/spf13/cobra"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
-
 	"knative.dev/client/lib/printing"
 	"knative.dev/client/pkg/kn/commands"
 	"knative.dev/client/pkg/printers"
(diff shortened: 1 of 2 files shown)


pkg/kn/commands/trigger/describe.go
```diff
@@ -123,4 +123,14 @@ func writeTrigger(dw printers.PrefixWriter, trigger *v1beta1.Trigger, printDetai
 			subWriter.WriteAttribute(key, value)
 		}
 	}
+	if len(trigger.Spec.Filters) > 0 {
```
dsimansk, line 126: This should handle all nested filters and display accordingly.

pkg/kn/commands/trigger/describe.go
```diff
@@ -123,4 +124,50 @@ func writeTrigger(dw printers.PrefixWriter, trigger *v1beta1.Trigger, printDetai
...
+				}
+			}
+		}
+
+		// Call write recursively for struct SubscriptionsAPIFilter of field: Not
+		if fieldValue.Kind() == reflect.Struct {
+			writeNestedFilters(dw, fieldValue.Interface().(v1beta1.SubscriptionsAPIFilter))
+		}
+	}
 }
```
pierDipi, line 173: you have the best context here whether reflection is the best way to go or not

pierDipi (commented):
It looks good to me overall, I agree with your comments that more unit tests would be necessary

pkg/kn/commands/trigger/describe.go
dsimansk: My intuition was that reflection will make it a bit shorter and cleaner than hardcoded fields. Plus hopefully it will be more future proof any of the fields is added/changed. 
At the end it got a bit out of hand and it's so much shorter etc., but the benefit of having it a bit more dynamic hopefully stayed.

rhuss: That would be also my question, whether it wouldn't be easier to use the fields directly and make it easier to comprehend. When new fields come up, one probably needs to check whether they fit in the layout anyway. Not sure how it looks like, sorry, I'm a bit disconnected. But let me ask GPT-4 ...

dsimansk: @rhuss here's the struct to show: https://github.com/knative/eventing/blob/main/pkg/apis/eventing/v1/trigger_types.go#L124

rhuss: > Can you tell me what the new method does and can give me a sample output what it would print ?

image
image

rhuss: This version doesn't entirely cut it, but I think I would prefer the simpler version that is more hardcoded as the reflect-based version, while flexible, is probably a bit over the top as I think one would need to adjust the output anyway if the filter gets more complex.

dsimansk: I've pushed revamped version with exact fields + unit test. https://github.com/knative/client/pull/1794/commits/0e2e46154ab35b24ff163ebd5c4f2c94f0c82124

dsimansk:
@pierDipi would you mind getting back to the PR?

pierDipi (approved):
This looks good to me!

/lgtm
/approve

PR #1796, opened 2023-03-23
fix: Fix image tag for nightly builds

dsimansk:
## Changes

* fix: Fix image tag for nightly builds

That's a follow up to my previous PR that added check around "latest" tag. We need those on release branches only, nightly release for main should always be tagged latest.

## Reference

Follow-up https://github.com/knative/client/pull/1792

**Release Note**

```release-note
NONE
```

/cc @rhuss

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -56,7 +56,7 @@ function build_release() {
       if ((current_minor >= latest_minor)); then
         echo "Newer patch release - publish latest image tag"
       else
-        echo "Patch release of older minor version - do not publish lates image tag"
+        echo "Patch release of older minor version - do not publish latest image tag"
         KO_FLAGS=$KO_FLAGS" --tags \"\""
       fi
     fi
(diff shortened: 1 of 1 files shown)


hack/release.sh
```diff
@@ -47,16 +47,18 @@ function build_release() {
...
+      echo "Newest .0 release - publish latest image tag"
     else
-      echo "Patch release of older minor version - do not publish lates image tag"
-      KO_FLAGS=$KO_FLAGS" --tags \"\""
+      local latest_minor=$(minor_version "$(latest_version)")
+      local current_minor=$(minor_version "${TAG}")
+      if ((current_minor >= latest_minor)); then
+        echo "Newer patch release - publish latest image tag"
+      else
+        echo "Patch release of older minor version - do not publish lates image tag"
```
rhuss, line 59:
```suggestion
        echo "Patch release of older minor version - do not publish latest image tag"
```

PR #1816, opened 2023-05-24
Fix plugin inlining that uses client-pkg dependency

dsimansk:
## Changes

* Fix plugin inlining that uses client-pkg dependency

This moves the initialization to new `client-pkg`, otherwise the inlined plugins depending on client-pkg aren't recognized as expected.

/cc @rhuss @vyasgun 

## Reference

Fixes #

**Release Note**

```release-note
Fix plugin inlining that uses client-pkg dependency
```


rhuss:
IIUC this is about moving the pluginmanager to `pkg`, so that plugins can add themselves to that global plugin array (because now they don't have a dependency to the client anymore, which is what we want).

If this is correct, the chance makes totally sense. However, I would also remove  moved parts from the client hier as well.

dsimansk:
@vyasgun sorry, rebased now. :)

dsimansk:
@vyasgun @rhuss if you may pls. I had to rebase due to other fix within e2e tests. Thanks!

dsimansk:
@vyasgun I went with a different approach, rather make both lists work for a while as a backward compatibility solution.

@rhuss per our chat, that's upstream fix.

cmd/kn/main.go
```diff
@@ -62,6 +64,11 @@ func run(args []string) error {
 
 	pluginManager := plugin.NewManager(config.GlobalConfig.PluginsDir(), config.GlobalConfig.LookupPluginsInPath())
 
+	// Merge plugins that implements client-pkg Plugin type to the same manager instance
+	for _, p := range pkgplugin.InternalPlugins {
+		pluginManager.AppendPlugin(p)
+	}
+
```
dsimansk, line 71: This part is important! 

There're 2 gloabal variables: 
- `client` `InternalPlugins`
- `client-pkg` `InternalPlugins`

Depending in the plugin impl both are still valid, either by using new depedency or former on client repo. Therefore there's list merge to a single one that is used onward. I.e. content of `client-pkg` list is merged into `client` list.

cmd/kn/main.go
```diff
@@ -62,6 +64,11 @@ func run(args []string) error {
 
 	pluginManager := plugin.NewManager(config.GlobalConfig.PluginsDir(), config.GlobalConfig.LookupPluginsInPath())
 
+	// Merge plugins that implements client-pkg Plugin type to the same manager instance
+	for _, p := range pkgplugin.InternalPlugins {
```
rhuss, line 68: I wonder whether we shouldn't put this internally in plugin.NewManager() where it feels more natural to me. It could be a lazy initialization (i.e. do only once regardless how often NewManager() is called). It would also keep `main()` a bit cleaner.

dsimansk: https://github.com/knative/client/pull/1816/commits/bfaf3f94b2acdf21505b7bccdfb19362d9eda66b

dsimansk: Does it look better now? Agreed on moving the scope to constructor.

dsimansk: @rhuss ping if you get to it.

PR #1820, opened 2023-06-22
Fix broker condition check

dsimansk:
## Description

This's changed on the Eventing end with latest nightly versions.

/cc @vyasgun @rhuss 

## Changes

* Fix broker condition check

## Reference

Fixes #

**Release Note**

```release-note
NONE
```

### DIFF
--- a/lib/test/eventtype.go
+++ b/lib/test/eventtype.go
@@ -44,7 +44,7 @@ func EventtypeList(r *KnRunResultCollector, eventtypes ...string) {
 func EventtypeDescribe(r *KnRunResultCollector, name string) {
 	out := r.KnTest().Kn().Run("eventtype", "describe", name)
 	r.AssertNoError(out)
-	assert.Check(r.T(), util.ContainsAll(out.Stdout, name, r.KnTest().Kn().Namespace(), "Ready", "BrokerReady"))
+	assert.Check(r.T(), util.ContainsAll(out.Stdout, name, r.KnTest().Kn().Namespace(), "Ready", "BrokerExists"))
 }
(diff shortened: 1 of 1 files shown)


lib/test/eventtype.go
```diff
@@ -44,7 +44,7 @@ func EventtypeList(r *KnRunResultCollector, eventtypes ...string) {
 func EventtypeDescribe(r *KnRunResultCollector, name string) {
 	out := r.KnTest().Kn().Run("eventtype", "describe", name)
 	r.AssertNoError(out)
-	assert.Check(r.T(), util.ContainsAll(out.Stdout, name, r.KnTest().Kn().Namespace(), "Ready", "BrokerReady"))
+	assert.Check(r.T(), util.ContainsAll(out.Stdout, name, r.KnTest().Kn().Namespace(), "Ready", "BrokerExists"))
```
matzew, line 47: It should be `ReferenceExists`:
https://github.com/knative/eventing/blob/main/pkg/apis/eventing/v1beta2/eventtype_lifecycle.go#L27

Since the `v1beta2` version is using a `KReference`, which basically allows us to point to any `KResource` in that sense

dsimansk: Yep, I'll do the switch in a follow-up PR.

dsimansk: I see the `default` broker is gone and that's probably another problem for this test. I'll refactor to v1beta2 a bit sooner than a follow-up PR. :)

dsimansk:
@vyasgun if you may pls.

PR #1821, opened 2023-06-26
Add default SecurityContext to every new ksvc

dsimansk:
## Description

This's handling a warning from API-server validation about SecurityContext being empty for `kn service create` and related cmds. 

I'll look into a few further improvements. E.g. either adding SecurityContext builder pattern or at least plain `--security-context seccon.yaml` flag to make it customizable.

 /cc @rhuss @vyasgun 

## Changes

* Add default SecurityContext to every new ksvc
*
*

## Reference

Fixes #

**Release Note**

```release-note
Add default SecurityContext to every new ksvc
```


pkg/kn/flags/podspec_helper.go
```diff
@@ -378,6 +380,26 @@ func UpdateImagePullPolicy(spec *corev1.PodSpec, imagePullPolicy string) error {
...
+	container := containerOfPodSpec(spec)
+	if container.SecurityContext == nil {
+		container.SecurityContext = DefaultSecCon()
+	}
+}
+
+func DefaultSecCon() *corev1.SecurityContext {
+	return &corev1.SecurityContext{
+		AllowPrivilegeEscalation: pointer.Bool(false),
+		RunAsNonRoot:             pointer.Bool(true),
```
rhuss, line 393: My question would be if this does not break the existing setup and whether this is not too restrictive. I mean, there are still tons of container images that can only run as root and/or need certain capabilities. Having such a strict default (that currently can't be changed), is quite drastic. I think we should only add this if there is at least an option to get back the previous behaviour, i.e. having no `SecurityContext` at all. This then could be documented as a change (which is still not backwards compatible). This option could be called `--no-security-context` or maybe even better something like `--security-context=none` (with the alternatives like `--security-context=strict` (the default) and `--security-context=privilegeEscalation:allow,runAsNonRoot:true,dropCapabilities:NET` for fine grained security context configuration (this we can add later but `strict` and `none` should be already along with this PR.

rhuss:
Please see my comment https://github.com/knative/client/pull/1821/files#r1242163943 that we should already add basic option handling here. Otherwise, I think it would break too much.

Also note the linter warnings.

pkg/kn/flags/podspec_helper.go
dsimansk: That's at least ready with on/off switch. https://github.com/knative/client/pull/1821/commits/0fcf8000185bdf24ed98893d8b103b494e846ade

pkg/kn/flags/podspec.go
```diff
@@ -234,6 +236,10 @@ func (p *PodSpecFlags) AddFlags(flagset *pflag.FlagSet) []string {
 	flagNames = append(flagNames, "pull-secret")
 	flagset.Int64VarP(&p.User, "user", "", 0, "The user ID to run the container (e.g., 1001).")
 	flagNames = append(flagNames, "user")
+
+	flagset.StringVar(&p.SecurityContext, "security-context", "strict", "Security Context definition to be added the service. Accepted values: strict | none.")
```
rhuss, line 240:
```suggestion
	flagset.StringVar(&p.SecurityContext, "security-context", "strict", "Predefined security context for the service. Accepted values: 'none' for no security context and 'strict' for dropping all capabilities, running as non-root, and no privilege escalation.")
```

pkg/kn/flags/podspec_helper.go
```diff
@@ -378,6 +380,42 @@ func UpdateImagePullPolicy(spec *corev1.PodSpec, imagePullPolicy string) error {
...
+	container := containerOfPodSpec(spec)
+	switch strings.ToLower(securityContext) {
+	case "none":
+		// Blank any Security Context defined
+		container.SecurityContext = &corev1.SecurityContext{}
+	case "strict":
+		// Add or update Security Context to default strict
+		container.SecurityContext = DefaultStrictSecCon()
+	case "":
+		// Add default strict SC flag is not used, hence empty value
```
rhuss, line 394: Can this be the case if a default value is set to "strict" ?

rhuss (approved):
Looks good to me, with some minor nits.

PR #1829, opened 2023-07-20
Add support for Eventing/v1beta2 EventTypes

### LINKED ISSUE #1818: CLI: Add support for EventType v1beta2
The EventType is now available in version `v1beta2`, and this needs to be reflected by the client

dsimansk:
## Changes

* 🎁 New Add support for Eventing/v1beta2 EventTypes

## Reference

Fixes #1818 

/cc @matzew @rhuss 

**Release Note**

```release-note
Add support for Eventing/v1beta2 EventTypes
```


pkg/kn/commands/flags/eventtype.go
```diff
@@ -25,6 +25,6 @@ type EventtypeFlags struct {
 func (e *EventtypeFlags) Add(cmd *cobra.Command) {
 	cmd.Flags().StringVarP(&e.Type, "type", "t", "", "Cloud Event type")
 	cmd.Flags().StringVar(&e.Source, "source", "", "Cloud Event source")
-	cmd.Flags().StringVarP(&e.Broker, "broker", "b", "", "Cloud Event broker")
+	cmd.Flags().StringVarP(&e.Broker, "broker", "b", "", "Cloud Event Broker. This flag is added for the convenience, since Eventing v1beta2 brokers as represented as KReference type.")
```
rhuss, line 28: I think this clarification is more confusing than it helps. I don't think that the user know ad hoc what a "KReference type" is. For the help message it would be better to use something like "Deprecated. Please use --reference instead", or if its not deprecated, I just would keep it as is. Anything else is just confusing,.

pkg/kn/commands/eventtype/create.go
```diff
@@ -82,6 +104,14 @@ func NewEventtypeCreateCommand(p *commands.KnParams) *cobra.Command {
 	}
 	commands.AddNamespaceFlags(cmd.Flags(), false)
 
+	referenceFlag.AddWithFlagName(cmd, "reference", "r")
+	flag := "reference"
+	cmd.Flag(flag).Usage = "Addressable Reference producing events. " +
+		"You can specify a broker, channel, or fully qualified GVK resource. " +
+		"Examples: '" + flag + " broker:nest' for a broker 'nest', " +
+		"'" + flag + " channel:pipe' for a channel 'pipe'."
+	//TODO: craft fqdn GVK example.
```
rhuss, line 113: Important to not forget it here, since using a GVK is the most complex way to specify this reference so there should be an example for sure.

dsimansk: From my e2e tests it seems there's an issue with GVK. At least it's not working for `EventTypes`. I'll be certainly taking a bit in-depth look.

pkg/kn/commands/flags/sink.go
```diff
@@ -49,26 +56,32 @@ func (i *SinkFlags) AddWithFlagName(cmd *cobra.Command, fname, short string) {
 		"'" + flag + " ksvc:mysvc:mynamespace' for a Knative service 'mysvc' in another namespace 'mynamespace', " +
 		"'" + flag + " https://event.receiver.uri' for an URI with an 'http://' or 'https://' schema, " +
 		"'" + flag + " ksvc:receiver' or simply '" + flag + " receiver' for a Knative service 'receiver' in the current namespace. " +
+		"'" + flag + " special.eventing.dev/v1alpha1/channels:pipe' for GroupVersionResource of v1alpha1 'pipe'. " +
+		"GroupVersionResource requires resource name in a lower case plural form. The inputs are sanitized on best effort basis, " +
+		"but in case of 'resource not found' error it may help to double check and correct the input. " +
```
dsimansk, line 61: @rhuss that's the case size is a culprit of my initial "bug" hunch. Can you please take a look at wording since you're master writer please?

dsimansk:
It seems very strange that `build-tests_client_main` keeps failing without any meaningful error being reported. I'll close this PR and try to create a fresh one, it seems like something is out sync in Prow for some reason.

PR #1831, opened 2023-07-21
Add support for Eventing/v1beta2 EventTypes

dsimansk:
## Changes

* Add support for Eventing/v1beta2 EventTypes

## Reference

Fixes #

**Release Note**

```release-note
Add support for Eventing/v1beta2 EventTypes with `--reference` option flag
```

/cc @rhuss @matzew


test/e2e/eventtype_test.go
```diff
@@ -37,6 +38,10 @@ const (
 )
 
 func TestEventtype(t *testing.T) {
+	//FIXME: enable after Eventing v1.11 is out
```
rhuss, line 41: How can we ensure that we don't forget this ?

dsimansk: Yep, it's a bit tedious. But the release is this week, I'll try not to foget to remove it immediately.

pkg/kn/commands/flags/sink.go
```diff
@@ -49,26 +57,32 @@ func (i *SinkFlags) AddWithFlagName(cmd *cobra.Command, fname, short string) {
 		"'" + flag + " ksvc:mysvc:mynamespace' for a Knative service 'mysvc' in another namespace 'mynamespace', " +
 		"'" + flag + " https://event.receiver.uri' for an URI with an 'http://' or 'https://' schema, " +
 		"'" + flag + " ksvc:receiver' or simply '" + flag + " receiver' for a Knative service 'receiver' in the current namespace. " +
+		"'" + flag + " special.eventing.dev/v1alpha1/channels:pipe' for GroupVersionResource of v1alpha1 'pipe'. " +
+		"GroupVersionResource requires resource name in a lower case plural form. The inputs are sanitized on best effort basis, " +
```
rhuss, line 61: "requires resource name in a lower case plural form" --> Can't we do at least the lower casing on our own so that that case doesn't mind ? Also the plural form is not intuitive if you don't know about resources. I would accept also both, naively assuming that plural form mean adding a `s` (or is there an easier way to find out ?)

The overall help message now became quite a bit lengthy and since it is used for every sink type and some commands have multiple sink options (eg `--sink`, `--sink-reply`, `--sink-dead-letter`), the help page becomes too long.

We might:

* Shorten it and reference to website for the full documentation.
* Try to keep it but only keep the absolutely necessary information.

Eg. instead of 

```
-s, --sink string               Addressable sink for events. You can specify a broker, channel, Knative service or URI. Examples: '--sink broker:nest' for a broker 'nest', '--sink channel:pipe' for a channel 'pipe', '--sink ksvc:mysvc:mynamespace' for a Knative service 'mysvc' in another namespace 'mynamespace', '--sink https://event.receiver.uri' for an URI with an 'http://' or 'https://' schema, '--sink ksvc:receiver' or simply '--sink receiver' for a Knative service 'receiver' in the current namespace. '--sink special.eventing.dev/v1alpha1/channels:pipe' for GroupVersionResource of v1alpha1 'pipe'. GroupVersionResource requires resource name in a lower case plural form. The inputs are sanitized on best effort basis, but in case of 'resource not found' error it may help to double check and correct the input. If a prefix is not provided, it is considered as a Knative service in the current namespace. If referring to a Knative service in another namespace, 'ksvc:name:namespace' combination must be provided explicitly.
```

we might use something like

```
-s, --sink string               Addressable sink for events. Use one of these formats: broker:<name>, channel:<name>, ksvc:<service>:<namespace>, <URI>, or simply <service>. Examples: '--sink broker:nest', '--sink ksvc:mysvc:mynamespace', '--sink https://event.receiver.uri', '--sink receiver'. If the prefix is omitted, it's assumed to be a Knative service in the current namespace. Also supports GroupVersionResource format like 'special.eventing.dev/v1alpha1/channels:pipe'. On errors like 'resource not found', verify your input format.
```

dsimansk: I agree, it was a bit long even without this change. Sure, the code tries to do both lower case of course and add naive `+s`. However, we never explained anything from that. I kind of like the idea of a little how to use sinks page with a couple of examples. 

I can move the plural form recommendation into error message maybe?

rhuss: Maybe we might leave of the error explanation, too ?

dsimansk: OK, let me remove the additional lines and keep just GVR example to check how that looks.

docs/cmd/kn_eventtype_create.md
```diff
@@ -21,9 +21,10 @@ kn eventtype create
 ### Options
 
 ```
-  -b, --broker string      Cloud Event broker
+  -b, --broker string      Cloud Event Broker. This flag is added for the convenience, since Eventing v1beta2 brokers as represented as KReference type.
   -h, --help               help for create
   -n, --namespace string   Specify the namespace to operate in.
+  -r, --reference string   Addressable Reference producing events. You can specify a broker, channel, or fully qualified GroupVersionResource (GVR). Examples: 'reference broker:nest' for a broker 'nest', 'reference channel:pipe' for a channel 'pipe', 'reference special.eventing.dev/v1alpha1/channels:pipe' for GroupVersionResource of v1alpha1 'pipe'.
```
rhuss, line 27: Here the intial `--` is missing for `reference`. Also maybe the short version would be nice (`-r channel:pipe`)

docs/cmd/kn_eventtype_create.md
```diff
@@ -21,9 +21,10 @@ kn eventtype create
 ### Options
 
 ```
-  -b, --broker string      Cloud Event broker
+  -b, --broker string      Cloud Event Broker. This flag is added for the convenience, since Eventing v1beta2 brokers as represented as KReference type.
```
rhuss, line 24: Still, I would remove the "is added for convenience" since this is an implementation detail that is confusing for a help page.

docs/cmd/kn_source_apiserver_create.md
```diff
@@ -26,7 +26,7 @@ kn source apiserver create NAME --resource RESOURCE --sink SINK
...
+  -s, --sink string               Addressable sink for events. You can specify a broker, channel, Knative service or URI. Examples: '--sink broker:nest' for a broker 'nest', '--sink channel:pipe' for a channel 'pipe', '--sink ksvc:mysvc:mynamespace' for a Knative service 'mysvc' in another namespace 'mynamespace', '--sink https://event.receiver.uri' for an URI with an 'http://' or 'https://' schema, '--sink ksvc:receiver' or simply '--sink receiver' for a Knative service 'receiver' in the current namespace. '--sink special.eventing.dev/v1alpha1/channels:pipe' for GroupVersionResource of v1alpha1 'pipe'. If a prefix is not provided, it is considered as a Knative service in the current namespace. If referring to a Knative service in another namespace, 'ksvc:name:namespace' combination must be provided explicitly.
```
rhuss, line 29: I would remove `If referring to a Knative service in another namespace, 'ksvc:name:namespace' combination must be provided explicitly.`

rhuss: `with an 'http://' or 'https://' schema` might be overkill. Maybe only "for an HTTP URI" ?

dsimansk: https://github.com/knative/client/pull/1831/commits/1a0fc480d5176b296bbb08fb16eba9c94ccd4af4

dsimansk: @rhuss can we merge it? asking because we're in a release day already.

dsimansk:
@knative/knative-release-leads FYI this should be merged for v1.11 release.

rhuss:
Looks good to me, I added a final comment to the wording. Feel free to merge nevertheless.

/approve
/lgtm
/hold

docs/cmd/kn_eventtype_create.md
dsimansk: I removed the whole sentence. Indeed it's an implementation detail.

PR #1855, opened 2023-08-22
Context Sharing POC

dsimansk:
## Description

This's a bit simplified approach from original Feature Track proposal. I wonder if we should simply start with a few well known keys, that can be consumed in `kn` core + plugins

## Changes

*
*
*

## Reference

Fixes #

**Release Note**

```release-note
TBD
```

/cc @rhuss

### DIFF
--- a/pkg/kn/commands/service/describe.go
+++ b/pkg/kn/commands/service/describe.go
@@ -19,7 +19,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"knative.dev/client/pkg/kn/plugin"
 	"sort"
 	"strconv"
 	"strings"
@@ -29,6 +28,7 @@ import (
 	"knative.dev/serving/pkg/apis/serving"
 
 	"knative.dev/client/pkg/kn/commands/revision"
+	"knative.dev/client/pkg/kn/plugin"
 	"knative.dev/client/pkg/printers"
 	clientservingv1 "knative.dev/client/pkg/serving/v1"
 
--- a/pkg/kn/commands/service/describe.go
+++ b/pkg/kn/commands/service/describe.go
@@ -19,7 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
-
+	"knative.dev/client/pkg/kn/plugin"
 	"sort"
 	"strconv"
 	"strings"
@@ -102,8 +102,12 @@ func NewServiceDescribeCommand(p *commands.KnParams) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			var serviceName string
 			if len(args) != 1 {
-				if cmd.Context() != nil && cmd.Context().Value("service") != nil {
-					serviceName = cmd.Context().Value("service").(string)
+				if plugin.CtxManager != nil {
+					data, err := plugin.CtxManager.FetchContextData()
+					if err != nil {
+						return err
+					}
+					serviceName = data["service"]
 				}
 				if serviceName == "" {
 					return errors.New("'service describe' requires the service name given as single argument")
--- a/pkg/kn/commands/service/describe.go
+++ b/pkg/kn/commands/service/describe.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+
 	"sort"
 	"strconv"
 	"strings"
@@ -99,10 +100,17 @@ func NewServiceDescribeCommand(p *commands.KnParams) *cobra.Command {
 		Example:           describe_example,
 		ValidArgsFunction: commands.ResourceNameCompletionFunc(p),
 		RunE: func(cmd *cobra.Command, args []string) error {
+			var serviceName string
 			if len(args) != 1 {
-				return errors.New("'service describe' requires the service name given as single argument")
+				if cmd.Context() != nil && cmd.Context().Value("service") != nil {
+					serviceName = cmd.Context().Value("service").(string)
+				}
+				if serviceName == "" {
+					return errors.New("'service describe' requires the service name given as single argument")
+				}
+			} else {
+				serviceName = args[0]
 			}
-			serviceName := args[0]
 
 			namespace, err := p.GetNamespace(cmd)
 			if err != nil {
@@ -182,7 +190,7 @@ func describe(w io.Writer, service *servingv1.Service, revisions []*revisionDesc
 	return nil
 }
 
-// Write out main service information. Use colors for major items.
+// WriteCache out main service information. Use colors for major items.
 func writeService(dw printers.PrefixWriter, service *servingv1.Service) {
 	commands.WriteMetadata(dw, &service.ObjectMeta, printDetails)
 	dw.WriteAttribute("URL", extractURL(service))
@@ -200,7 +208,7 @@ func writeService(dw printers.PrefixWriter, service *servingv1.Service) {
 	}
 }
 
-// Write out revisions associated with this service. By default only active
+// WriteCache out revisions associated with this service. By default only active
 // target revisions are printed, but with --verbose also inactive revisions
 // created by this services are shown
 func writeRevisions(dw printers.PrefixWriter, revisions []*revisionDesc, printDetails bool) {
--- a/pkg/kn/plugin/context_sharing.go
+++ b/pkg/kn/plugin/context_sharing.go
@@ -1,0 +1,225 @@
+// Copyright © 2023 The Knative Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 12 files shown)


pkg/kn/plugin/context.go
```diff
@@ -0,0 +1,59 @@
...
+		//file, err := os.Open("~/.config/kn/context-cache.json")
+		//if err != nil {
+		//	return nil, err
+		//}
+
+		mockData := `
+{  
+  "service":"hello",
+  "namespace":"default"
+}
```
dsimansk, line 41: This represents very minimal shared context between every Consumer and Producer to read/write. E.g. in a seems floating value changed on latest usage of `kn` or plugins.

dsimansk: This could be further split into plugin specific sub-context and shared sub-context. Where plugin specific is only accessed by it's owner. In addition with read/write permission field maybe?

pkg/kn/plugin/context.go
```diff
@@ -0,0 +1,59 @@
...
+			return nil, err
+		}
+		ctxManager = &ContextDataManager{
+			ContextData: contextData,
+		}
+	}
+	return ctxManager, nil
+}
+
+func (c *ContextDataManager) Find(key string) string {
```
dsimansk, line 57: Then defining the API as a set of access methods to retrieve and store the well-known values.

rhuss:
@dsimansk thanks for this initial work. I wouldn't mind to restrict ourselves to a set of fixed keys, but I'm not sure what this will buy us. Of course, the core commands themselves should check the Context for any mandatory argument that the user has not provided. These are are well known keys like "service" or "namespace". However, I would not prohibit that a plugin could add any key to the context manager, as this is not harmful as the core commands never will lookup this. However _other_ plugins might make use of it, so that e.g. the function plugin could communicate with the event plugin.

The tricky part really is, how external, non-inlined plugins can add to the context. For inline plugins, this is trivial as they just can add to a global map. For external plugins. that would need to an external call. For optimization, the core should know which plugin supports which context key, therefor the idea of the manifest that is obtained once during startup by calling every plugin that is reachable. This might be dangerous if plugins are looked up from the path and there are some dangling plugins or bogus plugins that would be called. So not really sure if this is the proper way. Alternatively we could keep a central registry remotely that contains this information in a plain file. We could fetch this once from a central place (e.g. GitHub) or fallback to a default one that we have compile into kn. This information ("which plugin supports which keys") should be quite static and don't changes very often.

rhuss:
on the other hand, we also could reconsider to implement the "sharing via a local file" first, too. Not sure how far we should go.

docs/cmd/kn.md
```diff
@@ -6,7 +6,7 @@ kn manages Knative Serving and Eventing resources
 
 kn is the command line interface for managing Knative Serving and Eventing resources
 
-Find more information about Knative at: https://knative.dev
+GetContext more information about Knative at: https://knative.dev
```
dsimansk, line 9: That's Goland refactore everything unwanted change.

pkg/kn/plugin/context_sharing.go
```diff
@@ -0,0 +1,181 @@
...
+			// Add manifest to map
+			c.Manifests[p.Name()] = *manifest
+		}
+	}
+	return nil
+}
+
+// TODO: We should cautiously execute external binaries
+// fetchExternalManifest returns Manifest from external plugin by exec `$plugin manifest get`
+func fetchExternalManifest(p Plugin) *Manifest {
```
dsimansk, line 151: That's a bit tricky part. Executing binary half blind if it's desired one even?

pkg/kn/commands/service/describe.go
```diff
@@ -99,10 +101,16 @@ func NewServiceDescribeCommand(p *commands.KnParams) *cobra.Command {
...
 		ValidArgsFunction: commands.ResourceNameCompletionFunc(p),
 		RunE: func(cmd *cobra.Command, args []string) error {
+			var serviceName string
 			if len(args) != 1 {
-				return errors.New("'service describe' requires the service name given as single argument")
+				c, _ := plugin.NewContextManager()
+				serviceName = c.GetDefault()["service"]
+				if serviceName == "" {
+					return errors.New("'service describe' requires the service name given as single argument")
+				}
```
dsimansk, line 110: @rhuss what would you say about using Golang's build-in Context since `map[string]string` is our context now? We can pass it through `ExecuteWithContext()` function that's already part of cobra.

Or shoud we be a bit more bold here and inject the params directly into args & flags?

pkg/kn/plugin/context_sharing.go
dsimansk: Preview of manifest flag: https://github.com/knative-extensions/kn-plugin-service-log/commit/c214f72c49c7da7179a6869a9af8e9bc38785f3f

That can be a standalone cmd. I.e. `kn service log manifest`, but it kind of makes single purpose plugins grow the list of cmds. :)

rhuss:
Thanks! I hope i can review it today, if not, Iwill do it early next week.

pkg/kn/plugin/context_sharing.go
rhuss: Indeed, the tricky part is not blindly calling each plugin in the path somehow, if possible. Of course *if* a plugin is in the path, it can be called anyway (maybe by accident), but chances are higher that a malicious plugin is called when it is called by itself. A simple solution could be to keep track of which plugins have been called explicitly and only can them when picking up all manifests (which might be hard to maintain, though ...)

dsimansk:
@rhuss I'll be adding unit test, but it's ready for a review and feedback I believe.

pkg/kn/plugin/context_sharing.go
```diff
@@ -0,0 +1,225 @@
...
+
+		if manifest.HasManifest {
+			c.populateDataKeys(manifest, pluginName)
+		}
+	}
+	plugins, err := pluginManager.ListPlugins()
+	if err != nil {
+		return err
+	}
+	for _, plugin := range plugins {
```
rhuss, line 157: I would be also cool, to remove plugins from the cache that are not listed anymore. So checking which keys of the c.Manifests[..] have been touched and remove at the end every entry with a key that is neither an internal or external plugin.

pkg/kn/plugin/manager.go
```diff
@@ -125,6 +105,10 @@ func (manager *Manager) AppendPlugin(plugin Plugin) {
 	InternalPlugins = append(InternalPlugins, plugin)
 }
 
+func (manager *Manager) GetInternalPlugins() PluginList {
```
rhuss, line 108: Usually you don't need a getter if you just return a exported variable. So, I would make either `InternalPlugins` private (and then call the method just `InternalPlugins()` which is probably more idiomatic go), or just access the public exported field when using the InternalPlugins (strange feeling for us Java guys, I know ;-)

rhuss (commented):
Thank, I did a first round of review. Beside the usual bike shedding some more substantial comments inline:

* Consider cleaning the cache, too, if some external plugin is not in the list anymore.
* Only do the lookup dance and external call to a plugin if really needed. E.g. it should be lazily done, e.g. when the `service describe` does not find a `service` name provided, instead of doing it always. That would keep the current behavior and only fallback to a context lookup when needed.

pkg/kn/plugin/manager.go
dsimansk: Yep, it still kind of nicer to have a getter. I'll get rid of it though. From the compatibility perspective we can't make `InternalPlugins` private due to inlined plugins appends.

dsimansk:
A corresponding client-pkg with types at least: https://github.com/knative/client-pkg/pull/127.

pkg/kn/commands/service/describe.go
```diff
@@ -99,10 +100,21 @@ func NewServiceDescribeCommand(p *commands.KnParams) *cobra.Command {
...
+					if err != nil {
+						return err
+					}
+					serviceName = data["service"]
+				}
+				if serviceName == "" {
+					return errors.New("'service describe' requires the service name given as single argument")
+				}
+			} else {
+				serviceName = args[0]
```
rhuss, line 116: I'm not sure, but I have the feeling that his access pattern will be repeated in several commands (all that require a service name), so maybe we could extract this handling in a utility method. Usually, I wouldn't opt for this, but this looks complex enough to avoid a C&P approach.

rhuss (commented):
This goes definitely in the write direction. See some comments inline (and the other comments I did before).

rhuss:
Let's get it merged and iterate from there on. Not sure how to preserve the unresolved comments for the next update. But I think they are still kind of valid.

/approve
/lgtm
/hold

PR #1860, opened 2023-08-30
Fix shellcheck warnings in hack/build.sh script

### LINKED ISSUE #1859: Fix shellcheck warnings in hack/build.sh script
There are a few warnings in our `./hack/build.sh` script that would be nice to fix. It might pop-up in case of changing the script in the future.

There are several more with `INFO` severity level. Those can be ignored for now. 

```
➜  client git:(main) shellcheck -S warning hack/build.sh

In hack/build.sh line 40:
  pushd $(basedir) >/dev/null 2>&1
        ^--------^ SC2046 (warning): Quote this to prevent word splitting.


In hack/build.sh line 43:
  if $(has_flag --help -h); then
     ^-------------------^ SC2091 (warning): Remove surrounding $() to avoid executing output (or use eval if intentional).


In hack/build.sh line 48:
  if $(has_flag --watch -w); then
     ^--------------------^ SC2091 (warning): Remove surrounding $() to avoid executing output (or use eval if intentional).


In hack/build.sh line 52:
    if $(has_flag --test -t); then
       ^-------------------^ SC2091 (warning): Remove surrounding $() to avoid executing output (or use eval if intentional).


In hack/build.sh line 61:
  if $(has_flag --fast -f); then
     ^-------------------^ SC2091 (warning): Remove surrounding $() to avoid executing output (or use eval if intentional).


In hack/build.sh line 64:
    if $(has_flag --test -t); then
       ^-------------------^ SC2091 (warning): Remove surrounding $() to avoid executing output (or use eval if intentional).


In hack/build.sh line 71:
  if $(has_flag --test -t); then
     ^-------------------^ SC2091 (warning): Remove surrounding $() to avoid executing output (or use eval if intentional).


In hack/build.sh line 77:
  if $(has_flag --codegen -c); then
     ^----------------------^ SC2091 (warning): Remove surrounding $() to avoid executing output (or use eval if intentional).


In hack/build.sh line 83:
  if $(has_flag --all -x); then
     ^------------------^ SC2091 (warning): Remove surrounding $() to avoid executing output (or use eval if intentional).


In hack/build.sh line 118:
  find $(echo "$source_dirs") -name "*.go" -print0 | xargs -0 gofmt -s -w
       ^--------------------^ SC2046 (warning): Quote this to prevent word splitting.


In hack/build.sh line 123:
  go_run "golang.org/x/tools/cmd/goimports@${GOIMPORTS_VERSION}" -w $(echo "$source_dirs")
                                                                    ^--------------------^ SC2046 (warning): Quote this to prevent word splitting.


In hack/build.sh line 124:
  find $(echo "$source_dirs") -name "*.go" -print0 | xargs -0 gofmt -s -w
       ^--------------------^ SC2046 (warning): Quote this to prevent word splitting.


In hack/build.sh line 139:
  if $(file kn | grep -q -i "Windows"); then
     ^-- SC2091 (warning): Remove surrounding $() to avoid executing output (or use eval if intentional).


In hack/build.sh line 199:
  $(basedir)/hack/update-deps.sh
  ^--------^ SC2046 (warning): Quote this to prevent word splitting.


In hack/build.sh line 211:
    local fswatch_opts='-e "^\..*$" -o pkg cmd'
                       ^----------------------^ SC2089 (warning): Quotes/backslashes will be treated literally. Use an array.


In hack/build.sh line 212:
    if $(has_flag --test -t); then
       ^-------------------^ SC2091 (warning): Remove surrounding $() to avoid executing output (or use eval if intentional).


In hack/build.sh line 215:
    if $(has_flag --verbose); then
       ^-------------------^ SC2091 (warning): Remove surrounding $() to avoid executing output (or use eval if intentional).


In hack/build.sh line 231:
    fswatch $fswatch_opts | xargs -n1 -I{} sh -c "$command && echo 👌 OK"
            ^-----------^ SC2090 (warning): Quotes/backslashes in this variable will not be respected.


In hack/build.sh line 261:
    filters="$@"
            ^--^ SC2124 (warning): Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.


In hack/build.sh line 321:
display_help() {
^-- SC2120 (warning): display_help references arguments, but none are ever passed.


In hack/build.sh line 326:
Usage: $(basename $BASH_SOURCE) [... options ...]
                  ^----------^ SC2128 (warning): Expanding an array without an index only gives the first element.


In hack/build.sh line 356:
if $(has_flag --debug); then
   ^-----------------^ SC2091 (warning): Remove surrounding $() to avoid executing output (or use eval if intentional).


In hack/build.sh line 362:
source $(basedir)/vendor/knative.dev/hack/library.sh
       ^--------^ SC2046 (warning): Quote this to prevent word splitting.


In hack/build.sh line 366:
  source "${file}"
         ^-------^ SC1090 (warning): ShellCheck can't follow non-constant source. Use a directive to specify location.


In hack/build.sh line 368:
source $(basedir)/hack/build-flags.sh
       ^--------^ SC2046 (warning): Quote this to prevent word splitting.

For more information:
  https://www.shellcheck.net/wiki/SC1090 -- ShellCheck can't follow non-const...
  https://www.shellcheck.net/wiki/SC2046 -- Quote this to prevent word splitt...
  https://www.shellcheck.net/wiki/SC2089 -- Quotes/backslashes will be treate...
```

/kind cleanup
/kind good-first-issue

xiangpingjiang:
## Changes

* build.sh script

## Reference

Fixes #1859 

**Release Note**

```release-note
Fix shellcheck warnings in `hack/build.sh` script
```

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# shellcheck source=/dev/null
+
 set -o pipefail
 set -o errexit
 set -o nounset
@@ -37,19 +39,19 @@ run() {
   export GO111MODULE=on
 
   # Jump into project directory
-  pushd $(basedir) >/dev/null 2>&1
+  pushd "$(basedir)" >/dev/null 2>&1
 
   # Print help if requested
-  if $(has_flag --help -h); then
-    display_help
+  if has_flag --help -h; then
+    display_help "$@"
     exit 0
   fi
 
-  if $(has_flag --watch -w); then
+  if has_flag --watch -w; then
     # Build and test first
     go_build
 
-    if $(has_flag --test -t); then
+    if has_flag --test -t; then
        go_test
     fi
 
@@ -58,29 +60,29 @@ run() {
   fi
 
   # Fast mode: Only compile and maybe run test
-  if $(has_flag --fast -f); then
+  if has_flag --fast -f; then
     go_build
 
-    if $(has_flag --test -t); then
+    if has_flag --test -t; then
        go_test
     fi
     exit 0
   fi
 
   # Run only tests
-  if $(has_flag --test -t); then
+  if has_flag --test -t; then
     go_test
     exit 0
   fi
 
   # Run only codegen
-  if $(has_flag --codegen -c); then
+  if has_flag --codegen -c; then
     codegen
     exit 0
   fi
 
   # Cross compile only
-  if $(has_flag --all -x); then
+  if has_flag --all -x; then
     cross_build || (echo "✋ Cross platform build failed" && exit 1)
     exit 0
   fi
@@ -115,13 +117,13 @@ codegen() {
 
 go_fmt() {
   echo "🧹 ${S}Format"
-  find $(echo "$source_dirs") -name "*.go" -print0 | xargs -0 gofmt -s -w
+  find "$(echo "$source_dirs")" -name "*.go" -print0 | xargs -0 gofmt -s -w
 }
 
 source_format() {
   set +e
-  go_run "golang.org/x/tools/cmd/goimports@${GOIMPORTS_VERSION}" -w $(echo "$source_dirs")
-  find $(echo "$source_dirs") -name "*.go" -print0 | xargs -0 gofmt -s -w
+  go_run "golang.org/x/tools/cmd/goimports@${GOIMPORTS_VERSION}" -w "$(echo "$source_dirs")"
+  find "$(echo "$source_dirs")" -name "*.go" -print0 | xargs -0 gofmt -s -w
   set -e
 }
 
@@ -136,7 +138,7 @@ go_build() {
   # Env var exported by hack/build-flags.sh
   go build -mod=vendor -ldflags "${KN_BUILD_LD_FLAGS:-}" -o kn ./cmd/...
 
-  if $(file kn | grep -q -i "Windows"); then
+  if file kn | grep -q -i "Windows"; then
     mv kn kn.exe
   fi
 }
@@ -196,7 +198,7 @@ check_license() {
 
 update_deps() {
   echo "🚒 Update"
-  $(basedir)/hack/update-deps.sh
+  "$(basedir)"/hack/update-deps.sh
 }
 
 generate_docs() {
@@ -209,10 +211,10 @@ generate_docs() {
 watch() {
     local command="./hack/build.sh --fast"
     local fswatch_opts='-e "^\..*$" -o pkg cmd'
-    if $(has_flag --test -t); then
+    if has_flag --test -t; then
       command="$command --test"
     fi
-    if $(has_flag --verbose); then
+    if has_flag --verbose; then
       fswatch_opts="$fswatch_opts -v"
     fi
     set +e
@@ -228,7 +230,7 @@ watch() {
     set -e
 
     echo "🔁 Watch"
-    fswatch $fswatch_opts | xargs -n1 -I{} sh -c "$command && echo 👌 OK"
+    fswatch "${fswatch_opts[@]}" | xargs -n1 -I{} sh -c "$command && echo 👌 OK"
 }
 
 # Dir where this script is located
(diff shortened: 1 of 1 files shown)


hack/build.sh
```diff
@@ -58,29 +60,29 @@ run() {
   fi
 
   # Fast mode: Only compile and maybe run test
-  if $(has_flag --fast -f); then
+  if has_flag --fast -f; then
```
rhuss, line 63: This is not as it is supposed to work: `has_flag` _echos_ `true` or `false` and this should be evaluated as a command. So running it in a subshell is the right way todo this construct. The warnings from the linter are bogus. As it is now, it won't work as `if` would evaluate the return code, which is always 0 (so all flags are set). I would recommend to keep as is or change `has_flag` to have a different semantics (but then, why ?).

dsimansk: Shellcheck would be happy with `if eval "$(has_flag --fast -f)"`. Ref: https://www.shellcheck.net/wiki/SC2091

It's a bit tricky in this case with `has_flag` design. But I'd opt for adding eval rather than rewriting the function. IMO it's beyond my initial intention to get rid off `shellcheck` warnings.

Indeed the `has_flag` parts are failing current integration tests execution, because of the flags must be evaluated bug @rhuss described above.

hack/build.sh
```diff
@@ -37,19 +39,19 @@ run() {
   export GO111MODULE=on
 
   # Jump into project directory
-  pushd $(basedir) >/dev/null 2>&1
+  pushd "$(basedir)" >/dev/null 2>&1
 
   # Print help if requested
-  if $(has_flag --help -h); then
+  if eval "$(has_flag --help -h)"; then
```
dsimansk, line 45: @rhuss are you OK with the proposed to solution to fix shellcheck's warning?

hack/build.sh
rhuss: tbh, I would prefer then to adapt `has_flag` to return 0 or 1 and then keep that `$(has_flag ...)` syntax (instead of adding the eval). The whole purpose of the original form was that is was just shorter. Adding an `eval` would counteract this.

hack/build.sh
rhuss: see my comment above, I would prefer to adapt `$has_flag` to return a 0 or 1 (instead of `true` or `false` that then is interpreted as a UNIX command)

dsimansk: @rhuss sure, thanks!

dsimansk: @xiangpingjiang sorry for misleading you with my previous comment. If you don't mind to rework it again a bit within Roland's direction.

hack/build.sh
xiangpingjiang: @rhuss 
If we adapt has_flag return 0 or 1 like below 

```bash
has_flag() {
    filters="$*"
    for var in "${ARGS[@]}"; do
        for filter in $filters; do
          if [ "$var" = "$filter" ]; then
              return 0
          fi
        done
    done
    return 1
}
```

this will work
```bash
if has_flag --help -h; then
    display_help
    exit 0
fi
```


Do we still need keep that $(has_flag ...) syntax ?

rhuss: nope, it's perfectly fine to change completely to your proposal. Thanks!

hack/build.sh
```diff
@@ -258,16 +260,15 @@ basedir() {
 
 # Checks if a flag is present in the arguments.
 has_flag() {
-    filters="$@"
+    filters="$*"
```
rhuss, line 263: I would like to keep "$@" as this will honour quotes for flags, two (e.g. `--flag "my input"`). I'm not sure why shellchecks recommends $* here.

hack/build.sh
```diff
@@ -323,7 +324,7 @@ display_help() {
     cat <<EOT
 Knative client build script
 
-Usage: $(basename $BASH_SOURCE) [... options ...]
+Usage: $(basename "${BASH_SOURCE[@]}") [... options ...]
```
rhuss, line 327: I don't think that makes much sense, as this script is never meant to be sourced  by another script (which only then makes a difference to the original version)

rhuss: So I would keep the original version or use "${BASH_SOURCE[0]}" if this makes shellcheck happier

hack/build.sh
```diff
@@ -352,20 +353,21 @@ Examples:
 * Build cross platform binaries: ..... build.sh --all
 EOT
 }
+display_help "$@"
```
rhuss, line 356: why should the help message always be displayed ? Maybe a debug left-over ?

rhuss:
Some minor comments, otherwise looks good to me. Thanks @xiangpingjiang !

xiangpingjiang:
> Some minor comments, otherwise looks good to me. Thanks @xiangpingjiang !
@rhuss , Thanks for your reviews, I corrected your comments, could you review  my latest  code ?

rhuss:
Thanks! Looks good to me, thanks a lot for patience :-)

/approve
/lgtm

PR #1868, opened 2023-09-20
Distinguish missing Serving/Eventing API from unsupported version

### LINKED ISSUE #1858: Distinguish missing Serving/Eventing API from unsupported version
### Feature request

In the current implementation when `kn` is unable to query Knative resource, we display combined "no or newer" error on missing API or incompatible API. 

Per feedback from users, that might be confusing how to approach the troubleshooting of such issue with the Knative installation. 

```
Error: no or newer Knative Serving API found on the backend, please verify the installation or update the 'kn' client
```

### Use case

There are 2 main scenarios to cover: 

1. There's no Knative Serving/Eventing API found on the cluster. This might be mangled installation, operators still installing, missing CRDs, controllers not up. 

```
Error: no Knative Serving API found on the backend, please verify the installation
```

2. The incompatible API version, e.g. removed `v1alpha1` and therefore backend x `kn` mismatch. 
```
Error: incompatible Knative Serving API found on the backend, please verify the installation or update the 'kn' client to matchin version
```

/cc @rhuss

xiangpingjiang:
## Description
Distinguish missing Serving/Eventing API from unsupported version

## Changes

*
*
*

## Reference

Fixes #1858 

**Release Note**

```release-note

```


pkg/errors/factory.go
```diff
@@ -52,15 +59,53 @@ func newStatusError(err error) error {
...
+			resourceName := getResourceNameFromErrMessage(errAPIStatus.Status().Message)
+			var kubeconfig *string
+			if home := homedir.HomeDir(); home != "" {
+				kubeconfig = flag.String("kubeconfig", filepath.Join(home, ".kube", "config"), "(optional) absolute path to the kubeconfig file")
+			} else {
+				kubeconfig = flag.String("kubeconfig", "", "absolute path to the kubeconfig file")
+			}
+			flag.Parse()
+			// use the current context in kubeconfig
+			config, _ := clientcmd.BuildConfigFromFlags("", *kubeconfig)
```
dsimansk, line 75: I'm not a huge fan of initializing new kube client instance in such a manner. We carry-over `KnParams` instance that holds all the clients commonly used, in addition it covers Kube's ClientConfig parsing, reflecting `--kubeconfig`  and `--context` flags.

I'd rather inject the KnParams into the new error functions and retrieve the needed clients from it.

pkg/errors/errors.go
```diff
@@ -21,12 +21,23 @@ import (
 	"unicode/utf8"
 )
 
-func NewInvalidCRD(apiGroup string) *KNError {
+func NewInvalidCRD(apiGroup string, canFindResource string) *KNError {
```
dsimansk, line 24: How about creating a specific cause functions per message. E.g. `NewInvalidCRDCauseUnknown`, or define cause as a const string? 

It might be easier to understand and consume whenever such new error is going to be added elsewhere etc.

PR #1888, opened 2023-10-31
fix Flaky E2E test: e2e/TestDomain - create domain with TLS

### LINKED ISSUE #1873: Flaky E2E test: e2e/TestDomain - create domain with TLS
The e2e tests are intermittently failing on e2e/TestDomain. In a specific `https` test case. When the URL should contain `https` but is empty. It's most likely a timing issue with empty value being checked prematurely. 

There's a wait loop to wait for URL to be populated before asserting it. It might be not working as expected.  
https://github.com/knative/client/blob/2d0415a7951b99130ccb514fa0e3c5343338225c/test/e2e/domain_mapping_test.go#L136C1-L144C3

Error log:
https://prow.knative.dev/view/gs/knative-prow/logs/continuous_client_main_periodic/1710944161479266304#

/kind good-first-issue

xiangpingjiang:
## Reference

Fixes #1873 

**Release Note**

```release-note

```

### DIFF
--- a/test/e2e/domain_mapping_test.go
+++ b/test/e2e/domain_mapping_test.go
@@ -134,7 +134,7 @@ func domainList(r *test.KnRunResultCollector, domainName string) {
 func domainDescribe(r *test.KnRunResultCollector, domainName string, tls bool) {
 	k := test.NewKubectl(r.KnTest().Kn().Namespace())
 	// Wait for Domain Mapping URL to be populated
-	for i := 0; i < 10; i++ {
+	for i := 0; i < 20; i++ {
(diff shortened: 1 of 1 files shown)


xiangpingjiang:
> /ok-to-test

hello @dsimansk 
Is this change ok, or I need do something else?

dsimansk:
> > /ok-to-test
> 
> hello @dsimansk Is this change ok, or I need do something else?

Well, that's a part to investigate. We can increase the timeout. However, I don't think it's a culprit here. It would be nice to verify that the wait loop is correct, not breaking too early in some edge case etc.

dsimansk:
Let's merge the change and monitor nightly/ other CI runs/

dsimansk (approved):

PR #1904, opened 2023-12-07
Added basic version of quiet mode to service create command

sreeram-venkitesh:
## Description

This PR was made as part of knative 48h hack. This is a proof of concept and there is no need to merge it until the work is complete.

## Changes

* Added QuietMode flag to root command
* Used QuietMode flag to conditionally print/hide output for the `kn service create` command

## Reference

Proof of concept for https://github.com/knative/client/issues/182

### DIFF
--- a/pkg/kn/commands/logger/logger.go
+++ b/pkg/kn/commands/logger/logger.go
@@ -23,11 +23,11 @@ import (
 )
 
 type Config struct {
-	QuietMode bool
+	Quiet bool
 }
 
 func NewLogger(config Config) io.Writer {
-	if config.QuietMode {
+	if config.Quiet {
 		return io.Discard
 	}
 
--- a/pkg/kn/commands/logger/logger.go
+++ b/pkg/kn/commands/logger/logger.go
@@ -1,0 +1,23 @@
+package logger
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+type Config struct {
+	QuietMode bool
+}
+
+func NewLogger(config Config) io.Writer {
+	if config.QuietMode {
+		return io.Discard
+	}
+
+	return io.Writer(os.Stdout)
+}
+
+func Log(writer io.Writer, message string) {
+	fmt.Fprintf(writer, message)
+}
--- a/pkg/kn/commands/logger/logger.go
+++ b/pkg/kn/commands/logger/logger.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2023 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package logger
 
 import (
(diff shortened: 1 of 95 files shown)


pkg/kn/root/root.go
```diff
@@ -92,6 +92,7 @@ Find more information about Knative at: https://knative.dev`, rootName),
 	rootCmd.PersistentFlags().StringVar(&p.KubeAsUID, "as-uid", "", "uid to impersonate for the operation")
 	rootCmd.PersistentFlags().StringArrayVar(&p.KubeAsGroup, "as-group", []string{}, "group to impersonate for the operation, this flag can be repeated to specify multiple groups")
 	flags.AddBothBoolFlags(rootCmd.PersistentFlags(), &p.LogHTTP, "log-http", "", false, "log http traffic")
+	rootCmd.PersistentFlags().BoolVarP(&p.QuietMode, "quiet-mode", "q", false, "run commands in quiet mode")
```
dsimansk, line 95: I prefer as short names as possible. :)

```suggestion
	rootCmd.PersistentFlags().BoolVarP(&p.QuietMode, "quiet", "q", false, "run commands in quiet mode")
```

dsimansk:
It seems `--quiet` flag is conflicting in some of the unit tests. Aka it's working and discarding the outputs.

pkg/kn/commands/logger/logger.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+type Config struct {
+	QuietMode bool
+}
```
dsimansk, line 27: `Config` struct should accept io.Writer as well. There might be other use cases that benefit from output redirect, not always being `os.Stdout`. Then it should be initialized with value of: 

```
cmd.OutOrStdout()
```

dsimansk: That's the case for failing unit tests, because output is redirected and captured to assert.

dsimansk (changes requested):

pkg/kn/commands/logger/logger.go
```diff
@@ -0,0 +1,39 @@
...
+
+func NewLogger(config Config) io.Writer {
```
rhuss, line 29: Instead of returning a writer, I probably would initialize a global variable here, that stores the write, so that subsequent calls to `Log` would need to pass back in the Writer again. E.g rename this just to `func InitLogger(conf Config) { ...` (without a return value) then store a static variable `outWriter` and use that `outWriter` in the `Log()` method directly.

pkg/kn/commands/logger/logger.go
```diff
@@ -0,0 +1,39 @@
...
+	if config.QuietMode {
+		return io.Discard
+	}
+
+	return io.Writer(os.Stdout)
+}
+
+func Log(writer io.Writer, message string) {
```
rhuss, line 37: I would change this to `Log(fmt string, args interface{})` string to match the signature of Fprintf() (which is also a vararg command, please check its signature). Then use the initialized `outWriter` from above directly in this Log method.

pkg/kn/commands/logger/logger.go
dsimansk: IMO this still seems like a problematic part, at least for unit tests. @sreeram-venkitesh pls let me know if you can take a look at it. Thanks!

rhuss:
@sreeram-venkitesh I'm so sorry that I'm not able to work on these (and oer client-related) issues currently, because of other work assignments. I can't forsee when this will change, but maybe @dsimansk or others can help out ?

What are currently the next steps needed to get this landed ?

Issue #1841, opened 2023-08-03
Ability to set toleration, affinity and node selector


Shashankft9:
### Feature request
I should be able to set toleration, affinity and node selector on my ksvc, when enabled through the featuregate mentioned here: https://knative.dev/docs/serving/configuration/feature-flags.

### Use case
I have a cluster where two nodes are having a different mtu set on them, so I don't want my ksvc pods to be scheduled there. Would have been helpful if i could set it globally through some configmap in knative-serving, but I dont think thats available.

rhuss:
That's a good idea, and it shouldn't be hard to implement it with some extra options. If you are fancy feel free to add it after we have discussed about the naming. We ❤️ contributions ;-)

dsimansk:
I wonder if the impl should check for feature flags to be enabled though. It might be really trick to provide appropriate feedback to user otherwise.

rhuss:
Good point, but I'm afraid that there is no Knative-only API for querying this information ... maybe just try and returning an error with a descriptive error message as feedback. In the help message we can indicate this (e..g with a footnote like add ing a "*" to the message text and then in the bottom something like "only if enabled in the Knative installation"

Shashankft9:
would the following work?
```
--tolerations stringArray    Tolerations to be set. Example: --tolerations Key="key1", Operator="Equal", Value="value1", Effect="NoSchedule"
--affinity stringArray       Node Affinity to be set. Examples: --affinity Type="Required", Key="topology.kubernetes.io/zone", Operator="In", Values="antarctica-east1, antarctica-west1"
--node-selector string       Node Selector to be set. Examples: --node-selector Disktype="ssd"
```

dsimansk:
@Shashankft9 yes that looks like a good start. Please make sure key values are case insensitive on user inputs, but correctly represented on `.PodSpec`. 

Do you need any hints wrt/ the implementation?  You can check similar flags in the `flags` package.

Btw I wonder how Serving validation webhook reacts to providing this value with disabled feature flags. Could you check it please?

Shashankft9:
/remove-lifecycle stale

I am working on it now, will put up a wip PR soon.

PR #1924, opened 2024-03-19
WIP: Assigning nodes

Shashankft9:
## Description
Adds ability to assign knative services to nodes

## Changes

* adds `nodeSelector`, `nodeAffinity` and `toleration` to podspec
* adds update functions in `podspec_helper.go` for each of those fields
* only supports ORed terms for required clause of node affinity
* supports previously added nodeselectors, but no removals in node affinity and toleration since there is no clear identifier.

## Reference

Fixes https://github.com/knative/client/issues/1841

**Release Note**

```release-note

```


Shashankft9:
@dsimansk hey, going to add some more tests, and maybe refactor few things, early feedback would be appreciated!

docs/cmd/kn_service_apply.md
```diff
@@ -77,8 +79,9 @@ kn service apply s0 --filename my-svc.yml
...
       --user int                          The user ID to run the container (e.g., 1001).
-      --volume stringArray                Add a volume from a ConfigMap (prefix cm: or config-map:) a Secret (prefix secret: or sc:), an EmptyDir (prefix ed: or emptyDir:) or a PersistentVolumeClaim (prefix pvc: or persistentVolumeClaim). Example: --volume myvolume=cm:myconfigmap, --volume myvolume=secret:mysecret or --volume emptyDir:myvol:size=1Gi,type=Memory. You can use this flag multiple times. To unset a ConfigMap/Secret reference, append "-" to the name, e.g. --volume myvolume-.
+      --volume stringArray                Add a volume from a ConfigMap (prefix cm: or config-map:) a Secret (prefix secret: or sc:), an EmptyDir (prefix ed: or emptyDir:) or a PersistentVolumeClaim (prefix pvc: or persistentVolumeClaim). PersistentVolumeClaim and EmptyDir only works if the feature gate is enabled in knative serving. Example: --volume myvolume=cm:myconfigmap, --volume myvolume=secret:mysecret or --volume emptyDir:myvol:size=1Gi,type=Memory. You can use this flag multiple times. To unset a ConfigMap/Secret reference, append "-" to the name, e.g. --volume myvolume-.
```
dsimansk, line 84: That's not correct entirely. `EmptyDir` is enabled. Per: https://github.com/knative/serving/blob/ba3f983855bc33555b8416bd745c773fbdd4079c/config/core/configmaps/features.yaml#L183

 I see, the PVC support is still not enabled by default. But I'd still opt changing the docs in this PR and rather create a new targeted one for it.

pkg/kn/commands/service/create.go
```diff
@@ -83,7 +83,16 @@ var create_example = `
...
+  kn service create profiletest --image knativesamples/helloworld --profile istio
+  
+  # Create a service with node selector (if feature flag is enabled here: https://knative.dev/docs/serving/configuration/feature-flags)
+  kn service create nodeselectortest --image knativesamples/helloworld --node-selector Disktype="ssd"
+
+  # Create a service with toleration (if feature flag is enabled here: https://knative.dev/docs/serving/configuration/feature-flags)
+  kn service create tolerationtest --image knativesamples/helloworld --toleration Key="node-role.kubernetes.io/master",Effect="NoSchedule",Operator="Equal",Value=""
+
+  # Create a service with node affinity (if feature flag is enabled here: https://knative.dev/docs/serving/configuration/feature-flags)
+  kn service create nodeaffinitytest --image knativesamples/helloworld --node-affinity Type="Required",Key="topology.kubernetes.io/zone",Operator="In",Values="antarctica-east1 antarctica-east2"`
```
dsimansk, line 95: Preferably those docs link should point to exact feature anchor, e.g. `https://knative.dev/docs/serving/configuration/feature-flags/#kubernetes-toleration`.

Shashankft9:
@dsimansk thanks for the feedback, I'll work on the pointers. But regarding the usefulness, we currently have one use case that needs something like this in kn cli, I will try to justify below:

So in some of edge cloud deployments and dev kubernetes clusters, we are using knative functions with tekton (on-cluster builds), but in addition to its current cli form, we have also used the function client and built a controller on top of it, that provides a CRD like UX for function creation. Currently on-cluster build has three tekton tasks - git clone, build and deploy, but this poses a problem when someone has to add things like scaling configurations, tls and other serving specific configurations. To do this with func cli, its quite easy, but when doing it through on-cluster builds, it implies that the user has to do the changes in [func.yaml](https://github.com/knative/func/blob/main/docs/reference/func_yaml.md#options) and then commit it in git and then run on-cluster builds, which goes quite against the UX that we are wanting to provide through CRD.
So, how do we solve this? 
We added another tekton task after deploy which we called as `kn-patch`, so now we accept scaling configurations in our CRD, and then those configurations we pass in this task, which essentially applies all required changes without the user ever doing any changes in git or even knowing that the function exists somewhere in git. 

But then, we came across one scenario in our edge cloud deployments where we had certain nodes in a kubernetes cluster dedicated to specific workloads and had a different mtu packet size set on them which had an affect on the usual service/pod networking for those nodes, essentially meaning that our functions _must not_ schedule on those nodes, so we have to make functions stick to some nodes, hence the need for these mechanisms through kn.

Does that make sense merely from the use-case perspective? I am not sure how others are using kn cli, but this is how we are currently using it.

>I wonder how the error message looks like from Serving's webhook. If it propagates a good hint for users why Ksvc creation failed.

For this, I can check and post here, but I think last when I tried, Serving's webhook gave a clear hint around it

Shashankft9:
here's what it looks like when i disable node affinity, node selector and toleration from feature flags:
```
root@faas-cluster-xnts4:~# kn service create nodeaffinitytest --image knativesamples/helloworld --node-affinity Type="Required",Key="topology.kubernetes.io/zone",Operator="In",Values="antarctica-east1 antarctica-east2"
Error: admission webhook "validation.webhook.serving.knative.dev" denied the request: validation failed: must not set the field(s): spec.template.spec.affinity
Run 'kn --help' for usage
```

```
root@faas-cluster-xnts4:~# kn service create tolerationtest --image knativesamples/helloworld --toleration Key="node-role.kubernetes.io/master",Effect="NoSchedule",Operator="Equal",Value=""
Error: admission webhook "validation.webhook.serving.knative.dev" denied the request: validation failed: must not set the field(s): spec.template.spec.tolerations
Run 'kn --help' for usage
```

```
root@faas-cluster-xnts4:~# kn service create nodeselectortest --image knativesamples/helloworld --node-selector Disktype="ssd"
Error: admission webhook "validation.webhook.serving.knative.dev" denied the request: validation failed: must not set the field(s): spec.template.spec.nodeSelector
Run 'kn --help' for usage
```

dsimansk:
> @dsimansk thanks for the feedback, I'll work on the pointers. But regarding the usefulness, we currently have one use case that needs something like this in kn cli, I will try to justify below:
> 
> So in some of edge cloud deployments and dev kubernetes clusters, we are using knative functions with tekton (on-cluster builds), but in addition to its current cli form, we have also used the function client and built a controller on top of it, that provides a CRD like UX for function creation. Currently on-cluster build has three tekton tasks - git clone, build and deploy, but this poses a problem when someone has to add things like scaling configurations, tls and other serving specific configurations. To do this with func cli, its quite easy, but when doing it through on-cluster builds, it implies that the user has to do the changes in [func.yaml](https://github.com/knative/func/blob/main/docs/reference/func_yaml.md#options) and then commit it in git and then run on-cluster builds, which goes quite against the UX that we are wanting to provide through CRD. So, how do we solve this? We added another tekton task after deploy which we called as `kn-patch`, so now we accept scaling configurations in our CRD, and then those configurations we pass in this task, which essentially applies all required changes without the user ever doing any changes in git or even knowing that the function exists somewhere in git.
> 
> But then, we came across one scenario in our edge cloud deployments where we had certain nodes in a kubernetes cluster dedicated to specific workloads and had a different mtu packet size set on them which had an affect on the usual service/pod networking for those nodes, essentially meaning that our functions _must not_ schedule on those nodes, so we have to make functions stick to some nodes, hence the need for these mechanisms through kn.
> 
> Does that make sense merely from the use-case perspective? I am not sure how others are using kn cli, but this is how we are currently using it. Maybe this is something that can be useful for the func's CRD and Operator story?
> 
> > I wonder how the error message looks like from Serving's webhook. If it propagates a good hint for users why Ksvc creation failed.
> 
> For this, I can check and post here, but I think last when I tried, Serving's webhook gave a clear hint around it

Thanks, for the extensive reply to support the usefulness concern. I'm getting a better picture now.

One idea how to address my concern might be introducing "experimental" or "advanced" section in the help message. I.e. to de-clutter current list of `kn service` flags and split them into sections. To clearly indicate that this flags require additional configuration like adding feature flag to Serving. 
I'll take a look how to achieve it in `spf13/cobra`, I recall there are a few options how to create a subsections. And of course descriptive sub-section names (naming game is always hardests part :)).

Shashankft9:
hello @dsimansk , this is ready for review - I have reverted the changes done for creating subsections in flags, and added tests

dsimansk:
@Shashankft9 could you pls try to rerun `./hack/build.sh -c` that should execute gofmt and fix formatting issues.

pkg/kn/flags/podspec.go
```diff
@@ -159,7 +162,7 @@ func (p *PodSpecFlags) AddFlags(flagset *pflag.FlagSet) []string {
 	flagset.StringArrayVarP(&p.Volume, "volume", "", []string{},
 		"Add a volume from a ConfigMap (prefix cm: or config-map:) a Secret (prefix secret: or sc:), "+
 			"an EmptyDir (prefix ed: or emptyDir:) or a PersistentVolumeClaim (prefix pvc: or persistentVolumeClaim). "+
-			"Example: --volume myvolume=cm:myconfigmap, --volume myvolume=secret:mysecret or --volume emptyDir:myvol:size=1Gi,type=Memory. "+
+			"PersistentVolumeClaim only works if the feature gate is enabled here: https://knative.dev/docs/serving/configuration/feature-flags/#kubernetes-persistentvolumeclaim-pvc. Example: --volume myvolume=cm:myconfigmap, --volume myvolume=secret:mysecret or --volume emptyDir:myvol:size=1Gi,type=Memory. "+
```
dsimansk, line 165: I'd rather refer to "Knative Serving feature flags congugration" without exact URL, as it might not age very well over time.

pkg/kn/flags/podspec_helper.go
```diff
@@ -398,6 +399,152 @@ func UpdateSecurityContext(spec *corev1.PodSpec, securityContext string) error {
...
+	var nodeAffinityType string
+	for key, value := range nodeAffinityAllMap {
+		if strings.ToLower(key) == "type" {
+			if value == "Required" {
+				nodeAffinityType = "Required"
+				if !reflect.ValueOf(spec.Affinity).IsZero() && !reflect.ValueOf(spec.Affinity.NodeAffinity).IsZero() &&
+					!reflect.ValueOf(spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution).IsZero() {
+					nodeSelectorTermsExisting = spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
+					//TODO: only supporting ORed terms, also support ANDed expressions in a single term
+					//TODO: only supporting matchExpressions, also support matchFields
```
dsimansk, line 482: @Shashankft9 would you like to capture those in a follow-up and address later?

pkg/kn/flags/podspec_helper.go
```diff
@@ -398,6 +399,152 @@ func UpdateSecurityContext(spec *corev1.PodSpec, securityContext string) error {
...
+	matchExpressionNew := v1.NodeSelectorRequirement{}
+	nodeAffinityAllMap, err := util.MapFromArray(nodeAffinity, "=")
+	if err != nil {
+		return err
+	}
+	var nodeAffinityType string
+	for key, value := range nodeAffinityAllMap {
+		if strings.ToLower(key) == "type" {
+			if value == "Required" {
+				nodeAffinityType = "Required"
```
dsimansk, line 477: It's a small nit. I would probably make it less restrictive and with `strings.ToLower(value) == "required"`. We have that at other places that parse such inputs. But it doesn't need to addressed immediately.

Shashankft9:
@dsimansk ack, I can work on those improvements in followup PR - thanks!

PR #1927, opened 2024-03-21
Populate container.SecurityContext.RunAsUser when --user flag is used

### LINKED ISSUE #1926: CLI: `--user` option does not work as expcted
### Bug report

Specifying `--user` flag does not populate the knative service yaml with the supplied value.

### Expected behavior

` kn service create s0 --image knativesamples/helloworld --user 1001` 

should set the value of `RunAsUser` field in container security context to 1001 but it remains unpopulated.

### kn version

```
gvyas-mac:knative gvyas$ kn version
Version:      v1.13.0
Build Date:   2024-01-24 16:48:59
Git Revision: 543522a3
Supported APIs:
* Serving
  - serving.knative.dev/v1 (knative-serving v1.13.0)
* Eventing
  - sources.knative.dev/v1 (knative-eventing v1.13.0)
  - eventing.knative.dev/v1 (knative-eventing v1.13.0)
```

> Nightly
> 0.12.x
> 0.11.x
> 0.10.x
> 0.9.x

vyasgun:
## Changes

* Setting container's RunAsUser field based on the `--user` flag at the end so it does not reinitialize the security context

## Reference

Fixes #1926 

**Release Note**

```release-note

```

### DIFF
--- a/pkg/kn/flags/podspec.go
+++ b/pkg/kn/flags/podspec.go
@@ -372,13 +372,6 @@ func (p *PodSpecFlags) ResolvePodSpec(podSpec *corev1.PodSpec, flags *pflag.Flag
 		UpdateImagePullSecrets(podSpec, p.ImagePullSecrets)
 	}
 
-	if flags.Changed("user") {
-		err = UpdateUser(podSpec, p.User)
-		if err != nil {
-			return err
-		}
-	}
-
 	if flags.Changed("containers") || flags.Changed("extra-containers") || p.ExtraContainers == "-" {
 		var fromFile *corev1.PodSpec
 		fromFile, err = decodeContainersFromFile(p.ExtraContainers)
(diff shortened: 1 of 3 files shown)


pkg/kn/flags/podspec.go
```diff
@@ -418,5 +411,12 @@ func (p *PodSpecFlags) ResolvePodSpec(podSpec *corev1.PodSpec, flags *pflag.Flag
 		}
 	}
 
+	if flags.Changed("user") {
+		err = UpdateUser(podSpec, p.User)
+		if err != nil {
+			return err
+		}
+	}
+
```
dsimansk, line 420: Is there a reason to change the order of evaluation here?

dsimansk: IMO it's very likely SC was override by `UpdateSecurityContext`.

dsimansk:
It would be nice to have a simple unit tests for the change.

dsimansk:
@vyasgun gentle ping. If you could check my comments, pls.

pkg/kn/flags/podspec.go
vyasgun: The order of evaluation is causing the value of the user flag to get overwritten. I have created an issue for this: https://github.com/knative/client/issues/1926

vyasgun:
Thanks for the review, @dsimansk! Updated the PR with a small change which will test both flags together.

dsimansk:
@vyasgun I still have a small hesitation how to fix the issue actually. Maybe the `--security-context` should behave differently. To be able to merge with existing values etc.

PR #1935, opened 2024-04-16
:broom: Dedup code by using client-pkg

cardil:
## Changes

* :broom: Dedup code by using client-pkg

## Reference

Fixes knative-extensions/kn-plugin-event#282

PR #1953, opened 2024-06-20
Fold in knative.dev/client-pkg as a submodule

### LINKED ISSUE #1941: Proposal: Fold back in client-pkg to Go submodule
## Summary

Move back the client-pkg code to this repository, and create `pkg` Go submodule, that could be consumed by plugins or wrappers.

## Problem

Currently, after extracting the client-pkg repository (see: https://github.com/knative/client/issues/1039), but not removing the code in the client, we end up with duplicated code both here and in the client-pkg. Most of the code is actually newer here, but plugins consume client-pkg. Of course, that's messy.

I tried to fix that in these PRs:

- https://github.com/knative/client/pull/1935
- https://github.com/knative/client-pkg/pull/177

These PRs effectively removed most of the code from the client in favor of the client-pkg. That made me realize, though, that it's probably going to make contributing to the project more cumbersome. Almost any change will need to be done in client-pkg, and then after the merge, the library reference in the client repo will need to be updated. Most likely in most cases, such a change would also require some refactoring in the client, so the automated auto-bump PRs will not be enough.

## Execution

We can reconcile the two versions of the code (here and in the client-pkg) by choosing the more modern one and putting it back here, in the client repo. At the same time, deprecate the client-pkg repo. To separate the code and allow plugins to use only the public, library part of this client, we could create a Go submodule for the `pkg/` directory here. The non-public code of the client can be moved to the `internal/` directory (following Go's suggested project structure). Also, by adding a `go.work` file, we can effectively make contributing to the client transparent (our tooling already supports this well).

The structure would look as follows:

```
knative.dev/client/
├── go.mod
├── go.work
├── cmd/
├── internal/
└── pkg/
    ├── go.mod
    └── vendor/
```

For the code in the main module, we need a relative replace stanza in the main `go.mod` file:

```gomod
replace knative.dev/client/pkg => ./pkg
```

The `go.work` file, and a separate go modules for the `knative.dev/client` package and `knative.dev/client/pkg` wouldn't be required, if we didn't use embedding of the plugins, in the downstream projects.

## Validation

I exercised such a repository structure in the following projects:

- [github.com/cardil/knative-multimodule-release-check](https://pkg.go.dev/github.com/cardil/knative-multimodule-release-check)
- [github.com/cardil/knative-multimodule-consume](https://pkg.go.dev/github.com/cardil/knative-multimodule-consume)

I was able to successfully release a multi-module repository using the Knative tools. I only had one minor hiccup, but it was not related to the structure, as I was releasing locally, not on the Prow. All of our tools, such as update-deps, testing work as well.

The consuming of such multi-module sub package, was also not a problem. One minor inconvenience was the versioning. In Go, the tag only works just for the main module, as @dsimansk mentioned in https://github.com/knative/client/issues/1039#issuecomment-852189944. Still, our tools work well with it. The update deps tool, the [buoy](https://pkg.go.dev/knative.dev/toolbox/buoy), can float submodules based on the release branch. And we don't currently tag the client package, so very little changes for the consumer.

If we really want the tags to work as well for the `pkg/` submodule, we could double-tag each release in this way:

```
v0.33.0
pkg/v0.33.0
```

Alternatively, maybe we could set up some redirects in a similar way as we support the `knative.dev` domain for go imports.

But personally, I think this is not necessary, as even without tags on the `pkg/` submodule, we are still on par with the current way of working.

The only real drawback to the proposed structure would be breaking Go's remote run feature for kn. Currently, people can run kn from anywhere Go is installed:

```shell
$ go run knative.dev/client/cmd/kn@latest
```

And that will download and execute the client. This breaks by adding the replace stanza in the main `go.mod` file. The `go install` breaks as well:

```shell
$ go run knative.dev/client/cmd/kn@latest              
go: downloading knative.dev/client v0.33.0
go: knative.dev/client/cmd/kn@latest (in knative.dev/client@v0.33.0):
	The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
```

We don't mention this support in any documentation or tutorials, but it's still a bit of a bummer. Personally, I think it's a bug in Golang, because why wouldn't Go support multi-module projects in this way?

### Embed plugins in a dedicated repo

We embed the plugins in kn, downstream. If we create a dedicated repo just for this, say `kn-blundle`, we could avoid adding multiple go modules (and workspace file) as there will be no cycles.

/kind proposal

cardil:
## Description

This PR merges code from the [knative.dev/client-pkg](https://pkg.go.dev/knative.dev/client-pkg), that is missing in this repo, onto the [knative.dev/client/pkg](https://pkg.go.dev/knative.dev/client/pkg). TBD a PR to deprecate the whole of the [knative.dev/client-pkg](https://pkg.go.dev/knative.dev/client-pkg).

## Changes

* :broom: This PR merges code from the `knative.dev/client-pkg`, that is missing in this repo, onto the `knative.dev/client/pkg`
* :broom: Publicize the code in `lib/` dir onto `pkg/`
* :broom: Remove the extra `kn` sub-package
* :broom: Makes `pkg/` as a Go submodule and introduces Go's workspace.

## Reference

Fixes #1941
Closes #1935
Closes knative/client-pkg#177
Requires knative/hack#386

**Release Note**

```release-note
The knative.dev/client-pkg package is now deprecated in favor of the knative.dev/client/pkg package.
```


go.mod
```diff
@@ -3,33 +3,22 @@ module knative.dev/client
...
 	knative.dev/networking v0.0.0-20240611072033-3b8764c0bb4c
 	knative.dev/pkg v0.0.0-20240614135239-339c22b8218c
 	knative.dev/serving v0.41.1-0.20240614080555-1f7cc4852a07
 	sigs.k8s.io/yaml v1.4.0
 )
 
-require k8s.io/utils v0.0.0-20240102154912-e7106e64919e
+replace knative.dev/client/pkg => ./pkg
+
+replace knative.dev/hack => github.com/cardil/knative-hack v0.0.0-20240702091202-6535957027e0
```
cardil, line 21: /hold

Requires https://github.com/knative/hack/pull/386

cardil:
The integration test fails because of net-istio bug: https://github.com/knative-extensions/net-istio/issues/1345

cardil:
Looks like both the github action and the prow tests run just for the main module, not for the `pkg`:

1. https://github.com/knative/client/actions/runs/10639800469/job/29498455266?pr=1953
2. https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_client/1953/unit-tests_client_main/1829644085225328640

dsimansk:
> Looks like both the github action and the prow tests run just for the main module, not for the `pkg`:
> 
> 1. https://github.com/knative/client/actions/runs/10639800469/job/29498455266?pr=1953
> 2. https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_client/1953/unit-tests_client_main/1829644085225328640

Should we go we default implementation of unit tests parsing per go module? 
https://github.com/knative/client/blob/main/test/presubmit-tests.sh#L54-L57

cardil:
> Should we go we default implementation of unit tests parsing per go module? https://github.com/knative/client/blob/main/test/presubmit-tests.sh#L54-L57

Great catch @dsimansk!

Apart from the above, we need to fix the GitHub Action, as it is the one to publish coverage to Codecov:

https://github.com/knative/actions/blob/8d6af971ff477fdd929367f4c33c865ba0082b13/.github/workflows/reusable-go-test.yaml#L55-L56

.github/workflows/knative-go-build.yaml
```diff
@@ -11,4 +11,4 @@ on:
 
 jobs:
   build:
-    uses: knative/actions/.github/workflows/reusable-go-build.yaml@main
+    uses: cardil/knative-actions/.github/workflows/reusable-go-build.yaml@feature/unit-test-each-go-module
```
cardil, line 14: /hold

Restore when https://github.com/knative/actions/pull/223 is merged

.github/workflows/knative-go-test.yaml
```diff
@@ -14,6 +14,6 @@ on:
 
 jobs:
   test:
-    uses: knative/actions/.github/workflows/reusable-go-test.yaml@main
+    uses: cardil/knative-actions/.github/workflows/reusable-go-test.yaml@feature/unit-test-each-go-module
```
cardil, line 17: /hold

Restore when https://github.com/knative/actions/pull/223 is merged

.github/workflows/knative-go-build.yaml
cardil: Looks like https://github.com/knative/actions/pull/223 helps:

- https://github.com/knative/client/pull/1953/checks?check_run_id=29579171848
- https://github.com/knative/client/actions/runs/10672089158/job/29579078173?pr=1953#step:7:49

.github/workflows/knative-go-test.yaml
dsimansk: Approved, the PR should be merged shortly.

dsimansk:
/approve
/lgtm

Thanks a lot @cardil!

cardil:
Maybe I'll do the last one test, before merging - to try using the PKG package in another repo.

cardil:
The https://github.com/knative-extensions/kn-plugin-event/pull/366 looks to work with this structure without any hiccups.

/hold cancel

PR #1954, opened 2024-06-22
added include flags to include security context

### LINKED ISSUE #1788: Service export should handle securityContext field
### Feature request

With more Pod Security Standards enabled clusters we need to address the `securityContext` being present in PodSpec and Containers definitions. Since the two cluster with different PSS config might not be compatible for the same export.

I'd suggest that we remove it in default behavior and add a flag to include it optionally

### UI Example

```

kn service export <ksvc> --with-securityContext
or
kn service export <ksvc> --with-sc
```

 /kind good-first-issue
 /kind proposal

/cc @rhuss @mvinkler

EraKin575:
## Changes

*
*
*

## Reference

Fixes #1788 

**Release Note**

made securityContext optional as a include flag
```release-note

```

### DIFF
--- a/pkg/kn/commands/service/export.go
+++ b/pkg/kn/commands/service/export.go
@@ -197,7 +197,7 @@ func exportLatestService(latestSvc *servingv1.Service, withRoutes bool, withSecu
 	if withRoutes {
 		exportedSvc.Spec.RouteSpec = latestSvc.Spec.RouteSpec
 	}
-	if withSecurityContext {
+	if !withSecurityContext {
 		exportedSvc.Spec.Template.Spec.SecurityContext = nil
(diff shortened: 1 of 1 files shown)


pkg/kn/commands/service/export.go
```diff
@@ -127,40 +127,59 @@ func NewServiceExportCommand(p *commands.KnParams) *cobra.Command {
 	}
 	flags := command.Flags()
 	commands.AddNamespaceFlags(flags, false)
+	flags.StringArray("include", nil, "Include certain info")
```
dsimansk, line 130: I'd prefer to keep the `--with-*` format of option flags. In this case `--with-securit-context` or a short form`--with-sc`.

PR #1968, opened 2024-09-13
:gift_heart: Extract reusable part of sink flag

cardil:
## Changes

* :gift_heart: Extract reusable part of sink flag

## Reference

Fixes #


pkg/output/logging/log_level.go
cardil, line 1: @dsimansk I added changes to the logging package, that I use in https://github.com/knative-extensions/kn-plugin-event/pull/368 (just for the continence of working on that kn-event PR)

I can extract those changes to separate PR if you like.

cardil: Last two commits, basically:

1. https://github.com/knative/client/pull/1968/commits/a38646840657ffa3023f3cbf43786a4ed4b9c3e4
1. https://github.com/knative/client/pull/1968/commits/3a2bcba04752e4cdbf3f600ea52aa20dcc93f272

dsimansk: Sure, that's fine. Thanks for highlighting it.

pkg/commands/subscription/list_test.go
```diff
@@ -69,7 +69,7 @@ func TestSubscriptionList(t *testing.T) {
 		assert.NilError(t, err)
 		ol := strings.Split(out, "\n")
 		assert.Check(t, util.ContainsAll(ol[0], "NAME", "CHANNEL", "SUBSCRIBER", "REPLY", "DEAD LETTER SINK", "READY", "REASON"))
-		assert.Check(t, util.ContainsAll(ol[1], "s0", "InMemoryChannel:imc0", "ksvc:ksvc0", "broker:b00", "broker:b01"))
+		assert.Check(t, util.ContainsAll(ol[1], "s0", "InMemoryChannel:imc0", "ksvc0", "broker:b00", "broker:b01"))
```
dsimansk, line 72: @cardil this `ksvc` prefix was dropped because of changes in `SinkToString` -> `AsText()` functions. It might have been midly useful to have a type reference in this list. But not necessary required.

pkg/commands/types.go
```diff
@@ -246,44 +243,3 @@ func (params *KnParams) RestConfig() (*rest.Config, error) {
 
 	return config, nil
 }
-
-// GetClientConfig gets ClientConfig from KubeCfgPath
-func (params *KnParams) GetClientConfig() (clientcmd.ClientConfig, error) {
```
dsimansk, line 251: Thanks, for cleaning this part. It seems like a very old piece, that doesn't make a lot of sense here.

pkg/commands/subscription/list_test.go
cardil: Yes. I wrote the `AsText` method to be as close as possible to the old one, but I thought that if we default to kservice in the input (`ksvc:`, and `kservice:` prefixes are optional), maybe we should also be consistent on the output side?

If you think, we should print those prefixes, even for kservice, then I can change it back.

BTW. Maybe we should deprecate the `SinkAsString` method (in favor of `AsText`), as it doesn't care about namespaces. If you target something from a different namespace, it will just print it by name, which is confusing to say the least.

pkg/commands/types.go
cardil: Yep. Now, the kubernetes related parsms are in their own struct, which makes far more sense.

BTW. I love the struct composition in Go. Such a great idea, they had...

dsimansk: Ops, in my first comment I thought it's located in different file. I.e. that an isolated command tried to override the config loading. Anyway, it makes sense here, but it's fine to have it extracted as you did. All good here! :)

pkg/commands/subscription/list_test.go
dsimansk: I see, OK let's keep it as optional. Apparently, judging by the test changes we haven't expected a lot of those types prefixes anyway.

